PR #7, opened 2025-11-12
Embedding gemma

dmunch:
## Proposed changes

Adding support for `EmbeddingGemma` to Embedders.

Basically cherry-picked all commits from https://github.com/ml-explore/mlx-swift-examples/pull/398 from to the new repository, made sure everything compiles and ran swift-format. 

Pretty new to MLX, and thought that would be a good learning opportunity and try to get something in. Let me know what kind of modifications etc. you'd still like to have to get this merged. 

## Checklist


Libraries/Embedders/Bert.swift
```diff
@@ -1,6 +1,7 @@
 // Copyright © 2024 Apple Inc.
 
 import MLX
+import MLXLMCommon
```
davidkoski, line 4: Is this just to pick up the quantization?  Think the Embedders should not require MLXLMCommon / MLXLLM if possible.  A copy of Quantization is OK.  If we end up with a _lot_ of duplication then I think MLXLMCommon might make sense.

Libraries/Embedders/Configuration.swift
```diff
@@ -76,6 +77,13 @@ private class ModelTypeRegistry: @unchecked Sendable {
             let model = Qwen3Model(configuration)
             return model
         },
+        "gemma3_text": {
+            url in
+            let configuration = try JSONDecoder().decode(
+                Gemma3TextConfiguration.self, from: Data(contentsOf: url))
```
davidkoski, line 83: I think it makes sense to copy this config type into Embedders rather than add new linkage.  Even sharing config types between models in the same library is rarely done.

Libraries/Embedders/EmbeddingGemma.swift
```diff
@@ -0,0 +1,111 @@
+import MLX
+import MLXLLM
+import MLXLMCommon
```
davidkoski, line 3: See elsewhere -- I think this should be done without adding linkage to additional libraries.

Libraries/MLXLLM/Models/Gemma3Text.swift
```diff
@@ -331,7 +430,7 @@ private class Gemma3Model: Module {
 public class Gemma3TextModel: Module, LLMModel {
 
     @ModuleInfo private var model: Gemma3Model
-    @ModuleInfo(key: "lm_head") var lmHead: Linear
+    @ModuleInfo(key: "lm_head") var lmHead: Module  // Can be Linear or QuantizedLinear
```
davidkoski, line 433: QuantizedLinear is a subtype of Linear so this should have been OK as-is -- did you see a problem here?

davidkoski (changes requested):
@dmunch this looks good overall, but see my comments about not adding a new dependency on MLXLMCommon and MLXLLM.

JaeminKim-amoz (changes requested):
**REQUEST CHANGES — protocol design issue**

Functional embedding gemma support, but the protocol extension approach needs work.

**Issues to address:**
- [HIGH] Three `fatalError("X does not support quantization")` stubs in Bert, NomicBert, and Qwen3 — runtime crashes instead of compile-time safety. Please use a protocol extension with a default implementation, or split into a separate protocol (`QuantizableEmbeddingModel`).
- [MEDIUM] `Gemma3TextConfiguration` gains `useBidirectionalAttention` without a default value — could break existing callers.
- [LOW] `createBidirectionalSlidingWindowMask` and `simpleSDPA` are free functions at module scope — should be namespaced.

davidkoski:
This has since been added in #136 .  Thank you!

PR #12, opened 2025-12-02
Add Arcee-AI's AfMoE

smdesai:
## Proposed changes

This PR adds support for Arcee-AI's AfMoE models. 

## Checklist

- [ x] I have read the [CONTRIBUTING](https://github.com/ml-explore/mlx/blob/main/CONTRIBUTING.md) document
- [ x] I have run `pre-commit run --all-files` to format my code / installed pre-commit prior to committing changes


smdesai:
https://github.com/user-attachments/assets/7471b812-6195-4e4a-ba39-aa753e47ae54

Libraries/MLXLLM/Models/AfMoE.swift
```diff
@@ -0,0 +1,621 @@
...
+            }
+        }
+    }
+}
+
+// MARK: - LoRA Extension
+
+extension AfMoEModel: LoRAModel {
+    public var loraLayers: [Module] {
+        model.layers
```
davidkoski, line 619: I wonder if this should be the attention layers?  Those have the various projection layers attached -- I suspect as-is this won't be LORA trainable.

davidkoski: Never mind, my mistake -- the default lora implementation looks at the linear layers underneath those layers.

PR #14, opened 2025-12-04
Fix many compiler warnings

DePasqualeOrg:
## Proposed changes

There were around 180 compiler warnings, which I've reduced to ~106~ 10. The remaining ones are due to concurrency issues that are more difficult to fix. I think these should be addressed separately. The large number of concurrency warnings probably led to many other warnings being overlooked, which can cause code quality to suffer.

## Checklist

### DIFF
--- a/Libraries/MLXLMCommon/Evaluate.swift
+++ b/Libraries/MLXLMCommon/Evaluate.swift
@@ -538,7 +538,11 @@ public enum GenerateDisposition: Sendable {
 ///   - tokenizer: tokenizer to convert tokens back into strings and recognize special tokens
 ///   - extraEOSTokens: any additional stop tokens
 ///   - didGenerate: visitor for the tokens as they are generated
-@available(*, deprecated, message: "Use the AsyncStream-based generate(input:cache:parameters:context:) instead for better Swift concurrency support")
+@available(
+    *, deprecated,
+    message:
+        "Use the AsyncStream-based generate(input:cache:parameters:context:) instead for better Swift concurrency support"
+)
 public func generate(
     promptTokens: [Int], parameters: GenerateParameters, model: any LanguageModel,
     tokenizer: Tokenizer,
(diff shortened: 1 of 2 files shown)


Libraries/Embedders/Load.swift
```diff
@@ -91,13 +91,6 @@ func loadSynchronous(modelDirectory: URL) throws -> EmbeddingModel {
         }
     }
 
-    if let quantization = baseConfig.quantization {
```
davidkoski, line 94: curious why this was removed -- it is needed for quantization

Libraries/MLXLLM/LLMModel.swift
```diff
@@ -24,12 +24,12 @@ extension LLMModel {
     {
         let prefillStepSize = windowSize ?? 512
         var y = input.text
-        var state: LMOutput.State? = nil
+        let state: LMOutput.State? = nil
```
davidkoski: this seems odd -- a `let` initialized to `nil` can probably be deleted?

Libraries/MLXLMCommon/Adapters/LoRA/LoRAContainer.swift
```diff
@@ -6,7 +6,7 @@
 //
 
 import Foundation
-import MLX
+@preconcurrency import MLX
```
davidkoski: I think this signals an issue -- MLX should have all the right concurrency annotations.

DePasqualeOrg: Okay, I've removed this for MLX. If there's an issue, it can be resolved later with the other concurrency warnings.

Libraries/Embedders/Load.swift
davidkoski: @DePasqualeOrg just looking for feedback on this, the rest looks good

davidkoski: @DePasqualeOrg not sure why this got resolved -- why is this deleted?

DePasqualeOrg:
I already left three comments in that thread with the same explanation, but maybe they're not showing on your end:

> perLayerQuantization happens in the block above the part that I deleted, and the compiler says to use perLayerQuantization instead, so I believe the part I deleted is redundant.

Take a look at the full context of that part, and I think it will be clear.

Libraries/Embedders/Load.swift
davidkoski: Per response it was also done above!

DePasqualeOrg:
Formatted ✅

PR #32, opened 2025-12-25
Expose inner model for all models

DePasqualeOrg:
## Proposed changes

A problem I've encountered while implementing some TTS models with LLM backbones in [mlx-swift-audio](https://github.com/DePasqualeOrg/mlx-swift-audio) is that mlx-swift-lm does not expose the inner model, as mlx-lm in Python does, which means that this part of the model needs to be re-implemented. Exposing the inner model can also be useful for customizing inference with techniques such as masking or speculative decoding.

The changes here mirror the public API of mlx-lm in Python, exposing the inner model via `model.model` (or `model.transformer` for OpenELM, NanoChat, and Lille130m).

This requires helper classes like `Attention` and `MLP` that were previously private to be given unique names to prevent collisions.

## Checklist

### DIFF
--- a/Libraries/MLXLLM/Models/AfMoE.swift
+++ b/Libraries/MLXLLM/Models/AfMoE.swift
@@ -123,7 +123,7 @@ public struct AfMoEConfiguration: Codable, Sendable {
 
 // MARK: - Attention
 
-private class Attention: Module {
+class AfMoEAttention: Module {
     let hiddenSize: Int
     let nHeads: Int
     let nKVHeads: Int
@@ -239,7 +239,7 @@ private class Attention: Module {
 
 // MARK: - MLP
 
-private class MLP: Module, UnaryLayer {
+class AfMoEMLP: Module, UnaryLayer {
     @ModuleInfo(key: "gate_proj") var gateProj: Linear
     @ModuleInfo(key: "down_proj") var downProj: Linear
     @ModuleInfo(key: "up_proj") var upProj: Linear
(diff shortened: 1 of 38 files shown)


Libraries/MLXLLM/Models/AfMoE.swift
```diff
@@ -123,7 +123,7 @@ public struct AfMoEConfiguration: Codable, Sendable {
 
 // MARK: - Attention
 
-private class Attention: Module {
+class AfMoEAttention: Module {
```
davidkoski, line 126: just a note as it took me a few minutes to figure out why this was changed: this has to move to `internal` (implicit visibility) because the property on `AfMoEModelInner` has that visibility.

davidkoski: I think this change is fine, but I wonder if we would prefer to put it inside an enum namespace?  Either way, this isn't making API (not `public`) so it is just an implementation detail.

DePasqualeOrg: So `AfMoE.Attention` instead of `AfMoEAttention`? I think that would be nice. I'll try to implement that for all the models and push a new commit.

davidkoski: Yeah, exactly.  It will end up indenting the whole file so this will be a big one :-)

davidkoski: It could be that the AfMoE is taken by the public model class name, but it can nest in that easily enough.

DePasqualeOrg: I tried this out, and it's quite a lot of work as well as noise in the commit history for a small aesthetic win. It would also result in an inconsistent naming scheme between the internal and public classes. Since the classes that would be renamed are all internal anyway, I would suggest deferring this for later, if at all.

PR #34, opened 2025-12-28
Optimize model loading performance

DePasqualeOrg:
### Parallel loading of tokenizer and weights

Tokenizer loading now runs concurrently with weight loading. For vision models, `preprocessor_config.json` is also loaded in parallel.

### Single `config.json` read

`config.json` was being read from disk twice during model loading. Now it's read once and the data is reused for both the base config and model-specific config decoding.

### Benchmark results

I've added model loading benchmarks in a separate test target that won't run in CI.

To show the total improvement of all my optimizations, I ran the benchmark with [#33](https://github.com/ml-explore/mlx-swift-lm/pull/33) in this repo, including pending optimizations [#302](https://github.com/huggingface/swift-transformers/pull/302), [#303](https://github.com/huggingface/swift-transformers/pull/303), and [#304](https://github.com/huggingface/swift-transformers/pull/304) in swift-transformers and the Hub client in offline mode to avoid a network call. After swift-transformers migrates to swift-huggingface for the Hub API, specify a model repo revision instead to avoid the network call:

| Model | Before | This PR | Improvement |
|-------|--------|-------|-------------|
| LLM (Qwen3-0.6B-4bit) | 345 ms | 297 ms | 48 ms |
| VLM (Qwen2-VL-2B-Instruct-4bit) | 465 ms | 361 ms | 104 ms |

These are the results of the improvement in this PR with swift-transformers in its current, unoptimized state:

| Model | Before | This PR | Improvement |
|-------|--------|-------|-------------|
| LLM (Qwen3-0.6B-4bit) | 3912 ms | 3879 ms | 33 ms |
| VLM (Qwen2-VL-2B-Instruct-4bit) | 4509 ms | 4480 ms | 29 ms |

The cumulative result of all my improvements is that **model loading time goes from ~3900–4500 ms to ~300–360 ms**.

### DIFF
--- a/Libraries/Embedders/EmbeddingModel.swift
+++ b/Libraries/Embedders/EmbeddingModel.swift
@@ -46,8 +46,7 @@ public actor ModelContainer {
     public init(
         hub: HubApi, modelDirectory: URL, configuration: ModelConfiguration
     ) async throws {
-        // Start tokenizer config loading asynchronously, then load model synchronously.
-        // Both operations run in parallel because async let begins execution immediately.
+        // Load tokenizer config and model in parallel using async let.
         async let tokenizerConfigTask = loadTokenizerConfig(
             configuration: configuration, hub: hub)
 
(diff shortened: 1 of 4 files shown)


davidkoski:
Great idea!  I will give this a look in the first week of Jan when I am back

PR #39, opened 2026-01-05
Fix GPTOSS Fatal error in getSlidingWindowMask

jeanmatthieu:
Fix GPTOSS `Fatal error: [broadcast_shapes] Shapes (1,64,512,641) and (1,64,512,640) cannot be broadcast`

 In GPTOSS's getSlidingWindowMask function, the mask size calculation for prefill (L > 1) was off by one compared to what RotatingKVCache actually produces after trimming.

  The Bug:
  - RotatingKVCache is initialized with maxSize = slidingWindow + 1 = 129
  - During prefill with chunk size L, after trimming, the cache contains maxSize + L - 1 entries
  - But the mask was computed with min(windowSize + 1, offset), producing L + min(windowSize + 1, offset) columns
  - This led to mask having more columns than k/v having entries

Tested using opencode and GPTOSS 20B

### DIFF
--- a/Libraries/MLXLLM/Models/GPTOSS.swift
+++ b/Libraries/MLXLLM/Models/GPTOSS.swift
@@ -256,11 +256,11 @@ private class AttentionBlock: Module {
 
         if L > 1 {
             _previousMask = nil
-            return makeMask(L, min(windowSize + 1, offset))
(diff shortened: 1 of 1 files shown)


davidkoski (approved):
Looks good, thank you!

PR #43, opened 2026-01-06
Fix Mistral3TextConfiguration parsing

jeanmatthieu:
Problem:
Loading mlx-community/Ministral-3-8B-Instruct-2512-4bit (and similar VLM-style Mistral3 models) via LLMModelFactory failed with:
```
Unable to set lm_head on Mistral3TextModel: none not compatible with [biases: [131072, 64], scales: [131072, 64]]
```
Root Cause:
The model's config.json has a VLM structure where tie_word_embeddings: false is at the top level, not inside text_config. The parser only checked text_config and defaulted to true, causing lmHead module to not be created, which then failed when loading the quantized lm_head weights.

Changes
  1. First, try to decode `tieWordEmbeddings` from container (which is either the top-level or text_config, depending on the model type).
      If not found there, fallback to topLevelContainer (always the root level).
      If still not found, default to false
  2. Changed tieWordEmbeddings default from true to false - Safer default because:
      - Creating an unused lmHead module is wasteful but works
      - Missing lmHead when loading quantized weights crashes 
      - Matches Mistral3VLMTextConfiguration which also defaults to false

### DIFF
--- a/Libraries/MLXLLM/Models/Mistral3Text.swift
+++ b/Libraries/MLXLLM/Models/Mistral3Text.swift
@@ -435,7 +435,6 @@ public struct Mistral3TextConfiguration: Codable, Sendable {
                 try decoder.container(keyedBy: CodingKeys.self)
             }
 
-
         modelType = try container.decodeIfPresent(String.self, forKey: .modelType) ?? "ministral3"
         hiddenSize = try container.decode(Int.self, forKey: .hiddenSize)
         hiddenLayers = try container.decode(Int.self, forKey: .hiddenLayers)


Libraries/MLXLLM/Models/Mistral3Text.swift
```diff
@@ -423,41 +423,56 @@ public struct Mistral3TextConfiguration: Codable, Sendable {
...
         // Handle layer_types with default to all full_attention
-        if let types = try container.decodeIfPresent([String].self, forKey: .layerTypes) {
+        if let types = try decodeWithFallback([String].self, forKey: .layerTypes) {
             layerTypes = types
         } else {
             layerTypes = Array(repeating: "full_attention", count: hiddenLayers)
         }
 
-        slidingWindow = try container.decodeIfPresent(Int.self, forKey: .slidingWindow)
+        slidingWindow = try decodeWithFallback(Int.self, forKey: .slidingWindow)
```
davidkoski: I wonder if this should be done more like how Gemma3n does it?  I think it is roughly the same problem.

```swift
    public init(from decoder: Decoder) throws {
        let nestedContainer = try decoder.container(keyedBy: VLMCodingKeys.self)

        // in the case of Gemma 3n model, the configuration matches VLMs and text config is under a text_config key
        let container =
            if nestedContainer.contains(.textConfig) {
                try nestedContainer.nestedContainer(keyedBy: CodingKeys.self, forKey: .textConfig)
            } else {
                try decoder.container(keyedBy: CodingKeys.self)
            }
```

jeanmatthieu: I've just adjusted the changes to only focus on `tieWordEmbeddings` decoding instead trying to generalise to any keys from `textConfig`. I couldn't find any specifications whether `tieWordEmbeddings` can actually be part of `textConfig` so perhaps we are doing an extra `decodeIfPresent` call that will always fail but at least we are looking it up in the top level container now

PR #45, opened 2026-01-07
Fix #44 Add support for chat re-hydration

### LINKED ISSUE #44: Feature Request: Support Chat History Injection (Re-hydration) in ChatSession
### Description

I am trying to build a persistent chat application using your library: **mlx-swift-lm**. Currently, ChatSession is designed as an ephemeral session; it initializes with an empty message list (just a system prompt) and stores messages as a private property.

This makes it difficult to implement standard "Prompt Re-hydration" patterns where a conversation is saved to a local database (like SwiftData) and restored upon app launch. Since I cannot inject the restored [Chat.Message] history into a new ChatSession, I cannot leverage the high-level convenience of this class for restored sessions.
I can see as well that the last version of the source code the ChatSession is marked as final class so I cannot even extend it to add this feature. 

### Proposed Solution

I propose adding a new initializer to ChatSession that accepts an existing history: [Chat.Message] array. This would allow developers to load a conversation from disk, pass it to ChatSession, and let the MLX engine handle the context "prefill" automatically upon the next generation request.

### Potential Implementation

Here is the proposed addition to ChatSession.swift. This overloads the init to accept a history parameter instead of just instructions.

```swift
// In Libraries/MLXLMCommon/ChatSession.swift

extension ChatSession {
    
    /// Initialize the `ChatSession` with an existing message history.
    /// This enables "Prompt Re-hydration" for persistent chat applications.
    ///
    /// - Parameters:
    ///   - model: the ``ModelContainer``
    ///   - history: The full array of messages to restore (including system prompt)
    ///   - generateParameters: parameters that control generation
    ///   - processing: media processing configuration for images/videos
    ///   - additionalContext: optional model-specific context
    public convenience init(
        _ model: ModelContainer,
        history: [Chat.Message],
        generateParameters: GenerateParameters = .init(),
        processing: UserInput.Processing = .init(resize: CGSize(width: 512, height: 512)),
        additionalContext: [String: any Sendable]? = nil
    ) {
        self.init(model, instructions: nil, generateParameters: generateParameters, processing: processing, additionalContext: additionalContext)
        
        // Overwrite the default empty messages with the restored history
        self.messages = history
        // The cache remains empty ([]); MLX will automatically trigger a prefill 
        // for the history when `respond()` is called.
    }

    /// Initialize the `ChatSession` with an existing message history (ModelContext variant).
    public convenience init(
        _ model: ModelContext,
        history: [Chat.Message],
        generateParameters: GenerateParameters = .init(),
        processing: UserInput.Processing = .init(resize: CGSize(width: 512, height: 512)),
        additionalContext: [String: any Sendable]? = nil
    ) {
        self.init(model, instructions: nil, generateParameters: generateParameters, processing: processing, additionalContext: additionalContext)
        self.messages = history
    }
}
```

### Use Case /  Benefit

Persistence: Allows apps to save chats to SwiftData and restore them seamlessly.

Stateless Architecture: Encourages the standard practice of using the database as the source of truth and treating the LLM context as transient.

aleroot:
## Proposed changes

Implement basic chat re-hydration to reload context from disk.

## Checklist

### DIFF
--- a/Tests/MLXLMTests/ChatSessionTests.swift
+++ b/Tests/MLXLMTests/ChatSessionTests.swift
@@ -90,7 +90,7 @@ public class ChatSessionTests: XCTestCase {
         print("Vision result:", result)
         XCTAssertTrue(result.lowercased().contains("red"))
     }
-    
+
     func testPromptRehydration() async throws {
         // Simulate a persisted history (e.g. loaded from SwiftData)
         let history: [Chat.Message] = [


aleroot:
> Changes look good, thank you!

Thanks to you for adding this.
I have seen there was a format issue, I have tried to fix.
Is there a new release coming out soon?

PR #50, opened 2026-01-09
Add MiniMax and MiMo v2 Flash models

ronaldmannak:
## Proposed changes

This PR includes two ports: MiniMax and MiMo v2 Flash. As with the previous PR, I don’t have enough memory available to run a full verification of the models. However, I reviewed the implementation and checked it against the reference Python code

## Checklist

### DIFF
--- a/Libraries/MLXLLM/Models/MiMoV2Flash.swift
+++ b/Libraries/MLXLLM/Models/MiMoV2Flash.swift
@@ -191,7 +191,7 @@ class MiMoV2FlashAttention: Module {
 
         return wo(output)
     }
-    
+
     override func updateMissing(
         parameter: String,
         verify: VerifyUpdate,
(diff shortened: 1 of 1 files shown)


ronaldmannak:
@davidkoski This PR builds without issues using MLX 0.30.2 and is ready for review and to test on a device with more memory than I have.

davidkoski:
> @davidkoski So MiniMax runs fine, but MiMo fails. In hindsight I should've created two separate PRs.

I don't know how many times I have said that myself :-)

> Is that correct?

Yes, exactly.

ronaldmannak:
> I don't know how many times I have said that myself :-)

Sorry about that. I was lazy and overconfident :)

I'll update the code

ronaldmannak:
@davidkoski override added. I'm calling super instead of throwing the error from within the override just in case the super implementation or error changes in the future

davidkoski:
OK, the corresponding change in mlx-swift is picked up, I think this just needs a rebase.

davidkoski (approved):
Looks great, thank you!

Issue #19, opened 2025-12-10
[BUG] Qwen3VL-4B-4bit is slow in generation


jimmycmlo:
**Describe the bug**
Running mlx-swift-example/VLMEval, using Qwen3-VL-4B-MLX-4bit, the generation speed at about 10 tokens/s. which is significantly slower than Qwen2.5-VL-3B-Instruct-4bit at 55 tokens/s and Qwen2-VL-2B-Instruct at 74 tokens/s. This is using a MacBook Pro M4 Pro on macOS Tahoe 26.2. I understand Qwen3-VL-4B-MLX-4bit is a bit bigger but 10 tokens/s is below expectation. I also ran Qwen3-VL-2B-MLX-4bit at 13 tokens/s, so still very slow.

Any idea why it is so much slower. I will be happy to investigate and debug if you can give me some idea where to look.

jimmycmlo:
So it looks like mlx-vlm is working fine at 181 tokens/second:

mlx_vlm.generate --model mlx-community/Qwen3-VL-2B-Instruct-4bit --max-tokens 100 --temperature 0.0 --image http://images.cocodataset.org/val2017/000000039769.jpg
Fetching 14 files: 100%|█████████████████████████████████████████████████████████████| 14/14 [00:00<00:00, 43755.78it/s]
Download complete: : 0.00B [00:00, ?B/s]                                                         | 0/14 [00:00<?, ?it/s]
==========
Files: ['http://images.cocodataset.org/val2017/000000039769.jpg'] 

Prompt: <|im_start|>user
<|vision_start|><|image_pad|><|vision_end|>What are these?<|im_end|>
<|im_start|>assistant

Based on the image provided, here is a description of the scene:

The image features two cats resting on a bright pink surface, which appears to be a blanket or a sheet on a couch. The cats are both sleeping or lounging on their backs.

- The cat on the left is a tabby with a pattern of dark stripes on a lighter, tan-colored body. It is lying on its back with its paws and tail visible.
- The cat on the right is a tabby with
==========
Prompt: 314 tokens, 219.465 tokens-per-sec
Generation: 100 tokens, 181.582 tokens-per-sec
Peak memory: 2.545 GB

Some tips on where to look?

scouzi1966:
@ronaldmannak Thanks!  By any chance, what do you use to test the Swift implementation? Do you use the Chat example (Applications/MLXChatExample) or MLXLMTests from this repo?

[...]

PR #51, opened 2026-01-09
GPT-oss performance optimizations

ronaldmannak:
## Proposed changes

I looked at the difference in the Python and Swift implementation of GPT-oss after seeing comment 
https://github.com/ml-explore/mlx-swift-lm/issues/19#issuecomment-3726753860 and found a few differences. With this PR the GPT-oss implementation is now on par with the Python implementation.
I'm also seeing a solid 10% performance boost on an M1 Max. It's not the 2-3x difference OP mentioned, but it is possible that different CPUs yield different results.

## Checklist


davidkoski (approved):
Looks good, thank you!

PR #55, opened 2026-01-09
fix thread safety issues

davidkoski:
## Proposed changes

- support for https://github.com/ml-explore/mlx-swift-examples/pull/454
- ModelContainer appeared to provide thread safe access to the KVCache and model
    - but in fact was not -- async token generation could use the KVCache concurrently
    - if you were to break the async stream early the previously call could still be running

## Checklist

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -26,7 +26,7 @@ let package = Package(
             targets: ["MLXEmbedders"]),
     ],
     dependencies: [
-        .package(url: "https://github.com/ml-explore/mlx-swift", .upToNextMinor(from: "0.30.2")),
+        .package(url: "https://github.com/ml-explore/mlx-swift", .upToNextMinor(from: "0.30.3")),
         .package(
             url: "https://github.com/huggingface/swift-transformers",
             .upToNextMinor(from: "1.1.6")


Libraries/MLXLMCommon/Utilities/SerialAccessContainer.swift
```diff
@@ -0,0 +1,118 @@
...
+        return try await body()
+    }
+}
+
+/// Provide serial exclusive access to state `<T>` to async callers.
+///
+/// Unlike an `actor`, this will guarantee exclusive access for the duration of the async
+/// call.  This is important for things like `ModelContainer` that have to perform async
+/// work but also need to prevent other callers for using _any_ of the internal state.
+final class SerialAccessContainer<T>: @unchecked Sendable {
```
davidkoski, line 44: This is the heart of the change.  In Swift an `actor` does provide exclusive access to a resource, but there are some caveats with `async` code -- it suspends the exclusive access during the `await`.  In the case of the `ChatSession` this means that the `KVCache` could be used concurrently by multiple tasks if a streaming response was interrupted (break the generation loop early) and a new stream is started immediately.

A different approach was required, thus this class.  You can't use locking mechanisms with asynchronous code, so the `AsyncMutex` provides something equivalent where callers will wait their turn.

Libraries/MLXLMCommon/ChatSession.swift
```diff
@@ -20,14 +20,15 @@ import MLX
...
-        case context(ModelContext)
+    enum Cache {
+        case empty
+        case kvcache([KVCache])
+        case history([Chat.Message])
     }
 
-    private let model: Model
-    private var messages: [Chat.Message]
-    private var cache: [KVCache]
```
davidkoski, line 30: On inspection of mlx-lm's chat implementation, it didn't keep the entire message history -- that is encapsulated in the KVCache.  This moves closer to that but has the ability to rebuild the KVCache by replaying a message history.

Libraries/MLXLMCommon/ChatSession.swift
```diff
@@ -20,14 +20,15 @@ import MLX
...
+    private let model: ModelContainer
+    private let instructions: String?
+    private let cache: SerialAccessContainer<Cache>
```
davidkoski, line 31: This use of SerialAccessContainer is they key piece that protects access to the KVCache (mutable per-session state).

Libraries/MLXLMCommon/ChatSession.swift
```diff
@@ -88,21 +89,20 @@ public final class ChatSession {
...
-            processing: processing,
-            additionalContext: additionalContext
-        )
-        self.messages = history
+        self.model = model
+        self.instructions = instructions
+        self.cache = .init(.history(history))
+        self.processing = processing
+        self.generateParameters = generateParameters
+        self.additionalContext = additionalContext
```
davidkoski, line 105: This is adapting changes from main that rebuild from a chat history.

Libraries/MLXLMCommon/ChatSession.swift
```diff
@@ -141,45 +140,13 @@ public final class ChatSession {
...
-        switch model {
-        case .container(let container):
-            output = try await container.perform { context in
-                try await generate(context: context)
-            }
-        case .context(let context):
-            output = try await generate(context: context)
+        var output = ""
+        for try await chunk in streamResponse(to: prompt, images: images, videos: videos) {
+            output += chunk
```
davidkoski, line 148: This code is getting more complex to manage the thread safety and we don't really need two implementations.  Just call the streaming version and build the result string.  It was doing roughly this before.

Libraries/MLXLMCommon/ChatSession.swift
```diff
@@ -211,16 +178,105 @@ public final class ChatSession {
...
     ) -> AsyncThrowingStream<String, Error> {
-        messages.append(.user(prompt, images: images, videos: videos))
-
         let (stream, continuation) = AsyncThrowingStream<String, Error>.makeStream()
 
+        // images and videos are not Sendable (MLXArray) but they are consumed
+        // and are only being sent to the inner async
+        let message = SendableBox<Chat.Message>(
+            .user(prompt, images: images, videos: videos)
+        )
```
davidkoski, line 226: For values that are not Sendable, we use SendableBox (internal to the library) to pass them between Tasks after proving that it is safe.  In this case we `consume` the inputs so the caller cannot use them after calling into this.

Libraries/MLXLMCommon/ChatSession.swift
```diff
@@ -211,16 +178,105 @@ public final class ChatSession {
...
+                    // should not be used in calling code.
+                    //
+                    // The benefit is that callers can be running multiple
+                    // ChatSessions in parallel, as long as the instances
+                    // are distinct.  In particular the KVCache cannot
+                    // be shared and that is the lock that is held here.
+
+                    let model = await model.perform { context in
+                        SendableBox(context.model)
+                    }.consume()
```
davidkoski, line 261: Previously we held a lock on the model during the prepare phase of the token iterator.  That isn't strictly necessary as the model weights are fully evaluated.  They are not Sendable, but we can provide this functionality inside the library.

Libraries/MLXLMCommon/Evaluate.swift
```diff
@@ -758,6 +758,12 @@ public func generate(
 /// instances of the `Generation` enum, which can represent text chunks, tool calls, or summary
 /// completion information.
 ///
+/// * Important: if the stream is terminated early (e.g. break from the loop) computation will continue
+/// using the model, parameters, KVCache, etc. for some time (typically a few ms).  This is typically OK for
+/// one-shot calls, but for "chat session" type calls consider using
+/// ``generateTask(promptTokenCount:context:iterator:)``
+/// so that the end of the generation task can be observed.
```
davidkoski, line 771: Documentation on when the stream processing actually ends and how callers can observe that if needed.

Libraries/MLXLMCommon/Evaluate.swift
```diff
@@ -796,52 +802,69 @@ public func generate(
...
-    input: LMInput, context: ModelContext,
-    iterator: TokenIterator
-) -> AsyncStream<Generation> {
+/// - Returns: An `AsyncStream` that emits `Generation` values and a `Task`
+public func generateTask(
+    promptTokenCount: Int,
+    modelConfiguration: ModelConfiguration,
+    tokenizer: Tokenizer,
+    iterator: consuming TokenIterator
+) -> (AsyncStream<Generation>, Task<Void, Never>) {
```
davidkoski, line 852: Adjusted the arguments to avoid non-Sendable values when they were not actually needed.

Libraries/MLXLMCommon/ModelContainer.swift
```diff
@@ -31,31 +31,55 @@ import Tokenizers
 ///     }
 /// }
 /// ```
-public actor ModelContainer {
-    var context: ModelContext
-    public var configuration: ModelConfiguration { context.configuration }
+public final class ModelContainer: Sendable {
+    private let context: SerialAccessContainer<ModelContext>
```
davidkoski, line 35: A complete reimplementation of ModelContainer based now on SerialAccessContainer instead of being an actor.  See elsewhere for details on why.

Libraries/MLXLMCommon/ModelContainer.swift
```diff
@@ -31,31 +31,55 @@ import Tokenizers
...
+        get async {
+            await context.read { $0.configuration }
+        }
+    }
+
+    public var processor: UserInputProcessor {
+        get async {
+            await context.read { $0.processor }
+        }
+    }
```
davidkoski, line 47: These are all Sendable values now so we can safely expose them like this.  They are _mutable_ (under lock) so they aren't just simple properties.

Libraries/MLXLMCommon/ModelContainer.swift
```diff
@@ -97,8 +127,9 @@ public actor ModelContainer {
     /// - Returns: Prepared language model input (transferred via `sending`)
     /// - Note: The `sending` keyword indicates the return value is transferred (not shared),
     ///   allowing non-Sendable types like `LMInput` to safely cross isolation boundaries.
-    public func prepare(input: UserInput) async throws -> sending LMInput {
-        try await context.processor.prepare(input: input)
+    public func prepare(input: consuming sending UserInput) async throws -> sending LMInput {
+        let processor = await self.processor
+        return try await processor.prepare(input: input)
```
davidkoski, line 132: Note that the prepare no longer requires an exclusive lock on the model (the processor is Sendable and can be used concurrently).

Tests/MLXLMTests/ChatSessionTests.swift
```diff
@@ -1,114 +1,163 @@
...
+            hiddenSize: 64, hiddenLayers: 8, intermediateSize: 128, attentionHeads: 8,
+            headDim: 256,
+            rmsNormEps: 0.00001, vocabularySize: 100, kvHeads: 4,
+            ropeTheta: 1_000_000, ropeLocalBaseFreq: 10_000,
+            ropeTraditional: false, queryPreAttnScalar: 256,
+            slidingWindow: 512, slidingWindowPattern: 6,
+            maxPositionEmbeddings: 32768
+        )
+        let model = Gemma3TextModel(config)
+        quantize(model: model, groupSize: 64, bits: 4)
```
davidkoski, line 27: This is actually a test for the `gemma-fixes` branch but it is a bit intertwined with the tests for the thread safety so just including it here.  This exercises the attention mask with a random weigh model.

Tests/MLXLMTests/ChatSessionTests.swift
```diff
@@ -1,114 +1,163 @@
...
+            // the next call can corrupt the state if not thread safe
+
+            var result2 = ""
+            for try await part in session.streamResponse(to: "hello again") {
+                result2 += part
+                if result2.count > 100 {
+                    break
+                }
+            }
+        }
```
davidkoski, line 93: This simulates the case where I initially discovered the thread safety issue.

Tests/MLXLMTests/EvalTests.swift
```diff
@@ -146,14 +148,15 @@ public class EvalTests: XCTestCase {
...
                     var taskResults: [Int] = []
                     let sampler = CategoricalSampler(temperature: 1.0)
 
                     for sampleId in 0 ..< samplesPerTask {
-                        let token = try withRandomState(
+                        let token = withRandomState(
                             MLXRandom.RandomState(seed: UInt64(samplerId * 1000 + sampleId))
                         ) {
-                            return sampler.sample(logits: sharedLogits)
+                            return sampler.sample(logits: logits)
```
davidkoski, line 159: Some of these tests were a little fast and loose with non Sendable data and tested illegal cases.

PR #57, opened 2026-01-11
Support task cancellation in ChatSession

aleroot:
## Proposed changes

I needed to implement a stop button in my in development macOS app but I could not find any stop or cancellation mechanism already built. 
The changes are my proposal for implementation of the cancellation checks in ChatSession.swift and add a corresponding test case in ChatSessionTests.swift.

I would have wanted to use the proper Atomic type instead of having to load os and the OSAllocatedUnfairLock, but since the target is not macOS 15, it is not possible. For the future the proper mechanism should be handled using: 

`private let cancelled = Atomic<Bool>(false)`

#### Note 

_I have removed even the calling of Stream().synchronize() again in ChatSession as it was redundant and blocking the consumer thread unnecessarily._

## Checklist

### DIFF
--- a/Tests/MLXLMTests/ChatSessionTests.swift
+++ b/Tests/MLXLMTests/ChatSessionTests.swift
@@ -111,4 +111,99 @@ public class ChatSessionTests: XCTestCase {
             "Model should recognize the name 'Bob' from the injected history, proving successful prompt re-hydration."
         )
     }
+
+    func testStopGeneration() async throws {
+        let session = ChatSession(Self.llmContainer)
+
+        var collectedOutput = ""
+        let prompt =
+            "Write a detailed 1000 word essay about the history of the Roman Empire. Please include dates and specific names."
+
+        let generationTask = Task {
+            do {
+                for try await chunk in session.streamResponse(to: prompt) {
+                    collectedOutput += chunk
+                    print(chunk, terminator: "")
+                }
+            } catch {
+                // Cancellation is expected
+                if !(error is CancellationError) {
+                    XCTFail("Unexpected error: \(error)")
+                }
+            }
+            return collectedOutput
+        }
+
+        try await Task.sleep(for: .seconds(1))
+
+        session.cancel()
+
+        let result = await generationTask.value
+
+        XCTAssertFalse(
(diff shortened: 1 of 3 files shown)


Tests/MLXLMTests/ChatSessionTests.swift
```diff
@@ -111,4 +111,99 @@ public class ChatSessionTests: XCTestCase {
...
+        var collectedOutput = ""
+        let prompt =
+            "Write a detailed 1000 word essay about the history of the Roman Empire. Please include dates and specific names."
+
+        let generationTask = Task {
+            do {
+                for try await chunk in session.streamResponse(to: prompt) {
+                    collectedOutput += chunk
+                    print(chunk, terminator: "")
+                }
```
davidkoski, line 127: None of this is necessary -- simply return early from this loop and the generation will stop.  However, see #55  -- there were some thread safety issues around the KVCache.

davidkoski: See also ChatModel in this PR:

- https://github.com/ml-explore/mlx-swift-examples/pull/454/files

The cancellation works without #55 but the KVCache is not protected and subsequent calls _can_ crash if they are done too quickly.

aleroot: Try to run these tests breaking just the loop, because in my case with my library version, the model didn’t stop and continue to generate till the end, I will to upgrade to the next release of the library once out, and try again

davidkoski: I just tried these tests with #55 and they seem to work as expected.  Check out the current concurrency related tests in:

- https://github.com/ml-explore/mlx-swift-lm/blob/main/Tests/MLXLMTests/ChatSessionTests.swift

Your tests can be made to work with these random weight models or it could be used with:

- https://github.com/ml-explore/mlx-swift-lm/blob/main/Tests/MLXLMIntegrationTests/ChatSessionIntegrationTests.swift

That is the code that this originally worked with.

I _think_ the interruption is working correctly and the KVCache is guarded against concurrent access (correctly now).

Issue #49, opened 2026-01-08
Model Request:  LFM2.5-VL


sandroandric:
Could we please add LiquidAI's  LFM2.5-VL 

It would be much much appreciated

adrgrondin:
@davidkoski this can be closed, since https://github.com/ml-explore/mlx-swift-lm/pull/58 was merged

PR #58, opened 2026-01-13
Add LFM2 VL model

adrgrondin:
## Proposed changes

- Add `lfm2_vl` architecture for LFM2-VL and LFM2.5-VL models (ported from https://github.com/Blaizzy/mlx-vlm/tree/main/mlx_vlm/models/lfm2_vl)

- Solves https://github.com/ml-explore/mlx-swift-lm/issues/49

## Checklist

### DIFF
--- a/Libraries/MLXLMCommon/InterpolationUtils.swift
+++ b/Libraries/MLXLMCommon/InterpolationUtils.swift
@@ -194,8 +194,8 @@ private func makeNearestInterpolationKernel() -> MLXFast.MLXFastKernel {
 // MARK: - Kernel Manager
 
 /// Manages Metal kernels for interpolation operations.
-public final class InterpolationKernelManager: Sendable {
-    public static let shared = InterpolationKernelManager()
+private final class InterpolationKernelManager: Sendable {
+    static let shared = InterpolationKernelManager()
 
     let bicubicKernel: MLXFast.MLXFastKernel
     let nearestKernel: MLXFast.MLXFastKernel


Libraries/MLXVLM/VLMModelFactory.swift
```diff
@@ -88,6 +88,8 @@ public enum VLMTypeRegistry {
         "llava_qwen2": create(FastVLMConfiguration.self, FastVLM.init),
         "pixtral": create(PixtralConfiguration.self, PixtralVLM.init),
         "mistral3": create(Mistral3VLMConfiguration.self, Mistral3VLM.init),
+        "lfm2_vl": create(LFM2VLConfiguration.self, LFM2VL.init),
+        "lfm2-vl": create(LFM2VLConfiguration.self, LFM2VL.init),
```
adrgrondin, line 92: `lfm2-vl` here is to support https://huggingface.co/mlx-community/LFM2-VL-1.6B-4bit and https://huggingface.co/mlx-community/LFM2-VL-450M-4bit that have `lfm2-vl` arch in their `config.json`. This looks like an error that happened during conversion but not sure, maybe @Blaizzy knows what happened.

Libraries/MLXVLM/Models/LFM2VL.swift
```diff
@@ -0,0 +1,1304 @@
...
+            self._patchEmbedding.wrappedValue = Linear(
+                config.numChannels * patchSize * patchSize,
+                embedDim
+            )
+            self._positionEmbedding.wrappedValue = Embedding(
+                embeddingCount: numPatches, dimensions: embedDim)
+        }
+
+        /// Resize positional embeddings using bilinear interpolation
+        static func resizePositionalEmbeddings(
```
adrgrondin: The Python port use `bicubic_interpolate` to do this using MLX's built-in interpolate function. It works with the current implementation (not sure about the effect on output) but needs a bicubic Swift implementation to match the Python port.

davidkoski: That looks like: https://github.com/Blaizzy/mlx-vlm/blob/fc8c92e31983a52761f37d503f903ec40bebbd62/mlx_vlm/models/kernels.py#L94

It isn't built-in, it is a custom kernel in mlx-vlm.  You can see `BitLinearKernelManager` for an example of how this is done in swift (pretty similar).

Libraries/MLXVLM/VLMModelFactory.swift
Blaizzy: Yes and no.

They finally got their HF integration and changed model type. 

I just kept it because of existing quants.

Blaizzy: They should had just kept the original but that ship has sailed

adrgrondin: Thanks. Will keep this like that then.

Libraries/MLXVLM/Models/LFM2VL.swift
adrgrondin: Thanks for the guidance with `BitLinearKernelManager`. Just pushed https://github.com/ml-explore/mlx-swift-lm/pull/58/changes/6abde21654d9f2981d9770efb0d4ceb9076840ac with `InterpolationUtils.swift` adding bicubic interpolation

Libraries/MLXLMCommon/InterpolationUtils.swift
```diff
@@ -0,0 +1,454 @@
...
+//
+//  Bicubic and Nearest interpolation using Metal kernels.
+//  Port of the Python MLX interpolation kernels.
+//
+
+import Foundation
+import MLX
+import MLXFast
+
+// MARK: - Kernel Sources
```
davidkoski, line 13: This should probably have a comment that points back to where it came from -- good to credit the original author and if we ever need to debug it, see what changes have been made to it.

Libraries/MLXLMCommon/InterpolationUtils.swift
```diff
@@ -0,0 +1,454 @@
...
+        inputNames: ["input", "dims"],
+        outputNames: ["output"],
+        source: source
+    )
+}
+
+// MARK: - Kernel Manager
+
+/// Manages Metal kernels for interpolation operations.
+public final class InterpolationKernelManager: @unchecked Sendable {
```
davidkoski: If you rebase on main I think this can be plain `Sendable` (not `@unchecked`)

Libraries/MLXLMCommon/InterpolationUtils.swift
davidkoski: Added #59 -- I think I haven't been diligent about making sure this is the case so I need to go back and add where missing!

Libraries/MLXLMCommon/InterpolationUtils.swift
davidkoski: Do you want this to be `public`?  It is only used inside the file so maybe it should be private for now.

Libraries/MLXLMCommon/InterpolationUtils.swift
adrgrondin: Updated credits with proper link to repo 👌

Libraries/MLXLMCommon/InterpolationUtils.swift
adrgrondin: Good catch, just rebased and removed the `@unchecked`

adrgrondin: Also converted `LFM2VLProcessor` to struct

adrgrondin: And made `InterpolationKernelManager` private too

PR #64, opened 2026-01-16
Feature/vlm external video frames support [Draft]

vade:
## Proposed changes

Please bear with me as this is my first MLX contribution!

This PR is a DRAFT [stemming from conversation here]( https://github.com/ml-explore/mlx-swift/discussions/335). 
The goal is to support VLM inference on arrays of frames that exist in memory from sources other than on disk video files. 

Today `UserInput` supports only `AVAsset` or a `url`, and the `MediaProcessing` enum  presumes those sources as well. 

This draft PR is aimed to open discussion to extend support for 'dense' in memory video frames (already decoded, at native  frame rate and resolution)

A review of the code implies we need to:
- Extending `UserInput` and `Video` enum to house in memory frames either as `[Image]` or `[VideoFrame]` objects
- Update / extened `MediaProcessing` with  new `asProcessedSequence` to handle sampling and conversion
- Update model processing entrypoints to account for the new logic, and optionality of `AVAsset`

## Checklist

### DIFF
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -1,6 +1,6 @@
 name: Build and Test
 
-on: pull_request  
+on: pull_request
 
 permissions:
   contents: read
@@ -14,13 +14,13 @@ jobs:
     steps:
       - uses: actions/checkout@v6
         with:
-          submodules: recursive           
+          submodules: recursive
 
       - name: Setup uv
         uses: astral-sh/setup-uv@v6
         with:
-            activate-environment: true
-     
+          activate-environment: true
+
       - name: Setup pre-commit
         shell: sh
         run: |
@@ -58,12 +58,12 @@ jobs:
         run: |
           cd /tmp/swift-format
           ln -s "$(swift build --show-bin-path -c release)/swift-format" /usr/local/bin/swift-format
-     
+
       - name: Configure safe directory for git
         shell: sh
         run: |
           git config --global --add safe.directory "$GITHUB_WORKSPACE"
-     
+
       - name: Run style checks
         shell: sh
         run: |
@@ -77,13 +77,16 @@ jobs:
       - uses: actions/checkout@v6
         with:
           submodules: recursive
-     
+
       - name: Verify MetalToolchain installed
         shell: bash
         run: xcodebuild -showComponent MetalToolchain
-            
+
       - name: Build (Xcode, macOS)
         shell: sh
+        env:
+          # workaround for CI failure
+          DEVELOPER_DIR: /Applications/Xcode-latest.app
         run: |
           xcodebuild -version
           xcrun --show-sdk-build-version
@@ -93,6 +96,9 @@ jobs:
 
       - name: Run Tests (Xcode, macOS)
         shell: sh
+        env:
+          # workaround for CI failure
+          DEVELOPER_DIR: /Applications/Xcode-latest.app
         run: |
           xcrun xctest ~/Library/Developer/Xcode/DerivedData/mlx-swift-lm-*/Build/Products/Debug/MLXLMTests.xctest
 
--- a/ACKNOWLEDGMENTS.md
+++ b/ACKNOWLEDGMENTS.md
@@ -7,6 +7,7 @@
 MLX Swift was developed with contributions from the following individuals:
 
 - [John Mai](https://github.com/johnmai-dev): Added support for multiple models (Qwen2, Starcoder2, InternLM2, Qwen3, Qwen3 MoE, GLM-4, MiMo, BitNet, SmolLM3, LFM2, Baichuan-M1).
+- [Christoph Rohde](https://github.com/CodebyCR): Improved Embedders documentation.
 
 <a href="https://github.com/ml-explore/mlx-swift-lm/graphs/contributors">
   <img class="dark-light" src="https://contrib.rocks/image?repo=ml-explore/mlx-swift-lm&anon=0&columns=20&max=100&r=true" />
--- a/Libraries/Embedders/BaseConfiguration.swift
+++ b/Libraries/Embedders/BaseConfiguration.swift
@@ -3,25 +3,47 @@
 import Foundation
 import MLX
 
-/// Base ``LanguageModel`` configuration -- provides `modelType`
-/// and `quantization` (used in loading the model).
+/// The fundamental configuration for any MLX-based model.
 ///
-/// This is used by ``ModelFactory/load(hub:configuration:progressHandler:)``
-/// to determine the type of model to load.
+/// `BaseConfiguration` provides the metadata necessary to identify the model architecture
+/// (`modelType`) and describes the quantization parameters used to compress the model's weights.
+/// It is designed to be decoded directly from a model repository's `config.json`.
 public struct BaseConfiguration: Codable, Sendable {
+
+    /// The architecture identifier (e.g., "bert", "roberta", "xlm-roberta").
     public let modelType: String
 
+    /// Configuration parameters for weight quantization.
+    ///
+    /// MLX uses group-wise quantization to reduce memory footprint. This struct
+    /// defines how weights are grouped and the precision (bits) used for each group.
     public struct Quantization: Codable, Sendable, Equatable {
+
+        /// Initializes a new quantization configuration.
+        /// - Parameters:
+        ///   - groupSize: The number of weights that share the same scale and bias.
+        ///   - bits: The bit-depth of the quantized weights (e.g., 4 or 8).
         public init(groupSize: Int, bits: Int) {
             self.groupSize = groupSize
             self.bits = bits
         }
 
+        /// The size of the quantization group.
         public let groupSize: Int
+
+        /// The number of bits per weight.
         public let bits: Int
+
+        /// Internal storage for the quantization mode.
         private var _mode: QuantizationMode? = nil
+
+        /// The quantization method to use (defaults to `.affine`).
+        ///
+        /// Affine quantization (asymmetric) uses both a scale and a zero-point
+        /// to map floating point values to integers.
         public var mode: QuantizationMode { _mode ?? .affine }
 
+        /// Converts the configuration into a tuple format compatible with `MLX.quantize`.
         public var asTuple: (Int, Int, QuantizationMode) { (groupSize, bits, mode) }
 
         enum CodingKeys: String, CodingKey {
@@ -31,15 +53,24 @@ public struct BaseConfiguration: Codable, Sendable {
         }
     }
 
-    /// handling instructions for ``PerLayerQuantization``
+    /// Instructions for handling individual layers during the quantization process.
     public enum QuantizationOption: Sendable {
+        /// Do not quantize this specific layer (keep it in high precision).
         case skip
+        /// Quantize this layer using the provided parameters.
         case quantize(Quantization)
     }
 
-    /// Per-layer ``Quantization`` values with optional default.
+    /// A container for per-layer quantization settings.
+    ///
+    /// This allows for "Mixed-Precision" or "Heterogeneous" quantization, where
+    /// sensitive layers (like the embedding head) can be kept at higher precision
+    /// while the rest of the model is compressed.
     public struct PerLayerQuantization: Sendable {
+        /// The default quantization for any layer not explicitly named in `perLayerQuantization`.
         public var quantization: Quantization? = nil
+
+        /// A dictionary mapping layer paths (e.g., "model.embed_tokens") to their quantization options.
         public var perLayerQuantization: [String: QuantizationOption]
 
         public init(
@@ -50,7 +81,9 @@ public struct BaseConfiguration: Codable, Sendable {
             self.perLayerQuantization = perLayerQuantization
         }
 
-        /// The quantization to apply for the given layer name or nil for no quantization.
+        /// Resolves the quantization parameters for a specific layer.
+        /// - Parameter layer: The path/name of the layer.
+        /// - Returns: The `Quantization` settings to apply, or `nil` if the layer should be skipped.
         public func quantization(layer: String) -> Quantization? {
             if let perLayer = perLayerQuantization[layer] {
                 switch perLayer {
@@ -65,26 +98,16 @@ public struct BaseConfiguration: Codable, Sendable {
         }
     }
 
-    /// Special codable to support a mixed key: Int / key: Quantization
-    /// structure for hereogenous quantization, e.g.
+    /// An internal container designed to handle the mixed JSON structure found in `config.json`.
     ///
-    /// ```
-    /// "quantization": {
-    ///     "group_size": 64,
-    ///     "bits": 4,
-    ///     "model.embed_tokens": {
-    ///         "group_size": 32,
-    ///         "bits": 4
-    ///     },
-    ///     "model.layers.0.self_attn.q_norm": false,
-    /// ```
-    ///
-    /// This mixed type structure requires manual decoding.
+    /// Quantization configs in MLX often interleave global keys (like `bits`) with
+    /// specific layer keys (like `model.layers.0...`). This container uses manual
+    /// decoding to separate these interleaved values.
     struct QuantizationContainer: Codable, Sendable {
         var quantization: Quantization
         var perLayerQuantization: PerLayerQuantization
 
-        // based on Dictionary's coding key
+        /// A custom CodingKey used to iterate over arbitrary layer names in JSON.
         internal struct _DictionaryCodingKey: CodingKey {
             internal let stringValue: String
             internal let intValue: Int?
@@ -101,23 +124,27 @@ public struct BaseConfiguration: Codable, Sendable {
         }
 
         init(from decoder: any Decoder) throws {
-            // handle the embedded Quantization
+            // 1. Decode global quantization (bits/group_size) from the current level
             self.quantization = try Quantization(from: decoder)
 
-            // and the interleaved per-layer values
+            // 2. Decode interleaved per-layer values
             var perLayerQuantization = [String: QuantizationOption]()
             let container = try decoder.container(keyedBy: _DictionaryCodingKey.self)
+
             for key in container.allKeys {
                 switch key.stringValue {
                 case Quantization.CodingKeys.groupSize.rawValue: continue
                 case Quantization.CodingKeys.bits.rawValue: continue
+                case Quantization.CodingKeys._mode.rawValue: continue
 
                 default:
+                    // If the value is a boolean 'false', we treat it as .skip
                     if let f = try? container.decode(Bool.self, forKey: key) {
                         if !f {
                             perLayerQuantization[key.stringValue] = .skip
                         }
                     } else {
+                        // Otherwise, try to decode a specific Quantization object for this layer
                         perLayerQuantization[key.stringValue] = .quantize(
                             try container.decode(Quantization.self, forKey: key))
                     }
@@ -142,13 +169,16 @@ public struct BaseConfiguration: Codable, Sendable {
         }
     }
 
+    /// Internal storage for quantization details extracted from `config.json`.
     var quantizationContainer: QuantizationContainer?
 
+    /// The default quantization settings.
     @available(*, deprecated, message: "Please use perLayerQuantization instead")
     public var quantization: Quantization? {
         quantizationContainer?.quantization
     }
 
+    /// The per-layer quantization settings, including the default fallback.
     public var perLayerQuantization: PerLayerQuantization? {
         quantizationContainer?.perLayerQuantization
     }
--- a/Libraries/Embedders/Bert.swift
+++ b/Libraries/Embedders/Bert.swift
@@ -69,7 +69,6 @@ private class TransformerBlock: Module {
     func callAsFunction(_ inputs: MLXArray, mask: MLXArray? = nil) -> MLXArray {
         let attentionOut = attention(inputs, keys: inputs, values: inputs, mask: mask)
         let preNorm = preLayerNorm(inputs + attentionOut)
-
         let mlpOut = down(gelu(up(preNorm)))
         return postLayerNorm(mlpOut + preNorm)
     }
@@ -133,11 +132,11 @@ public class BertModel: Module, EmbeddingModel {
     }
 
     public func callAsFunction(
-        _ inputs: MLXArray, positionIds: MLXArray? = nil, tokenTypeIds: MLXArray? = nil,
+        _ inputs: MLXArray,
+        positionIds: MLXArray? = nil,
+        tokenTypeIds: MLXArray? = nil,
         attentionMask: MLXArray? = nil
-    )
-        -> EmbeddingModelOutput
-    {
+    ) -> EmbeddingModelOutput {
         var inp = inputs
         if inp.ndim == 1 {
             inp = inp.reshaped(1, -1)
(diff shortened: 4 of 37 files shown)


vade:
So I want to discuss a few points

Firstly - All model implementations presume the existence of an AVAsset as produced by `UserInput.Video.asAVAsset()` 

I assume its taken for granted that the model processing across all VLM models at `LMInput.ProcessedVideo` need to be updated to handle the case where the asset is nil and we have incoming video frames? 

Secondly, it appears the right thing to do would be to promote `VideoFrame` from MediaProcessing to UserInput, as we likely need / want a way to provide timing input from user space into the pipeline? 

Thanks for any guidance on this process!

Libraries/MLXLMCommon/UserInput.swift
```diff
@@ -39,21 +39,37 @@ public struct UserInput {
             }
         }
     }
-
+    
+    public struct VideoFrame {
+        public let frame: CIImage
```
davidkoski, line 44: I wonder if this should be `Image`?  I think we probably want to consume CIImage for most image processing needs, but if somebody has an MLXArray (maybe they used a model to generate them?) then the would allow us to pass an array of MLXArray and have an easy way to convert to CIImage when needed.

It doesn't _have_ to be that way, just a thought.

vade: I thought about that, but it occurred to me that external clients would need to provide sample times so that the sparse temporal sampling could be done accurately and if a VLM is sensitive to timing data or supports multiple frame rates having time provided would be important.

Libraries/MLXLMCommon/UserInput.swift
```diff
@@ -39,21 +39,37 @@ public struct UserInput {
...
+    
     /// Representation of a video resource.
     public enum Video {
         case avAsset(AVAsset)
         case url(URL)
+        /// Useful for decoded frames held in memory
+        case frames([VideoFrame])
 
-        public func asAVAsset() -> AVAsset {
+        public func asAVAsset() -> AVAsset? {
```
davidkoski: I agree this one is a problem.  We have a couple of approaches:

- make a breaking API change (probably OK here, but not ideal)
- keep it as non-optional but fatalError if we have `.frames`, probably with a deprecation of this property
- build an AVAsset programmatically using AVMutableComposition and friends (a lot of work and maybe nobody cares)

I think the middle approach might be a good middle ground, but the first might be good enough.

vade: some additional thoughts on the above proposal : 

theres other gotchas about AVAsset that might be worth making it nil-able internally:

* assets actually let you initialize with a bunk URL, and simply dont return an obvious error, ie last i checked, an AVURLAsset(URL(string:"https://foo.bar/baz.mov"))) is a non nil AVAsset that you can't do anything with

* assets whose property for `decodable` or `playable` is false, which vends a valid asset, but not one that you can do anything with

* assets with no video track

so I think theres some cases where even a valid asset isnt necessarily valid in the context of a VLM, let alone some assets which wont decode, and semantically returning nil there (or a new function signature?) might make sense?

I'd ***highly suggest*** not adding a dynamically allocated AVMutableComposition, the overhead alone to latency would be a bad idea, we've already got frames "decoded" :)

Libraries/MLXLMCommon/UserInput.swift
davidkoski: Yes, timing is important, I meant more like this:

```swift
struct VideoFrame {
    let frame: Image
    let let timeStamp: CMTime
```

for sure we need the timeStamp

Libraries/MLXLMCommon/UserInput.swift
davidkoski: I wonder if this should be `throws` also/instead?  I agree the URL one suggests it should at least be `throws`.

If the known callers are using the frame iteration or can be converted into doing so I think deprecation might be the right way to go.  Then we can worry less about the issues in here as we don't intend to support it (much) going forward.

Libraries/MLXLMCommon/UserInput.swift
vade: Oh, i misunderstood, i thought you meant promoting VideoFrame vs Image :) Apologies. 

Im open to either! 

Actually having it as `Image` could be interesting, as for VFX tooling often you have sequences of images on disk, as an array of URLs. Thats interesting.

Libraries/MLXVLM/MediaProcessing.swift
```diff
@@ -5,11 +5,6 @@ import CoreImage.CIFilterBuiltins
 import MLX
 import MLXLMCommon
 
-public struct VideoFrame {
```
davidkoski, line 8: Maybe this should have a deprecated `typealias` to `UserInput.VideoFrame` -- one less API break.

vade: Thats a good call!

Libraries/MLXVLM/Models/Qwen25VL.swift
```diff
@@ -794,6 +780,65 @@ public struct Qwen25VLProcessor: UserInputProcessor {
...
+    {
+        switch video
+        {
+        case .frames(let frames):
+            return try await self.processFrames(frames, processing: processing)
+            
+        default:
+            return try await self.processAsset(video.asAVAsset()!, processing: processing)
+        }
+    }
```
davidkoski: See my comment on MediaProcessing -- this function would go away because the function on MediaProcessing would take `Video`

Libraries/MLXVLM/Models/Qwen25VL.swift
```diff
@@ -794,6 +780,65 @@ public struct Qwen25VLProcessor: UserInputProcessor {
...
+            let resizedImage = MediaProcessing.apply(
+                frame.frame, processing: processing)
+            if resizedSize == .zero {
+                let size = resizedImage.extent.size
+                let (resizedHeight, resizedWidth) = try QwenVL.targetSize(
+                    height: Int(size.height), width: Int(size.width),
+                    factor: config.patchSize * config.mergeSize,
+                    minPixels: config.minPixels, maxPixels: config.maxPixels)
+                resizedSize = CGSize(width: resizedWidth, height: resizedHeight)
+            }
```
davidkoski: And if it took a `Video` I think these two methods would also fold down into one as they do the same frame processing.

Libraries/MLXVLM/MediaProcessing.swift
```diff
@@ -389,6 +386,51 @@ public enum MediaProcessing {
             totalDuration: duration
         )
     }
+    
+    static public func asProcessedSequence(_ videoFrames:[VideoFrame],
+                                           targetFPS: (CMTime) -> Double,
+                                           frameProcessing: (VideoFrame) throws -> VideoFrame = { $0 }
+    ) async throws -> ProcessedFrames {
```
davidkoski: It doesn't have to be done in this PR, but it would be nice if we could rewrite the previous implementations as composition of this one.  Additionally it would be nice if the caller could pass a `Video` rather than one of the cases.

That could be done by making this function have this signature:

```swift
    static public func asProcessedSequence(_ video: Video,
                                           targetFPS: (CMTime) -> Double,
                                           frameProcessing: (VideoFrame) throws -> VideoFrame = { $0 }
    ) async throws -> ProcessedFrames {
```

Internally it would switch on it being `frames` or `avasset` and call the appropriate internal function.

I think that would be sufficient for this PR -- it would give the API we want: something that consumes `Video` and doesn't require the caller to switch on the various cases that they probably don't care about.

This might be for another PR or for never, but let me throw it out there (I thought I had filed an issue for this, but I can't find it).  What if we did this with some composition -- what if we had a generator of frames and a filter of frames.

Right now we want to filter frames:

- trim start/end
- target frame rate
- maximum number of frames
- whatever

One way we might do that is a protocol:

```swift
protocol VideoFrameFilter: Sendable {
    func filterFrames(_ timestamps: [CMTime]) -> [CMTime]
}
```

Then we could have a couple of these that we compose:

```swift
struct TrimVideoFrameFilter: VideoFrameFilter {
    let start: CMTime?
    let end: CMTime?

    func filterFrames(_ timestamps: [CMTime]) -> [CMTime] {
        var result = timestamps
        if let start {
            result = result.filter { $0 >= start }
        }
        if let end {
            result = result.filter { $0 <= end }
        }
        return result
    }
}
```

You could imagine something that did maxFrame, targetFrameRate, etc. and maybe something that would combine them in sequence.

Then the generator could look like something like this (I am saying it is Sendable but I don't know if it can be):

```swift
protocol VideoFrameGenerator: Sendable {
    var frameTimes: [CMTime] { get }

    // I don't know which of these is best, but something in this shape
    func generateFrames(frameTimes: [CMTime], frameProcessing: (UserInput.VideoFrame) throws -> UserInput.VideoFrame) -> [UserInput.VideoFrame]

    func generateFrames(frameTimes: [CMTime]) -> AsyncSequence<VideoFrame>
    func generateFrames(frameTimes: [CMTime]) -> Sequence<VideoFrame>
}
```

I think the latter two might be better:  more flexible and we can accumulate the results however we want.

Anyway, the AVAsset implementation sketched:

```swift
struct AVAssetVideoFrameGenerator: VideoFrameGenerator {
    let avAsset: AVAsset

    var frameTimes: [CMTime] {
        // something with AVAssetReaderTrackOutput to read the frame times OR just compute when we think they will be based on the start/end time and nominal frame rate
    }

    func generateFrames(frameTimes: [CMTime]) -> AsyncStream<VideoFrame> {
            let (stream, continuation) = AsyncStream<Generation>.makeStream()
            
            let task = Task {
                let generator = AVAssetImageGenerator(asset: asset)
                ...
                for  frame in generator.images(for: frameTimes) {
                    continuation.yield(frame)
            }
            
            // some stream cancellation code

            return stream
    }
}
```

Maybe the Generator is overkill and we only need the `VideoFrameFilter`.  Anyway, that would let us use the same code to iterate the temporally filtered frames and apply the `frameProcessing`.

I think the important part of all this blah blah is that we want API that takes a `Video` and encapsulates the production of video frames.  Current callers should migrate to this and we can deprecate the functions that take an `AVAsset`.

vade: Ah, thats a nicer implementation that also simplifies the call sites, ill def do the first suggestion for this PR :)

As for the second suggestion, I was thinking of something similar, but it seemed a much larger change! 

At first pass, i think something like this is doable, and it also allows for 'just in time' decode which likely will save on memory usage / memory pressure in some cases. 

Might i suggest saving that for a second PR? I think its a great idea.

davidkoski: I think doing the time filtering in a second PR is fine -- we haven't had it so far so we are no worse off :-). It will let us simplify the arguments here and we can just have these functions cover the new implementation when it is ready.

My idea doesn't have to be the way we do it, but something that would encapsulate the time filtering would be great.  And I love composition when it fits :-)

Libraries/MLXLMCommon/UserInput.swift
davidkoski: OK, how about this:  we don't really want/need this method any more if the new processing method takes a `Video` directly.

First we mark this as deprecated:

```swift
@available(*, deprecated, message: "Use MediaProcessing.asProcessedSequence() with the Video directly")
```

Then this can either return `AVAsset?` (breaking API but we want people to move off anyway) OR it can stay `AVAsset` and `fatalError` inside if it just has frames.  I think the API break might be better in this case as it pushes people gently (but not too gently) to the new API.

vade: makes sense! ill take a gander at these changes. Thanks for the feedback!

Libraries/MLXVLM/MediaProcessing.swift
vade: Thanks, so i implemented a singular public path for `asProcessedSequence` and cleaned up all the VLM implementations to not violate DRY. 

I dont see any unit tests for VLM functionality, and ive yet to write any for the new [VideoFrame] path, so id be VERY leery just yet i havent broken anything or done anything too dumb™

Ill do some manual validation, but do you have any thoughts on what a unit test for the VLMs should look like? Also how should we test without requiring downloading all the models on the test system? 

thanks for the input so far.

vade:
I resolved some threads, so my comments on testing might be lost but

* marked asAVAsset() as deprecated
* I also did make the existing asProcessedSequence functions private in lieu of a singular entry point against asProcessedSequence with a UserInput.Video - not sure if that will be frowned upon or breaking?

Note ive not done a ton of testing, - and while Unit tests pass, im not sure i saw anything specific for VLM functionality or video decode - not sure how testing in MLX is handled, and ive yet to do serious manual vetting of the above code, so lets test this a bit. Thanks again for the guidance!

vade:
Thanks, i fixed the breaking API changed marked the existing api deprecated in MediaProcessing and, mirrored the functions internally to not have warnings on internal code. 

As for the testing, thanks for guidance there. 

I dont personally have resources to run the all models test, but i will take the opportunity to mock some tests out for MediaProcessing paths as is w/o model integration

If i include some small quicktime movies into the testing apparatus it would be helpful to validate some of the other paths, ie 
* num frames in sequence with various frame rates vs sample per second rates
* asset validation cases
* raw frame vs asset based decode paths.

davidkoski:
> I dont personally have resources to run the all models test, but i will take the opportunity to mock some tests out for MediaProcessing paths as is w/o model integration

No worries -- it takes a while to run them even once all the weights are downloaded.  I would just pick one of the VLMs and make sure it works before/after the change.

Libraries/MLXVLM/MediaProcessing.swift
```diff
@@ -327,16 +324,82 @@ public enum MediaProcessing {
...
+            video,
+            targetFPS: { _ in Double(samplesPerSecond) },
+            frameProcessing: frameProcessing,
+            maxFrames: Int.max)
+    }
+
+    static public func asProcessedSequence(
+        _ video: UserInput.Video,
+        targetFPS: (CMTime) -> Double,
+        frameProcessing: (VideoFrame) throws -> VideoFrame = { $0 },
```
davidkoski: I would put `frameProcessing` as the last argument for the "trailing closure" syntax.  It also groups maxFrames with targetFPS (both filtering on the frames).

Libraries/MLXVLM/MediaProcessing.swift
```diff
@@ -327,16 +324,82 @@ public enum MediaProcessing {
...
+
+        case .url(let url):
+            let asset = AVAsset(url: url)
+            try await Self.validateAsset(asset)
+            return try await _asProcessedSequence(
+                asset, maxFrames: maxFrames, targetFPS: targetFPS, frameProcessing: frameProcessing)
+
+        case .frames(let videoFrames):
+            return try await _asProcessedSequence(
+                videoFrames, targetFPS: targetFPS, frameProcessing: frameProcessing)
```
davidkoski, line 376: This ignores the maxFrames parameter -- this could be handled in a followup where the frame filtering is factored out (so that the same logic applies to both).

The code this calls looks very similar to the AVAsset case so maybe it is just missing?

Libraries/MLXVLM/Models/Qwen25VL.swift
```diff
@@ -1,5 +1,6 @@
 // Port of https://github.com/Blaizzy/mlx-vlm/tree/main/mlx_vlm/models/qwen2_5_vl
 
+import AVFoundation
```
davidkoski: I think this is not needed.

Libraries/MLXVLM/Models/Qwen2VL.swift
```diff
@@ -2,6 +2,7 @@
 
 // port of https://github.com/Blaizzy/mlx-vlm/tree/main/mlx_vlm/models/qwen2_vl
 
+import AVFoundation
```
davidkoski: Same -- perhaps this was used with an earlier version of the API?

Tests/MLXLMTests/MediaProcessingTests.swift
```diff
@@ -0,0 +1,128 @@
...
+        let input = inputFilter.outputImage!.cropped(
+            to: CGRect(x: 0, y: 0, width: 1536, height: 1106))
+
+        let target = CGSize(width: 1540, height: 1120)
+        let output = MediaProcessing.resampleBicubic(input, to: target)
+
+        XCTAssertEqual(output.extent.size, target)
+    }
+
+    func testVideoFileAsSimpleProcessedSequence() async throws {
```
davidkoski, line 26: These look great!

Libraries/MLXVLM/Models/Qwen2VL.swift
vade: Ah, great catch, yes those were lingering from previous approaches! Good eye.

Libraries/MLXVLM/MediaProcessing.swift
vade: Appreciate the ergonomics there!

Libraries/MLXVLM/MediaProcessing.swift
vade: Ah, my presumption was the caller was passing a specific set of frames to be processed, and all should be attended to. If you want a diff max, set a diff number of frames for input?

vade:
Hey there. Thanks for the feedback. Ive resolved the imports and the trailing closure issue. I left a note about my thoughts on not needing to attend to max frames (the number of input raw frames should be the max and we dont need two sources of truth there)

I feel pretty good about where this has landed!

Libraries/MLXVLM/MediaProcessing.swift
davidkoski: That is reasonable, but I think the maxFrames is actually used as a dynamic FPS (look at the linspace call).  The same argument could be made for the `targetFPS` -- if you wanted a different FPS the caller could arrange that.

I think (opinion) we should honor whatever parameters we pass.  The caller can arrange the frames and leave these as default values or maybe they loaded the frames from files and do want some processing.

Libraries/MLXVLM/MediaProcessing.swift
```diff
@@ -389,6 +452,54 @@ public enum MediaProcessing {
...
+        let finalFrameCount = max(desiredFrames, 1)
+
+        let sampledTimeValues = MLXArray.linspace(
+            0, duration.value, count: Int(finalFrameCount)
+        ).asArray(Int64.self)
+
+        // Construct a CMTime using the sampled CMTimeValue's and the asset's timescale
+        let timescale = duration.timescale
+        let sampledTimes = sampledTimeValues.map { CMTime(value: $0, timescale: timescale) }
+        let videoFramesToSample = videoFrames.filter { sampledTimes.contains($0.timeStamp) }
```
davidkoski: This is going for exact match -- I wonder if we need to use "nearest before"?  If we had real video frames (say from a camera or we read them from a video file) the timing isn't always perfect.

vade: Hrm, interesting. Thats a valid concern. 

I was following the spirit of the `AVAssetGenerator` code path which uses a `requestedTimeToleranceBefore` / after of `.zero` - i suppose the same concern might apply there as well? 

(and come to think of it, i have come across files in the wild where `.zero` tolerance on `AVAssetGenerator` has caused some issues on occasion)

vade: Although I guess, in retrospect, `AVAssetGenerator` is likely smart enough to understand that a CMTime that arrives within the range of a frame's on screen presentation  (ie CMTimeRange(start: thisFramesPresentationTimeStamp, end:nextFramesPresentationTimeStamp) or some such)  - whereas my code does in fact rely on hitting the presentationTimeStamp on the nose.

davidkoski: At least CMTimes are not floats and _can_ have exact match semantics :-)

I think the requestedTimeToleranceBefore tells it how much slop is allowed, e.g. if you allow some it might go a little wide to pick up a P frame which is cheaper to decode.  I think a tolerance of zero might still give you all the frames but it won't allow any cheaper paths.

davidkoski: I think something like this is what we want:

```swift
        var frameIndex = videoFrames.startIndex
        for value in sampledTimeValues {
            let targetTime = CMTime(value: value, timescale: timescale)
            
            // find the last frame <= the targetTime
            var targetIndex: Int?
            while frameIndex < videoFrames.endIndex {
                if videoFrames[frameIndex].timeStamp > targetTime {
                    break
                } else {
                    targetIndex = frameIndex
                    frameIndex += 1
                }
            }
            
            if let targetIndex {
                let videoFrame = videoFrames[targetIndex]
                let frame = try frameProcessing(
                    .init(frame: videoFrame.frame, timeStamp: videoFrame.timeStamp))
                ciImages.append(frame.frame)
                timestamps.append(frame.timeStamp)
            }
        }
```

I was testing like this:

```swift
        let videoFrames: [VideoFrame] =
        (0..<50)
            .map {
                let value = $0 * 2 + (Bool.random() ? 1 : 0)
                return .init(timeStamp: .init(value: CMTimeValue(value), timescale: 10))
            }
        
        let duration = CMTime(value: 95, timescale: 10)
        let fps = 2.0
```

so here were my source frames:

```
[0.0, 0.3, 0.5, 0.6, 0.9, 1.1, 1.3, 1.5, 1.7, 1.9, 2.0, 2.3, 2.4, 2.7, 2.8, 3.1, 3.3, 3.4, 3.7, 3.9, 4.0, 4.2, 4.4, 4.6, 4.9, 5.1, 5.2, 5.5, 5.7, 5.9, 6.0, 6.2, 6.5, 6.6, 6.9, 7.0, 7.3, 7.4, 7.6, 7.9, 8.0, 8.3, 8.5, 8.7, 8.9, 9.1, 9.2, 9.4, 9.6, 9.8]
```

and the targets:

```
[0.0, 0.5, 1.0, 1.5, 2.1, 2.6, 3.1, 3.6, 4.2, 4.7, 5.2, 5.8, 6.3, 6.8, 7.3, 7.9, 8.4, 8.9, 9.5]
```

And here is the target frame -> actual frame:

```
0.0 -> 0.0
0.5 -> 0.5
1.0 -> 0.9
1.5 -> 1.5
2.1 -> 2.0
2.6 -> 2.4
3.1 -> 3.1
3.6 -> 3.4
4.2 -> 4.2
4.7 -> 4.6
5.2 -> 5.2
5.8 -> 5.7
6.3 -> 6.2
6.8 -> 6.6
7.3 -> 7.3
7.9 -> 7.9
8.4 -> 8.3
8.9 -> 8.9
9.5 -> 9.4
```

vade: Ah, i see. Thanks, thats a great illustration of the issue. Ill add this to the PR asap.

davidkoski:
The CI build failure requires a workaround, current on main -- if you rebase this I can trigger it again.  Thanks!

Libraries/MLXVLM/MediaProcessing.swift
vade: checking in - do you want this resolved to move the PR forward?

davidkoski: No, I don't have a strong preference here -- I think it can be addressed if we refactor how the trim/scale/frame parameters are handled.

PR #69, opened 2026-01-20
Use EOS tokens from config files

ronaldmannak:
## Proposed changes

As discussed in #68 , this PR brings EOS token logic on par with MLX-ML and will automatically use EOS tokens from `config.json`, and `generation_config.json` while keeping the existing path that uses `tokenizer_config.json` intact

## Checklist

### DIFF
--- a/Libraries/MLXLMCommon/ModelConfiguration.swift
+++ b/Libraries/MLXLMCommon/ModelConfiguration.swift
@@ -44,7 +44,6 @@ public struct ModelConfiguration: Sendable {
         tokenizerId: String? = nil, overrideTokenizer: String? = nil,
         defaultPrompt: String = "hello",
         extraEOSTokens: Set<String> = [],
-        eosTokenIds: Set<Int> = [],
         preparePrompt: (@Sendable (String) -> String)? = nil
     ) {
         self.id = .id(id, revision: revision)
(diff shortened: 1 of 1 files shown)


Libraries/MLXLMCommon/ModelConfiguration.swift
```diff
@@ -33,34 +33,41 @@ public struct ModelConfiguration: Sendable {
...
 
+    /// EOS token IDs loaded from config.json/generation_config.json
+    public var eosTokenIds: Set<Int> = []
+
     public init(
         id: String, revision: String = "main",
         tokenizerId: String? = nil, overrideTokenizer: String? = nil,
         defaultPrompt: String = "hello",
         extraEOSTokens: Set<String> = [],
+        eosTokenIds: Set<Int> = [],
```
davidkoski: The `eosTokenIds` works a bit different than `extraEOSTokens` or `defaultPrompt` -- no matter what value is passed here, it will be replaced during the load.  Perhaps this should be removed from the `init`.  I was thinking the setter could be `internal(set)` but it is called between libraries.  Maybe just document that this isn't an override that is passed in, it is set during load.

davidkoski (changes requested):
Looks good, thank you for tracking that down!

Please look at my comment on `eosTokenIds` -- I think it is slightly misleading that a user could init a ModelConfiguration thinking they are providing an override there but it will be reset during load.

Libraries/MLXLMCommon/ModelConfiguration.swift
ronaldmannak: You're right. The property already has a default value, so we can remove it completely from the init

PR #72, opened 2026-01-21
Added wired memory control

robertmsale:
## Proposed changes

Adds an enum to generate functions that allow wired memory tuning, respects system default setting, and utilizes the preexisting task for actor isolation.

Relevant upstream:
- mlx/mlx/backend/metal/allocator.cpp- set_wired_limit(...) and residency handling 
- mlx/mlx/backend/metal/resident.cpp - ResidencySet implementation

It looks like upstream already noops any wired memory adjustments on unsupported platforms so this code compiles on iOS and on unsupported platforms (pre iOS 18 and macOS 15) without any feature gating. This enables granular control as well as maximum working set.

Since Qwen3-Next-80b is right at the limit of what my system can handle with a full KV, I ran some tests and found that wiring the memory of the generate function results in the same performance as wiring the weights + generate, so wrapping the model is unnecessary. Generate function is the perfect place for this

Bench:
```swift
import Foundation
import MLX
import MLXLLM
import MLXLMCommon

@main
struct mlxperiments {
    enum BenchError: Error {
        case missingInfo
    }

    static func main() async {
        do {
            let prompt = "Explain why wired memory can speed up large-model inference."
            let config = ModelConfiguration(
                directory: URL(
                    fileURLWithPath:
                        "/Users/robertsale/.lmstudio/models/lmstudio-community/Qwen3-Next-80B-A3B-Instruct-MLX-4bit"
                )
            )

            let container = try await LLMModelFactory.shared.loadContainer(configuration: config)
            let parameters = GenerateParameters(
                maxTokens: 32,
                maxKVSize: 262_144,
                kvBits: 8,
                kvGroupSize: 64,
                quantizedKVStart: 0,
                temperature: 0.6,
                topP: 0.9,
                repetitionPenalty: nil,
                repetitionContextSize: 20,
                prefillStepSize: 2048
            )
            let wiredModes: [(String, WiredMemoryLimit)] = [
                ("default", .default),
                ("fixed", .fixed(bytes: 51_514_999_440)), // weights + 256k KV Int8 + Mamba + Attention @ 2048 chunks
                ("max", .max),
            ]

            for (label, wiredMemory) in wiredModes {
                GPU.resetPeakMemory()
                let startSnapshot = Memory.snapshot()
                var totalPromptTokens = 0
                var totalGenerationTokens = 0
                var totalPromptTime: TimeInterval = 0
                var totalGenerationTime: TimeInterval = 0

                for iteration in 1...3 {
                    let info: GenerateCompletionInfo = try await container.perform { @Sendable context in
                        let input = UserInput(prompt: prompt)
                        let prepared = try await context.processor.prepare(input: input)
                        let stream = try MLXLMCommon.generate(
                            input: prepared,
                            parameters: parameters,
                            context: context,
                            wiredMemory: wiredMemory
                        )
                        var finalInfo: GenerateCompletionInfo?
                        for await generation in stream {
                            if case .info(let info) = generation {
                                finalInfo = info
                            }
                        }
                        guard let finalInfo else {
                            throw BenchError.missingInfo
                        }
                        return finalInfo
                    }

                    totalPromptTokens += info.promptTokenCount
                    totalGenerationTokens += info.generationTokenCount
                    totalPromptTime += info.promptTime
                    totalGenerationTime += info.generateTime

                    print("[\(label)][\(iteration)] \(info.summary())")
                }

                let endSnapshot = Memory.snapshot()
                let avgPromptTokens = Double(totalPromptTokens) / 3.0
                let avgGenerationTokens = Double(totalGenerationTokens) / 3.0
                let avgPromptTime = totalPromptTime / 3.0
                let avgGenerationTime = totalGenerationTime / 3.0
                let avgPromptTokensPerSecond = Double(totalPromptTokens) / totalPromptTime
                let avgGenerationTokensPerSecond = Double(totalGenerationTokens) / totalGenerationTime

                print(
                    "[\(label)][avg] Prompt:     \(Int(avgPromptTokens)) tokens, \(avgPromptTokensPerSecond.formatted()) tokens/s, \(avgPromptTime.formatted())s"
                )
                print(
                    "[\(label)][avg] Generation: \(Int(avgGenerationTokens)) tokens, \(avgGenerationTokensPerSecond.formatted()) tokens/s, \(avgGenerationTime.formatted())s"
                )
                print("[\(label)] Memory start: \(startSnapshot)")
                print("[\(label)] Memory end:   \(endSnapshot)")
                print("[\(label)] Memory delta: \(startSnapshot.delta(endSnapshot))")
                Memory.clearCache()
            }
        } catch {
            fputs("mlxperiments failed: \(error)\n", stderr)
        }
    }
}
```

Results:
```
[default][1] Prompt:     20 tokens, 6.000112 tokens/s, 3.333271s
Generation: 32 tokens, 0.619567 tokens/s, 51.648991s
[default][2] Prompt:     20 tokens, 6.084363 tokens/s, 3.287115s
Generation: 32 tokens, 0.621679 tokens/s, 51.473488s
[default][3] Prompt:     20 tokens, 6.107697 tokens/s, 3.274557s
Generation: 32 tokens, 0.618731 tokens/s, 51.718766s
[default][avg] Prompt:     20 tokens, 6.063704 tokens/s, 3.298314s
[default][avg] Generation: 32 tokens, 0.61999 tokens/s, 51.613748s
[default] Memory start: Peak:   0K           (0)
Active: 42766M       (44844101616)
Cache:  1K           (1924)
[default] Memory end:   Peak:   42976M       (45063987254)
Active: 42766M       (44844101696)
Cache:  228M         (239555134)
[default] Memory delta: Peak:   42976M       (45063987254)
Active: 0K           (80)
Cache:  228M         (239553210)

[fixed][1] Prompt:     20 tokens, 12.289435 tokens/s, 1.627414s
Generation: 32 tokens, 40.853011 tokens/s, 0.783296s
[fixed][2] Prompt:     20 tokens, 112.502345 tokens/s, 0.177774s
Generation: 32 tokens, 41.185421 tokens/s, 0.776974s
[fixed][3] Prompt:     20 tokens, 111.909433 tokens/s, 0.178716s
Generation: 32 tokens, 41.068706 tokens/s, 0.779182s
[fixed][avg] Prompt:     20 tokens, 30.243397 tokens/s, 0.661301s
[fixed][avg] Generation: 32 tokens, 41.03525 tokens/s, 0.779817s
[fixed] Memory start: Peak:   0K           (0)
Active: 42766M       (44844101616)
Cache:  0K           (0)
[fixed] Memory end:   Peak:   42976M       (45063987254)
Active: 42811M       (44890746972)
Cache:  183M         (192908646)
[fixed] Memory delta: Peak:   42976M       (45063987254)
Active: 44M          (46645356)
Cache:  183M         (192908646)

[max][1] Prompt:     20 tokens, 112.481528 tokens/s, 0.177807s
Generation: 32 tokens, 40.866103 tokens/s, 0.783045s
[max][2] Prompt:     20 tokens, 113.459151 tokens/s, 0.176275s
Generation: 32 tokens, 41.063228 tokens/s, 0.779286s
[max][3] Prompt:     20 tokens, 114.211291 tokens/s, 0.175114s
Generation: 32 tokens, 41.032108 tokens/s, 0.779877s
[max][avg] Prompt:     20 tokens, 113.379562 tokens/s, 0.176399s
[max][avg] Generation: 32 tokens, 40.986964 tokens/s, 0.780736s
[max] Memory start: Peak:   0K           (0)
Active: 42804M       (44883439704)
Cache:  0K           (0)
[max] Memory end:   Peak:   42976M       (45064019542)
Active: 42811M       (44890746972)
Cache:  183M         (192884610)
[max] Memory delta: Peak:   42976M       (45064019542)
Active: 7136K        (7307268)
Cache:  183M         (192884610)
```

Results were awesome! It turns out precomputing memory requirements upfront benefits from the same performance as just allocating max (but only of you clear cache)! Lemme know what you think @davidkoski 

## Checklist

### DIFF
--- a/Libraries/MLXLMCommon/WiredMemoryUtils.swift
+++ b/Libraries/MLXLMCommon/WiredMemoryUtils.swift
@@ -32,7 +32,7 @@ public enum WiredMemoryUtils {
     ///
     /// This does not attempt to generate semantically meaningful text; it only ensures
     /// a valid token sequence of the requested length for memory sizing purposes.
-    public static func makeTokenIDs(
+    private static func makeTokenIDs(
         count: Int,
         tokenizer: Tokenizer,
         seedText: String = " hello"
@@ -70,7 +70,7 @@ public enum WiredMemoryUtils {
     ///
     /// - Note: This is intended for text-only models. Multimodal models should
     ///   supply a fully prepared `LMInput` via their processor instead.
-    public static func makeTokenInput(
+    private static func makeTokenInput(
         count: Int,
         tokenizer: Tokenizer,
         seedText: String = " hello"
@@ -89,7 +89,7 @@ public enum WiredMemoryUtils {
     ///   - model: The language model to prefill.
     ///   - parameters: Generation parameters that control prefill behavior.
     /// - Returns: The cache array after prefill, suitable for KV sizing.
-    public static func prefillOnly(
+    private static func prefillOnly(
         input: LMInput,
         model: any LanguageModel,
         parameters: GenerateParameters


Libraries/MLXLMCommon/WiredMemory.swift
```diff
@@ -0,0 +1,57 @@
...
+        guard Device.defaultDevice().deviceType == .gpu else {
+            return try await body()
+        }
+
+        guard let limit = resolvedLimitBytes() else {
+            return try await body()
+        }
+
+        #if os(macOS) || os(iOS) || os(tvOS)
+            if #available(macOS 15, iOS 18, tvOS 18, *) {
```
davidkoski: Are either of these guards necessary?  The `Memory` API is available in mlx-swift 0.30.2 and that only requires macOS 14.0.

Looking at the MLX underpinnings this does seem specific to Metal, but it is defined in `memory.c` so appears to be generic (e.g. should be supported across different backends even if it ends up being a NOP).

Libraries/MLXLMCommon/Evaluate.swift
```diff
@@ -815,78 +819,80 @@ public func generate(
 /// - Returns: An `AsyncStream` that emits `Generation` values
 public func generate(
     input: LMInput, context: ModelContext,
-    iterator: TokenIterator
+    iterator: TokenIterator, wiredMemory: WiredMemoryLimit = .default
 ) -> AsyncStream<Generation> {
```
davidkoski, line 835: I wonder if we need this for the synchronous generation too?

```
public func generate(
    input: LMInput, context: ModelContext,
    iterator: TokenIterator,
    didGenerate: ([Int]) -> GenerateDisposition
) -> GenerateResult {
```

Also see #55 -- that has some touches in this file too that will need to be merged with this.

Libraries/MLXLMCommon/WiredMemory.swift
robertmsale: > Are either of these guards necessary?

You're absolutely right. I tried removing the guards and there isn't even an optimization reason for having it there so I'll go ahead and remove them.

Libraries/MLXLMCommon/Evaluate.swift
davidkoski: I think taking this after #55 might be the right way to go because that clarifies (and makes usable) the exact lifetime of the inference.

robertmsale: > I think taking this after #55 might be the right way to go

Sounds good! I'll hold off on the synchronous wired memory limit code as well, so once #55 comes through I can resolve conflicts as its own tiny commit and a more reviewable sync wired generate separately.

Libraries/MLXLMCommon/WiredMemory.swift
```diff
@@ -0,0 +1,579 @@
+// Copyright © 2024 Apple Inc.
+
+import Darwin
+import Dispatch
+import Foundation
+import MLX
+
+@_silgen_name("mlx_set_wired_limit")
+private func mlx_set_wired_limit(_ res: UnsafeMutablePointer<size_t>, _ limit: size_t) -> Int32
```
davidkoski: I think this code should live in mlx-swift -- I don't think the current wired memory implementation is really usable except in a single threaded app.

Wo don't want to break current API so I think the existing code should use a globally available "sum" policy (that seems safest) implicitly (but documented).  We can then deprecate the current methods (still callable, but don't!)

Then we can add the new entry points that deal with the tickets/policy.

Libraries/MLXLMCommon/WiredMemory.swift
```diff
@@ -0,0 +1,579 @@
...
+
+private enum WiredMemoryBackend {
+    static var isSupported: Bool {
+        Device.defaultDevice().deviceType == .gpu
+    }
+
+    static func readCurrentLimit() -> Int? {
+        guard isSupported else { return nil }
+        var previous: size_t = 0
+        let result = mlx_set_wired_limit(&previous, 0)
```
davidkoski: I don't think we should support this operation, or if we do we need a non-destructive read from the mlx core.  This temporarily sets the limit to 0 and I don't know what effect that might have (now or in the future).

I don't think we really need this -- I think we want to _tell_ it what the wire memory size is, not ask.  If we need to track it for policy reasons, then I think the policy should track it internally.  I don't really like shadow state like that, but I suspect we don't need to know what it currently is or if it is just advisory we can tolerate.

Libraries/MLXLMCommon/WiredMemory.swift
```diff
@@ -0,0 +1,579 @@
...
+        guard isSupported else { return false }
+        guard limit >= 0 else { return false }
+        var previous: size_t = 0
+        let result = mlx_set_wired_limit(&previous, size_t(limit))
+        return result == 0
+    }
+}
+
+private func maxRecommendedWorkingSetBytes() -> Int? {
+    let maxBytes = GPU.deviceInfo().maxRecommendedWorkingSetSize
```
davidkoski: It might be worth measuring the cost here or perhaps using the MTL API directly.  I don't know if this is a static or dynamic value, but assuming dynamic is safer (so having a function to call like this is probably a good idea).

Just thinking out loud here :-)

Libraries/MLXLMCommon/WiredMemory.swift
```diff
@@ -0,0 +1,579 @@
...
+/// Policy for computing a process-global wired memory limit.
+///
+/// - Note: To group tickets correctly, policies should be pure and either Hashable
+///   or reference types. Non-hashable value types are treated as distinct policies.
+public protocol WiredMemoryPolicy: Sendable {
+    /// Compute the desired wired limit in bytes for the current active set.
+    func limit(baseline: Int, activeSizes: [Int]) -> Int
+
+    /// Decide whether a new ticket can be admitted. Defaults to allowing all tickets.
+    func canAdmit(baseline: Int, activeSizes: [Int], newSize: Int) -> Bool
```
davidkoski: I am not clear how this is used -- what if it can't admit?  It looks like it just ignores the ticket?  What is the use case where we need this?

If we don't have any implementation today (I guess we do in the sum policy) we could leave this out and add later (via the default implementation).

What is the difference between this and just claming against the max size?

Libraries/MLXLMCommon/WiredMemory.swift
```diff
@@ -0,0 +1,579 @@
...
+    }
+}
+
+/// Policy for computing a process-global wired memory limit.
+///
+/// - Note: To group tickets correctly, policies should be pure and either Hashable
+///   or reference types. Non-hashable value types are treated as distinct policies.
+public protocol WiredMemoryPolicy: Sendable {
+    /// Compute the desired wired limit in bytes for the current active set.
+    func limit(baseline: Int, activeSizes: [Int]) -> Int
```
davidkoski: So this computes baseline + some aggregation of activeSizes?  I was thinking that the representation of the ask would be private between the ticket and the policy.  That way we don't limit the implementation if it wants to model shared wired memory (which I understand is represented here in the activeSizes, I think).

I think the operations the policy should support are:

- activate ticket -> Int
- deactivate ticket -> Int

and (critically) these would be functions on the ticket itself -- that way the type agreement is internal to the ticket.  The ticket would be a concrete implementation of a protocol (see note below on the ticket).

The details of the bookkeeping are entirely internal to the policy.  If this is too vague (and you like the idea) I can write a small example.

Just an idea -- I think what you have works for the use cases we know of, but I wonder about future cases.

On further reading, I think my idea would encapsulate the manager inside the policy (for the most part) and only allow one policy active at a time.  It may be that the multiple policy format you have is better and the way to go, see what you think.

Libraries/MLXLMCommon/WiredMemory.swift
```diff
@@ -0,0 +1,579 @@
...
+        self.bytes = limit
+    }
+
+    public func limit(baseline: Int, activeSizes: [Int]) -> Int {
+        bytes
+    }
+}
+
+/// Handle for coordinating wired memory changes across concurrent work.
+public struct WiredMemoryTicket: Sendable, Identifiable {
```
davidkoski: If you like the ticket/policy idea I mentioned above then this would turn into a protocol and there would be a ticket per policy type (`any WiredMemoryTicket`).

Libraries/MLXLMCommon/WiredMemory.swift
```diff
@@ -0,0 +1,579 @@
...
+    public static func withWiredLimitSync<R>(
+        _ ticket: WiredMemoryTicket,
+        _ body: () throws -> R
+    ) rethrows -> R {
+        let startSemaphore = DispatchSemaphore(value: 0)
+        Task {
+            _ = await ticket.start()
+            startSemaphore.signal()
+        }
+        startSemaphore.wait()
```
davidkoski: We shouldn't do this -- it is an anti-pattern to do semaphore signaling to turn async back to sync.  I know it marks all the API as asynchronous but it is the correct way to do it.

Since this sits on top of an `actor` that forces it to be only async.  The other option is to not use an `actor` and use locks to implement `@unchecked Sendable`.  That gives thread safety and synchronous calls.  But locks don't work with async calls, see `SerialAccessContainer`.  Well, locks won't work in an `async` context, but ought to work inside a private synchronous function, which may be enough -- we just need to guard shared mutable state.

TLDR:

- we can't use semaphores in this way
- either make this call `async` only (built on the `actor`)
- OR use locks inside the manager to make synchronous and async implementations

Libraries/MLXLMCommon/WiredMemory.swift
```diff
@@ -0,0 +1,579 @@
...
+
+    private enum PolicyKey: Hashable {
+        case hash(AnyHashable)
+        case object(ObjectIdentifier)
+        case unique(UUID)
+    }
+
+    private var baseline: Int?
+    private var active: [UUID: TicketState] = [:]
+    private var policies: [PolicyKey: any WiredMemoryPolicy] = [:]
```
davidkoski: Ah, interesting -- this looks like you would allow multiple policies to be active at once?  I think we could get away with forcing the application to own this state (which policy, but not static if the application has other needs), but this is a good idea too.

Libraries/MLXLMCommon/WiredMemory.swift
```diff
@@ -0,0 +1,579 @@
...
+        private var eventSequence: UInt64 = 0
+    #endif
+
+    private struct TicketState {
+        let size: Int
+        let policyKey: PolicyKey
+        let policyLabel: String
+    }
+
+    private enum PolicyKey: Hashable {
```
davidkoski: We could just require policies to be Identifiable.

Libraries/MLXLMCommon/WiredMemory.swift
```diff
@@ -0,0 +1,579 @@
...
+            }
+            endSemaphore.wait()
+        }
+
+        return try body()
+    }
+}
+
+public actor WiredMemoryManager {
+    public static let shared = WiredMemoryManager()
```
davidkoski: Does this need to be `nonisolated`?

Libraries/MLXLMCommon/WiredMemory.swift
```diff
@@ -0,0 +1,579 @@
...
+                }
+            }
+        #else
+            return AsyncStream { continuation in
+                continuation.finish()
+            }
+        #endif
+    }
+
+    public func start(id: UUID, size: Int, policy: any WiredMemoryPolicy) async -> Int {
```
davidkoski: Suggestion:  `id` -> `ticketId`

davidkoski: I think this method in particular needs some documentation.  It looks like it returns the new wired size?  What does the caller do with that?  It isn't to apply the value as that seems to happen internally.  Should this just be void?

Libraries/MLXLMCommon/WiredMemory.swift
```diff
@@ -0,0 +1,579 @@
...
+        }
+
+        var baselineValue = ensureBaseline()
+        if active[id] != nil {
+            emit(
+                kind: .ticketStartIgnored,
+                ticketID: id,
+                size: normalizedSize,
+                baseline: baselineValue
+            )
```
davidkoski: This is an interesting idea -- I have thought about adopting https://github.com/apple/swift-log (I think that is probably the right pick) for observability.  Do you have a use case in mind here?  Should this be done using logs instead and this is the forcing function for adding them?  I feel like we might add quite a few logs once we go that path.  I am a little hesitant to have a bunch of one-off observability mechanisms.

Libraries/MLXLMCommon/WiredMemory.swift
davidkoski: OK, I feel like I am not very clear here and perhaps pushing my own mental model.  Let me restate this as requirements and goals.

- this should live in the mlx-swift layer because all types of models might want to make use of it
- there should be no way to use the wired memory without going through some sort of thread safe layer -- the current methods should sit on top of this with an implied policy (and be deprecated in favor of using explicit policies, but still work and "safe")
- different situations will call for different policies and algorithms for determining how much memory to wire
    - known cases are at least: sum, max, and something a bit fancier that deals with shared weights + context (KVCache and intermediates in computation).  there could be more, the point is there is diversity
- this is managing a process global resource so it either has to be managed at the application level (only) OR it has to have the flexibility to layer in multiple policies simultaneously
- task aware, thread safe
- ideally can be called from both sync and async contexts -- the usual withX pattern

I think that covers the needs.  Aside from being in mlx-lm-swift, I think your implementation covers this.

The entities in the model are something like:

- ticket -- a request for resource.  it identifies an amount of memory but may contain more context like an identifier it is shared with (which model for shared weights)
- policy -- given N active tickets, how much memory do we wire down
- manager -- given N active policies, how much memory do we wire down

The policy and manager could be the same thing:  you could imagine a policy that used composition, but I think the other key part of the manager is that it is a singleton and global.  No matter how you reach it, it will combine all active policies.  If we were to use composition in a policy there would still need to be coordination in injecting the different sub-policies.  A singleton isn't usually the first choice in designs these days, but it is modeling a singleton resource at the process level.

I think the API wants to deal in tickets -- the generate method takes an optional ticket and aside from its existence it could be opaque.  The generate method calls a withWiredMemory style method to scope its use of the wired memory safely.

I think there is a question on how the bookkeeping works -- is it internal to the policy?  That makes the policy implementation heavier but much more flexible.  If it is in the manager then something like what you have is flexible up to a point (and maybe that is fine).  

So consider what level of abstraction would be needed at the mlx-swift level.

Tests/MLXLMIntegrationTests/WiredMemoryIntegrationTests.swift
```diff
@@ -0,0 +1,496 @@
...
+import XCTest
+
+final class WiredMemoryIntegrationTests: XCTestCase {
+    enum TestError: Error {
+        case missingInfo
+        case timeout
+        case missingBaseline
+    }
+
+    struct DoubleSumPolicy: WiredMemoryPolicy, Hashable, Sendable {
```
davidkoski: Are these tests redundant with the ones in mlx-swift?  We might want to test the policies that are added in mlx-swift-lm but any testing of the core functionality should be in mlx-swift.

Libraries/MLXLMCommon/Evaluate.swift
```diff
@@ -4,6 +4,12 @@ import Foundation
 import MLX
 import Tokenizers
 
+private func withOptionalWiredLimitSync<R>(_ limit: Int?, _ body: () -> R) -> R {
```
davidkoski: Is this still needed?

Libraries/MLXLMCommon/Evaluate.swift
```diff
@@ -608,57 +626,60 @@ public func generate(
 public func generate(
     input: LMInput, context: ModelContext,
     iterator: TokenIterator,
+    wiredMemoryLimit: Int? = nil,
     didGenerate: ([Int]) -> GenerateDisposition
 ) -> GenerateResult {
-    var start = Date.timeIntervalSinceReferenceDate
-    var promptTime: TimeInterval = 0
+    return withOptionalWiredLimitSync(wiredMemoryLimit) {
```
davidkoski: This is a NOP in the implementation -- is this change out of date?

At this layer the call is synchronous so the wired memory API can't support it.  I am not sure which level makes the most sense.  Probably `ChatSession` (being the top level).  I wonder if lower level uses should just use the mlx-swift API along with an LLM-focused policy?

Ah, but maybe it _has_ to be down at the lowest level because that is where we know the lifetime.  Which would suggest it should fit here:

```
public func generateTask(
    promptTokenCount: Int,
    tokenizer: Tokenizer,
    iterator: consuming TokenIterator

    let (stream, continuation) = AsyncStream<Generation>.makeStream()

    let iterator = SendableBox(iterator)

    // Launch a Task to perform iteration asynchronously.
    let task = Task {
        let iterator = iterator.consume()
```

Tests/MLXLMIntegrationTests/WiredMemoryIntegrationTests.swift
```diff
@@ -0,0 +1,300 @@
+// Copyright © 2025 Apple Inc.
+
+import Foundation
+import MLX
+import MLXLMCommon
+import XCTest
+
+final class WiredMemoryIntegrationTests: XCTestCase {
```
davidkoski: Not a blocker for the PR, but ideally these would not be in the integration tests target -- those are not run in CI (as they require large downloads on every run).  I think the LLM inference isn't actually required to exercise this (or it could probably use the random weight versions from EvalTests).

Libraries/MLXLMCommon/Evaluate.swift
```diff
@@ -598,6 +608,8 @@ public func generate(
 ///   - input: prepared language model input
 ///   - context: model context (model and tokenizer)
 ///   - iterator: token iterator
+///   - wiredMemoryLimit: Optional wired limit (bytes). This synchronous path no
+///     longer applies wired limits; use ticket-based APIs for wiring control.
```
davidkoski: Should this just not have the parameter then?  Adding a parameter that doesn't do anything doesn't seem to help much.  I don't think you can just switch back and forth between the sync and async version.

Libraries/MLXLMCommon/Evaluate.swift
```diff
@@ -669,6 +684,8 @@ public func generate(
 ///   - input: prepared language model input
 ///   - parameters: parameters controlling the token generation
 ///   - context: model context (model and tokenizer)
+///   - wiredMemoryLimit: Optional wired limit (bytes). This synchronous path no
+///     longer applies wired limits; use ticket-based APIs for wiring control.
```
davidkoski: Same question as above.

Libraries/MLXLMCommon/Evaluate.swift
```diff
@@ -678,12 +695,15 @@ public func generate(
 )
 public func generate(
     input: LMInput, parameters: GenerateParameters, context: ModelContext,
+    wiredMemoryLimit: Int? = nil,
```
davidkoski: Same sync case.

Libraries/MLXLMCommon/Evaluate.swift
```diff
@@ -694,6 +714,10 @@ public func generate(
 ///   - input: prepared language model input
 ///   - context: model context (model and tokenizer)
 ///   - iterator: token iterator
+///   - wiredMemoryLimit: Optional wired limit (bytes). This synchronous path no
+///     longer applies wired limits directly and executes as a no-op wrapper. Use
+///     async generation with `WiredMemoryTicket.withWiredLimit` to coordinate
+///     policy-driven wired memory management.
```
davidkoski: Same (I will stop marking them but note that not all of these sync cases have clear docs that it is ignored).  My vote is to not add the parameter if it can't be used.

Libraries/MLXLMCommon/Evaluate.swift
```diff
@@ -857,86 +893,96 @@ public func generateTask(
...
 
-        // Finalize the stream
-        continuation.finish()
+        if let ticket = wiredMemoryTicket {
+            await WiredMemoryTicket.withWiredLimit(ticket) {
+                performIteration()
+            }
+        } else {
+            performIteration()
+        }
```
davidkoski, line 954: Interesting -- perhaps it should have allowed a `nil` ticket and then we don't need this dance.  A NOP ticket might also be a good way to handle it as well.  I didn't think of this complication :-)

I don't think it is a blocker for this PR, but maybe something we should come back and improve later?  I think having to do this dance obfuscates the code a little.

Libraries/MLXLMCommon/WiredMemoryUtils.swift
```diff
@@ -0,0 +1,249 @@
...
+    }
+}
+
+/// Helpers for deriving wired memory budgets from real runtime measurements.
+public enum WiredMemoryUtils {
+    /// Produce a token ID array of exactly `count` tokens using the given tokenizer.
+    ///
+    /// This does not attempt to generate semantically meaningful text; it only ensures
+    /// a valid token sequence of the requested length for memory sizing purposes.
+    public static func makeTokenIDs(
```
davidkoski: Should these (next couple utility functions) be public?  They are called by `tune` and seem unlikely to be of general use outside of that.

robertmsale:
Hey @davidkoski I addressed the `wiredMemoryLimit: Int?` params (completely removed), made the non-public WiredMemoryUtils helpers private, and reduced the wired-memory tests here down to policy math only. Core policy/ticket manager behavior and event-stream coverage now lives in `mlx-swift`, and this avoids unnecessary heavy work on CI machines.

For the NOP/nil ticket case, I think keeping the explicit branch is the safest shape for now. If we later add optional-ticket ergonomics upstream in mlx-swift, the semantics here would still be valid. If no ticket is provided, we do not touch the wired-memory management pipeline, the manager does not emit ticket lifecycle events, and generation behaves as usual.

PR #73, opened 2026-01-21
Update glm4_moe_lite To Store KV Latent in Cache

ronaldmannak:
## Proposed changes

This PR includes the optimizations proposed in https://github.com/ml-explore/mlx-lm/pull/780.

The last two commits incorporate @awni’s suggested improvements from that PR. Please double-check them, I’m not fully confident I interpreted all of the proposed changes correctly.

## Checklist

### DIFF
--- a/Libraries/MLXLLM/Models/GLM4MOELite.swift
+++ b/Libraries/MLXLLM/Models/GLM4MOELite.swift
@@ -544,7 +544,9 @@ public class GLM4MoELiteModel: Module, LLMModel, KVCacheDimensionProvider {
                     // Infer bits and group size
                     inferredBits = (v.dim(-1) * 32) / dims
                     inferredGroupSize = dims / scales.dim(-1)
-                    v = dequantized(v, scales: scales, biases: biases, groupSize: inferredGroupSize, bits: inferredBits)
+                    v = dequantized(
+                        v, scales: scales, biases: biases, groupSize: inferredGroupSize,
+                        bits: inferredBits)
                 }
 
                 let numHeads = configuration.attentionHeads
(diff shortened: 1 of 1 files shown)


ronaldmannak:
(I sneaked in one small cache fix)

adrgrondin:
I was working on the changes to the safetensors files too and got something running but inference is extremely slow with `mlx-community/GLM-4.7-Flash-4bit` on M5 24GB, probably around 1tk/20s, with `mlx-lm` I have no issues at 40tk/s.

Reproducing the same with your latest commit @ronaldmannak.

Anyone here experiencing the same? Maybe something deeper about M5 and `mlx-swift`? Not really sure where to look for this and I can't test on MX Pro right now.

@davidkoski any idea?

ronaldmannak:
@adrgrondin did you try the commit I pushed 2 minutes before your comment? That one fixes a cache issue that slowed down subsequent generation (meaning that the first answer was fast, after that everything slowed down)

Also, I'm sure you've thought of that, but 24GB may be a bit small for GLM. You might want to try #72 to avoid paging

Libraries/MLXLLM/Models/GLM4MOELite.swift
```diff
@@ -8,9 +8,123 @@
 
 import Foundation
 import MLX
+import MLXFast
```
adrgrondin: Not building with the `import MLXFast`. I think it can be removed.

ronaldmannak: You're right. I removed it

davidkoski: FWIW MLXFast was folded into MLX proper and lives under an enum by the same name -- the calls look the same.  This no longer builds because the MLXLLM doesn't link MLXFast (a stub library remains).

davidkoski:
I can test it but it won't be until Thursday.  If needed I can profile both this and mlx-lm and see where the difference lies.  

When doing performance tests make sure it is running either outside of Xcode completely or make sure to disable:

- attach debugger
- validate Metal shaders

I wouldn't expect 40x difference though.

adrgrondin:
@ronaldmannak @davidkoski I did some tests with #72 and was able to get the model running like `mlx-lm`. So it is liked to memory, it was just not obvious for me since no tuning to the wired memory was needed with `mlx_lm.generate` but also I was not really familiar with all that.

Thanks again @ronaldmannak for the recommendation

ronaldmannak:
@adrgrondin that's great to hear!

ronaldmannak:
That makes sense. I hope you're able to merge #72 asap, it will be a big boost for people with limited memory (and that's everyone in this community except @ivanfioravanti ) ;)

davidkoski:
> That makes sense. I hope you're able to merge #72 asap, it will be a big boost for people with limited memory (and that's everyone in this community except @ivanfioravanti ) ;)

It isn't ready yet -- we uncovered a reasonable size gap.  The wired memory is global for the process, which is fine in python.  In swift we need to use that mechanism but simulate task scoping.  The author has an implementation but it needs to live at the mlx-swift layer.  Maybe we can do it in two steps and live with the gap for a short time.

ronaldmannak:
@davidkoski any chance we can get this merged soon?

PR #78, opened 2026-01-26
Add configurable tool call parsing with support for multiple model formats

### LINKED ISSUE #47: ToolCallProcessor has hard coded tags
I noticed tool calling isn't working for LFM2 models, and after investigation, found out there's a mismatch in tool calling tags.

https://github.com/ml-explore/mlx-swift-lm/blob/5064b8c5d8ed3b0bbb71385c4124f0fc102e74a2/Libraries/MLXLMCommon/Tool/ToolCallProcessor.swift#L21-L22

This would be fine for llama/qwen-based models, but LFM2 uses `<|tool_call_start|>` and `<|tool_call_end|>` which are completely ignored.

tpae:
## Proposed changes

Fixes #47 - Tool call tags are hardcoded and don't support models like LFM2.

This PR adds a flexible tool call parsing system that supports multiple formats used by different LLM families, bringing parity with the Python `mlx-lm` library's tool parser implementations.

### Supported Formats

| Format | Models | Reference |
|--------|--------|-----------|
| JSON (default) | Llama, Qwen, Mistral | [json_tools.py](https://github.com/ml-explore/mlx-lm/blob/main/mlx_lm/tool_parsers/json_tools.py) |
| JSON (LFM2 tags) | LFM2, LFM2-MoE | [json_tools.py](https://github.com/ml-explore/mlx-lm/blob/main/mlx_lm/tool_parsers/json_tools.py) |
| XML Function | Qwen3 Coder | [qwen3_coder.py](https://github.com/ml-explore/mlx-lm/blob/main/mlx_lm/tool_parsers/qwen3_coder.py) |
| GLM4 | GLM4, GLM4-MoE | [glm47.py](https://github.com/ml-explore/mlx-lm/blob/main/mlx_lm/tool_parsers/glm47.py) |
| Gemma | Gemma | [function_gemma.py](https://github.com/ml-explore/mlx-lm/blob/main/mlx_lm/tool_parsers/function_gemma.py) |
| Kimi K2 | Kimi K2 | [kimi_k2.py](https://github.com/ml-explore/mlx-lm/blob/main/mlx_lm/tool_parsers/kimi_k2.py) |
| MiniMax M2 | MiniMax M2 | [minimax_m2.py](https://github.com/ml-explore/mlx-lm/blob/main/mlx_lm/tool_parsers/minimax_m2.py) |

### Key Changes

- **New `ToolCallFormat` enum** with configurable formats and convenience properties
- **New `ToolCallParser` protocol** for pluggable parsing implementations  
- **Auto-detection** of format from `model_type` in config.json via `ToolCallFormat.infer(from:)`
- **Schema-aware type conversion** matching Python behavior for integer, boolean, object types
- **Backwards compatible** - existing code works without changes

### API Usage

// Automatic (recommended) - format detected from model_type
let container = try await LLMModelFactory.shared.loadContainer(configuration: config)

// Explicit format
let processor = ToolCallProcessor(format: .lfm2)

// Custom JSON tags
let processor = ToolCallProcessor(format: .json(startTag: "<custom>", endTag: "</custom>"))

## Checklist

### DIFF
--- a/Libraries/MLXLMCommon/Tool/ToolCallFormat.swift
+++ b/Libraries/MLXLMCommon/Tool/ToolCallFormat.swift
@@ -34,15 +34,21 @@ public protocol ToolCallParser: Sendable {
 /// This enum defines the various tool call formats used by different LLM families.
 /// Each format has its own syntax for encoding function names and arguments.
 ///
+/// The raw string values can be used for JSON serialization or CLI parameters.
+///
 /// Reference: https://github.com/ml-explore/mlx-lm/tree/main/mlx_lm/tool_parsers
-public enum ToolCallFormat: Sendable, Equatable {
-    /// JSON format with configurable start/end tags.
+public enum ToolCallFormat: String, Sendable, Codable, CaseIterable {
+    /// Default JSON format used by Llama, Qwen, and most models.
     /// Example: `<tool_call>{"name": "func", "arguments": {...}}</tool_call>`
-    case json(startTag: String, endTag: String)
+    case json
+
+    /// LFM2 JSON format with model-specific tags.
+    /// Example: `<|tool_call_start|>{"name": "func", "arguments": {...}}<|tool_call_end|>`
+    case lfm2
 
     /// XML function format used by Qwen3 Coder.
     /// Example: `<function=name><parameter=key>value</parameter></function>`
-    case xmlFunction
+    case xmlFunction = "xml_function"
 
     /// GLM4 format with arg_key/arg_value tags.
     /// Example: `func<arg_key>k</arg_key><arg_value>v</arg_value>`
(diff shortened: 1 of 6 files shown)


Libraries/MLXLMCommon/Tool/ToolCallFormat.swift
```diff
@@ -0,0 +1,130 @@
...
+    public static let glm4Moe = ToolCallFormat.glm4
+
+    /// Gemma function call format.
+    public static let gemmaFunction = ToolCallFormat.gemma
+
+    /// Kimi K2 format.
+    public static let kimi = ToolCallFormat.kimiK2
+
+    /// MiniMax M2 format.
+    public static let minimax = ToolCallFormat.minimaxM2
```
davidkoski: Are these effectively aliases?  Otherwise it is a little odd to see both `case` and `public static let` in the same namespace (this enum).

I wonder about making all of these cases and making the type:

```swift
enum ToolCallFormat: String
```

in case we want to write to JSON or use as command line parameters in llm-tool?

I think it wold just require changing the `createParser` function fold these together.

tpae: I agree, making it a String-backed enum would be cleaner and enable JSON serialization / CLI usage. I updated the code

PR #87, opened 2026-01-29
[WIP] Attempt to run on Linux

Joannis:
## Proposed changes

## Checklist


Joannis:
Sorry for the maintainers, I pushed this WIP code here for public tracking of a bug I'm sorting first. I'm assuming you're on the same team

davidkoski:
Yeah, the VLM code probably needs some thinking to separate out the image processing and image representation.  The models should all be dealing with MLXArray, hopefully we can just replace MediaProcessing.

Libraries/MLXLLM/Models/BaichuanM1.swift
```diff
@@ -7,6 +7,7 @@
 
 import Foundation
 import MLX
+import MLXFast
```
davidkoski: We shouldn't need this -- MLXFast is an empty package with some stubs to forward back to MLX.  MLX has an enum `MLXFast` so without `import MLXFast` the same code should still compile.

Joannis: Good one yeah. It's an artifact since this code has been on my machine for a while. Will rid of it

Libraries/MLXVLM/Models/QwenVL.swift
```diff
@@ -1,258 +1,266 @@
...
-}
-
-public struct QwenVL {
-    /// Rotates half the hidden dims of the input
-    static func rotateHalf(_ x: MLXArray) -> MLXArray {
-        let index = x.dim(-1) / 2
-        let x1 = x[.ellipsis, 0 ..< index]
-        let x2 = x[.ellipsis, index...]
-        return concatenated([-x2, x1], axis: -1)
+#if os(macOS) || os(iOS) || os(tvOS) || os(watchOS) || os(visionOS)
```
Joannis, line 1: This doesn't work on Linux yet

JaeminKim-amoz (commented):
**COMMENT — WIP, not mergeable yet**

The direction is good — conditionally compiling Metal-dependent code is the right approach. CI addition is clean. The `@Sendable` closure fixes are genuine improvements.

**Blockers:**
- [HIGH] ~4K lines of whitespace churn from re-indenting files inside `#if` blocks makes review extremely difficult and hides real changes. Please avoid re-indenting, or split whitespace changes into a separate commit.
- [MEDIUM] WIP status — no tests, pre-commit not run, incomplete checklist.
- [MEDIUM] `MediaProcessing.swift` and `UserInput.swift` changes (500+ and 120+ lines) need closer inspection to confirm they are whitespace-only.

GoodOlClint:
Hi @Joannis, @davidkoski — I've been working through Linux/CUDA enablement for this package on an aarch64 box (GB10, sm_121, CUDA 13.0) and ended up covering ground next to this PR, so rather than open anything competing I wanted to share what I found and offer to help here.

**What I've already sent separately:** #483 covers only the non-VLM gaps left by the merged #321 — the missing `autoreleasepool` shim, an unguarded `import os` in MLXGuidedGeneration, the incomplete `Logger` shim, and the unconditional `-lc++` link. Those were chosen specifically because they sit outside this PR's scope. The one file that does overlap is `Package.swift`, and only on the `.linkedLibrary("c++")` line.

**Three things that might be useful for this PR:**

1. *The surface grew.* On `main` today there are 16 unguarded Apple-framework imports across 15 files in `Libraries/MLXVLM/` — `AVFoundation` in `MediaProcessing.swift`, `CoreImage` in 14 model files, plus `CoreMedia` in `SmolVLM2.swift`. This PR touches 5 of those 15. The other 10 (`FastVLM`, `Gemma4`, `GlmOcr`, `Idefics3`, `LFM2VL`, `Mistral3`, `Pixtral`, `Qwen25VL`, `Qwen3VL`, `SmolVLM2`) all landed after this branch was cut.

2. *#321 established a different idiom that would shrink the diff.* It merged after this PR was last pushed and put `#if !canImport(X)` type shims in `Libraries/MLXLMCommon/Linux/` — `CoreGraphics.swift` (a `CGSize`/`CGFloat` shim), `CoreMedia.swift` (`CMTime`), `Logger.swift`, `String+Linux.swift`. Because the shims supply the *types*, call sites need no platform conditionals at all. That directly addresses the thing you flagged as blowing up this diff ("I've added a fair amount of `#if os(macOS) || os(iOS) || ...` checks ... Formatting made sure that this blows up the diff") — the same coverage in the shim style should be substantially smaller and wouldn't need whitespace-blind review.

3. *It also lines up with what @davidkoski suggested.* `CIImage` usage really is concentrated: 32 references in `MediaProcessing.swift` versus 1–6 in each model file, and in the models it's almost entirely `preprocess(image: CIImage, ...)` signatures and `asCIImage()` calls. So a Linux `CIImage`/media shim plus a Linux `MediaProcessing` backend looks like it would cover most of the surface without per-model edits, which is close to "we can just replace MediaProcessing".

**Offer:** this PR is now `mergeable: CONFLICTING` against `main`, and refreshing it across 6 months of new VLM models is real work. I'm happy to do that — rebase onto current `main`, convert the platform lists to the `canImport` shim style, and extend to the 10 newer model files — either as a PR into your branch or as a co-authored branch, whichever you'd prefer, with authorship kept on your original commits. I can also actually run it: I have the Linux/CUDA box, though the SPM CUDA path needed some follow-ups to the merged mlx-swift #413 first (ml-explore/mlx-swift#442, #443, #444).

Happy to just leave this here as notes if you'd rather carry it yourself — let me know either way.

Joannis:
I'd be happy to have you take this up, it's still important to me just not urgent.

Joannis:
@GoodOlClint feel free to email me or message me elsewhere if you need input or eyes.

Issue #61, opened 2026-01-14
GPT-oss requires Harmony format (render + parse)


ronaldmannak:
### Summary
`mlx-swift-lm` includes a GPT-OSS model definition, but GPT-OSS needs **Harmony** formatting end-to-end (prompt rendering + stop tokens + parsing streamed/decoded output). The OpenAI cookbook explicitly states:

> “gpt-oss should not be used without using the harmony format, as it will not work correctly.”  
Source: https://cookbook.openai.com/articles/openai-harmony

Today, `mlx-swift-lm` doesn’t render/parse Harmony transcripts for GPT-OSS, which can lead to incorrect behavior (e.g., tool call boundaries, `analysis` vs `final`, stop conditions).

### Context / prior work
- Harmony FFI already exists in Swift via **PicoHarmony**:  
  https://github.com/PicoMLX/PicoHarmony/tree/master
- Harmony parsing is already integrated in **Pico AI Server** (shipping app):  
  https://apps.apple.com/us/app/pico-ai-server-llm-vlm-mlx/id6738607769?mt=12
- Related discussion (tooling + behavior differences):  
  https://github.com/dinoki-ai/osaurus/issues/282#issuecomment-3732855725

### What I’m seeing
When GPT-OSS is run without Harmony formatting/parsing, the output behavior differs significantly compared to Harmony-enabled flows.

**Screenshots (same prompt):**
- Without Harmony enabled:  
<img width="1292" height="1059" alt="Image" src="https://github.com/user-attachments/assets/be4dbdfd-a090-430f-8cf7-159ed1ba1bae" />
 
- With Harmony enabled:  
<img width="1292" height="1059" alt="Image" src="https://github.com/user-attachments/assets/04b673ae-aece-4703-86f6-d5e0f51944d4" />

### Expected behavior
For GPT-OSS models, `mlx-swift-lm` should provide a path that:
- renders conversations in **Harmony format** (including required headers/roles/channels),
- uses correct **stop tokens** for assistant completion vs tool/action segments,
- parses completion output back into structured messages (and supports streaming parsing).

### Proposal
Add **first-class Harmony integration for GPT-OSS** in `mlx-swift-lm`, ideally by leveraging PicoHarmony rather than re-implementing Harmony logic.

Concretely:
1. Add an optional Harmony dependency (e.g., `PicoHarmony`) gated by platform / feature flag if needed.
2. In the GPT-OSS chat pipeline:
[...]

PR #88, opened 2026-02-01
Add raw token streaming (TokenGeneration) and generateTokens convenience APIs

ronaldmannak:
**Summary**
This PR introduces a non-deprecated, low-level token streaming API that yields **raw token IDs** for projects that need token-level access (e.g. Harmony parsing #61 ), plus high-level convenience wrappers mirroring the existing `generate(...)` helpers.

**What’s included**
- New `TokenGeneration` enum (mirrors `Generation`, but yields `.token(Int)` and `.info(GenerateCompletionInfo)`).
- New `generateTokenTask(...) -> (AsyncStream<TokenGeneration>, Task<Void, Never>)`:
  - Streams raw token IDs from a `TokenIterator`
  - Uses the same EOS/unknown stopping semantics as `generateTask`
  - Supports `includeStopToken` for debugging / boundary-sensitive consumers
  - Cancels underlying task on stream termination
- New convenience APIs:
  - `generateTokens(...) throws -> AsyncStream<TokenGeneration>`
  - `generateTokensTask(...) throws -> (AsyncStream<TokenGeneration>, Task<Void, Never>)`

**Why**
Some downstream parsers require access to the model’s output at the **token ID** level. The existing `generateTask` detokenizes and parses tool calls, which loses the raw token stream needed for Harmony-style parsing. This PR provides a clean, non-deprecated path for raw token consumption without changing existing generation behavior.

**Notes**
- Default behavior does **not** emit the stop token; `includeStopToken` is opt-in.
- Existing `generateTask` / `generate` APIs are unchanged.

### DIFF
--- a/skills/mlx-swift-lm/SKILL.md
+++ b/skills/mlx-swift-lm/SKILL.md
@@ -1,6 +1,6 @@
 ---
 name: swift-mlx-lm
-description: MLX Swift LM - Run LLMs and VLMs on Apple Silicon using MLX. Covers local inference, streaming, tool calling, LoRA fine-tuning, embeddings, and porting models from Python MLX-LM.
+description: MLX Swift LM - Run LLMs and VLMs on Apple Silicon using MLX. Covers local inference, streaming, wired memory coordination, tool calling, LoRA fine-tuning, embeddings, and model porting.
 triggers:
   - mlx
   - mlx-swift
@@ -9,7 +9,9 @@ triggers:
   - local llm swift
   - vision language model swift
   - lora training swift
-  - port model
+  - wired memory
+  - wiredmemory
+  - wired memory ticket
   - model porting
   - add model support
 ---
@@ -18,12 +20,12 @@ triggers:
 
 ## 1. Overview & Triggers
 
-mlx-swift-lm is a Swift package for running Large Language Models (LLMs) and Vision-Language Models (VLMs) on Apple Silicon using MLX. It supports local inference, fine-tuning via LoRA/DoRA, and embeddings generation.
+mlx-swift-lm is a Swift package for running Large Language Models (LLMs) and Vision-Language Models (VLMs) on Apple Silicon using MLX. It supports local inference, streaming generation, wired-memory coordination, tool calling, LoRA/DoRA fine-tuning, and embeddings.
 
 ### When to Use This Skill
 - Running LLM/VLM inference on macOS/iOS with Apple Silicon
 - Streaming text generation from local models
-- Vision tasks with images/video (VLMs)
+- Coordinating concurrent inference with wired-memory policies and tickets
 - Tool calling / function calling with models
 - LoRA adapter training and fine-tuning
 - Text embeddings for RAG/semantic search
@@ -31,10 +33,10 @@ mlx-swift-lm is a Swift package for running Large Language Models (LLMs) and Vis
 
 ### Architecture Overview
 ```
-MLXLMCommon     - Core infrastructure (ModelContainer, ChatSession, KVCache, etc.)
-MLXLLM          - Text-only LLM support (Llama, Qwen, Gemma, Phi, DeepSeek, etc. - examples, not exhaustive)
-MLXVLM          - Vision-Language Models (Qwen2-VL, PaliGemma, Gemma3, etc. - examples, not exhaustive)
-Embedders       - Embedding models (BGE, Nomic, MiniLM)
+MLXLMCommon     - Core infra (ModelContainer, ChatSession, Evaluate, KVCache, wired memory helpers)
+MLXLLM          - Text-only LLM support (Llama, Qwen, Gemma, Phi, DeepSeek, etc.)
+MLXVLM          - Vision-Language Models (Qwen-VL, PaliGemma, Gemma3, etc.)
+MLXEmbedders    - Embedding models and pooling utilities
 ```
 
 ## 2. Key File Reference
@@ -43,8 +45,10 @@ Embedders       - Embedding models (BGE, Nomic, MiniLM)
 |---------|-----------|
 | Thread-safe model wrapper | `Libraries/MLXLMCommon/ModelContainer.swift` |
 | Simplified chat API | `Libraries/MLXLMCommon/ChatSession.swift` |
-| Generation & streaming | `Libraries/MLXLMCommon/Evaluate.swift` |
+| Generation & streaming APIs | `Libraries/MLXLMCommon/Evaluate.swift` |
 | KV cache types | `Libraries/MLXLMCommon/KVCache.swift` |
+| Wired-memory policies | `Libraries/MLXLMCommon/WiredMemoryPolicies.swift` |
+| Wired-memory measurement helpers | `Libraries/MLXLMCommon/WiredMemoryUtils.swift` |
 | Model configuration | `Libraries/MLXLMCommon/ModelConfiguration.swift` |
 | Chat message types | `Libraries/MLXLMCommon/Chat.swift` |
 | Tool call processing | `Libraries/MLXLMCommon/Tool/ToolCallFormat.swift` |
@@ -62,20 +66,16 @@ Embedders       - Embedding models (BGE, Nomic, MiniLM)
 import MLXLLM
 import MLXLMCommon
 
-// Load model (downloads from HuggingFace automatically)
 let modelContainer = try await LLMModelFactory.shared.loadContainer(
     configuration: .init(id: "mlx-community/Qwen3-4B-4bit")
 )
 
-// Create chat session
 let session = ChatSession(modelContainer)
 
-// Single response
 let response = try await session.respond(to: "What is Swift?")
 print(response)
 
-// Streaming response
-for try await chunk in session.streamResponse(to: "Explain concurrency") {
+for try await chunk in session.streamResponse(to: "Explain structured concurrency") {
     print(chunk, terminator: "")
 }
 ```
@@ -91,13 +91,12 @@ let modelContainer = try await VLMModelFactory.shared.loadContainer(
 )
 
 let session = ChatSession(modelContainer)
-
-// With image (video is also an optional parameter)
 let image = UserInput.Image.url(imageURL)
+
 let response = try await session.respond(
     to: "Describe this image",
     image: image,
-    video: nil  // Optional video parameter
+    video: nil
 )
 ```
 
@@ -106,7 +105,6 @@ let response = try await session.respond(
 ```swift
 import Embedders
 
-// Note: Embedders uses loadModelContainer() helper (not a factory pattern)
 let container = try await loadModelContainer(
     configuration: ModelConfiguration(id: "mlx-community/bge-small-en-v1.5-mlx")
 )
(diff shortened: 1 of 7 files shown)


Libraries/MLXLMCommon/Evaluate+Token.swift
```diff
@@ -0,0 +1,198 @@
...
+/// without detokenization or tool-call parsing.
+///
+/// - Parameters:
+///   - promptTokenCount: number of tokens in the prompt
+///   - modelConfiguration: model configuration (for EOS/extra EOS tokens)
+///   - tokenizer: tokenizer (for EOS id and unknown token id)
+///   - iterator: token iterator
+///   - includeStopToken: when true, the terminating EOS/unknown token is yielded before finishing
+/// - Returns: An `AsyncStream` that emits token IDs and a final `.info`, plus a `Task`.
+public func generateTokenTask(
```
davidkoski: I am a little concerned about having two copies of this -- it isn't just packaging up the arguments, this is the core of the logic.  That means any changes to the core have to be applied in two places (e.g. the new wired memory PR #72 or any bug fixes).  Risky.

What if we made this:

```swift
public func generateTask(
    promptTokenCount: Int,
    tokenizer: Tokenizer,
    iterator: consuming TokenIterator
```

into:

```swift
func generateTask<Output>(
    promptTokenCount: Int,
    tokenizer: Tokenizer,
    iterator: consuming TokenIterator,
    transform: (Int) -> Output
) -> (AsyncStream<Output>, Task<Void, Never>) {
```

Then the current public API can cover that and this new can use it as well.  It isn't quite as easy as I state here because there is some state that is owned by the `transform`:

```swift
        var detokenizer = NaiveStreamingDetokenizer(tokenizer: tokenizer)
        let toolCallProcessor = ToolCallProcessor(
            format: modelConfiguration.toolCallFormat ?? .json
        )
```

What do you think?  If you are interested in trying this approach, go for it.  I can also try it and generate a patch.

I _think_ this will be better but until we see it can't say for sure.

ronaldmannak: That's an excellent point. Instead of the transform closure we could do something like so to make it concurrency friendly:

```swift
private protocol TokenLoopHandler {
    associatedtype Output

    /// Value used for GenerateCompletionInfo.generationTokenCount
    var generationTokenCount: Int { get }

    /// Return false to stop the loop early.
    mutating func onToken(
        _ token: Int,
        emit: (Output) -> AsyncStream<Output>.Continuation.YieldResult
    ) -> Bool

    /// Called only when includeStopToken == true and a stop token was hit.
    mutating func onStopToken(
        _ token: Int,
        emit: (Output) -> AsyncStream<Output>.Continuation.YieldResult
    ) -> Bool

    func infoEvent(_ info: GenerateCompletionInfo) -> Output
}

private func generateLoopTask<H: TokenLoopHandler>(
    promptTokenCount: Int,
    tokenizer: Tokenizer,
    iterator: consuming TokenIterator,
    includeStopToken: Bool = false,
    handler: consuming H
) -> (AsyncStream<H.Output>, Task<Void, Never>) {
    let (stream, continuation) = AsyncStream<H.Output>.makeStream()
    let iterator = SendableBox(iterator)
    let handler = SendableBox(handler)

    let task = Task {
        let iterator = iterator.consume()
        var handler = handler.consume()

        var start = Date.timeIntervalSinceReferenceDate
        var promptTime: TimeInterval = 0

        var eosTokenIds = modelConfiguration.eosTokenIds
        if let tokenizerEos = tokenizer.eosTokenId { eosTokenIds.insert(tokenizerEos) }
        for token in modelConfiguration.extraEOSTokens {
            if let id = tokenizer.convertTokenToId(token) { eosTokenIds.insert(id) }
        }

        for token in iterator {
            if Task.isCancelled { break }

            if promptTime == 0 {
                let now = Date.timeIntervalSinceReferenceDate
                promptTime = now - start
                start = now
            }

            let isStop = token == tokenizer.unknownTokenId || eosTokenIds.contains(token)
            if isStop {
                if includeStopToken,
                   !handler.onStopToken(token, emit: continuation.yield) { break }
                break
            }

            if !handler.onToken(token, emit: continuation.yield) { break }
        }

        let now = Date.timeIntervalSinceReferenceDate
        let info = GenerateCompletionInfo(
            promptTokenCount: promptTokenCount,
            generationTokenCount: handler.generationTokenCount,
            promptTime: promptTime + iterator.promptPrefillTime,
            generationTime: now - start
        )

        _ = continuation.yield(handler.infoEvent(info))
        Stream().synchronize()
        continuation.finish()
    }

    continuation.onTermination = { _ in task.cancel() }
    return (stream, task)
}
```

plus two handlers:

```swift
private struct TextToolHandler: TokenLoopHandler {
    typealias Output = Generation

    var generationTokenCount = 0
    var detokenizer: NaiveStreamingDetokenizer
    let toolCallProcessor: ToolCallProcessor

    init(tokenizer: Tokenizer, format: ToolCallFormat) {
        self.detokenizer = NaiveStreamingDetokenizer(tokenizer: tokenizer)
        self.toolCallProcessor = ToolCallProcessor(format: format)
    }

    mutating func onToken(_ token: Int, emit: (Generation) -> AsyncStream<Generation>.Continuation.YieldResult) -> Bool {
        generationTokenCount += 1
        detokenizer.append(token: token)

        if let chunk = detokenizer.next(),
           let text = toolCallProcessor.processChunk(chunk),
           case .terminated = emit(.chunk(text)) { return false }

        if let call = toolCallProcessor.toolCalls.popLast(),
           case .terminated = emit(.toolCall(call)) { return false }

        return true
    }

    mutating func onStopToken(_ token: Int, emit: (Generation) -> AsyncStream<Generation>.Continuation.YieldResult) -> Bool {
        true
    }

    func infoEvent(_ info: GenerateCompletionInfo) -> Generation { .info(info) }
}

  private struct RawTokenHandler: TokenLoopHandler {
      typealias Output = TokenGeneration
      var generationTokenCount = 0

      mutating func onToken(
      _ token: Int,
      emit: (TokenGeneration) -> AsyncStream<TokenGeneration>.Continuation.YieldResult
  ) -> Bool {
      generationTokenCount += 1
      if case .terminated = emit(.token(token)) {
          return false
      }
      return true
  }

  mutating func onStopToken(
      _ token: Int,
      emit: (TokenGeneration) -> AsyncStream<TokenGeneration>.Continuation.YieldResult
  ) -> Bool {
      generationTokenCount += 1
      if case .terminated = emit(.token(token)) {
          return false
      }
      return true
  }


    func infoEvent(_ info: GenerateCompletionInfo) -> TokenGeneration { .info(info) }
}
```
and we'll keep the two public methods:
```swift
public func generateTask(...) -> (AsyncStream<Generation>, Task<Void, Never>) {
    generateLoopTask(..., handler: TextToolHandler(tokenizer: tokenizer, format: modelConfiguration.toolCallFormat ?? .json))
}

public func generateTokenTask(..., includeStopToken: Bool = false) -> (AsyncStream<TokenGeneration>, Task<Void, Never>) {
    generateLoopTask(..., includeStopToken: includeStopToken, handler: RawTokenHandler())
}
```
(this code also solves a smaller issue that tokens get undercounted in the current implementation when chunks consist of multiple tokens)

Let me know what you think of this approach.

davidkoski: I like it!  I agree the protocol is a better fit here and allows for nice lifecycle handling.

ronaldmannak: Great! The blast radius was a bit bigger than I anticipated, but both paths work.

Libraries/MLXLMCommon/Evaluate.swift
```diff
@@ -610,54 +688,19 @@ public func generate(
     iterator: TokenIterator,
     didGenerate: ([Int]) -> GenerateDisposition
 ) -> GenerateResult {
-    var start = Date.timeIntervalSinceReferenceDate
-    var promptTime: TimeInterval = 0
```
davidkoski, line 616: Nice, I noticed we had a couple versions of the sync code when reviewing #72.  Given that it is all deprecated an even stronger reason not to have multiples :-)

Libraries/MLXLMCommon/Evaluate.swift
```diff
@@ -850,37 +859,132 @@ public func generateTask(
...
+/// without detokenization or tool-call parsing.
+///
+/// - Parameters:
+///   - promptTokenCount: number of tokens in the prompt
+///   - modelConfiguration: model configuration (for EOS/extra EOS tokens)
+///   - tokenizer: tokenizer (for EOS id and unknown token id)
+///   - iterator: token iterator
+///   - includeStopToken: when true, the terminating EOS/unknown token is yielded before finishing
+/// - Returns: An `AsyncStream` that emits token IDs and a final `.info`, plus a `Task`.
+public func generateTokenTask(
```
davidkoski, line 970: This (the three versions) makes me realize we have a bit of a mess with the many variants of token generation :-)

I think this is the right approach -- it parallels the text version.

Libraries/MLXLMCommon/Evaluate.swift
```diff
@@ -892,27 +996,20 @@ public func generateTask(
...
-            detokenizer.append(token: token)
-            if let chunk = detokenizer.next() {
-                tokenCount += 1
-
-                // Process chunk through the tool call processor
-                if let textToYield = toolCallProcessor.processChunk(chunk) {
-                    if case .terminated = continuation.yield(.chunk(textToYield)) {
+            // Check for end-of-sequence tokens
+            if token == tokenizer.unknownTokenId || stopTokenIDs.contains(token) {
+                if includeStopToken {
```
davidkoski: Should this be handled here or in the `handler`?  It did generate the stop token (whether we use it or not) so the count including it makes sense.  Perhaps the `onStopToken` should determine what to do with it?

Libraries/MLXLMCommon/Evaluate.swift
```diff
@@ -1038,9 +1144,130 @@ public enum Generation: Sendable {
...
+    mutating func onToken(
+        _ token: Int,
+        emit: (sending Output) -> AsyncStream<Output>.Continuation.YieldResult
+    ) -> Bool
+
+    /// Called only when includeStopToken == true and a stop token was hit.
+    mutating func onStopToken(
+        _ token: Int,
+        emit: (sending Output) -> AsyncStream<Output>.Continuation.YieldResult
+    ) -> Bool
```
davidkoski, line 1271: Does this need a `didStop` handler?  E.g. if we hit the maxToken limit or the stream was cancelled -- it would never observe a stop token but maybe the handler needs to observe that?

Or is that covered with `infoEvent` (that is always called at the end).

davidkoski:
It looks like this needs a run of `swift-format` and now that #72 is merged a rebase on that.  I can help with that merge if needed,  I think the same code applies and just fits inside the `performIteration` closure.

Other than that I think this is ready to go!

Libraries/MLXLMCommon/Evaluate.swift
ronaldmannak: I think you want to keep this in the shared loop so `generationTokenCount` matches emitted stream tokens so, for example, EOS is not emitted and not counted if `includeStopToken` is `false`. What did you have in mind with the `onStopToken`?

Libraries/MLXLMCommon/Evaluate.swift
ronaldmannak: I've added a `stopReason` with a subset of the values the OpenAI API uses. Since this is the only layer where we can distinguish between EOS and hitting maxTokens it should be here and it's probably the most convenient way for devs. Does this work for you?

ronaldmannak:
@davidkoski  done!

Libraries/MLXLMCommon/Evaluate.swift
davidkoski: Just wondering I guess on what side of the call the logic on handling the stop token should live.  In particular, if the `includeStopToken` is false the `onStopToken()` will not be called -- should it (to finish the lifecycle)?

To some extent it doesn't matter -- this is a private protocol and the implementation only needs to be consistent within this file.  If we decide later that it does need to observe it then no problem, we can change it.

Libraries/MLXLMCommon/Evaluate.swift
davidkoski: OK, I see the stopReason in GenerateCompletionInfo.  I think this is good and like I mentioned in the comment earlier in this file I realized this is a private protocol -- we don't need to plan ahead too much for future expansion as we can just change the implementation if needed.

PR #89, opened 2026-02-01
Update GLM4MoELite attention to use embed_q and unembed_out

KingboardMa:
Replace kv_b_proj with embed_q and unembed_out MultiLinear layers to match the official mlx-lm implementation and support GLM-4.7-Flash model weights.

Changes:
- Add MultiLinear and QuantizedMultiLinear classes for per-head linear projections
- Replace kv_b_proj with embed_q (projects q_nope to kv_lora_rank) and unembed_out (projects attention output from kv_lora_rank to v_head_dim)
- Update attention computation to use latent KV representation
- Derive values from keys by slicing off rope dimensions

## Proposed changes

## Checklist

### DIFF
--- a/Libraries/MLXLMCommon/Tool/Parsers/PythonicToolCallParser.swift
+++ b/Libraries/MLXLMCommon/Tool/Parsers/PythonicToolCallParser.swift
@@ -1,0 +1,100 @@
+// Copyright © 2025 Apple Inc.
+
+import Foundation
+
+/// Parser for Pythonic tool call format: [function_name(arg1='value1', arg2='value2')]
+/// Used by LFM2.5 and similar models that output tool calls in Python function call syntax.
+/// Reference: LiquidAI LFM2.5 chat template format
+public struct PythonicToolCallParser: ToolCallParser, Sendable {
+    public let startTag: String?
+    public let endTag: String?
+
+    public init(startTag: String? = nil, endTag: String? = nil) {
+        self.startTag = startTag
+        self.endTag = endTag
+    }
+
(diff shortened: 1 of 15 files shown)


Libraries/MLXLLM/Models/GLM4MOELite.swift
```diff
@@ -11,6 +11,117 @@ import MLX
 import MLXLMCommon
 import MLXNN
 
+// MARK: - MultiLinear
+
+/// Multi-head linear layer where each head has its own weight matrix
+public class MultiLinear: Module, Quantizable {
```
davidkoski, line 17: I think these classes should probably not be `public` (unless they need to be).  They should also get the `GLM4` prefix if they are not `private`.

Libraries/MLXLLM/Models/GLM4MOELite.swift
```diff
@@ -33,7 +144,8 @@ class GLM4MoELiteAttention: Module {
     @ModuleInfo(key: "o_proj") var oProj: Linear
     @ModuleInfo(key: "kv_a_proj_with_mqa") var kvAProjWithMqa: Linear
     @ModuleInfo(key: "kv_a_layernorm") var kvALayerNorm: RMSNorm
-    @ModuleInfo(key: "kv_b_proj") var kvBProj: Linear
```
davidkoski, line 36: So these keys are not going to be present?  Do we need to handle either set?

davidkoski:
Oh, I think this is essentially the same as #73 -- closing in favor of that one.

PR #91, opened 2026-02-01
LFM2.5 outputs [func(arg='value')] Pythonic syntax, not JSON

tpae:
## Proposed changes

I was trying hard to debug why LFM2.5 still didn't call the tools correctly. After studying this: https://github.com/Liquid4All/cookbook/tree/main/examples/flight-search-assistant

It turns out LFM2 uses Pythonic syntax, not JSON.

I was able to verify this to be fully working now.

## Checklist


tpae:
cc @davidkoski @awni 

Looks like this fix should be added to `mlx-lm` also.

davidkoski (approved):

PR #92, opened 2026-02-02
Add skill.md

ronaldmannak:
## Proposed changes

Add `skill.md` for tools like Claude Code, Codex, and Droid describing workflows for working with MLX-Swift-LM

## Checklist


skills/mlx-swift-lm/references/concurrency.md
```diff
@@ -0,0 +1,345 @@
...
+
+| Type | Purpose |
+|------|---------|
+| `SerialAccessContainer<T>` | Exclusive async access to wrapped state |
+| `AsyncMutex` | Lock that works with async blocks |
+| `SendableBox<T>` | Transfer non-Sendable values across isolation |
+| `ModelContainer` | Thread-safe model wrapper (uses SerialAccessContainer) |
+| `ChatSession` | NOT thread-safe (single task only) |
+
+## SerialAccessContainer
```
davidkoski, line 19: This makes me curious -- are the skills for developers working _in_ mlx-swift-lm or _using_ mlx-swift-lm? This information is mostly for working in the libraries as these types are not publicly exposed.  Used of the libraries shouldn't have to worry about this.

The information is really good and useful for people working in the library.  I don't think it exists anywhere else (except maybe comments in the implementation).

I don't think we need to change anything, just an observation -- a curiosity.

ronaldmannak: Actually, I meant it to cover both use-cases, as many contributors are adding models and improving mlx-swift-lm itself. What I can do is explicitly mention when info is internal mlx-swift-lm as opposed to relevant for building on top of it. That said, if you think it's okay now, we can always wait for more feedback. I'm sure there are many things that we can improve in terms of a better fit for LLMs or readability.

davidkoski: Perhaps the llms will pick and choose the right files anyway depending on what you are doing.  I think it is ok as-is, just a thing I hadn't considered (and that is true of the human-facing docs as well, which since they document only public API are largely excluding the internal developers).

davidkoski (approved):
Changes look good, thank you for this contribution -- excited to see how this helps people!

PR #95, opened 2026-02-07
Remove consuming keyword from LoRAContainer.init parameter

### LINKED ISSUE #94: [BUG] LoRAContainer init fails to compile on Swift 6.1 — consuming parameter used twice
**Describe the bug**

The `LoRAContainer.init` marks its `parameters` argument as `consuming`, but then uses it twice — once passed to `eval()` and once assigned to `self.parameters`. Swift 6.1 correctly diagnoses this as a double consumption.

https://github.com/ml-explore/mlx-swift-lm/blob/f032515ca4b81ce8f689b138926e4ff825e48f17/Libraries/MLXLMCommon/Adapters/LoRA/LoRAContainer.swift#L84-L93

This regression was introduced in f032515.

**To Reproduce**

Run `swift build` with Swift 6.1 on `main`.
 
**Expected behavior**

 - OS Version: macOS 26.2 (25C56)
 - Device: Mac Studio (Mac16,9)
 - Version 2.30.3

First encountered in [this GitHubActions run](https://github.com/mattt/AnyLanguageModel/actions/runs/21780704423/job/62844214349) for https://github.com/mattt/AnyLanguageModel/pull/116

mattt:
## Proposed changes

Fixes #94 

This PR removes `consuming` from `LoRAContainer.init`’s `parameters` argument. 

The initializer needs to `eval` the parameters to force MLX arrays to be fully evaluated, and then store them. Marking the argument as `consuming` makes that a double-consume under Swift 6.1 ownership checking. (I'm not sure why this doesn't appear to be a problem in Swift 6.2)

Dropping `consuming` preserves the original intent while restoring correct semantics.

## Checklist

### DIFF
--- a/Libraries/MLXLMCommon/Adapters/LoRA/LoRAContainer.swift
+++ b/Libraries/MLXLMCommon/Adapters/LoRA/LoRAContainer.swift
@@ -83,7 +83,7 @@ public struct LoRAContainer: ModelAdapter, @unchecked Sendable {
 
     public init(
         configuration: LoRAConfiguration,
-        parameters: consuming ModuleParameters
+        parameters: ModuleParameters
     ) {
         // ensure that the parameters are fully evaluated before we promise
         // that they are Sendable


Libraries/MLXLMCommon/Adapters/LoRA/LoRAContainer.swift
atdrendel, line 93: @mattt This is a fine fix, but, personally, I think keeping the original ownership transfer intension and just moving the `eval(parameters)` down would be better. We'd still be able to ensure `LoRAContainer` maintains its `Sendable` requirements while also preventing the caller from re-using the parameters later on, which I guess was the reason for adding `consuming` to the `parameters` parameter.

Here's the change I made in my [fork](https://github.com/shareup/mlx-swift-lm):

```swift
    public init(
        configuration: LoRAConfiguration,
        parameters: consuming ModuleParameters
    ) {
        self.parameters = parameters
        // ensure that the parameters are fully evaluated before we promise
        // that they are Sendable
        eval(self.parameters)
    }
```

davidkoski: Yeah, I think this sounds right.  My thought when adding the `consuming` was I wanted to transfer ownership of the `MLXArray` into the container.  MLXArray is not `Sendable` exactly (at all, mostly, but ... 😅  ), but if we are careful it can be used that way (and it is _really_ useful to do so).

If the caller were careful it be ok to not `consume` the parameters, but I am not sure there is a use case where we need that (other than we like it to compile without errors, which is the reason for the PR of course).  Since we can't count on the caller to be careful I think it is better to be `consuming`.  If reordering the use fixes the issue I think that is the way to go.

PR #100, opened 2026-02-10
switch to current mlx-swift

davidkoski:
- incorporate change from #95 (fixing #94)

## Proposed changes

## Checklist

### DIFF
--- a/Libraries/MLXLMCommon/Adapters/LoRA/LoRAContainer.swift
+++ b/Libraries/MLXLMCommon/Adapters/LoRA/LoRAContainer.swift
@@ -85,11 +85,11 @@ public struct LoRAContainer: ModelAdapter, @unchecked Sendable {
         configuration: LoRAConfiguration,
         parameters: consuming ModuleParameters
     ) {
-        // ensure that the parameters are fully evaluated before we promise
-        // that they are Sendable
-        eval(parameters)
         self.configuration = configuration
         self.parameters = parameters
+        // ensure that the parameters are fully evaluated before we promise
+        // that they are Sendable
+        eval(self.parameters)
     }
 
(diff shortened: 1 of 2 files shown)


Libraries/MLXLMCommon/Adapters/LoRA/LoRAContainer.swift
```diff
@@ -85,11 +85,11 @@ public struct LoRAContainer: ModelAdapter, @unchecked Sendable {
...
         parameters: consuming ModuleParameters
     ) {
-        // ensure that the parameters are fully evaluated before we promise
-        // that they are Sendable
-        eval(parameters)
         self.configuration = configuration
         self.parameters = parameters
+        // ensure that the parameters are fully evaluated before we promise
+        // that they are Sendable
+        eval(self.parameters)
```
davidkoski, line 92: @mattt @atdrendel -- per discussion in #95 I think this is what we want or feel free to update #95 and I will take that.  Thanks!

Package.swift
```diff
@@ -26,7 +26,7 @@ let package = Package(
             targets: ["MLXEmbedders"]),
     ],
     dependencies: [
-        .package(url: "https://github.com/ml-explore/mlx-swift", .upToNextMinor(from: "0.30.3")),
+        .package(url: "https://github.com/ml-explore/mlx-swift", .upToNextMinor(from: "0.30.6")),
```
davidkoski, line 29: Use the new mlx-swift.

Libraries/MLXLMCommon/Adapters/LoRA/LoRAContainer.swift
atdrendel: @davidkoski Yeah, that looks right to me. Passing `self.parameters` to `eval()` keeps the `Sendable` conformance and removes the ownership error. Thanks for doing that.

PR #103, opened 2026-02-12
Fix MLXEmbedders README.md example

CodebyCR:
* move '.eval()' call in own line to get expected result
* adjust README code for Test case
* add MLXEmbedders directory to Test Target
* add test case for README example

## Proposed changes

## Checklist


Tests/MLXLMTests/EmbedderExampleTests.swift
```diff
@@ -0,0 +1,92 @@
+// Copyright © 2026 Apple Inc.
+
+import Foundation
+import MLX
+import MLXEmbedders
+import Testing
+import Tokenizers
+
+struct EmbedderExampleTests {
```
davidkoski: Since this requires a download it should go in `MLXLMIntegrationTests`.  These aren't run in CI but are useful for developers.

Look at `EvalTests` -- would something like that (random weights) be interesting?

CodebyCR: You are right this is a better evaluation approach.

PR #105, opened 2026-02-12
ensure models have links back to where they were ported from

davidkoski:
## Proposed changes

- fix #59
- adds links mostly and corrects a couple

## Checklist

Put an `x` in the boxes that apply.

- [x] I have read the [CONTRIBUTING](https://github.com/ml-explore/mlx/blob/main/CONTRIBUTING.md) document
- [x] I have run `pre-commit run --all-files` to format my code / installed pre-commit prior to committing changes
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] I have updated the necessary documentation (if needed)

PR #106, opened 2026-02-12
use ReerCodable macro to allow for default values

davidkoski:
## Proposed changes

Looking at moving to a Codable macro that makes it easier to integrate with JSON configuration files.  Specifically:

- easy default values
- easy to override json keys

These are both possible with standard Swift Coding but are unwieldy in the quantity we need here.   See:

- https://github.com/ml-explore/mlx-swift-lm/blob/main/Libraries/MLXLMCommon/Documentation.docc/porting.md#configuration

For an example.  This changes:

```swift
public struct YourModelConfiguration: Codable, Sendable {
    public let hiddenSize: Int

    // use this pattern for values that need defaults
    public let _layerNormEps: Float?
    public var layerNormEps: Float { _layerNormEps ?? 1e-6 }

    enum CodingKeys: String, CodingKey {
        case hiddenSize = "hidden_size"
        case _layerNormEps = "layer_norm_eps"
    }
}
```

to this:

```swift
@Codable
public struct YourModelConfiguration: Sendable {
    @CodingKey("hidden_size") public var hiddenSize: Int
    @CodingKey("layer_norm_eps") public var layerNormEps: Float = 1e-6
}
```

The downside is that we are using macros (there is some additional build cost).

There was a previous attempt here: https://github.com/ml-explore/mlx-swift-examples/pull/316. This is roughly the same approach but doesn't attempt to port all the configuration at once.  Keeping that up-to-date was too much.  Once this is in we can migrate the configuration in chunks.

## Checklist

### DIFF
--- a/Libraries/MLXLLM/Models/Cohere.swift
+++ b/Libraries/MLXLLM/Models/Cohere.swift
@@ -2,6 +2,7 @@ import Foundation
 import MLX
 import MLXLMCommon
 import MLXNN
+import ReerCodable
 
 // port of https://github.com/ml-explore/mlx-lm/blob/main/mlx_lm/models/cohere.py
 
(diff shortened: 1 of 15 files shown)


Libraries/MLXLLM/Models/Cohere.swift
```diff
@@ -2,8 +2,9 @@ import Foundation
 import MLX
 import MLXLMCommon
 import MLXNN
+import ReerCodable
 
-// port of https://github.com/ml-explore/mlx-examples/blob/main/llms/mlx_lm/models/cohere.py
+// port of https://github.com/ml-explore/mlx-lm/tree/main/mlx_lm/models/cohere.py
```
davidkoski: Once #105 is in, this should be a NOP.

Libraries/MLXLLM/Models/Llama.swift
```diff
@@ -340,108 +341,37 @@ public class LlamaModel: Module, LLMModel, KVCacheDimensionProvider {
...
-        {
-            self.tieWordEmbeddings = tieWordEmbeddings
-        }
-        if let attentionBias = try container.decodeIfPresent(Bool.self, forKey: .attentionBias) {
-            self.attentionBias = attentionBias
-        }
-        if let mlpBias = try container.decodeIfPresent(Bool.self, forKey: .mlpBias) {
-            self.mlpBias = mlpBias
-        }
+    public func didDecode(from decoder: any Decoder) throws {
```
davidkoski, line 367: This is an interesting one because it has some logic to post-process the configuration.

angeloskath (approved):
Just 🚀

JaeminKim-amoz (changes requested):
**REQUEST CHANGES — needs discussion on migration strategy**

Compelling boilerplate reduction (-383 lines). The `@Codable`/`@CodingKey` macro usage is clean, and `LlamaConfiguration.didDecode` preserves rope_scaling validation correctly.

**Issues to discuss:**
- [MEDIUM] `-skipMacroValidation` in CI bypasses macro sandboxing — acceptable for a trusted dependency, but worth documenting the tradeoff.
- [MEDIUM] Incremental migration (only 3 LLM + 4 VLM configs) means two Codable patterns coexist — contributors may be confused. Consider a full migration or clearer docs on which pattern to use for new models.
- [LOW] `Gemma2Configuration` changed from explicit `Codable` to implicit (via macro) + `Sendable` — this is correct but implicit.

davidkoski:
Not going to take this as-is -- the idea is good but I am concerned about supply chain issues around macros.  I may revisit this by embedding a minimal macro package that would support this.  Closing for now.

PR #113, opened 2026-02-18
fix Sendable issues, unused code, deprecation warnings

davidkoski:
## Proposed changes

fix Sendable issues, unused code, deprecation warnings -- just some warnings that have crept back in.

## Checklist

### DIFF
--- a/Libraries/MLXEmbedders/Load.swift
+++ b/Libraries/MLXEmbedders/Load.swift
@@ -109,11 +109,6 @@ func prepareModelDirectory(
     }
 }
 
-/// Default instance of HubApi to use.  This is configured to save downloads into the caches directory.
-public let defaultHubApi: HubApi = {
-    HubApi(downloadBase: FileManager.default.urls(for: .cachesDirectory, in: .userDomainMask).first)
-}()
-
 /// Asynchronously loads the `EmbeddingModel` and its associated `Tokenizer`.
 ///
 /// This is the primary high-level function for initializing an embedding pipeline.
--- a/Libraries/MLXLMCommon/ModelContainer.swift
+++ b/Libraries/MLXLMCommon/ModelContainer.swift
@@ -112,7 +112,7 @@ public final class ModelContainer: Sendable {
     /// Callers _must_ eval any `MLXArray` before returning as
     /// `MLXArray` is not `Sendable`.
     public func perform<V, R: Sendable>(
-        sending values: consuming V, _ action: @Sendable (ModelContext, V) async throws -> R
+        nonSendable values: consuming V, _ action: @Sendable (ModelContext, V) async throws -> R
     ) async rethrows -> sending R {
         let values = SendableBox(values)
         return try await context.read {
--- a/Libraries/MLXLMCommon/Tool/Parsers/GemmaFunctionParser.swift
+++ b/Libraries/MLXLMCommon/Tool/Parsers/GemmaFunctionParser.swift
@@ -69,7 +69,7 @@ public struct GemmaFunctionParser: ToolCallParser, Sendable {
                 ? String(argsStr[argsStr.index(after: commaIdx)...]) : ""
 
             // Try JSON decode, fallback to string
-            arguments[key] = deserialize(value) ?? value
+            arguments[key] = tryParseJSON(value) ?? value
         }
 
         return ToolCall(function: .init(name: funcName, arguments: arguments))
--- a/Libraries/MLXLMCommon/Tool/Parsers/ParserUtilities.swift
+++ b/Libraries/MLXLMCommon/Tool/Parsers/ParserUtilities.swift
@@ -4,25 +4,26 @@ import Foundation
 
 // MARK: - Basic Deserialization
 
-private func toSendable(_ value: Any) -> (any Sendable)? {
+private func asSendable(_ value: Any) -> (any Sendable)? {
     switch value {
     case let dict as [String: Any]:
-        return dict.compactMapValues { toSendable($0) }
+        return dict.compactMapValues { asSendable($0) }
     case let array as [Any]:
-        return array.compactMap { toSendable($0) }
+        return array.compactMap { asSendable($0) }
     case let sendable as any Sendable:
         return sendable
     default:
         return nil
     }
 }
+
 /// Deserialize a string value to JSON or return as string.
 ///
 /// Attempts JSON parsing first, falling back to the original string value.
 /// Reference: Python's `ast.literal_eval` / `json.loads` pattern
-func deserialize(_ value: String) -> (any Sendable)? {
+func tryParseJSON(_ value: String) -> (any Sendable)? {
     if let data = value.data(using: .utf8) {
-        return try? toSendable(JSONSerialization.jsonObject(with: data))
+        return try? asSendable(JSONSerialization.jsonObject(with: data))
     }
     return nil
 }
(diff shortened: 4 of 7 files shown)


Libraries/MLXLLM/Models/Apertus.swift
```diff
@@ -411,7 +411,7 @@ private class ApertusModelInner: Module {
     ) -> MLXArray {
         var h = embedTokens(inputs)
 
-        let mask = createAttentionMask(h: inputs, cache: cache)
+        let mask = createAttentionMask(h: inputs, cache: cache?.first)
```
davidkoski, line 416: Use the newer version of the method -- this pattern matches all the other models.

Libraries/MLXLMCommon/Tool/Parsers/GemmaFunctionParser.swift
```diff
@@ -69,13 +69,7 @@ public struct GemmaFunctionParser: ToolCallParser, Sendable {
...
 
             // Try JSON decode, fallback to string
-            if let data = value.data(using: .utf8),
-                let json = try? JSONSerialization.jsonObject(with: data)
-            {
-                arguments[key] = json
-            } else {
-                arguments[key] = value
-            }
+            arguments[key] = deserialize(value)
```
davidkoski: `jsonObject() -> Any` and we need `any Sendable`.  The `deserialize()` function does this string -> data -> recast as `any Sendable` dance, so use that.

This pattern is repeated several times.

Libraries/MLXLMCommon/Tool/Parsers/ParserUtilities.swift
```diff
@@ -4,17 +4,27 @@ import Foundation
 
 // MARK: - Basic Deserialization
 
+private func toSendable(_ value: Any) -> (any Sendable)? {
```
davidkoski: How to convert `Any` to a value that is `any Sendable` (from a JSON).

Libraries/MLXLMCommon/ModelContainer.swift
```diff
@@ -108,6 +108,18 @@ public final class ModelContainer: Sendable {
         }
     }
 
+    /// Perform an action on the ``ModelContext`` with additional (non `Sendable`) context values.
+    /// Callers _must_ eval any `MLXArray` before returning as
+    /// `MLXArray` is not `Sendable`.
+    public func perform<V, R: Sendable>(
+        sending values: consuming V, _ action: @Sendable (ModelContext, V) async throws -> R
```
davidkoski: Add a variant that consumes, wraps V and sends it to the inner async context.

Libraries/MLXVLM/Models/LFM2VL.swift
```diff
@@ -987,7 +987,7 @@ public class LFM2VL: Module, VLMModel, KVCacheDimensionProvider {
         var spatialShapes: MLXArray? = nil
         var pixelAttentionMask: MLXArray? = nil
 
-        if let pixels = pixelValues, let frames = input.image?.frames, !frames.isEmpty {
+        if pixelValues != nil, let frames = input.image?.frames, !frames.isEmpty {
```
davidkoski, line 990: Not used

Tests/MLXLMIntegrationTests/ToolCallIntegrationTests.swift
```diff
@@ -203,10 +203,7 @@ public class ToolCallIntegrationTests: XCTestCase {
         input: UserInput,
         maxTokens: Int
     ) async throws -> (text: String, toolCalls: [ToolCall]) {
-        var collectedText = ""
-        var collectedToolCalls: [ToolCall] = []
-
-        let result = try await container.perform { (context: ModelContext) in
+        let result = try await container.perform(sending: input) { (context: ModelContext, input) in
```
davidkoski: Use the new `sending` variant.

Libraries/MLXEmbedders/Load.swift
```diff
@@ -108,6 +108,11 @@ func prepareModelDirectory(
     }
 }
 
+/// Default instance of HubApi to use.  This is configured to save downloads into the caches directory.
+public let defaultHubApi: HubApi = {
```
davidkoski: Make this work like the LLM hub uses -- by default download to the .caches directory, otherwise it ends up in ~/Documents which bloats backups (and makes it inaccessible to unit tests!)

Tests/MLXLMIntegrationTests/EmbedderIntegrationTests.swift
```diff
@@ -76,7 +76,7 @@ struct EmbedderIntegrationtests {
 
         for (index, resultSimilarity) in similarities.enumerated() {
             #expect(
-                resultSimilarity.isEqual(to: expectedSimilarities[index]),
+                abs(resultSimilarity - expectedSimilarities[index]) < 0.01,
```
davidkoski, line 79: Needed a threshold compare

Libraries/MLXEmbedders/Load.swift
angeloskath: There may be a valid reason to avoid this but I think it would be neat to download in `~/.cache/` instead of `~/Library/Caches`.

By using `~/.cache/` then the downloaded models are shared between python and swift.

Tests/MLXLMIntegrationTests/EmbedderIntegrationTests.swift
angeloskath: nit: the comment should probably be one line up

ronaldmannak: @angeloskath Note that in sandboxed apps, that URL will point to a sandboxed location, not to where Python models are actually cached

Libraries/MLXEmbedders/Load.swift
davidkoski: The format is different:

```
ls ~/.cache/huggingface/hub/models--Qwen--Qwen2.5-VL-3B-Instruct 

blobs     refs      snapshots
```

vs

```
ls ~/Downloads/huggingface/models/Qwen/Qwen3-VL-8B-Instruct 

chat_template.json               model-00002-of-00004.safetensors preprocessor_config.json         vocab.json
config.json                      model-00003-of-00004.safetensors tokenizer_config.json
generation_config.json           model-00004-of-00004.safetensors tokenizer.json
model-00001-of-00004.safetensors model.safetensors.index.json     video_preprocessor_config.json
```

I agree that would be better if we could share them.

Using ~/Downloads has two things that we want:

- it isn't backed up with the application, e.g. on iOS -- we don't need to back up models because we can just download them again (and ~/.cache in a sandboxed application would be)
- sandbox-wise it is an area that an app can be granted access to, not so with ~/.cache
    - to be clear in the sandbox if you use ~/.cache it will be a non-shared directory inside the container.  you can get use the actual home directory but sandboxes apps won't have access to it (sandbox-wise)

Tests/MLXLMIntegrationTests/EmbedderIntegrationTests.swift
davidkoski: > nit: the comment should probably be one line up

Good spot, moved.

angeloskath: @ronaldmannak yeah indeed I read that which is why I wrote it in this weird way. Basically what I meant is, how awesome if it would default to `~/.cache` when not running in the sandbox. Probably pointless but 🤷‍♂️

ronaldmannak: Got it :) It isn't pointless for non-sandboxed apps (but I don't know how many of those still exist)

PR #114, opened 2026-02-19
add optional toolCall dispatch and tool output injection

### LINKED ISSUE #112: [BUG] ChatSession supports tools but doesn't have a good way to hook them up
**Describe the bug**

ChatSession supports tools but doesn't have a good way to hook them up

**To Reproduce**

In #107 tool spec support was added to ChatSession and this works, but it effectively stops when the tool call is made -- there is no mechanism to hook it to the tool or to supply the tool output back into the session.

We need a way to provide that.

Note: from the integration tests you _can_ of course do it if you examine the chunks, but not if you are consuming the text:

```swift
        for try await generation in session.streamDetails(
            to: "What is the weather in San Francisco?",
            images: [],
            videos: []
        ) {
            switch generation {
            case .chunk(let text):
                responseText += text
            case .toolCall(let toolCall):
                toolCalls.append(toolCall)
            case .info:
                break
            }
        }
```

There is no way to inject the `.tool()` response back into the stream.

davidkoski:
- fix #112

## Proposed changes

Adds:

- optional tool call dispatch handler for ChatSession sting streaming
- a way to inject the tool response if using ChatSession details (caller handles dispatch)

## Checklist

### DIFF
--- a/Libraries/MLXLMCommon/ChatSession.swift
+++ b/Libraries/MLXLMCommon/ChatSession.swift
@@ -51,9 +51,9 @@ public final class ChatSession {
         instructions: String? = nil,
         generateParameters: GenerateParameters = .init(),
         processing: UserInput.Processing = .init(resize: CGSize(width: 512, height: 512)),
+        additionalContext: [String: any Sendable]? = nil,
         tools: [ToolSpec]? = nil,
-        toolDispatch: (@Sendable (ToolCall) async throws -> String)? = nil,
-        additionalContext: [String: any Sendable]? = nil
+        toolDispatch: (@Sendable (ToolCall) async throws -> String)? = nil
     ) {
         self.model = model
         self.instructions = instructions
@@ -80,9 +80,9 @@ public final class ChatSession {
         instructions: String? = nil,
         generateParameters: GenerateParameters = .init(),
         processing: UserInput.Processing = .init(resize: CGSize(width: 512, height: 512)),
+        additionalContext: [String: any Sendable]? = nil,
         tools: [ToolSpec]? = nil,
-        toolDispatch: (@Sendable (ToolCall) async throws -> String)? = nil,
-        additionalContext: [String: any Sendable]? = nil
+        toolDispatch: (@Sendable (ToolCall) async throws -> String)? = nil
     ) {
         self.model = ModelContainer(context: model)
         self.instructions = instructions
@@ -144,9 +144,9 @@ public final class ChatSession {
         history: [Chat.Message],
         generateParameters: GenerateParameters = .init(),
         processing: UserInput.Processing = .init(resize: CGSize(width: 512, height: 512)),
+        additionalContext: [String: any Sendable]? = nil,
         tools: [ToolSpec]? = nil,
-        toolDispatch: (@Sendable (ToolCall) async throws -> String)? = nil,
-        additionalContext: [String: any Sendable]? = nil
+        toolDispatch: (@Sendable (ToolCall) async throws -> String)? = nil
     ) {
         self.model = ModelContainer(context: model)
         self.instructions = instructions


Libraries/MLXLMCommon/ChatSession.swift
```diff
@@ -42,13 +43,16 @@ public final class ChatSession {
     ///   - instructions: optional system instructions for the session
     ///   - generateParameters: parameters that control generation
     ///   - processing: media processing configuration for images/videos
+    ///   - tools: optional tool specifications
+    ///   - toolDispatch: optional tool dispatch -- required for toolcalls if streaming strings rather than details
```
davidkoski, line 47: @louis-jan FYI take a look at this update, see also #107

Libraries/MLXLMCommon/ChatSession.swift
```diff
@@ -34,6 +34,7 @@ public final class ChatSession {
     public var generateParameters: GenerateParameters
     public var additionalContext: [String: any Sendable]?
     public var tools: [ToolSpec]?
+    public var toolDispatch: (@Sendable (ToolCall) async throws -> String)?
```
davidkoski, line 37: We can't do a dictionary of tools because they have type parameters.

Libraries/MLXLMCommon/ChatSession.swift
```diff
@@ -152,11 +167,14 @@ public final class ChatSession {
     /// - Returns: the model's response
     public func respond(
         to prompt: String,
+        role: Chat.Message.Role = .user,
```
davidkoski, line 170: Let callers pass in an optional role (backward compatible).  See tests later in the PR.

Libraries/MLXLMCommon/ChatSession.swift
```diff
@@ -291,35 +315,49 @@ public final class ChatSession {
...
+                        for await item in stream {
+                            // if there is no toolDispatch then the caller must
+                            // handle the toolcall
+                            if let toolCall = item.toolCall, let toolDispatch {
+                                let toolResult = try await toolDispatch(toolCall)
+                                messages = [.tool(toolResult)]
+                                task.cancel()
+                                await task.value
+                                continue restart
+                            }
```
davidkoski, line 347: This is the real change

Tests/MLXLMIntegrationTests/ChatSessionIntegrationTests.swift
```diff
@@ -132,6 +132,36 @@ public class ChatSessionIntegrationTests: XCTestCase {
         // The model should either produce a tool call or mention the tool/weather
         let hasContent = responseText.count > 0 || !toolCalls.isEmpty
         XCTAssertTrue(hasContent, "Response should contain either text or tool calls")
+
+        let weather = try await session.respond(
+            to: "Foggy with a high in the low 60s, clearing later in the day", role: .tool)
```
davidkoski, line 137: Inject a response to the tool call and make sure it responds appropriately

Tests/MLXLMIntegrationTests/ChatSessionIntegrationTests.swift
```diff
@@ -132,6 +132,36 @@ public class ChatSessionIntegrationTests: XCTestCase {
...
+            let time: String
+        }
+
+        let timeTool = Tool<EmptyInput, TimeOutput>(
+            name: "get_time",
+            description: "Get the current date and time including day of week.",
+            parameters: []
+        ) { _ in
+            TimeOutput(time: "Wed Feb 18 17:50:43 PST 2026")
+        }
```
davidkoski, line 154: A variant with the dispatch dealing only in strings.

Libraries/MLXLMCommon/ChatSession.swift
davidkoski: @mattt this might be of interest to you as well, depending on how you integrate -- let me know if this direction causes any issues.

Libraries/MLXLMCommon/ChatSession.swift
```diff
@@ -103,6 +113,7 @@ public final class ChatSession {
         generateParameters: GenerateParameters = .init(),
         processing: UserInput.Processing = .init(resize: CGSize(width: 512, height: 512)),
         tools: [ToolSpec]? = nil,
+        toolDispatch: (@Sendable (ToolCall) async throws -> String)? = nil,
         additionalContext: [String: any Sendable]? = nil
```
louis-jan: It would be better to move tools and toolDispatch to the bottom so it supports trailing closure better.
```suggestion
        additionalContext: [String: any Sendable]? = nil,
        tools: [ToolSpec]? = nil,
        toolDispatch: (@Sendable (ToolCall) async throws -> String)? = nil
```

Libraries/MLXLMCommon/ChatSession.swift
louis-jan: This is great!

Libraries/MLXLMCommon/ChatSession.swift
davidkoski: Good idea!  It is a minor API breakage but since the tool calling couldn't complete the loop with the current API, I think this is probably safe.

Libraries/MLXLMCommon/ChatSession.swift
mattt: @davidkoski I like it! In fact, this would unblock streaming tool calling in MLX for AnyLanguageModel.

Also, +1 for @louis-jan's suggestion (https://github.com/ml-explore/mlx-swift-lm/pull/114#discussion_r2825484319) to move the tool dispatch to the end so it can be a trailing closure.

angeloskath (approved):
Looks great!

Just to be clear and improve my understanding (I do need to read more of the Swift code), given the `role` addition output injection could also be done via the streamDetails and an outer loop that feeds any tool call back in.

PR #115, opened 2026-02-20
audit RoPE use across models

### LINKED ISSUE #110: [BUG] Llama 3.1 RoPE scaling skipped when rope_scaling values decode as Int
**Describe the bug**
When loading [Llama 3.1 8B Instruct 4-bit](https://huggingface.co/mlx-community/Meta-Llama-3.1-8B-4bit), RoPE scaling is silently skipped because `rope_scaling` values in `config.json` decode as `Int` (e.g. `original_max_position_embeddings: 8192`). The Swift implementation only accepted `.float` values, so the guard failed and the llama3 RoPE scaling path was disabled. This causes token/logit divergence vs Python mlx-lm (e.g., first visible divergence around token ~92 on a fixed prompt/seed).

**To Reproduce**

1. Build `llm-tool`:
```bash
xcodebuild -scheme llm-tool -configuration Debug build
```

2. Run with a fixed prompt/seed and greedy decoding:
```bash
./mlx-run --debug llm-tool eval \
  --model "/path/to/Meta-Llama-3.1-8B-Instruct-4bit" \
  --prompt "Hello, Today is Feb 5 2026, How are you doing? It's a beautifull day, lets play the game, shall we? Hide and seek time." \
  --seed 42 --temperature 0 --max-tokens 1024
```

3. Compare to Python mlx-lm for the same model/prompt/seed:
```bash
python -m mlx_lm generate \
  --model "/path/to/Meta-Llama-3.1-8B-Instruct-4bit" \
  --prompt "Hello, Today is Feb 5 2026, How are you doing? It's a beautifull day, lets play the game, shall we? Hide and seek time." \
  --seed 42 --temp 0 --max-tokens 1024
```

**Code Snippet**
```swift
// Minimal example showing that Llama 3.1 RoPE scaling can be skipped
// when rope_scaling values decode as Int.
import Foundation
import MLXLLM

let modelPath = "/path/to/Meta-Llama-3.1-8B-Instruct-4bit"
let configURL = URL(filePath: modelPath).appending(path: "config.json")
let data = try Data(contentsOf: configURL)
let config = try JSONDecoder().decode(LlamaConfiguration.self, from: data)

// Prior to the fix, rope_scaling values that decode as Int caused the
// llama3 RoPE scaling path to be skipped.
print("rope_scaling =", String(describing: config.ropeScaling))
```

**Expected behavior**
Llama 3.1 RoPE scaling should always be applied when `rope_scaling` is present in `config.json`, regardless of whether values are encoded as ints or floats. Output should match Python mlx-lm for identical prompt/seed/params.

 - OS Version: macOS 15.5
 - Device: M4 Mac Mini
 - Version: mlx-swift-examples 0.29.1 (mlx-swift 0.29.1)

**Additional context**
`config.json` contains:
```json
"rope_scaling": {
  "factor": 8.0,
  "low_freq_factor": 1.0,
  "high_freq_factor": 4.0,
  "original_max_position_embeddings": 8192,
  "rope_type": "llama3"
}
```
Because `StringOrNumber` decodes ints before floats, `original_max_position_embeddings` becomes `.int`, and the previous `case .float` guard fails.

---

### Related issue: chat template whitespace normalization

**Describe the issue**
swift-jinja can insert extra whitespace when rendering Llama 3.1 chat templates (e.g., extra newlines). This changes the prompt tokenization and causes divergence from Python. A normalization step is required to match Python’s token counts and outputs.

**Symptoms**
- Prompt token count differs (e.g., 74 vs 72 for the same prompt).
- Output differs even with identical model/seed/params.

**Expected behavior**
Rendered chat template tokens should match Python’s `apply_chat_template` output without extra whitespace.

davidkoski:
## Proposed changes

- fix bugs related to rope config decode
- use common RoPE implementation if mlx-lm did
- fix #110 -- full details of the investigation there

## Checklist

### DIFF
--- a/Libraries/MLXLLM/Models/DeepseekV3.swift
+++ b/Libraries/MLXLLM/Models/DeepseekV3.swift
@@ -153,6 +153,17 @@ class DeepseekV3Attention: Module {
         self._oProj.wrappedValue = Linear(
             numHeads * vHeadDim, hiddenSize, bias: config.attentionBias)
 
+        if let ropeScaling = config.ropeScaling {
+            let mScaleAllDim = ropeScaling["mscale_all_dim"]?.asFloat() ?? 0.0
+            if mScaleAllDim != 0 {
+                let scalingFactor = ropeScaling["factor"]?.asFloat() ?? 1.0
+                if scalingFactor > 1 {
+                    let s = 0.1 * mScaleAllDim * log(scalingFactor) + 1.0
+                    self.scale = self.scale * s * s
+                }
+            }
+        }
+
         self.rope = initializeRope(
             dims: qkRopeHeadDim, base: ropeTheta, traditional: true,
             scalingConfig: config.ropeScaling, maxPositionEmbeddings: maxPositionEmbeddings)
--- a/Libraries/MLXLMCommon/RoPEUtils.swift
+++ b/Libraries/MLXLMCommon/RoPEUtils.swift
@@ -90,14 +90,6 @@ public class Llama3RoPE: Module, OffsetLayer, ArrayOffsetLayer {
 public class YarnRoPE: Module, OffsetLayer, ArrayOffsetLayer {
     let dimensions: Int
     let traditional: Bool
-    let maxPositionEmbeddings: Int
-    let base: Float
-    let scalingFactor: Float
-    let originalMaxPositionEmbeddings: Int
-    let betaFast: Float
-    let betaSlow: Float
-    let mscale: Float
-    let mscaleAllDim: Float
 
     private let _mscale: Float
     private let _freqs: MLXArray
@@ -118,14 +110,6 @@ public class YarnRoPE: Module, OffsetLayer, ArrayOffsetLayer {
 
         self.dimensions = dimensions
         self.traditional = traditional
-        self.maxPositionEmbeddings = maxPositionEmbeddings
-        self.base = base
-        self.scalingFactor = scalingFactor
-        self.originalMaxPositionEmbeddings = originalMaxPositionEmbeddings
-        self.betaFast = betaFast
-        self.betaSlow = betaSlow
-        self.mscale = mscale
-        self.mscaleAllDim = mscaleAllDim
 
         func yarnFindCorrectionDim(numRotations: Float) -> Float {
             return Float(dimensions)
@@ -179,8 +163,13 @@ public class YarnRoPE: Module, OffsetLayer, ArrayOffsetLayer {
     }
 
     public func callAsFunction(_ x: MLXArray, offset: Int = 0) -> MLXArray {
+        // "copy" of x as we are going to write through it and don't want to update
+        // through the reference
+        // https://github.com/ml-explore/mlx-swift/issues/364
+        var x = x
         if _mscale != 1.0 {
-            x[.ellipsis, 0 ..< dimensions] = _mscale * x[.ellipsis, 0 ..< dimensions]
+            x = x[0..., .ellipsis]
+            x[.ellipsis, 0 ..< dimensions] *= _mscale
         }
 
         return MLXFast.RoPE(
@@ -195,8 +184,10 @@ public class YarnRoPE: Module, OffsetLayer, ArrayOffsetLayer {
     }
 
     public func callAsFunction(_ x: MLXArray, offset: MLXArray) -> MLXArray {
+        var x = x
         if _mscale != 1.0 {
-            x[.ellipsis, 0 ..< dimensions] = _mscale * x[.ellipsis, 0 ..< dimensions]
+            x = x[0..., .ellipsis]
+            x[.ellipsis, 0 ..< dimensions] *= _mscale
         }
 
         return MLXFast.RoPE(
--- a/Libraries/MLXLMCommon/SuScaledRoPE.swift
+++ b/Libraries/MLXLMCommon/SuScaledRoPE.swift
@@ -2,8 +2,9 @@ import Foundation
 import MLX
 import MLXNN
 
+// port of https://github.com/ml-explore/mlx-lm/blob/main/mlx_lm/models/rope_utils.py
+
 /// Su Scaled Rotary Position Embedding.
-/// Switches between short and long factors based on sequence length.
 public class SuScaledRoPE: Module, OffsetLayer, ArrayOffsetLayer {
     let dimensions: Int
     let originalMaxPositionEmbeddings: Int
@@ -20,6 +21,9 @@ public class SuScaledRoPE: Module, OffsetLayer, ArrayOffsetLayer {
         shortMScale: Float? = nil,
         longMScale: Float? = nil
     ) {
+        // Note: per python source shortFactor and shortMScale are unused.
+        // https://github.com/ml-explore/mlx-lm/pull/707
+
         precondition(dimensions % 2 == 0, "Dimensions must be even")
 
         self.dimensions = dimensions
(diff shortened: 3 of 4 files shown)


davidkoski:
> This is great @davidkoski I'm working on batching and this PR will do a lot of the work, and significantly decrease the blast radius of my batching PR. Since you're already editing the files, do you think you can make `initializeRope()` conform `OffsetLayer & ArrayOffsetLayer`? (Maybe a typealias like `public typealias RoPELayer = OffsetLayer & ArrayOffsetLayer` would make it easier to read?)

Yeah, that is a good idea -- I will give this a try today!

Libraries/MLXEmbedders/Models/NomicBert.swift
```diff
@@ -186,18 +187,14 @@ func computeBaseFrequency(
...
-    guard case .float(let factor) = ropeScaling["factor"],
-        case .float(let lowFreqFactor) = ropeScaling["low_freq_factor"]
-            ?? .float(1.0),
-        case .float(let highFreqFactor) = ropeScaling["high_freq_factor"]
-            ?? .float(4.0),
-        case .float(let oldContextLen) = ropeScaling[
-            "original_max_position_embeddings"]
-            ?? .float(8192)
-    else {
+    guard let factor = ropeScaling["factor"]?.asFloat() else {
```
davidkoski, line 190: The original issue in #110 are for a different piece of code, but same issue -- the bug is that this is asking for "config that is a float" vs "config _as_ a float".

Libraries/MLXEmbedders/Models/Qwen3.swift
```diff
@@ -56,28 +56,10 @@ private class Attention: Module {
...
-                break
-            }
-        }
-
-        self.rope = RoPE(
-            dimensions: headDim,
-            traditional: false,
-            base: args.ropeTheta,
-            scale: ropeScale
+        self.rope = initializeRope(
```
davidkoski, line 59: Using `initializeRope()` (matches `initialize_rope()` from mlx-lm) is great because:

- it avoids the config decode bugs as it is centralized and done correctly
- you can use other RoPE styles
- and mlx-lm does it this way, but not for every model

In the places where this is changed like this it is also done that way in mlx-lm.  It may be that others _should_ be this way in mlx-lm, but for now I am just going for parity.

Libraries/MLXEmbedders/Configuration.swift
```diff
@@ -2,30 +2,6 @@
 
 import Foundation
 
-public enum StringOrNumber: Codable, Equatable, Sendable {
```
davidkoski, line 5: We can use the one from MLXLMCommon

Libraries/MLXLLM/Models/Apertus.swift
```diff
@@ -152,88 +152,6 @@ private class XIELU: Module, UnaryLayer {
     }
 }
 
-private class DynamicNTKScalingRoPE: Module {
```
davidkoski, line 155: I presume some copy & pasting -- this was actually a LlamaRope not DynamicNTKScaling.  Several of the deleted RoPE implementations follow this pattern, including the one originally reported.

Libraries/MLXLLM/Models/DeepseekV3.swift
```diff
@@ -84,72 +84,6 @@ private func yarnFindCorrectionRange(
...
-    return scale <= 1 ? 1.0 : 0.1 * mscale * log(scale) + 1.0
-}
-
-private func yarnLinearRampMask(minVal: Float, maxVal: Float, dim: Int) -> MLXArray {
-    let updatedMaxVal = minVal == maxVal ? maxVal + 0.001 : maxVal
-    let linearFunc = (MLXArray(0 ..< dim) - minVal) / (updatedMaxVal - minVal)
-    return clip(linearFunc, min: 0, max: 1)
-}
-
-class DeepseekV3YarnRotaryEmbedding: Module {
```
davidkoski, line 97: In mlx-lm it uses the common RoPE factory and the yarn implementation there.

Libraries/MLXLLM/Models/Llama.swift
```diff
@@ -8,130 +8,6 @@ import Tokenizers
 
 // port of https://github.com/ml-explore/mlx-examples/blob/main/llms/mlx_lm/models/llama.py
 
-func computeBaseFrequency(
```
davidkoski, line 11: This was unused code and also had the bug

Libraries/MLXLLM/Models/Llama.swift
```diff
@@ -8,130 +8,6 @@ import Tokenizers
...
-        let wavelen = 2 * .pi / freq
-        let smooth = max(
-            0, min(1, (wavelen - highFreqWavelen) / (lowFreqWavelen - highFreqWavelen)))
-        return freq * ((1 - smooth) * factor + smooth)
-    }
-
-    return newBaseFreqs.reduce(0, +) / Float(newBaseFreqs.count)
-}
-
-class LlamaDynamicNTKScalingRoPE: Module {
```
davidkoski, line 53: Again the misnamed RoPE, we can use the rope factory.  This is the original report for #110

Libraries/MLXLMCommon/RoPEUtils.swift
```diff
@@ -9,11 +9,11 @@ import Foundation
 import MLX
 import MLXNN
 
-public class Llama3RoPE: Module, OffsetLayer {
+public class Llama3RoPE: Module, OffsetLayer, ArrayOffsetLayer {
     let dims: Int
     let maxPositionEmbeddings: Int
     let traditional: Bool
-    let freqs: MLXArray
+    let _freqs: MLXArray
```
davidkoski, line 16: Without the `_` it will give an error on loading the weights -- I presume we never hit this path as the factory wasn't used much (yet).

Libraries/MLXLMCommon/RoPEUtils.swift
```diff
@@ -69,12 +69,25 @@ public class Llama3RoPE: Module, OffsetLayer {
             base: nil,
             scale: 1.0,
             offset: offset,
-            freqs: freqs
+            freqs: _freqs
         )
     }
+
+    public func callAsFunction(_ x: MLXArray, offset: MLXArray) -> MLXArray {
```
davidkoski, line 76: Add the array offset variants, @ronaldmannak

Libraries/MLXLMCommon/SuScaledRoPE.swift
```diff
@@ -4,13 +4,11 @@ import MLXNN
 
 /// Su Scaled Rotary Position Embedding.
 /// Switches between short and long factors based on sequence length.
-public class SuScaledRoPE: Module, OffsetLayer {
+public class SuScaledRoPE: Module, OffsetLayer, ArrayOffsetLayer {
```
davidkoski, line 8: I re-ported this from python as it had changed a bit and the code we had on the swift side made it hard to do the ArrayOffset

Tests/MLXLMIntegrationTests/ChatSessionIntegrationTests.swift
```diff
@@ -24,13 +24,21 @@ public class ChatSessionIntegrationTests: XCTestCase {
         let vlmExpectation = XCTestExpectation(description: "Load VLM")
 
         Task {
-            llmContainer = try await IntegrationTestModels.shared.llmContainer()
-            llmExpectation.fulfill()
+            do {
+                llmContainer = try await IntegrationTestModels.shared.llmContainer()
+                llmExpectation.fulfill()
+            } catch {
+                fatalError("Unable to load llm: \(error)")
```
davidkoski, line 31: I was using a different model and it was swallowing errors that I needed to see :-)

Libraries/MLXLMCommon/RoPEUtils.swift
```diff
@@ -180,15 +193,36 @@ public class YarnRoPE: Module, OffsetLayer {
             freqs: self._freqs
         )
     }
+
+    public func callAsFunction(_ x: MLXArray, offset: MLXArray) -> MLXArray {
+        if _mscale != 1.0 {
+            x[.ellipsis, 0 ..< dimensions] = _mscale * x[.ellipsis, 0 ..< dimensions]
```
angeloskath, line 199: `let x = x[0.., .ellipsis]`

The proper way to fix this is to make sure that `x.at[.ellipsis, 0 ..< dimensions].multiply(_mscale)` maps to a slice update instead of a scatter product. But the above should be good for now anyway.

Issue #79, opened 2025-07-06
Proposal: Add thinking Case to Generation Enum


ronaldmannak:
Currently, the `Generation` enum has three cases: `chunk`, `info`, and `toolCall`. 

Many newer APIs (such as Ollama’s `thinking` property in `Message`) now include special properties for "thinking" directly in their response data structures, rather than encoding such tokens in text.

### Rationale

Different models may use varying tokens to represent "thinking," making it complicated to detect or filter these tokens at the application layer. Moving the responsibility for handling these special tokens to the inference engine would simplify integration and keep application code cleaner.

### Proposed Change

Add a new `.thinking` case to the `Generation` enum:

```swift
public enum Generation: Sendable {
    /// A generated token represented as a String.
    case chunk(String)

    /// A generated "thinking" token, represented as a String.
    case thinking(String)

    /// Completion information summarizing token counts and performance metrics.
    case info(GenerateCompletionInfo)

    /// A tool call from the language model.
    case toolCall(ToolCall)
    ...
}
```

### Considerations

- **Breaking Change:** Adding a new enum case will require updates to any exhaustive `switch` statements that handle `Generation` in both the mlx-swift-examples code and third party apps using MLX-Swift.

Looking for feedback!

PR #116, opened 2026-02-20
Add harmony tool call parser

louis-jan:
As mentioned in issue #61, mlx-swift-lm includes a GPT-OSS model definition, but GPT-OSS requires Harmony formatting end-to-end. However, since thinking case is not yet standardized (see #79), it can be parsed from applications. For tool calls, as they are available for most supported models, this PR adds the tool call parser for gpt-oss model.

## Proposed changes

Adds:
- HarmonyToolCallParser to parse tool call for GPT-OSS model
- Move `includeStopToken` to ToolHandler impl, as it's used only in `RawTokenLoopHandler`. generateLoopTask should not know this setting as it's ToolHandler impl-specific logic.
- In Harmony, <|call|> is a tool call end tag but also an EOS chunk, so this PR added a check for the current tool call parser state and detokenize to complete the tool parse and emit.

## Checklist

### DIFF
--- a/Libraries/MLXLMCommon/Evaluate.swift
+++ b/Libraries/MLXLMCommon/Evaluate.swift
@@ -979,8 +979,7 @@ public func generateTokenTask(
         modelConfiguration: modelConfiguration,
         tokenizer: tokenizer,
         iterator: iterator,
-        includeStopToken: includeStopToken,
-        handler: RawTokenLoopHandler()
+        handler: RawTokenLoopHandler(includeStopToken)
     )
 }
 
@@ -990,7 +989,6 @@ private func generateLoopTask<Handler: TokenLoopHandler>(
     tokenizer: Tokenizer,
     iterator: consuming TokenIterator,
     wiredMemoryTicket: WiredMemoryTicket? = nil,
-    includeStopToken: Bool = false,
     handler: consuming Handler
 ) -> (AsyncStream<Handler.Output>, Task<Void, Never>) {
 
@@ -1030,13 +1028,15 @@ private func generateLoopTask<Handler: TokenLoopHandler>(
 
                 // Check for end-of-sequence tokens
                 if token == tokenizer.unknownTokenId || stopTokenIDs.contains(token) {
-                    if includeStopToken {
+                    // Increase tokenCount only if handler emits end token
+                    if !handler.onStopToken(token, emit: { value in
                         tokenCount += 1
-                        if !handler.onStopToken(token, emit: continuation.yield) {
-                            stopReason = .cancelled
-                            break
-                        }
+                        return continuation.yield(value)
+                    }) {
+                        stopReason = .cancelled
+                        break
                     }
+                    
                     stopReason = .stop
                     break
                 }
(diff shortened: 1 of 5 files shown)


Libraries/MLXLMCommon/Evaluate.swift
```diff
@@ -1030,13 +1028,15 @@ private func generateLoopTask<Handler: TokenLoopHandler>(
...
-                        if !handler.onStopToken(token, emit: continuation.yield) {
-                            stopReason = .cancelled
-                            break
-                        }
+                        return continuation.yield(value)
+                    }) {
+                        stopReason = .cancelled
+                        break
                     }
+
```
louis-jan, line 1039: @ronaldmannak FYI, take a look at this update. I've refactored so that includeStopToken is encapsulated within RawTokenHandler only. Now, whenever a token needs to be emitted, the count increments automatically without requiring this function to be aware of the includeStopToken flag.
cc @davidkoski

Libraries/MLXLMCommon/Evaluate.swift
```diff
@@ -1312,7 +1312,24 @@ private struct TextToolTokenLoopHandler: TokenLoopHandler, @unchecked Sendable {
...
+                    let _ = toolCallProcessor.processChunk(chunk)
+                    
+                    // Emit completed tool call if needed
+                    if let toolCall = toolCallProcessor.toolCalls.popLast() {
+                        let _ = emit(.toolCall(toolCall))
+                    }
+                }
+            }
+        }
+        return true
```
louis-jan, line 1332: @davidkoski, gpt-oss model will return <|call|> as tool call endtag but also stop token EOS. This change is to check whether it needs to complete the tool parsing process instead of just ending the stream.

Libraries/MLXLMCommon/Evaluate.swift
```diff
@@ -1337,8 +1359,10 @@ private struct RawTokenLoopHandler: TokenLoopHandler {
         _ token: Int,
         emit: (sending TokenGeneration) -> AsyncStream<TokenGeneration>.Continuation.YieldResult
     ) -> Bool {
-        if case .terminated = emit(.token(token)) {
-            return false
+        if(self.includeStopToken) {
+            if case .terminated = emit(.token(token)) {
+                return false
+            }
         }
```
louis-jan, line 1366: So generateLoopTask would not need to care about EOS token emitting (as this is a specific logic for this raw token handler)

Libraries/MLXLMCommon/Tool/ToolCallProcessor.swift
```diff
@@ -26,17 +26,17 @@ public class ToolCallProcessor {
 
     // MARK: - Properties
 
-    private let parser: any ToolCallParser
     private let tools: [[String: any Sendable]]?
-    private var state = State.normal
     private var toolCallBuffer = ""
+    var state = State.normal
+    let parser: any ToolCallParser
```
louis-jan, line 32: I don't really like this change, but exposing these changes so TextToolTokenLoopHandler can check whether the tool parser is processing or not, also for endTag check.
https://github.com/ml-explore/mlx-swift-lm/pull/116/changes#diff-d636823870a24e57e1e1d2c6b125edc68ba15e35c0122cb749fab0fd714d601fR1315

ronaldmannak:
Hi @louis-jan — thanks for tagging me, I appreciate it.
There are a couple of Harmony-specific issues, which is why I didn’t add Harmony parsing in the Generate Token PR:
- **Parsing needs to happen at the token level**, not on the tokenized string.  
- **Even with token-level parsing**, non–tool-calling Harmony output doesn’t map cleanly onto MLX-Swift-LM’s `Generation` model. That leaves you with two rough options:
  - **Lose important nuance** (which matters for things like [Open Responses](https://www.openresponses.org)), or
  - **Expose Harmony tags** in the generated text.
  
Neither is an acceptable option IMO.
For now, the solution I am using is doing the harmony parsing in the app layer. OpenAI provides a [Rust library](https://github.com/openai/harmony) for Harmony parsing. I built an FFI wrapper for it: [Pico Harmony](https://github.com/picomlx/picoharmony). The catch is that the FFI currently ships with a **precompiled** version of the Rust library, and I totally understand why folks would be hesitant to depend on that.
I’ve offered for OpenAI to take ownership of the library, but I haven’t heard back yet (hint hint @stephancasas). I also haven’t followed up :)
I’d be happy to transfer ownership to **ml-explore** as well, but integrating Harmony properly at the MLX-Swift-LM layer would require a fairly substantial refactor.
I guess it depends on where the MLX team wants to take the library

PR #117, opened 2026-02-24
Add wiredMemoryTicket to GenerateTokens

ronaldmannak:
## Proposed changes

This PR adds wired memory to token generation

## Checklist

### DIFF
--- a/skills/mlx-swift-lm/references/wired-memory.md
+++ b/skills/mlx-swift-lm/references/wired-memory.md
@@ -6,7 +6,7 @@
 
 - Policies: `WiredSumPolicy`, `WiredMaxPolicy`, `WiredFixedPolicy`, `WiredBudgetPolicy`
 - Measurement helpers: `WiredMemoryUtils` and `WiredMemoryMeasurement`
-- Integration point: `wiredMemoryTicket` parameter on all generation APIs (text-level and token-level)
+- Integration point: `wiredMemoryTicket` parameter on generation APIs (text-level and token-level)
 
 Core implementation:
 


Libraries/MLXLMCommon/Evaluate.swift
```diff
@@ -910,13 +910,17 @@ public func generateTask(
...
+///     memory control (macOS 15 / iOS 18 / tvOS 18 or newer).
 /// - Returns: An `AsyncStream` that emits `TokenGeneration` values.
 public func generateTokens(
     input: LMInput,
     cache: [KVCache]? = nil,
     parameters: GenerateParameters,
     context: ModelContext,
-    includeStopToken: Bool = false
+    includeStopToken: Bool = false,
+    wiredMemoryTicket: WiredMemoryTicket? = nil
```
davidkoski, line 923: Interesting -- the wiredMemoryTicket is added to many (most?) of the generate calls but it looks like a couple were missed.  This is potentially a sign that there are too many `generate()` functions.  I guess I knew that :-)

skills/mlx-swift-lm/references/wired-memory.md
```diff
@@ -6,7 +6,7 @@
 
 - Policies: `WiredSumPolicy`, `WiredMaxPolicy`, `WiredFixedPolicy`, `WiredBudgetPolicy`
 - Measurement helpers: `WiredMemoryUtils` and `WiredMemoryMeasurement`
-- Integration point: `wiredMemoryTicket` parameter on generation APIs
+- Integration point: `wiredMemoryTicket` parameter on all generation APIs (text-level and token-level)
```
davidkoski: FWIW the purely synchronous call does not support it:

```
public func generate(
    input: LMInput, context: ModelContext,
    iterator: TokenIterator,
    didGenerate: (Int) -> GenerateDisposition
) -> GenerateCompletionInfo {
```

This is because the Task scoped ticket can't be used in a sync context.  This is deprecated anyway, so this documentation is probably fine, but `all` might be too strong a word -- see what you think.

Libraries/MLXLMCommon/Evaluate.swift
ronaldmannak: Yeah, sorry. I created this PR late last night after a bad flu and I ran into a build error. My memory is a but blurry but I do have a feeling we had this conversation already :) I actually thought I already had created it last week and was sure it was already accepted, but it wasn't and I can't find it. If we discussed this before, I apologize. 
What's missing was wired memory support for token generation (needed for Harmony models) because the the token generation and wired memory PRs crossed each other. That's fixed with this PR.
I would welcome a more unified approach as discussed in #116

davidkoski: OK, understand the crossing of PRs -- there have been a couple in this area for sure.

My comment was more me talking to myself and lamenting how it looks like I didn't get the generate API quite right the first through N times so it had to keep growing :-)

ronaldmannak: Gotcha. It's kind of impossible to get it right when model output evolves from plain text to reasoning to something as complex as Harmony in the span of, what is it, 12 or 18 months or so

Issue #41, opened 2026-01-05
Release a version


nezhyborets:
Hi guys, we really need this fix in our project https://github.com/ml-explore/mlx-swift-lm/pull/11
Could you please release a new version of the package? It would include the code from PR. Thanks!

davidkoski:
https://github.com/ml-explore/mlx-swift-lm/releases/tag/2.29.3

PR #118, opened 2026-02-24
Decouple from tokenizer and downloader packages

DePasqualeOrg:
MLX Swift LM currently has two fundamental problems:

- Model loading is tightly coupled to the Hugging Face Hub. A Hub client is required even when loading models from a local directory.
- Model loading performance with Swift Transformers lags far behind the Python equivalent, typically taking several seconds in Swift versus a few hundred milliseconds in Python.

This PR implements the following solutions:

- Swift Transformers is replaced with [Swift Tokenizers](https://github.com/DePasqualeOrg/swift-tokenizers), a streamlined and optimized fork that focuses purely on tokenizer functionality, with no Hugging Face dependency and no extraneous Core ML code. **This unlocks a 10x to 15x speedup in model loading times.**
- The `Downloader` protocol abstracts away the model hosting provider, making it easy to use other providers such as ModelScope or define custom providers such as downloading from storage buckets.
- Swift Hugging Face, a dedicated client for the Hub, is used in an optional module. No Hugging Face Hub code is bundled for users who don't need it.

## Benchmarks

Model loading times on M3 MacBook Pro:

| Benchmark | Before | After | Speedup |
|-----------|--------|-------|---------|
| LLM | 3228 ms | 319 ms | **10.1x** |
| VLM | 3788 ms | 365 ms | **10.4x** |
| Embedding | 1479 ms | 95 ms | **15.6x** |

To run the benchmarks before the changes in this PR, check out commit `3752cc2`.

You can run the benchmarks in a separate scheme in Xcode with `RUN_BENCHMARKS=1`, or from the command line:

`TEST_RUNNER_RUN_BENCHMARKS=1 xcodebuild test -scheme mlx-swift-lm-Package -destination 'platform=macOS' -only-testing:Benchmarks`

## Usage

Loading from a local directory:

```swift
import MLXLLM
import MLXLMCommon

let modelDirectory = URL(filePath: "/path/to/model")
let container = try await loadModelContainer(from: modelDirectory)
```

Convenience method from `MLXLMHuggingFace` module (uses default Hub client):

```swift
import MLXLLM
import MLXLMHuggingFace

let container = try await loadModelContainer(id: "mlx-community/Qwen3-4B-4bit")
```

Using a custom Hugging Face Hub client:

```swift
import MLXLLM
import MLXLMHuggingFace

let hub = HubClient(token: "hf_...")
let container = try await loadModelContainer(
    from: hub,
    id: "mlx-community/Qwen3-4B-4bit"
)
```

Using a custom downloader:

```swift
import MLXLLM
import MLXLMCommon

struct S3Downloader: Downloader {
    func download(
        id: String, revision: String?, matching patterns: [String],
        useLatest: Bool, progressHandler: @Sendable @escaping (Progress) -> Void
    ) async throws -> URL {
        // Download model files and return a local directory URL
    }
}

let container = try await loadModelContainer(
    from: S3Downloader(),
    id: "my-bucket/my-model"
)
```

Embedding models and adapters follow the same patterns.

## Cache strategy

The `Downloader` protocol includes a `useLatest` parameter (default `false`) that controls whether to check the network for updates:

- **`useLatest: false`**: Resolves refs (e.g. "main") to commit hashes locally via the cache's `refs/` directory and returns cached files immediately, with no network call. This avoids 100–200ms of latency on every model load.
- **`useLatest: true`**: Always checks the network for the latest commit, then downloads any missing or updated files.

This improves on the Python `huggingface_hub` in two ways: Python always makes an `api.repo_info()` network call before returning cached files, even for commit hashes. Swift skips the network entirely for commit hashes (which are immutable, so cached files are always valid) and additionally resolves branch names locally via `resolveCachedSnapshot()` when freshness isn't needed. Users who want the latest files can opt in to the network call explicitly.

In Swift Hugging Face, this is implemented as a two-method design:
- `resolveCachedSnapshot()` resolves refs locally using cached metadata
- `downloadSnapshot()` only uses the fast path on commit hashes (which are immutable), while branch names always trigger a network call

### Loading API

The `hub` parameter (previously `HubApi`) has been replaced with `from` (any `Downloader` or `URL` for a local directory). Functions that previously defaulted to `defaultHubApi` no longer have a default – callers must either pass a `Downloader` explicitly or use the convenience methods in `MLXLMHuggingFace` / `MLXEmbeddersHuggingFace`, which default to `HubClient.default`.

For most users who were using the default Hub client, adding `import MLXLMHuggingFace` or `import MLXEmbeddersHuggingFace` and using the convenience overloads is sufficient.

Users who were passing a custom `HubApi` instance should create a `HubClient` instead and pass it as the `from` parameter. `HubClient` conforms to `Downloader` via `MLXLMHuggingFace`.

### `ModelConfiguration`

- `tokenizerId` and `overrideTokenizer` have been replaced by `tokenizerSource: TokenizerSource?`, which supports `.id(String)` for remote sources and `.directory(URL)` for local paths.
- `preparePrompt` has been removed. This shouldn't be used anyway, since support for chat templates is available.
- `modelDirectory(hub:)` has been removed. For local directories, pass the `URL` directly to the loading functions. For remote models, the `Downloader` protocol handles resolution.

### Tokenizer loading

`loadTokenizer(configuration:hub:)` has been removed. Tokenizer loading now uses `AutoTokenizer.from(directory:)` from Swift Tokenizers directly.

`replacementTokenizers` (the `TokenizerReplacementRegistry`) has been removed. Use `AutoTokenizer.register(_:for:)` from Swift Tokenizers instead.

### `defaultHubApi`

The `defaultHubApi` global has been removed. Hugging Face Hub access is now provided by `HubClient.default` from the `HuggingFace` module.

### Low-level APIs

- `downloadModel(hub:configuration:progressHandler:)` → `Downloader.download(id:revision:matching:useLatest:progressHandler:)`
- `loadTokenizerConfig(configuration:hub:)` → `AutoTokenizer.from(directory:)`
- `ModelFactory._load(hub:configuration:progressHandler:)` → `_load(configuration: ResolvedModelConfiguration)`
- `ModelFactory._loadContainer`: removed (base `loadContainer` now builds the container from `_load`)

## Maintainership of Swift Tokenizers

I'm currently maintaining Swift Tokenizers, but I think a better home for it would be the ml-explore organization. Hugging Face's packages are tightly coupled to their platform, while Swift Tokenizers is designed for a clean separation of concerns and is more closely related to the model code in MLX Swift LM.

## To do

- [ ] Review and merge [PR #41](https://github.com/huggingface/swift-huggingface/issues/41) in swift-huggingface, which improves cache hit performance by avoiding unnecessary network calls, then switch to a tagged release (currently pinned to branch in my fork)
- [ ] Review [changes](https://github.com/DePasqualeOrg/swift-tokenizers/pulls?q=is%3Apr) in Swift Tokenizers since fork from Swift Transformers
- [ ] Decide on maintainership of Swift Tokenizers

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,5 +1,10 @@
 # MLX Swift LM
 
+> [!IMPORTANT]
+> The `main` branch is a _new_ major version number: 3.x.  In order
+> to decouple from tokenizer and downloader packages some breaking
+> changes were introduced. See [Breaking Changes](#breaking-changes) for more information.
+
 MLX Swift LM is a Swift package to build tools and applications with large language models (LLMs) and vision language models (VLMs) in [MLX Swift](https://github.com/ml-explore/mlx-swift).
 
 Some key features include:
@@ -257,3 +262,37 @@ Developers can use these examples in their own programs -- just import the swift
 - [MLXLLM](https://swiftpackageindex.com/ml-explore/mlx-swift-lm/main/documentation/mlxllm): Large language model example implementations
 - [MLXVLM](https://swiftpackageindex.com/ml-explore/mlx-swift-lm/main/documentation/mlxvlm): Vision language model example implementations
 - [MLXEmbedders](https://swiftpackageindex.com/ml-explore/mlx-swift-lm/main/documentation/mlxembedders): Popular encoders and embedding models example implementations
+
+## Breaking Changes
+
+### Loading API
+
+The `hub` parameter (previously `HubApi`) has been replaced with `from` (any `Downloader` or `URL` for a local directory). Functions that previously defaulted to `defaultHubApi` no longer have a default – callers must either pass a `Downloader` explicitly or use the convenience methods in `MLXLMHuggingFace` / `MLXEmbeddersHuggingFace`, which default to `HubClient.default`.
+
+For most users who were using the default Hub client, adding `import MLXLMHuggingFace` or `import MLXEmbeddersHuggingFace` and using the convenience overloads is sufficient.
+
+Users who were passing a custom `HubApi` instance should create a `HubClient` instead and pass it as the `from` parameter. `HubClient` conforms to `Downloader` via `MLXLMHuggingFace`.
+
+### `ModelConfiguration`
+
+- `tokenizerId` and `overrideTokenizer` have been replaced by `tokenizerSource: TokenizerSource?`, which supports `.id(String)` for remote sources and `.directory(URL)` for local paths.
+- `preparePrompt` has been removed. This shouldn't be used anyway, since support for chat templates is available.
+- `modelDirectory(hub:)` has been removed. For local directories, pass the `URL` directly to the loading functions. For remote models, the `Downloader` protocol handles resolution.
+
+### Tokenizer loading
+
+`loadTokenizer(configuration:hub:)` has been removed. Tokenizer loading now uses `AutoTokenizer.from(directory:)` from Swift Tokenizers directly.
+
+`replacementTokenizers` (the `TokenizerReplacementRegistry`) has been removed. Use `AutoTokenizer.register(_:for:)` from Swift Tokenizers instead.
+
+### `defaultHubApi`
+
+The `defaultHubApi` global has been removed. Hugging Face Hub access is now provided by `HubClient.default` from the `HuggingFace` module.
+
+### Low-level APIs
+
+- `downloadModel(hub:configuration:progressHandler:)` → `Downloader.download(id:revision:matching:useLatest:progressHandler:)`
+- `loadTokenizerConfig(configuration:hub:)` → `AutoTokenizer.from(directory:)`
+- `ModelFactory._load(hub:configuration:progressHandler:)` → `_load(configuration: ResolvedModelConfiguration)`
+- `ModelFactory._loadContainer`: removed (base `loadContainer` now builds the container from `_load`)
+


davidkoski:
I think the approach is good overall but there is one problem we will have to figure out:

> ```swift
> // Package.swift
> dependencies: [
>     .package(url: "https://github.com/ml-explore/mlx-swift-lm", from: "2.0.0"),
>     .package(url: "https://github.com/ml-explore/mlx-swift-lm-tokenizers", from: "1.0.0"),
>     .package(url: "https://github.com/ml-explore/mlx-swift-lm-huggingface", from: "1.0.0"),
> ]
> ```

The same "logistics" issue appears here -- we cannot easily add new repositories.  All of the functionality will have to go into `mlx-swift-lm`.

I think this might be a place where the traits would be useful.  If you can use them, it could select which back ends you actually want to pull.  If not, you will pull more dependencies than you need but the build process should only build, link and copy the ones you use.

DePasqualeOrg:
I think that approach is not fair or ideal for the following reasons:

- It unfairly allows one company's library to exclusively use the name `Tokenizers` in the context of this package.
- It requires people to pull libraries that they're not using during package resolution.

For those reasons, I think the integration packages should be separate. Anyone can make and host one, and they're just a few lines of code for protocol conformance.

davidkoski:
OK, I have a proof of concept using macros.  I have a stand-in for the real thing that looks like this:

```swift
public protocol MLXTokenizer: Sendable {
    func encode(text: String) -> [Int]
    func decode(tokens: [Int]) -> String
}

public func generate(tokenizer: MLXTokenizer) -> String {
    let tokens = tokenizer.encode(text: "testing")
    return tokenizer.decode(tokens: tokens)
}
```

**Note: there is no hard dependency on _any_ concrete Tokenizer**

We want to call it along these lines:

```swift
let tokenizer = PreTrainedTokenizer(...) // e.g. the HuggingFace Tokenizer
print(generate(tokenizer: tokenizer))
```

That won't work as-is because `PreTrainedTokenizer` doesn't implement `MLXTokenizer`, at least not type-wise.

If we added:

```swift
extension PreTrainedTokenizer: @retroactive MLXTokenizer { }
```

Then it would work, but we are conforming a type we don't own to a protocol.

OK, so try 1 with a macro looks like this:

```swift
enum Tokenizers {
    #MLXTokenizer(PreTrainedTokenizer.self)
}

let tokenizer = try Tokenizers.MLXPreTrainedTokenizer()
print(generate(tokenizer: tokenizer))
```

The enum is needed because the macro can't generate a top level type (unless it has a static name).  The macro ends up generating a simple wrapper for the type (assuming it looks like a HuggingFace Tokenizer API-wise) and forwards the protocol methods.

Try 2 looks like this:

```swift
#TokenizerFactory(PreTrainedTokenizer.self)

let tokenizer = try makeTokenizer()
print(generate(tokenizer: tokenizer))
```

The factory generates a function with a fixed name so it can appear at the top level.  Assuming you could build/link it would allow multiple providers of tokenizers if you did this in different files.

Try 3:

```swift
let tokenizer = try #MakeTokenizer(PreTrainedTokenizer.self)
print(generate(tokenizer: tokenizer))
```

No top level function, just an inline expression.

For all of these the `import Tokenizers` is in the application, not the library.  I have handwaved a bunch of stuff about how to actually generate the wrapper knowing that the underlying type looks like the current HuggingFace Tokenizer API.  I am ignoring how we get the configuration, etc. but I presume this could be worked out.

This wouldn't block nicer integrations that actually implemented `loadModelContainer()`, but might let us break the dependency in a different way.

DePasqualeOrg:
@davidkoski, I think we would need to see a working code example of that macro approach, but I suspect that it won't be able to do everything that we need to do to make this work. Check how I've set things up in the integration packages to see what I mean.

I really think the integration packages are the happy, simple path, and they should be hosted alongside the respective tokenizer/downloader packages.

Tests/MLXLMIntegrationTests/ChatSessionIntegrationTests.swift
davidkoski, line 1: It is very unfortunate that we lose these -- they are _very_ useful in testing the model integrations.  From messing around it looks like we could reference the required integrations in Package.swift and they would only be built for tests, but Xcode seeing the duplicate targets is back to the original trouble.

I wonder if we could work around this by having a `/IntegrationTests` top level directory with its own Package.swift referencing mlx-swift-lm AND the HuggingFace libraries/integrations?  That would leave it still usable for its stated purpose but keep it out of builds using mlx-swift-lm.

Package.swift
```diff
@@ -1,4 +1,4 @@
-// swift-tools-version: 5.12
+// swift-tools-version: 6.0
```
davidkoski: I wonder if this will cause any problems.  I think the key is less the swift-tools-version and more the `platforms` -- app developers want to support older versions of macOS/iOS.  This may require people to use newer Xcode but would still be able to deploy to the older versions.

Just thinking out loud here.  Maybe we should do the same in mlx-swift.

Tests/MLXLMIntegrationTests/ChatSessionIntegrationTests.swift
DePasqualeOrg: They're not lost. They moved to `Libraries/IntegrationTestHelpers/IntegrationTestHelpers.swift` so that they can be run in the integration packages with the respective tokenizer and downloader packages. This allows for consistent integration testing across the different tokenizer and downloader options.

Libraries/MLXLMCommon/Tool/Parsers/ParserUtilities.swift
```diff
@@ -2,6 +2,29 @@
 
 import Foundation
 
+// MARK: - JSON to Sendable Bridge
+
+/// Convert a JSON-deserialized value to `any Sendable`.
+///
+/// `JSONSerialization` returns `Any`, but all JSON types it produces
+/// (String, NSNumber, NSNull, Array, Dictionary) are Sendable.
+func asSendable(_ value: Any) -> any Sendable {
```
davidkoski, line 11: I am not sure why this isn't a merge conflict, see https://github.com/ml-explore/mlx-swift-lm/blob/main/Libraries/MLXLMCommon/Tool/Parsers/ParserUtilities.swift#L7 -- very similar code in the same spot.

Tests/MLXLMIntegrationTests/ChatSessionIntegrationTests.swift
davidkoski: Ah, nice!  I do think a Package.swift (separate from the one at the top level) that would let us run these in Xcode would be ideal.  That would let us depend on specific integrations without adding them to the top level Package.swift.

These are primarily for testing the _models_ not the downloading and tokenizers (but of course they are integration tests and the whole thing gets tested).

Libraries/MLXLMCommon/Tool/Parsers/ParserUtilities.swift
davidkoski: Oh, actually it will be -- it just doesn't show that in the diff view.

Package.swift
DePasqualeOrg: I originally switched to 6.0 due to a cyclical dependency issue, which is supported in that version, but that is no longer an issue now that the test logic is exported for use in the integration packages. I switched it back to 5.12, although I suspect 6.0 wouldn't have been too much of a problem.

Libraries/BenchmarkHelpers/BenchmarkHelpers.swift
```diff
@@ -0,0 +1,466 @@
...
+// MARK: - Download Benchmarks
+
+/// Benchmark download cache hit performance. Ensures the model is cached with a warm-up
+/// download, then measures repeated cache lookups.
+public func benchmarkDownloadCacheHit(
+    from downloader: any Downloader,
+    modelId: String = "mlx-community/Qwen3-0.6B-4bit",
+    runs: Int = BenchmarkDefaults.downloadRuns
+) async throws -> BenchmarkStats {
+    let patterns = ["*.safetensors", "*.json", "*.jinja"]
```
davidkoski: I wonder if the actual library integrations should use these same patterns?  e.g. `resolveDirectories()` and `load()` use `["*.safetensors", "*.json"]`.

Libraries/MLXEmbedders/Load.swift
```diff
@@ -70,43 +63,48 @@ public enum EmbedderError: LocalizedError {
...
-///   - hub: The `HubApi` instance for managing downloads.
+///   - downloader: The downloader to use for fetching remote resources.
 ///   - configuration: The configuration identifying the model.
+///   - useLatest: When true, always checks the provider for updates.
 ///   - progressHandler: A closure to monitor download progress.
-/// - Returns: A `URL` pointing to the directory containing model files.
-func prepareModelDirectory(
-    hub: HubApi,
+/// - Returns: A tuple of (modelDirectory, tokenizerDirectory).
+func resolveDirectories(
```
davidkoski, line 75: This code looks nearly identical to `ModelAdapterFactory.load()`.  I wonder if these can/should share implementation?

It looks like the MLXEmbedders retains the top level `load()` function while the LLM/VLM side migrate this into `ModelFactory`.  Comparing against `main` it looks like that is the pattern there too, but maybe we should take this opportunity to consolidate some of this.

Tests/MLXLMIntegrationTests/ChatSessionIntegrationTests.swift
DePasqualeOrg: They can be run in Xcode in the integration packages, and tokenization is an important part of testing the models. I've been through a lot of pain with Swift tooling choking on package naming conflicts, and the best solution seemed to be to keep anything that could conflict out of this package. It's also possible that something could work here but cause problems for consumers.

Libraries/MLXLMCommon/Tool/Parsers/ParserUtilities.swift
DePasqualeOrg: I've resolved the merge conflicts.

Libraries/BenchmarkHelpers/BenchmarkHelpers.swift
DePasqualeOrg: Good catch. I resolved this in the latest commit by using package-level constants for tokenizer and model file patterns.

Libraries/MLXEmbedders/Load.swift
DePasqualeOrg: I investigated this and it seems it's not so straightforward, since MLXEmbedders has its own `ModelConfiguration` type.

davidkoski: Yeah, I guess it does -- the ModelConfiguration in MLXLMCommon has properties for detokenizing, so that kind of makes sense.  Perhaps a task for after this PR.  With a major version bump we could take a look at various APIs and see if any could be simplified or cleaned up.

davidkoski:
@DePasqualeOrg check out:

Xcode 26.4 beta 3
March 9, 2026

it has a working traits feature in the project dependencies!

davidkoski:
> @davidkoski, I experimented with using traits today, and it still fails at the build planning step (before compilation) due to the module name collision. And if we only include one of those packages here, the same problem will occur for anyone who wants to use the other by importing it in their project.
> 
> I think this further supports having a clean separation of concerns, which is the current state of this PR.

Ah, that is unfortunate -- traits are supposed to be for cases like this.  Well, perhaps not colliding packages.

I want to play around with it a bit so I can file some internal issues if needed.

davidkoski:
OK, I think I am getting my head wrapped around traits enough to see how they won't work for this case (some tests on the trait-test branch and more changes I have locally).

## swift-log

So a case where this might work:  let's say we wanted to use swift-log as an opt-in:

- we define a `SwiftLog` trait
- we have `#if SwiftLog` in the code to import and call functions
- we have a conditional dependency on the `MLXLMCommon` target to add `swift-log` if the trait is enabled

This would always check out swift-log because the Package dependencies can't depend on traits (not ideal).  swift-log would not build if it were not being used and it would not be linked (good).

Note the `#if SwiftLog` is _not_ available in Package.swift -- it is a setting on building the products.

## SDKTest (proxy for fancy linkage)

Another variant, a little closer to what we need with the tokenizers/hub:  what if we wanted some library that required a different SDK?  I made an SDKTest directory with a Packages.swift with its own `platforms` setting.  That seems to work.  I can refer to this from the top level Package.swift.  But:

- same issue with traits -- I can control linkage in a target but not the package level dependency
- the products/targets in the sub-package are not visible to code that depends on the top level Package -- in other words SDKTest is not visible
- worse, Xcode doesn't show the contents of SDKTest -- it is like the files do not exist

We could hoist the API surface from SDKTest into e.g. MLXLMCommon and `#if SDKTest` around code that forwarded to the library.

## The Issue at Hand

This leaves us with:

- we can't dynamically control Package level dependencies with traits, only Target level
- that means we will pull all the adaptors and tokenizers, etc. (not great, but not fatal yet)
- Xcode won't build with the duplicate names, even using the [swiftpm module aliasing](https://docs.swift.org/swiftpm/documentation/packagemanagerdocs/modulealiasing/)
- at best we could make a generic library that used traits to conditionally compile code that exposed the various integrations, but Xcode can't actually build such a thing

davidkoski:
> I don't quite understand the path you have in mind, but it sounds like you want to keep the Hugging Face dependency, which would prevent anyone from using my faster Swift Tokenizers package.

Not exactly.  I have a few things in mind:

- can we get the non-api-breaking part of this merged right away to avoid merge conflicts
- I am leery of making an API break that would require people to use a new repository outside of what they may currently trust (supply chain attack worries) -- maybe I am being overly cautious for nothing, but it is something I think about.  It is a reason I haven't merged #106 
- we don't currently have a solution that deals with that

I think people _should_ try your improved tokenizers package and in time it may become the de-facto standard, but I don't want to force it on anyone yet.

Right now your integration with the HF implementations are around 160 lines of code -- I don't think it is reasonable to have people copy that into their projects.

> It has been an enormous amount of work to get this to this point, and I would really like to get this merged so that we can move on. I think it's in an ideal state, with a clean separation of concerns and a straightforward way for users to migrate (whenever you do a major version bump) and pick what dependencies they want to use.

I agree, I like what you have done.  If the integration repos were in place (above) I would be preparing to merge.

davidkoski:
> > * can we get the non-api-breaking part of this merged right away to avoid merge conflicts
> 
> This is the part I don't understand. Even though I've taken great care to keep breaking changes to a minimum, the protocol is a breaking change.

Ah, I am not explaining myself well then and looking at it closer, I think you are correct.

On the LLM side the tokenizer is separate from the model.  The fact that the type changes from `Tokenizers.Tokenizer` to `MLXLMCommon.Tokenizer` is an API linkage break but not a source break (which is generally OK for swiftpm packages since you build from source).

The VLM side it tougher because the `UserInputProcessor` (which all the VLMs specialize) takes a `Tokenizers.Tokenizer` as input.  We can change all of them in MLXVLM but if somebody had a custom model in their application / library it would be a breaking change.

My idea (which I think is incorrect now) is that we would supply the implementation of `MLXLMCommon.Tokenizer` to keep it building on top of HF, but the custom VLM inits would all break (outside of MLXVLM).

> > * I am leery of making an API break that would require people to use a new repository outside of what they may currently trust (supply chain attack worries)
> 
> That's a valid concern, and it's why I suggested that people can copy ~100 lines of code instead of importing my packages.

I don't think copying 100 lines is a reasonable upgrade path (perhaps number of lines is not the metric as you would likely just copy a file into your repository).  But perhaps here we can add an HF specific macro to build the integration?

The minimum change to keep as-is would be:

```swift
import MLXLMCommon
import Tokenizers
import HuggingFaceAdaptorMacro

// named TBD, but let's say
let container = try await #huggingFaceLoadModelContainer(
    id: "mlx-community/Qwen3-4B-4bit"
)
```

Plus a change in their Package.swift.  The macro would let us inject the code at build-time and could ship as part of mlx-swift-lm without requiring a hugging face dependency (I think, though I have thought other things that turned out to be false).  This would give a couple of lines change while breaking the hard link dependency in mlx-swift-lm.

So you would have two ways of integrating:

- use the repositories that you provide that allow a choice between Hub and Tokenizer implementations
- use the macro and link (at the app level) to the hugging face libraries

If this works I think it would solve my concerns and let us move toward the conformance repositories.

What do you think about this?

PR #120, opened 2026-02-25
Adding Support for Qwen3.5 & Qwen3.5 MoE (Vision)

johnmai-dev:
## Proposed changes

1. Add Qwen3.5 and Qwen3.5 MoE (Vision)
2. Skip sanitize when __metadata__.format == "mlx" to avoid reapplying HF-only transforms (e.g., norm + 1) that cause gibberish output. https://github.com/Blaizzy/mlx-vlm/blob/a673abf2e91c0afbb0fd53d8cf6cd7eda9bd487f/mlx_vlm/utils.py#L215

Ported
- https://github.com/ml-explore/mlx-lm/pull/869
- https://github.com/Blaizzy/mlx-vlm/pull/722

## Checklist

### DIFF
--- a/Libraries/MLXLLM/Models/Qwen35.swift
+++ b/Libraries/MLXLLM/Models/Qwen35.swift
@@ -264,8 +264,12 @@ final class Qwen35GatedDeltaNet: Module {
         var state = cache?[1]
         let dtype = q.dtype
         let invScale = pow(Float(headKDim), -0.5)
-        let qNormed = MLXArray(pow(invScale, 2)).asType(dtype) * MLXFast.rmsNorm(q, weight: MLXArray.mlxNone, eps: 1e-6)
-        let kNormed = MLXArray(invScale).asType(dtype) * MLXFast.rmsNorm(k, weight: MLXArray.mlxNone, eps: 1e-6)
+        let qNormed =
+            MLXArray(pow(invScale, 2)).asType(dtype)
+            * MLXFast.rmsNorm(q, weight: MLXArray.mlxNone, eps: 1e-6)
+        let kNormed =
+            MLXArray(invScale).asType(dtype)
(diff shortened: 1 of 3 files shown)


Libraries/MLXLMCommon/Load.swift
```diff
@@ -6,6 +6,45 @@ import MLX
 import MLXNN
 import Tokenizers
 
+private func isMLXFormatSafetensor(url: URL) -> Bool {
```
johnmai-dev: Skip sanitize when __metadata__.format == "mlx" to avoid reapplying HF-only transforms (e.g., norm + 1) that cause gibberish output.

tpae:
Would love this to be merged soon @davidkoski 🙏

Libraries/MLXLMCommon/Load.swift
```diff
@@ -80,7 +122,11 @@ public func loadWeights(
     }
 
     // per-model cleanup
-    weights = model.sanitize(weights: weights)
+    safetensorURLs.sort { $0.path < $1.path }
+    let isMLXFormat = safetensorURLs.first.map(isMLXFormatSafetensor(url:)) ?? false
+    if !isMLXFormat {
```
davidkoski: This makes sense but I wonder if we should be using `loadArraysAndMetadata()` to load the metadata rather than manually decoding the header as isMLXFormatSafetensor does?

I wonder also if we should add a `metadata:` parameter to `sanitize(weights:)`.  Perhaps the default implementation could examine the `format` and call the existing method if not `mlx`.  I think the general rule of thumb that loading `mlx` format shouldn't go through sanitize is fine but I would prefer to defer it to the model (where developers could customize the behavior).

johnmai-dev: I’ve already used `loadArraysAndMetadata`.

PR #121, opened 2026-02-25
Fix KVCache serialization

petrukha-ivan:
## Proposed changes

Fixed multiple KVCache serialization problems. 

I noticed that a simple save/load round trip doesn’t work.

```swift
try savePromptCache(url: url, cache: cache, metadata: [:])
try loadPromptCache(url: url) // KVCacheError(message: "Mismatch in cache counts")
```

The root issue is the empty `metaState` representation. Returning an empty array is different from returning an empty string in the Python version, because an empty array doesn’t populate the required metadata. As a result, this part of saving never writes any values, but we still expect at least something when loading.

```swift
// Flatten cache_info as "0.i.j" (first element of cache_metadata)
for (i, info) in cacheInfo.enumerated() {
    for (j, metaValue) in info.enumerated() {
        flattenedMetadata["0.\(i).\(j)"] = metaValue // This won't be called, as we have nothing to iterate
    }
}
```

This results in the following metadata in Swift:

```
["2.7": "KVCache", "2.8": "KVCache", "2.2": "KVCache", "2.9": "KVCache", "2.1": "KVCache", "2.6": "KVCache", "2.4": "KVCache", "2.5": "KVCache", "2.3": "KVCache", "2.0": "KVCache"]
```

Compared to the Python version:
```
{'0.0': '', '0.1': '', '0.2': '', '0.3': '', '0.4': '', '0.5': '', '0.6': '', '0.7': '', '0.8': '', '0.9': '', '2.0': 'KVCache', '2.1': 'KVCache', '2.2': 'KVCache', '2.3': 'KVCache', '2.4': 'KVCache', '2.5': 'KVCache', '2.6': 'KVCache', '2.7': 'KVCache', '2.8': 'KVCache', '2.9': 'KVCache'}
```

Returning not just an empty array, but an array containing an empty string, fixes the problem. I added unit tests to ensure correct behavior. The tests also highlighted more issues, which have been fixed as well:
- `ChunkedKVCache` was recognized as `KVCacheSimple` because of the wrong type matching order
- `ArraysCache` handling was missing

## Checklist

### DIFF
--- a/Tests/MLXLMTests/KVCacheTests.swift
+++ b/Tests/MLXLMTests/KVCacheTests.swift
@@ -3,16 +3,18 @@ import MLX
 import MLXLMCommon
 import Testing
 
-@Test(.serialized, arguments: [
-    ({ KVCacheSimple() }),
-    ({ RotatingKVCache(maxSize: 32) }),
-    ({ QuantizedKVCache() }),
-    ({ ChunkedKVCache(chunkSize: 16) }),
-    ({ ArraysCache(size: 2) }),
-    ({ MambaCache() })
-])
+@Test(
+    .serialized,
+    arguments: [
+        ({ KVCacheSimple() }),
+        ({ RotatingKVCache(maxSize: 32) }),
+        ({ QuantizedKVCache() }),
+        ({ ChunkedKVCache(chunkSize: 16) }),
+        ({ ArraysCache(size: 2) }),
+        ({ MambaCache() }),
+    ])
 func testCacheSerialization(creator: (() -> any KVCache)) async throws {
-    let cache = (0..<10).map { _ in creator() }
+    let cache = (0 ..< 10).map { _ in creator() }
     let keys = MLXArray.ones([1, 8, 32, 64], dtype: .bfloat16)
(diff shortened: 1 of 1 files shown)


Tests/MLXLMTests/KVCacheTests.swift
```diff
@@ -0,0 +1,40 @@
...
+
+@Test(.serialized, arguments: [
+    ({ KVCacheSimple() }),
+    ({ RotatingKVCache(maxSize: 32) }),
+    ({ QuantizedKVCache() }),
+    ({ ChunkedKVCache(chunkSize: 16) }),
+    ({ ArraysCache(size: 2) }),
+    ({ MambaCache() })
+])
+func testCacheSerialization(creator: (() -> any KVCache)) async throws {
```
davidkoski: This looks really nice!

davidkoski (changes requested):
Change looks good but swift-format is failing -- please run that.

I also filed #123 because it might be nice to round trip these from ChatSession instead of just the lower level APIs.

Issue #109, opened 2026-02-17
[BUG] Nemotron 4H support broken


ronaldmannak:
Has anyone been able to use any of the Nvidia Nemotron 3 Nano models? NemotronH support was added in #75 but the JSON parsing of config.json seems to fail at the `infinity` line:

```json
    "time_step_limit": [
        0.0,
        Infinity
    ],
```

ronaldmannak:
Thanks! I wasn't aware that property existed. Let me try

davidkoski:
#125 fixes this on the mlx-swift-lm side.  Waiting for:

- https://github.com/huggingface/swift-transformers/pull/324

and a tag before we can call this closed.

PR #125, opened 2026-03-02
Add JSON5 support

ronaldmannak:
## Proposed changes

This PR fixes JSON5 issue in #109 that prevented Nvidia Nemotron to load in MLX-Swift-LM.

This PR can be merged independently of a similar PR in Swift-Transformers ( https://github.com/huggingface/swift-transformers/pull/324 ) but Nemotron will only work when both PRs are applied.

## Checklist

### DIFF
--- a/Libraries/MLXLMCommon/Extensions/JSONDecoder+JSON5.swift
+++ b/Libraries/MLXLMCommon/Extensions/JSONDecoder+JSON5.swift
@@ -2,8 +2,8 @@
 
 import Foundation
 
-public extension JSONDecoder {
-    static func json5() -> JSONDecoder {
+extension JSONDecoder {
+    public static func json5() -> JSONDecoder {
         let decoder = JSONDecoder()
         decoder.allowsJSON5 = true
         return decoder


Libraries/MLXLLM/LLMModelFactory.swift
```diff
@@ -6,12 +6,18 @@ import MLX
 import MLXLMCommon
 import Tokenizers
 
+private func JSON5Decoder() -> JSONDecoder {
+    let d = JSONDecoder()
+    d.allowsJSON5 = true
```
davidkoski: Looks good.  I suspect MLXEmbedders needs it too -- do you want to update that one?

ronaldmannak: I missed those. Will fix

ronaldmannak:
> Yeah, that is what we discussed in #109 -- I thought so too, but you pointed out this was converting strings not bare literals. I do agree that API _looks like_ what we want, like _exactly_ what we want :-)

Oops yes, I already forgot about that 😂

PR #128, opened 2026-03-03
fix unreliable tests

### LINKED ISSUE #119: [BUG] MLXLMTests.ChatSessionTests testChatSessionSync are unreliable
I am seeing intermittent CI failures on MLXLMTests.ChatSessionTests testChatSessionSync (and maybe some of the others).  I think the testing tokenizer is a bit too random.

davidkoski:
## Proposed changes

- fixes https://github.com/ml-explore/mlx-swift-lm/issues/119
- make the random weight model and tokenizer behave in a predictable way

The TLDR on that is there are some integration tests that use a fake tokenizer and a real model with random weights.  If you are unlucky the model would generate a valid token (0) that would be taken as EOS and the test would not get the expected length string (basically > 0 length).  It _mostly_ works, but mostly isn't great for CI.  This makes sure the model won't accidentally generate EOS and we just use a cap on the number of tokens.

## Checklist

### DIFF
--- a/Tests/MLXLMIntegrationTests/EmbedderIntegrationTests.swift
+++ b/Tests/MLXLMIntegrationTests/EmbedderIntegrationTests.swift
@@ -51,13 +51,7 @@ struct EmbedderIntegrationtests {
 
     @Test("MLXEmbedders README.md example")
     func testReadMeExample() async throws {
-        guard let (searchInputs, resultEmbeddings) = try? await readeMeExampleResult() else {
-            throw NSError(
-                domain: "EmbedderIntegrationtests",
-                code: 1,
-                userInfo: [NSLocalizedDescriptionKey: "Failed to get example results"]
-            )
-        }
+        let (searchInputs, resultEmbeddings) = try await readeMeExampleResult()
 
         // Compute similarities
         let searchQueryEmbedding = resultEmbeddings[0]
--- a/Tests/MLXLMTests/ChatSessionTests.swift
+++ b/Tests/MLXLMTests/ChatSessionTests.swift
@@ -15,8 +15,8 @@ public class ChatSessionTests: XCTestCase {
     private func model() -> ModelContext {
         let config = Gemma3TextConfiguration(
             modelType: "text",
-            hiddenSize: 64, hiddenLayers: 8, intermediateSize: 128, attentionHeads: 8,
-            headDim: 256,
+            hiddenSize: 64, hiddenLayers: 8, intermediateSize: 64, attentionHeads: 4,
+            headDim: 64,
             rmsNormEps: 0.00001, vocabularySize: 100, kvHeads: 4,
             ropeTheta: 1_000_000, ropeLocalBaseFreq: 10_000,
             ropeTraditional: false, queryPreAttnScalar: 256,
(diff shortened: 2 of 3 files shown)


Tests/MLXLMIntegrationTests/EmbedderIntegrationTests.swift
```diff
@@ -51,13 +51,7 @@ struct EmbedderIntegrationtests {
...
     @Test("MLXEmbedders README.md example")
     func testReadMeExample() async throws {
-        guard let (searchInputs, resultEmbeddings) = try? await readeMeExampleResult() else {
-            throw NSError(
-                domain: "EmbedderIntegrationtests",
-                code: 1,
-                userInfo: [NSLocalizedDescriptionKey: "Failed to get example results"]
-            )
-        }
+        let (searchInputs, resultEmbeddings) = try await readeMeExampleResult()
```
davidkoski, line 54: Noticed while in here -- this previously ate the error and replaced it with a generic one.  Keep the error, it is useful.

Tests/MLXLMTests/ChatSessionTests.swift
```diff
@@ -15,8 +15,8 @@ public class ChatSessionTests: XCTestCase {
     private func model() -> ModelContext {
         let config = Gemma3TextConfiguration(
             modelType: "text",
-            hiddenSize: 64, hiddenLayers: 8, intermediateSize: 128, attentionHeads: 8,
-            headDim: 256,
+            hiddenSize: 64, hiddenLayers: 8, intermediateSize: 64, attentionHeads: 4,
+            headDim: 64,
```
davidkoski, line 19: Reduce the size a little -- runs faster.

Tests/MLXLMTests/ChatSessionTests.swift
```diff
@@ -39,11 +39,13 @@ public class ChatSessionTests: XCTestCase {
             tokenizer: processor.tokenizer)
     }
 
+    private var generationParameters = GenerateParameters(maxTokens: 50)
```
davidkoski: This is half the key -- the new EOS token is outside the range that the model will generate so we don't have to worry about it stopping early.  In fact it won't stop at all!

Tests/MLXLMTests/ChatSessionTests.swift
```diff
@@ -209,8 +215,8 @@ public class ChatSessionTests: XCTestCase {
             try await Task.sleep(for: .milliseconds(10))
         }
 
-        // try another message, wait for full completion
-        model.respond("message2")
+        // try another message, wait for full completion (but cap the length)
+        model.session.generateParameters = .init(maxTokens: 50)
```
davidkoski: Apply the maxTokens here so the second loop will terminate (we interrupt the first loop).

Tests/MLXLMTests/TestTokenizer.swift
```diff
@@ -9,11 +9,18 @@ import Tokenizers
 struct TestTokenizer: Tokenizer {
 
     let length = 8
+    let maxLength = 50
 
-    var vocabulary: [Int: String]
+    /// a token outside the range that the model will generate, see vocabularySize
+    let _eosTokenId = 101
+    let _unknownTokenId = 102
```
davidkoski, line 16: This is really the key to the fix -- if these are out of range of the vocabularySize then the model won't accidentally generate them and produce e.g. empty strings.

Tests/MLXLMTests/TestTokenizer.swift
```diff
@@ -33,8 +40,12 @@ struct TestTokenizer: Tokenizer {
...
 
     func encode(text: String) -> [Int] {
-        (0 ..< length).map { _ in
-            Int.random(in: 0 ..< 100)
+        (0 ..< length).enumerated().map { (index, _) in
+            if index > length {
+                _eosTokenId
+            } else {
+                Int.random(in: 1 ..< vocabularySize)
+            }
```
davidkoski: The rest of this is just some cleanup.

angeloskath (approved):
👍

PR #131, opened 2026-03-03
Fix XMLFunctionParser regex to match newlines

### LINKED ISSUE #130: [BUG] XMLFunctionParser regex fails on multiline output (Qwen3.5)
**Describe the bug**
XMLFunctionParser regex uses .*? which doesn't match newlines in Swift/ICU. Qwen3.5 models generate newlines between XML function tags, so parse() silently returns nil and tool calls  are dropped. 

**To Reproduce**

Include code snippet
```swift
import MLXLMCommon

let parser = XMLFunctionParser()

// Qwen3.5 generates newlines between tags
let content = """
    <tool_call>
    <function=get_current_datetime>
    </function>
    </tool_call>
    """

let result = parser.parse(content: content, tools: nil)
print(result == nil) // true — should be a valid tool call
```

**Expected behavior**
parse() should return a ToolCall with name: "get_current_datetime". Instead it returns nil because the regex <function=(.*?)</function> can't match across newlines.

 - OS Version: [macOS 26 (latest as of March 3)]
 - Device: [M1 Pro Max]
 - Version [main (commit 0ef0e10)]

**Additional context**

pixelsoccupied:
## Proposed changes

Swift/ICU regex `.` does not match `\n`. Models like Qwen3.5 generate newlines between XML function tags, so the parser silently fails. Use `[\s\S]` instead. 

fixed #130

## Checklist

- [x ] I have read the [CONTRIBUTING](https://github.com/ml-explore/mlx/blob/main/CONTRIBUTING.md) document
- [x ] I have run `pre-commit run --all-files` to format my code / installed pre-commit prior to committing changes
- [x ] I have added tests that prove my fix is effective or that my feature works
- [x ] I have updated the necessary documentation (if needed)

### DIFF
--- a/Libraries/MLXLMCommon/Tool/Parsers/XMLFunctionParser.swift
+++ b/Libraries/MLXLMCommon/Tool/Parsers/XMLFunctionParser.swift
@@ -11,10 +11,10 @@ public struct XMLFunctionParser: ToolCallParser, Sendable {
     public init() {}
 
     public func parse(content: String, tools: [[String: any Sendable]]?) -> ToolCall? {
-        // Pattern: <function=(.*?)</function>
+        // Pattern: <function=(content)</function> — [\s\S] matches newlines
         guard
             let funcMatch = content.range(
-                of: #"<function=(.*?)</function>"#, options: .regularExpression)
+                of: #"<function=([\s\S]*?)</function>"#, options: .regularExpression)
         else { return nil }
 
         let funcContent = String(content[funcMatch])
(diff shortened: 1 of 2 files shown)


Libraries/MLXLMCommon/Tool/Parsers/XMLFunctionParser.swift
```diff
@@ -11,10 +11,10 @@ public struct XMLFunctionParser: ToolCallParser, Sendable {
     public init() {}
 
     public func parse(content: String, tools: [[String: any Sendable]]?) -> ToolCall? {
-        // Pattern: <function=(.*?)</function>
+        // Pattern: <function=(content)</function> — [\s\S] matches newlines
         guard
             let funcMatch = content.range(
-                of: #"<function=(.*?)</function>"#, options: .regularExpression)
+                of: #"<function=([\s\S]*?)</function>"#, options: .regularExpression)
```
davidkoski, line 17: From what I read this is single-line mode vs not (the default).  Nice find!

PR #132, opened 2026-03-03
Fix tool calling for Mistral 3

atdrendel:
## Proposed changes

The Mistral 3 text and vision models, used by Ministral 3 and Devstral 2, support tool calling, but mlx-swift-lm did not send tools to the model and did not properly process tool calls from the model. This pull request addresses both of those issues.

### Prior to these changes

Tools never even made it to the model. So, the Mistral3 models just hallucinated a response.

Mistral3 unsuccessful tool calling before changes

### After applying these changes

The Mistral3 models receive the tool, and the newly-introduced `MistralToolCallParser` parses the tool calls correctly.

Mistral3 successful tool calling

## Checklist

### DIFF
--- a/Libraries/MLXLMCommon/Tool/Parsers/MistralToolCallParser.swift
+++ b/Libraries/MLXLMCommon/Tool/Parsers/MistralToolCallParser.swift
@@ -51,10 +51,8 @@ public struct MistralToolCallParser: ToolCallParser, Sendable {
 
         guard !namePart.isEmpty else { return nil }
 
-        // Parse arguments as JSON using deserialize from ParserUtilities
-        let arguments = deserialize(argsPart)
-
-        guard let argsDict = arguments as? [String: any Sendable] else {
+        // Parse arguments as JSON using tryParseJSON from ParserUtilities
+        guard let argsDict = tryParseJSON(argsPart) as? [String: any Sendable] else {
             return nil
         }
 


Libraries/MLXLMCommon/Tool/Parsers/MistralToolCallParser.swift
```diff
@@ -0,0 +1,66 @@
...
+/// Examples:
+/// - `[TOOL_CALLS]get_weather[ARGS]{"location": "Tokyo"}`
+/// - `[TOOL_CALLS]fn1[ARGS]{...}[TOOL_CALLS]fn2[ARGS]{...}` (multiple calls)
+///
+/// The end tag is `</s>` (EOS token). Since stop tokens are intercepted at the
+/// token ID level before detokenization, the EOS text never reaches the processor
+/// — tool calls are extracted via `ToolCallProcessor.flush()` at generation end.
+public struct MistralToolCallParser: ToolCallParser, Sendable {
+    public let startTag: String? = "[TOOL_CALLS]"
+    public let endTag: String? = "</s>"
```
atdrendel: The Mistral V11/13 tool call formats do **not** include an `endTag`. The tools end at the end of generation. However, `ToolCallProcessor.isInlineFormat` requires both `parser.startTag` and `parser.endTag` to be non-nil to use the tagged path. So, I set the EOS token as `endTag` to fit that requirement. I didn't want to change the implementation of `ToolCallProcessor.isInlineFormat`.

```swift
/// Whether this processor uses inline format (no start/end tags).
private var isInlineFormat: Bool {
    parser.startTag == nil || parser.endTag == nil
}
```

Libraries/MLXLMCommon/Tool/ToolCallProcessor.swift
```diff
@@ -67,6 +67,46 @@ public class ToolCallProcessor {
...
 
+    /// Flush any buffered content, attempting to parse it as tool call(s).
+    ///
+    /// Call this when generation ends (e.g., on EOS token) to handle formats
+    /// whose end tag is never delivered as text (e.g., Mistral where `</s>`
+    /// is intercepted at the token ID level).
+    ///
+    /// For formats with end tags that appear in the text stream, the buffer
+    /// will already be empty at generation end, making this a no-op.
+    public func flush() {
```
atdrendel: As the comment says, this should be a no-op for any existing tool call processors, but it's the least-invasive way I could come up with to flush the tool calls for Mistral3.

Libraries/MLXVLM/Models/Mistral3.swift
atdrendel, line 1: The changes in this file were the initial changes I made. Without these, the Mistral3 models never even received the tools from the caller.

Libraries/MLXVLM/VLMModelFactory.swift
```diff
@@ -329,6 +329,11 @@ public final class VLMModelFactory: ModelFactory {
         var mutableConfiguration = configuration
         mutableConfiguration.eosTokenIds = eosTokenIds
 
+        // Auto-detect tool call format from model type if not explicitly set
+        if mutableConfiguration.toolCallFormat == nil {
+            mutableConfiguration.toolCallFormat = ToolCallFormat.infer(from: baseConfig.modelType)
+        }
```
atdrendel, line 335: This seemed like an oversight. It should always have been present, as far as I could tell.

Libraries/MLXLMCommon/Tool/Parsers/MistralToolCallParser.swift
```diff
@@ -0,0 +1,66 @@
...
+
+    public init() {}
+
+    public func parse(content: String, tools: [[String: any Sendable]]?) -> ToolCall? {
+        var text = content
+
+        // Strip [TOOL_CALLS] prefix if present
+        if let range = text.range(of: "[TOOL_CALLS]") {
+            text = String(text[range.upperBound...])
```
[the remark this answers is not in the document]
atdrendel: I pushed a change to fix this issue.

Libraries/MLXLMCommon/Evaluate.swift
```diff
@@ -1337,6 +1344,16 @@ private struct TextToolTokenLoopHandler: TokenLoopHandler, @unchecked Sendable {
...
     }
 
+    mutating func onGenerationEnd(
+        emit: (sending Generation) -> AsyncStream<Generation>.Continuation.YieldResult
+    ) {
+        toolCallProcessor.flush()
+        while let toolCall = toolCallProcessor.toolCalls.first {
+            toolCallProcessor.toolCalls.removeFirst()
+            _ = emit(.toolCall(toolCall))
+        }
```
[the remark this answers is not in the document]
atdrendel: I fixed this.

Libraries/MLXLMCommon/Tool/Parsers/MistralToolCallParser.swift
davidkoski: Interesting.  We do want it to use `processTaggedChunk()` as it handles the start of the tool call.  Ideally it could handle no `endTag` but it doesn't right now.  I am not sure how the EOS shows up here.

I wonder if it would be better to make this work with a nil `endTag` and not need the `flush()` (if I am reading that right)?

Libraries/MLXLMCommon/Tool/ToolCallProcessor.swift
davidkoski: I think this function should move below `processChunk()` in the file (it comes at the end of processing).  I wonder if something like `processEOS()` would be a better name (to fit in with `processChunk()`)?

Libraries/MLXLMCommon/Tool/ToolCallProcessor.swift
```diff
@@ -67,6 +67,46 @@ public class ToolCallProcessor {
...
+        guard state == .collectingToolCall || state == .potentialToolCall else { return }
+        guard !toolCallBuffer.isEmpty else {
+            state = .normal
+            return
+        }
+
+        if let startTag = parser.startTag {
```
davidkoski: Do you think this should be something that a tool call parser opts into?  Or is only engaged when it doesn't have an `endTag` set?

Actually I think this splitting probably belongs in the `ToolCallParser`, something like:

```swift
/// handle remaining tool call data at EOS
func parseEOS(_ toolCallBuffer: String) -> [ToolCall]
```

I am not sure if it should `throws` -- it doesn't right now and the tool call parser seems "resilient" (best effort), so probably no.

I think we can't have an `endTag` in the string because the parser would have already handled that, so we could just use this code here as the default implementation but it would leave the ability for tools to override it.

Libraries/MLXLMCommon/Tool/Parsers/MistralToolCallParser.swift
davidkoski: OK, I see how `flush()` works here -- we can't avoid it because we need to handle the EOS and there was nothing in the lifecycle for that.  How about:

- endTag = nil
- change `isInlineFormat` to only look at `startTag`

Specifically:

```diff
diff --git a/Libraries/MLXLMCommon/Tool/Parsers/MistralToolCallParser.swift b/Libraries/MLXLMCommon/Tool/Parsers/MistralToolCallParser.swift
index 5b85009..daef613 100644
--- a/Libraries/MLXLMCommon/Tool/Parsers/MistralToolCallParser.swift
+++ b/Libraries/MLXLMCommon/Tool/Parsers/MistralToolCallParser.swift
@@ -20,7 +20,7 @@ import Foundation
 /// — tool calls are extracted via `ToolCallProcessor.flush()` at generation end.
 public struct MistralToolCallParser: ToolCallParser, Sendable {
     public let startTag: String? = "[TOOL_CALLS]"
-    public let endTag: String? = "</s>"
+    public let endTag: String? = nil
 
     public init() {}
 
diff --git a/Libraries/MLXLMCommon/Tool/ToolCallProcessor.swift b/Libraries/MLXLMCommon/Tool/ToolCallProcessor.swift
index 0f84a99..ed661c8 100644
--- a/Libraries/MLXLMCommon/Tool/ToolCallProcessor.swift
+++ b/Libraries/MLXLMCommon/Tool/ToolCallProcessor.swift
@@ -57,7 +57,7 @@ public class ToolCallProcessor {
 
     /// Whether this processor uses inline format (no start/end tags).
     private var isInlineFormat: Bool {
-        parser.startTag == nil || parser.endTag == nil
+        parser.startTag == nil
     }
 
     /// The first character of the start tag for quick detection.
```

I tried that locally and the tests pass -- I think we should try to express what we mean.

Libraries/MLXLMCommon/Evaluate.swift
```diff
@@ -1337,6 +1344,23 @@ private struct TextToolTokenLoopHandler: TokenLoopHandler, @unchecked Sendable {
...
 
+    mutating func onGenerationEnd(
+        emit: (sending Generation) -> AsyncStream<Generation>.Continuation.YieldResult
+    ) {
+        toolCallProcessor.flush()
+
+        guard !toolCallProcessor.toolCalls.isEmpty else { return }
+
+        let pendingToolCalls = toolCallProcessor.toolCalls
+        toolCallProcessor.toolCalls.removeAll(keepingCapacity: true)
```
davidkoski: Why clear this?  The processor is local to the eval loop, why not just iterate the `toolCalls`?

Libraries/MLXLMCommon/Evaluate.swift
```diff
@@ -1337,6 +1344,23 @@ private struct TextToolTokenLoopHandler: TokenLoopHandler, @unchecked Sendable {
...
+
+        for toolCall in pendingToolCalls {
```
davidkoski: I wonder if the caller will actually handle multiple tool calls (one of the cases in the tests)?  If using registered tools it will immediately call the tool and restart with the `<tool>...</tool>` output (I think).  If the caller is iterating details then they may do the same.

It might not matter as the model may just emit a second tool call after it only has one result, or maybe we need to file an issue on this -- handle multiple sequential tool calls.

Libraries/MLXLMCommon/Tool/Parsers/MistralToolCallParser.swift
atdrendel: Yeah, that's better.

Libraries/MLXLMCommon/Evaluate.swift
atdrendel: At least in our app's implementation, we definitely handle support multiple tool calls inside of the same turn. I'll paste our actual streaming response loop below so you can see how it looks in practice, but the key thing to note is we immediately stream reasoning and text responses to the UI and collect the tool calls. The tool calls are then executed, their results collected, and then those results are re-applied to the conversation, and then we await the next response from the model. So, at least for us, we'd want to support multiple tool calls.

```swift
for try await chunk in llm {
    guard !Task.isCancelled else {
        throw CancellationError()
    }

    let chunkStart = r.text.endIndex

    switch chunk {
    case let .reasoning(text):
        r.text += text
        if let current = currentReasoningRange {
            currentReasoningRange =
                current.lowerBound ..< r.text.endIndex
        } else {
            currentReasoningRange =
                chunkStart ..< r.text.endIndex
        }

        try SmartDocumentNeedingLocalLLM
            .appendStreamingContent(
                chunk: text,
                for: needingLLM.id,
                in: db
            )

    case let .text(text):
        pushCurrentReasoningRange()
        r.text += text
        try SmartDocumentNeedingLocalLLM
            .appendStreamingContent(
                chunk: text,
                for: needingLLM.id,
                in: db
            )

    case let .toolCall(call):
        pushCurrentReasoningRange()
        r.toolCalls.append(call)
    }
}
```

davidkoski: OK that makes sense -- it wasn't a case I knew about so it doesn't work that way in `ChatSession` internal tool handling.  Added:

- https://github.com/ml-explore/mlx-swift-lm/issues/134

Libraries/MLXLMCommon/Tool/Parsers/MistralToolCallParser.swift
atdrendel: Pushed.

Libraries/MLXLMCommon/Tool/ToolCallProcessor.swift
atdrendel: Pushed.

Libraries/MLXLMCommon/Tool/ToolCallProcessor.swift
atdrendel: I pushed a change that I think matches your idea.

PR #133, opened 2026-03-03
Fixed tool calling for qwen3.5

tpae:
## Proposed changes

Fixed tool calling detection logic for qwen3.5 series

Screenshot 2026-03-03 at 12 26 51 PM

## Checklist

### DIFF
--- a/Libraries/MLXLMCommon/Tool/ToolCallFormat.swift
+++ b/Libraries/MLXLMCommon/Tool/ToolCallFormat.swift
@@ -70,8 +70,8 @@ public enum ToolCallFormat: String, Sendable, Codable, CaseIterable {
     /// Example: `<|tool_call_start|>[func(arg='value')]<|tool_call_end|>`
     case lfm2
 
-    /// XML function format used by Qwen3 Coder.
-    /// Example: `<function=name><parameter=key>value</parameter></function>`
+    /// XML function format used by Nemotron, Qwen3 Coder, Qwen3.5, and similar models.
+    /// Example: `<tool_call><function=name><parameter=key>value</parameter></function></tool_call>`
     case xmlFunction = "xml_function"
 
     /// GLM4 format with arg_key/arg_value tags.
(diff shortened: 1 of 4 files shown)


atdrendel:
> Some overlap with #132 -- should be easy to resolve, I think :-)

😄 Yeah, I feel pretty good about our chances at being able to resolve the conflicts.

Libraries/MLXLMCommon/Tool/ToolCallFormat.swift
```diff
@@ -66,6 +66,10 @@ public enum ToolCallFormat: String, Sendable, Codable, CaseIterable {
     /// Example: `<invoke name="f"><parameter name="k">v</parameter></invoke>`
     case minimaxM2 = "minimax_m2"
 
+    /// Qwen3.5 format: XML function syntax wrapped in tool_call tags.
+    /// Example: `<tool_call><function=name><parameter=key>value</parameter></function></tool_call>`
+    case qwen35 = "qwen3_5"
```
atdrendel: I think this is the same format required for Nemotron, for which tool calling is currently broken, and Qwen3 Coder, isn't it? (Honestly, I can't think of a single model that uses the XML function style that doesn't use `<tool_call>` and `</tool_call>`.) I like your change to add the initializer with custom `startTag` and `endTag`, but I'd probably set the default to `<tool_call>` and `</tool_call>`. Even though it's kind of a breaking change, I don't think the current behavior is working for any models, unless there are ones that I haven't tested and can't think of.

We may want to remove the default `startTag` and `endTag` arguments entirely. Explicit is better than implicit, in my opinion. We could get rid of the separate `qwen3_5` tool call format and just have this:

```swift
case .xmlFunction:
    return XMLFunctionParser(startTag: "<tool_call>", endTag: "</tool_call>")
```

- [Nemotron](https://huggingface.co/nvidia/NVIDIA-Nemotron-3-Nano-30B-A3B-BF16/blob/main/chat_template.jinja#L92)
- [Qwen3 Coder](https://huggingface.co/Qwen/Qwen3-Coder-480B-A35B-Instruct/blob/main/qwen3coder_tool_parser.py#L37)
- [Qwen3.5](https://huggingface.co/Qwen/Qwen3.5-27B/blob/main/chat_template.jinja#L105)

tpae: Thanks for the feedback, and I agree. I can make the changes since I have reports of issues with Qwen3 Coder

atdrendel:
I just wanted to verify everything worked. So, I checked out this branch, added some Nemotron tests, downloaded , and everything worked as expected. Nice work.

CleanShot 2026-03-10 at 23 00 52@2x

PR #136, opened 2026-03-06
Add gemma 3 embedding model

CodebyCR:
## Proposed changes

## Checklist

### DIFF
--- a/Tests/MLXLMIntegrationTests/EmbedderIntegrationTests.swift
+++ b/Tests/MLXLMIntegrationTests/EmbedderIntegrationTests.swift
@@ -1,7 +1,7 @@
 // Copyright © 2026 Apple Inc.
 
 import Foundation
-@preconcurrency import MLX
+import MLX
 import MLXEmbedders
 import MLXLMCommon
 import MLXVLM


Libraries/MLXEmbedders/Extensions/MLXArray+Helper.swift
```diff
@@ -0,0 +1,36 @@
+// Copyright © 2026 Apple Inc.
+
+import MLX
+
+extension MLXArray {
+
+    public static func arange(_ size: Int) -> MLXArray {
+        return MLXArray(Array(0 ..< size))
+    }
```
davidkoski: See https://github.com/ml-explore/mlx-swift/blob/main/Source/MLX/Factory.swift#L1056

I think we don't need this at all.

Libraries/MLXEmbedders/Extensions/MLXArray+Helper.swift
```diff
@@ -0,0 +1,36 @@
...
+    /// norm is smaller than `eps`, it is clamped to `eps` to ensure numerical
+    /// stability and prevent division by zero.
+    ///
+    /// - Parameters:
+    ///   - axis: The axis along which to compute the norm. Defaults to `-1`.
+    ///   - eps: A small epsilon value to prevent division by zero. Defaults to `1e-12`.
+    /// - Returns: An `MLXArray` with the same shape as the original, normalized along `axis`.
+    ///
+    /// - Complexity: O(n), where n is the total number of elements in the array.
+    public func l2Normalized(axis: Int = -1, eps: Float = 1e-12) -> MLXArray {
```
davidkoski, line 19: Where did this come from?  I wonder if this should be in mlx-swift or even in mlx?

Libraries/MLXEmbedders/Extensions/MLXArray+Helper.swift
```diff
@@ -0,0 +1,36 @@
...
+
+    public static func arange(_ size: Int) -> MLXArray {
+        return MLXArray(Array(0 ..< size))
+    }
+
+    /// Returns a new array normalized by its L2 norm along the specified axis.
+    ///
+    /// The L2 normalization of an array $x$ is calculated as:
+    ///
+    /// $$x_{norm} = \frac{x}{\max(\|x\|_2, \epsilon)}$$
```
davidkoski: The swift documentation can't render math sadly.

Libraries/MLXEmbedders/Models/Bert.swift
```diff
@@ -3,12 +3,6 @@
 import MLX
 import MLXNN
 
-extension MLXArray {
-    public static func arange(_ size: Int) -> MLXArray {
-        return MLXArray(Array(0 ..< size))
-    }
-}
```
davidkoski, line 10: Good!

Tests/MLXLMIntegrationTests/EmbedderIntegrationTests.swift
```diff
@@ -1,8 +1,10 @@
 // Copyright © 2026 Apple Inc.
 
 import Foundation
-import MLX
+@preconcurrency import MLX
```
davidkoski: Why is this needed?  It is either a sign of a bug in mlx-swift or in this test code.

Libraries/MLXEmbedders/Extensions/MLXArray+Helper.swift
CodebyCR: In my opinion it can vanish too. I think there was only one call in all.
I moved this from Bert to avoid multiple MLXArray extensions.

Libraries/MLXEmbedders/Extensions/MLXArray+Helper.swift
CodebyCR: I think the core idea was to create a proxy Normalizer Scope to generalize and make this more obvious L2Normalisation (and other layers maybe) which should always inlined or inlinable, but it shifts more and more to this extension approach.

~I take a look into Swift-MLX there are [Linalg.norm()](https://github.com/ml-explore/mlx-swift/blob/main/Source/MLX/Linalg.swift#L61) which could solve this.~

Libraries/MLXEmbedders/Extensions/MLXArray+Helper.swift
CodebyCR: Okay I think DocC can handle this.

Libraries/MLXEmbedders/Extensions/MLXArray+Helper.swift
CodebyCR: @davidkoski this method depends on the [Linalg.norm()](https://github.com/ml-explore/mlx-swift/blob/main/Source/MLX/Linalg.swift#L61) method, so moving this to MLX comes with a lot of code movement, but I could make an pull request for Swift- MLX and refactor this occurrence after its merged.

Tests/MLXLMIntegrationTests/EmbedderIntegrationTests.swift
CodebyCR: This happened as I wrote the test. I had imported a deprecated method and the linter judged me, but it is not longer used.

PR #140, opened 2026-03-10
add missing context/toolcall parameters

davidkoski:
## Proposed changes

- see #139

## Checklist

### DIFF
--- a/Libraries/MLXLMCommon/ModelContainer.swift
+++ b/Libraries/MLXLMCommon/ModelContainer.swift
@@ -215,6 +215,7 @@ public final class ModelContainer: Sendable {
     ///
     /// - Parameter messages: Array of message dictionaries with "role" and "content" keys
     /// - Returns: Array of token IDs
+    @available(*, deprecated, message: "Use applyChatTemplate directly on tokenizer")
     public func applyChatTemplate(messages: [[String: String]]) async throws -> [Int] {
         let tokenizer = await self.tokenizer
         return try tokenizer.applyChatTemplate(messages: messages)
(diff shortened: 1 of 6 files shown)


Libraries/MLXVLM/Models/FastVLM.swift
```diff
@@ -989,7 +989,9 @@ public struct FastVLMProcessor: UserInputProcessor {
 
         if input.images.isEmpty {
             // No image scenario
-            let promptTokens = try tokenizer.applyChatTemplate(messages: messages)
+            let promptTokens = try tokenizer.applyChatTemplate(
+                messages: messages, tools: input.tools,
+                additionalContext: input.additionalContext)
```
davidkoski, line 994: I think the VLMs got copy & pasted without the tool and additional context (e.g. thinking) -- tool calling was probably not in place when the first VLM was built.

Anyway, see also #139 which is the same issue.

Libraries/MLXLMCommon/ModelContainer.swift
```diff
@@ -215,6 +215,7 @@ public final class ModelContainer: Sendable {
     ///
     /// - Parameter messages: Array of message dictionaries with "role" and "content" keys
     /// - Returns: Array of token IDs
+    @available(*, deprecated, message: "Use applyChatTemplate directly on tokenizer")
```
davidkoski, line 218: This one is unused in the library -- warn people not to call it because it doesn't have the extra parameters.

angeloskath (approved):
👍

PR #146, opened 2026-03-13
Add parser for GPT-OSS Harmony tool call format

aleroot:
## Proposed changes

Add parser for GPT-OSS Harmony tool call format

## Checklist

### DIFF
--- a/Libraries/MLXLMCommon/ChatSession.swift
+++ b/Libraries/MLXLMCommon/ChatSession.swift
@@ -157,6 +157,11 @@ public final class ChatSession {
         /// nonzero progress is an invalidated ledger and forces a rebuild.
         var cachedTokens: [Int]
 
+        /// Tokens returned by the previous generation but not yet represented
+        /// by `cachedTokens`. This is normally empty; a speculative round can
+        /// leave its final verifier sample here.
+        var uncommittedTokens: [Int] = []
+
         @discardableResult
         mutating func record(
             _ assistant: AssistantGeneration,
@@ -169,6 +174,7 @@ public final class ChatSession {
                 // contain generated or lookahead tokens, so invalidate
                 // the ledger and rebuild from the retained messages.
                 cachedTokens.removeAll()
+                uncommittedTokens.removeAll()
                 return false
             }
 
@@ -176,14 +182,17 @@ public final class ChatSession {
             if processedTokenCount >= cachedTokenCount,
                 processedTokenCount - cachedTokenCount <= generatedTokens.count
             {
+                let committedGeneratedTokenCount = processedTokenCount - cachedTokenCount
                 cachedTokens.append(
-                    contentsOf: generatedTokens.prefix(
-                        processedTokenCount - cachedTokenCount))
+                    contentsOf: generatedTokens.prefix(committedGeneratedTokenCount))
+                uncommittedTokens = Array(
+                    generatedTokens.dropFirst(committedGeneratedTokenCount))
             } else {
                 // The iterator/cache relationship could not be proven
                 // (for example, a speculative iterator retained
                 // un-emitted lookahead).
                 cachedTokens.removeAll()
+                uncommittedTokens.removeAll()
             }
 
             messages.append(
@@ -822,8 +831,18 @@ public final class ChatSession {
 
                     var pendingMessages = inputMessages.consume()
 
+                    // How this model's generated token stream relates to a chat
+                    // template re-render is a property of its response protocol,
+                    // not of the session.
+                    let promptCachePolicy = PromptCacheReusePolicy(
+                        protocolRules: modelConfiguration.toolCallFormat?
+                            .promptCacheReuseRules(tokenizer: tokenizer) ?? [])
+
                     // loop can restart on tool calls
                     restart: while !pendingMessages.isEmpty {
+                        let isToolResultContinuation =
+                            pendingMessages.contains { $0.role == .tool }
+                            && conversation?.messages.last?.tool?.calls?.isEmpty == false
                         let templateMessages: [Chat.Message]
                         let conversationMessageCountBeforePending: Int?
                         if var currentConversation = conversation {
@@ -841,6 +860,17 @@ public final class ChatSession {
                         }
                         let containsNewMedia = pendingMessages.contains {
                             !$0.images.isEmpty || !$0.videos.isEmpty || !$0.audios.isEmpty
+                        }
+                        let structuredToolCallCount = templateMessages.reduce(into: 0) {
+                            count, message in
+                            // GPT-OSS renders at most one call from each
+                            // assistant message. Other protocols ignore this
+                            // model-specific boundary count.
+                            if message.role == .assistant,
+                                message.tool?.calls?.isEmpty == false
+                            {
+                                count += 1
+                            }
                         }
 
                         let userInput = UserInput(
@@ -869,8 +899,9 @@ public final class ChatSession {
                             } ?? false
 
                         var reusedMainCacheWithoutDraft = false
+                        var requiresMainOnlyContinuation = false
                         if var currentConversation = conversation {
-                            let fullTokenIds = input.text.tokens.asArray(Int.self)
+                            let promptTokenIds = input.text.tokens.asArray(Int.self)
                             let cachedTokenIds = currentConversation.cachedTokens
                             assert(
                                 kvCache.nativeAttentionOffsetsAreAligned,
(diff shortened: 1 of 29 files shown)


JaeminKim-amoz (commented):
**COMMENT — good feature, minor concerns**

Well-architected GPT-OSS Harmony parser. The `Mode` enum in `TextToolTokenLoopHandler` cleanly separates concerns. Strong test coverage (600+ lines). `normalizedSendableJSON` handles NSNumber edge cases well.

**Notes:**
- [MEDIUM] `@unchecked Sendable` on `TextToolTokenLoopHandler` with mutating `Mode` enum — value-type semantics should be safe, but worth a second look.
- [LOW] `Swift.Decoder` disambiguation in GPTOSS.swift suggests a namespace collision — consider a typealias.

DePasqualeOrg:

aleroot:
> I've addressed the parsing problems in this library in [Swift LM Response Parser](https://github.com/DePasqualeOrg/swift-lm-response-parser), which parses reasoning and tool-use output from all models implemented in MLX Swift LM and emits a stream of [Open Responses](https://www.openresponses.org) events. I created a demo of it in [this PR](https://github.com/ml-explore/mlx-swift-examples/pull/475) in MLX Swift Examples and submitted a [proposal](https://github.com/ml-explore/mlx-swift-lm/issues/260) for how it could be used with MLX Swift LM.
> 
> If you're able to test it for your use case, please post your feedback here or in an issue in the repo.

I don’t know why we would need another non official custom library for this. Honestly I do not like the idea at all. I won’t tie my project to a random custom library like that

DePasqualeOrg:
That's fine, and it's your choice to make. I created Swift LM Response Parser because I wanted robust parsing and streaming for my own apps, and I think the incremental-improvement path is not going to get to parity within this library anytime soon. MLX LM in Python faces the same problems and is far from parity.

But to address your point about why a separate library would be needed: Fully solving this problem requires tens of thousands of lines of code, including hundreds of tests. Swift LM Response Parser is currently around 40,000 lines of code, half of which are tests mainly ported from vLLM and SGLang. There's no way that volume of code is going to land here.

aleroot:
@davidkoski this is a bit old, but this morning I have rebased it on the current main and reorganised a bit. If you can have a look, I am still using GPT OSS and it would be useful.

Libraries/MLXLMCommon/ChatSession.swift
```diff
@@ -845,8 +849,36 @@ public final class ChatSession {
...
+                                    tokenizer: tokenizer),
+                                let callIndex = fullTokenIds.indices.dropFirst(commonPrefixCount)
+                                    .first(where: { fullTokenIds[$0] == callTokenID })
+                            {
+                                // The generated cache already contains the exact
+                                // analysis and tool-call frame, except for the sampled
+                                // `<|call|>` stop token. The rendered transcript cannot
+                                // reproduce private analysis, so splice it at that commit
+                                // token instead of trimming away the live Harmony context.
+                                let suffix = Array(fullTokenIds[callIndex...])
```
davidkoski: I think there might be a gap here in interaction with the TokenIterator.  The standard token iterator does something like this:

```swift
func next() -> Int? {
    let previousY = y
    let token = step(previousY)
    asynEval(token)
    return previousY.item()
}
```

the goal is to keep the metal pipeline full -- we async eval the next token so it is either ready or partially done when we come to pick up the next token.

The problem is that `step(previousY)` actually commits the _next_ token to the KVCache, even if we don't consume it.  I think hitting the token limit is similar -- it has called step on a token it will never produce.

Do we need to define the contract and give it a set of stop tokens or a callback to check?  I think we would need some tests to check and enforce that.

The speculative token iterators are perhaps even more complicated.

I think.

davidkoski: This seems a bit like https://github.com/ml-explore/mlx-swift-lm/pull/414#discussion_r3693924993 -- the asynchronous and pipelined nature of the iterator needs to be handled precisely by the caller and it is tricky.

I think we can't afford to remove the asyncEval -- we get a good chunk of performance from that.  We might be able to move the `item()` call before we call `step()` so a stop would be a hard stop, but I think the cancel of the iterator case from #414 is still an issue.

Libraries/MLXLMCommon/ChatSession.swift
```diff
@@ -845,8 +849,36 @@ public final class ChatSession {
...
+                            if isHarmonyToolContinuation,
+                                cacheOffset == cachedTokenIds.count,
+                                allMainCachesAreAligned,
+                                allDraftCachesAreAligned,
+                                !containsNewMedia,
+                                input.text.mask == nil,
+                                let callTokenID = HarmonyFrameParser.callTokenID(
+                                    tokenizer: tokenizer),
+                                let callIndex = fullTokenIds.indices.dropFirst(commonPrefixCount)
+                                    .first(where: { fullTokenIds[$0] == callTokenID })
```
davidkoski: This might be an issue:

```swift
let commonPrefixCount = zip(fullTokenIds, cachedTokenIds).prefix { $0 == $1 }.count
```

So this will find the callIndex _after_ the common prefix.  `fullTokenIds` contains the chat template's render of the transcript.  `cachedTokenIds` contains the raw tokenIds from this current turn.  However, `cachedTokenIds` contains (I think) the analysis tokens and those are not in the `fullTokenIds`, so the common prefix stops there -- the token streams are not equivalent.

Again, I think :-)

Libraries/MLXLMCommon/Tool/ToolCallFormat.swift
```diff
@@ -132,6 +140,10 @@ public enum ToolCallFormat: String, Sendable, Codable, CaseIterable {
             return MistralToolCallParser()
         case .llama3:
             return Llama3ToolCallParser()
+        case .gptOSS:
+            preconditionFailure(
+                "GPT-OSS uses the token-level Harmony protocol and cannot be parsed by ToolCallProcessor"
+            )
```
davidkoski: This works with the various calls in Evaluate that use the `TextToolTokenLoopHandler`, but caller that use `RawTokenLoopHandler` (to do their own tool call handling) will run into problems here.  For example the FoundationModel stuff.

Previously if you tried to use a gptOSS model with MLXFoundationModels, there was no `gptOSS` case and it would get `nil` and default to the `json` processor:

```swift
        handler: TextToolTokenLoopHandler(
            tokenizer: tokenizer,
            stopStrings: modelConfiguration.effectiveStopStrings,
            format: modelConfiguration.toolCallFormat ?? .json,
            tools: tools
        )
```

It wouldn't work for tool calls, but if you didn't register tools the model itself would work.

Now that it has a case for gptOSS, these alternate paths won't work -- it will be a fatalError to use the model outside the normal `generateTask()` loop.

Ideally all of this would just work without a special case, but harmony is a special case and it needs to be handled as such.  I do think that alternate callers should still be able to use the model.

Some ideas (some combination or maybe none of these!):

- if gptOSS can't use a normal tool parser, perhaps it shouldn't be a `ToolCallFormat`
    - but it is convenient and it is the same high level concept
- perhaps the tool parsing has to encapsulate the larger scope of cache management
    - and whatever else harmony requires
    - this would allow it to be a real ToolCallParser
- a short term fix: log a warning and return a JSON parser
    - this would leave the model usable and in the same state that it was before

At a minimum I think this PR should allow alternate callers to use the model (without tool calling) as before.

@thechriswebb FYI

Libraries/MLXLMCommon/ChatSession.swift
```diff
@@ -772,6 +772,10 @@ public final class ChatSession {
 
                     // loop can restart on tool calls
                     restart: while !pendingMessages.isEmpty {
+                        let isHarmonyToolContinuation =
+                            modelConfiguration.toolCallFormat == .gptOSS
+                            && pendingMessages.contains { $0.role == .tool }
+                            && conversation?.messages.last?.tool?.calls?.isEmpty == false
                         let templateMessages: [Chat.Message]
```
davidkoski: Not entirely related to this PR, but it is highlighting the issue:

The fact that ChatSession's `streamMap()` is so long suggests a few things:

- there is code in there that _all_ token generation loops need, e.g. handling the tool calling
- the KVCache/session manipulation may be specific to ChatSession but it looks like it might be factored into a policy

I don't think it is required for this PR to solve all of that, but it would be nice if we can come up with a way to refactor this code in the future such that it was done (perhaps) via composition or another layer that encapsulated the handling.

It would be nice if that made it testable such that we could feed it tokens and see the machinery work.  I suppose ChatSession _is_ testable, after a fashion, but it would be nice to be able to unit test the machinery.  It might help with some of the token iterator interaction & complexity.

Again, not for this PR, but thoughts for the future.

davidkoski (changes requested):
Two bugs, maybe.  See comments and let me know what you think.  This is via inspection + claude analysis so I might be off base.

I think we need to make sure gptOSS is still usable e.g. by MLXFoundationModels, which uses RawTokenLoopHandler.  It doesn't need to handle tool calls, but should still be callable, see comment.

Libraries/MLXLMCommon/ChatSession.swift
aleroot: I agreed with this direction and used the feedback to make a scoped refactor in this PR. Token-stream processing is now encapsulated behind a composable TokenStreamDecoder, with standard and Harmony-specific implementations. This removes protocol/tool-call handling from the generic generation loop and allows the machinery to be tested directly by feeding it tokens.  
I left KV-cache and session-restart policy within ChatSession for now, since extracting that cleanly deserves a separate focused change.

aleroot:
> Two bugs, maybe. See comments and let me know what you think. This is via inspection + claude analysis so I might be off base.
> 
> I think we need to make sure gptOSS is still usable e.g. by MLXFoundationModels, which uses RawTokenLoopHandler. It doesn't need to handle tool calls, but should still be callable, see comment.

Thanks for the detailed review. All requested issues should now be addressed, and I also completed the suggested scoped refactor:

- Generic generation no longer depends on Harmony or GPT-OSS details.
- Token parsing, stop handling, and tool-call emission are composed through TokenStreamDecoder.
- Speculative-generation finalization is now an explicit capability, including unreturned lookahead.
- Harmony framing uses token IDs and stricter recipient parsing.

The token-processing machinery has direct regression tests, including raw generation, stop-string ordering, and speculative cache behavior.

This turned out to be much more work than I expected but some of the refactoring can turn out to be useful in the future...

davidkoski (approved):
I like the direction you went with this!

PR #147, opened 2026-03-14
perf: eliminate CPU←GPU sync in penalty processors, optimize TopPSampler

spokvulcan:
## Summary

- **GPU-only penalty processors**: Replace `[Int]` token buffers with GPU-resident `TokenRing` struct using `MLXArray.arange` + `MLX.where` mask writes. Eliminates `token.item(Int.self)` CPU←GPU sync in `didSample()`, restoring `asyncEval()` pipelining.
- **Align sampling with Python mlx-lm**: Replace `softmax` with `logSoftmax`, work in log-prob space throughout. Apply filters in the same order as `mlx_lm.sample_utils.make_sampler`: top_p → min_p → top_k. Each filter operates on the full vocabulary in original token order.
- **Remove `Sendable` from `LogitProcessor`**: `MLXArray` is not `Sendable`; aligns with existing `LogitSampler` precedent.

### Penalty processors
- `TokenRing` struct encapsulates shared ring buffer logic (was duplicated 3×)
- `positions` array via `MLXArray.arange()` computed once in `init` (was CPU-allocated per token)
- `ring` property marked `private` on all penalty structs

### TopPSampler
- `logSoftmax(logits)` instead of `softmax(logits)` — matches Python's `logits - logsumexp(logits)`
- Mask with `-inf` in log-space instead of `0` in linear-prob space — no more `log(0)` round-trip
- Separate `applyTopP`, `applyMinP`, `applyTopK` methods matching `mlx_lm/sample_utils.py`
- `applyTopP` uses `putAlong` for O(V) scatter-back instead of double `argSort`
- `categorical` on `[B, V]` returns `[B]` directly — batch-safe by construction

## Test plan

- [x] Verified correct generation output matches original implementation
- [x] Benchmarked with presence penalty + topK sampling
- [x] swift-format clean, builds with no warnings

### DIFF
--- a/Libraries/MLXLMCommon/Evaluate.swift
+++ b/Libraries/MLXLMCommon/Evaluate.swift
@@ -280,16 +280,14 @@ public struct TopPSampler: LogitSampler {
     }
 
     /// Keep only the top-k highest-probability tokens.
-    /// Matches `apply_top_k` from `mlx_lm/sample_utils.py`.
+    /// Mirrors `apply_top_k` from `mlx_lm/sample_utils.py`.
     private func applyTopK(_ logprobs: MLXArray, topK: Int) -> MLXArray {
         let vocabularySize = logprobs.dim(-1)
         guard topK < vocabularySize else { return logprobs }
-        // O(V) partition: the element at position kth is the top-K boundary.
-        let kth = vocabularySize - topK
-        let partitioned = argPartition(logprobs, kth: kth, axis: -1)
-        let thresholdIdx = partitioned[0..., kth ..< (kth + 1)]
-        let threshold = takeAlong(logprobs, thresholdIdx, axis: -1)
-        return MLX.where(logprobs .>= threshold, logprobs, negInf)
+        // O(V) partition on negated logprobs so top-k land at [0, topK).
+        // Indices at [topK, V) are the tokens to mask out.
+        let maskIndices = argPartition(-logprobs, kth: topK - 1, axis: -1)[0..., topK...]
+        return putAlong(logprobs, maskIndices, values: negInf, axis: -1)
     }
 }
 


Libraries/MLXLMCommon/Evaluate.swift
```diff
@@ -235,10 +238,59 @@ public struct TopPSampler: LogitSampler {
             // Match mlx-lm Python behavior:
             // apply filtering on the base distribution, then apply temperature at sampling time.
             let probs = softmax(logits, axis: -1)
-            let sortedIndices = argSort(probs, axis: -1)
+            let vocabularySize = probs.dim(-1)
```
davidkoski: I am looking at `apply_top_p` from mlx-lm and noticed:

- it has logProbs and has to use exp to get them back to linear space (I think)
- it doesn't use softmax at all

I wonder if these are a problem, softmax especially?  softmax was here before -- I am not sure if that was a problem then or if it was from an earlier version of the code from mlx-lm

Libraries/MLXLMCommon/Evaluate.swift
```diff
@@ -291,64 +332,83 @@ public struct CategoricalSampler: LogitSampler {
...
-        if prompt.shape[0] <= repetitionContextSize {
-            self.tokens = prompt.asArray(Int.self)
+        let n = prompt.dim(0)
+        let promptTokens = prompt.asType(.int32)
+        if n <= repetitionContextSize {
+            if n < repetitionContextSize {
+                let padding = MLXArray.zeros([repetitionContextSize - n], type: Int32.self)
+                tokenRing = concatenated([promptTokens.reshaped(-1), padding])
+            } else {
+                tokenRing = promptTokens.reshaped(-1)
```
davidkoski: This token ring code is duplicated a few times -- I wonder if this should be in a TokenRing struct to encapsulate?

Libraries/MLXLMCommon/Evaluate.swift
```diff
@@ -291,64 +332,83 @@ public struct CategoricalSampler: LogitSampler {
...
+/// Processor that implements a `repetitionPenalty`.
+///
+/// Token history is kept as a GPU-resident ring buffer to avoid CPU←GPU
+/// synchronization in `didSample()`. This preserves `asyncEval()` pipelining
+/// in `TokenIterator`.
 public struct RepetitionContext: LogitProcessor {
-    /// tokens in the repetition context sliding window
-    var tokens = [Int]()
+    /// GPU-resident ring buffer of recent token IDs.
+    var tokenRing: MLXArray
```
davidkoski: This may be a problem:

```swift
public protocol LogitProcessor: Sendable
```

`MLXArray` is not sendable.  It could be treated as such if it were 1) evaluated and 2) never mutated, but that is not the case here (or the other spots that use it like this).

`LogitSampler` is not Sendable -- it is owned by the token iteration and doesn't need to be.  I think that is also true of `LogitProcessor`

- [ ] remove `Sendable` from `LogitProcessor`

Libraries/MLXLMCommon/Evaluate.swift
```diff
@@ -291,64 +332,83 @@ public struct CategoricalSampler: LogitSampler {
...
     mutating public func didSample(token: MLXArray) {
-        if tokens.count >= repetitionContextSize {
-            tokens[index] = token.item(Int.self)
-            index = (index + 1) % repetitionContextSize
-        } else {
-            tokens.append(token.item(Int.self))
-        }
+        // GPU-only: update ring buffer position without CPU←GPU sync.
+        // writeIndex is a CPU-side counter (just an integer position, not a token value).
+        let positions = MLXArray(0 ..< Int32(repetitionContextSize))
```
davidkoski: This could be computed once in `init()` and should probably use `arange()` as that will be done on the GPU.  There are a couple copies of this -- it should probably go into TokenRing or be updated in all 3 places.

Libraries/MLXLMCommon/Evaluate.swift
```diff
@@ -235,10 +238,59 @@ public struct TopPSampler: LogitSampler {
...
+                        let keepMask = sortedProbs .>= (maxProbs * minP)
+                        filteredProbs = MLX.where(
+                            keepMask, filteredProbs, zeros(like: filteredProbs))
+                    }
+
+                    // Always keep the max-probability token as a valid sampling candidate.
+                    filteredProbs[0..., -1] = sortedProbs[0..., -1]
+
+                    let sampledLocal = categorical(log(filteredProbs) * (1 / temp))
+                    return sortedIndices.squeezed(axis: 0)[sampledLocal]
```
davidkoski: I think this assumes batch size of 1, which will change in #150 (I presume, I haven't looked at that in detail yet).

Claude suggests:

```swift
  return takeAlong(sortedIndices, sampledLocal.expandedDimensions(axis: -1), axis: -1)
      .squeezed(axis: -1)                                                                                                   
```
                  
>   expandedDimensions(axis: -1) adds a trailing dim ([B] → [B, 1]) without assuming anything about batch size, takeAlong
>  gathers the one index per batch row, then squeezed(axis: -1) removes only that trailing dim — not the batch dim.

Libraries/MLXLMCommon/Evaluate.swift
spokvulcan: Aligned sampling with Python mlx-lm

Libraries/MLXLMCommon/Evaluate.swift
spokvulcan: I restructured the sampler to match Python's composable filter chain (top_p → min_p → top_k), batch safe by construction

PR #149, opened 2026-03-15
Fix Qwen35 VLM crash on text-only inference

### LINKED ISSUE #148: Qwen3.5 MoE VLM crashes on text-only inference (SmallVector out of range)
### Description

Loading `mlx-community/Qwen3.5-35B-A3B-4bit` via `VLMModelFactory.shared.loadContainer()` works fine — the container loads and weights map correctly. But the first text-only inference call crashes with:

```
MLX/ErrorHandler.swift:343: Fatal error: SmallVector out of range.
at mlx-c/mlx/c/array.cpp:335
```

The same model loads and runs fine through `LLMModelFactory`.

### Root Cause

`Qwen3VLLanguage.getRopeIndex()` ([Qwen3VL.swift L1341](https://github.com/ml-explore/mlx-swift-lm/blob/main/Libraries/MLXVLM/Models/Qwen3VL.swift#L1341)) assumes `inputIds` is always 2D `[batch, seq]`:

```swift
let (batchSize, seqLength) = (inputIds.dim(0), inputIds.dim(1))
```

When the VLM container is used for text-only generation (no images), the token tensor can be 1D `[seq]`. `inputIds.dim(1)` on a 1D array triggers the SmallVector fatal error.

The call path is:
1. `Qwen35.prepare(input:cache:windowSize:)` — no image/video, so `pixelValues` is nil
2. `languageModel(inputIds, ...)` — forwards raw token IDs
3. `Qwen35Language.LanguageModel.callAsFunction` — `positionIds == nil`, `ropeDeltas == nil`, so enters the `getRopeIndex` branch
4. `Qwen3VLLanguage.getRopeIndex(inputIds:...)` — crashes on `inputIds.dim(1)` becasue the tensor is 1D

This shows up concretely when `WiredMemoryUtils.tune()` runs its prefill measurement — it creates a plain `LMInput(tokens:)` with 1D tokens, calls `model.prepare()`, and the VLM path crashes.

### Reproduction

```swift
import MLXLMCommon
import MLXVLM

let config = ModelConfiguration(id: "mlx-community/Qwen3.5-35B-A3B-4bit")
let container = try await VLMModelFactory.shared.loadContainer(configuration: config)

// This crashes:
try await container.perform { context in
    try await WiredMemoryUtils.tune(
        context: context,
        tokenCount: 512,
        parameters: GenerateParameters(maxTokens: 1)
    )
}
```

### Suggested Fix

The simplest fix would be to ensure `inputIds` is 2D before calling `getRopeIndex`. Something like:

```swift
// In Qwen35Language.LanguageModel.callAsFunction, before calling getRopeIndex:
var inputsForRope = inputs
if inputsForRope.ndim == 1 {
    inputsForRope = inputsForRope.expandedDimensions(axis: 0)
}
```

Or alternatively, `getRopeIndex` itself could handle 1D input gracefully.

### Environment

- mlx-swift-lm: main branch (HEAD = bc3c20e)
- mlx-swift: main branch (HEAD = b6e128c)  
- Model: `mlx-community/Qwen3.5-35B-A3B-4bit` (model_type: `qwen3_5_moe`)
- macOS 26.0, Apple M4 Max 96GB

dirvine:
## Problem

Loading a Qwen3.5 MoE model (e.g. `mlx-community/Qwen3.5-35B-A3B-4bit`) via `VLMModelFactory` works fine — the containr loads successfully. But the first text-only inference call crashes with:

```
Fatal error: SmallVector out of range.
at mlx-c/mlx/c/array.cpp:335
```

The same model loads and runs without issue through `LLMModelFactory`.

## Root Cause

`Qwen35Language.LanguageModel.callAsFunction` passes `inputs` directly to `Qwen3VLLanguage.getRopeIndex()`, which does:

```swift
let (batchSize, seqLength) = (inputIds.dim(0), inputIds.dim(1))
```

When the VLM container is used for text-only generation (no images), callers like `WiredMemoryUtils.tune()` and `TokenIterator` pass 1D `[seq]` token tensors. `dim(1)` on a 1D array is out of range → fatal crash.

The same issue affects `inputs.dim(1)` at lines 946 and 966 in the same function.

## Fix

Add an `ndim` check at the top of `callAsFunction` to expand 1D inputs to 2D `[1, seq]` before any dimension-dependent logic runs. This matches what the LLM path does via `tokens[text: .newAxis]` in `prefillOnly`.

## Testing

Verified locally with `mlx-community/Qwen3.5-35B-A3B-4bit` on M4 Max 96GB:
- Container loads via `VLMModelFactory` ✅
- `WiredMemoryUtils.tune()` completes (was crashing) ✅  
- Text inference produces correct output ✅
- Full e2e test suite passes ✅

Fixes #148

### DIFF
--- a/Libraries/MLXVLM/Models/Qwen35.swift
+++ b/Libraries/MLXVLM/Models/Qwen35.swift
@@ -921,6 +921,10 @@ enum Qwen35Language {
             imageGridTHW: [THW]? = nil,
             videoGridTHW: [THW]? = nil
         ) -> LMOutput {
+            // Ensure inputs is 2D [batch, seq]. Text-only callers (e.g.
+            // WiredMemoryUtils, TokenIterator) may pass 1D token arrays.
+            let inputs = inputs.ndim == 1 ? inputs.expandedDimensions(axis: 0) : inputs
+
             if pixelValues != nil {
                 precomputedPositionIds = nil
                 ropeDeltas = nil


Libraries/MLXVLM/Models/Qwen35.swift
```diff
@@ -921,6 +921,10 @@ enum Qwen35Language {
             imageGridTHW: [THW]? = nil,
             videoGridTHW: [THW]? = nil
         ) -> LMOutput {
+            // Ensure inputs is 2D [batch, seq]. Text-only callers (e.g.
+            // WiredMemoryUtils, TokenIterator) may pass 1D token arrays.
```
davidkoski, line 925: I wonder if that is incorrect (as in this is the bug).  I would expect the shape to always be `[B, S]`.

So:

- are these paths buggy in that they allow 1d arrays?
- if not, perhaps the callers should ensure the correct shape -- otherwise this is a one-off fix for a single model

davidkoski (changes requested):
Please look at my comment on the shape -- let me know what you think.  I suspect it is a caller issue, but not 100% sure.

JaeminKim-amoz (commented):
**COMMENT (effectively APPROVE)**

Excellent surgical fix — 4 lines with a clear comment explaining "why." The `ndim` check mirrors how the LLM path handles this. Tested on hardware with the exact model that crashed. No concerns.

Libraries/MLXVLM/Models/Qwen35.swift
dirvine: I rechecked the shape question on this branch (`d536384`). Concrete findings:

- The normal processor path is already batched: `Qwen3VLProcessor.prepare(input:)` creates `MLXArray(promptTokens).expandedDimensions(axis: 0)` plus a matching mask for text-only/no-media input.
- However, the generic text-only APIs can still deliver 1D tokens to `Qwen35.prepare` before any caller-side batching happens:
  - `LMInput(tokens:)` just stores the `MLXArray` it is given; it does not add a batch axis.
  - `WiredMemoryUtils.makeTokenInput(...)` creates `LMInput(tokens: MLXArray(tokenIDs))`, i.e. `[seq]`.
  - `WiredMemoryUtils.prefillOnly(...)` calls `model.prepare(input, ...)` first; its `tokens[text: .newAxis]` batching is only in the `.tokens` result branch after `prepare` returns.
  - `TokenIterator` has the same prepare-first structure: `step(previous:)` adds `.newAxis`, but only after `model.prepare(...)`. `Qwen35.prepare` returns `.logits` and does the prefill inside `prepare`, so it never benefits from that later `step` normalization for the crashing first call.
- In `Qwen35.prepare`, `let inputIds = input.text.tokens` is forwarded directly to `languageModel(...)`. From there, the first-position path calls `Qwen3VLLanguage.getRopeIndex(...)`, whose first line reads `inputIds.dim(1)`. With a 1D `[seq]` array, that is the `SmallVector out of range` crash.

So I don't think this is just an external caller bug: the package itself has public/generic text-only paths that create or accept 1D `LMInput`, and Qwen35 is the VLM path that consumes the prefill inside `prepare` before the usual generation code can add the batch dimension. Normalizing at the Qwen35 language boundary covers both `Qwen35.prepare(...)` and the simplified `Qwen35.callAsFunction(_ inputs:cache:)` path.

Scope note: I also checked #157. That is a separate mutable-position-state issue (`precomputedPositionIds` / `ropeDeltas` leaking across evaluations) and can still produce broadcast-shape failures even when inputs are already 2D. This PR is only the narrow #148 fix for the 1D `dim(1)` / `SmallVector` crash; #157 still needs the broader state fix.

Validation from my local check today:
- PR checks are green (`lint`, `mac_build_and_test`).
- `swift build` passes locally on this branch.
- Full `swift test` builds successfully locally, then the runtime test run fails in my environment with MLX's `Failed to load the default metallib` error; that appears environment/toolchain related rather than caused by this four-line change.

davidkoski: Thanks for the followup!

FWIW I think Qwen3VL has the same issue

davidkoski (approved):
Thank you for the fix!

PR #150, opened 2026-03-16
Add continuous batching support for concurrent LLM inference

ronaldmannak:
## Proposed changes

- **Continuous batching engine** that transparently serves multiple concurrent requests in a shared decode loop, with zero overhead for single requests
- **BatchKVCache & BatchRotatingKVCache** implementations with left-padding strategy for batched attention across sequences of different lengths
- **InferenceScheduler** actor that starts in single-request mode (`TokenIterator`) and automatically upgrades to `BatchTokenIterator` when a second request arrives; 3rd+ requests join the existing batch on the fly
- **LRUPromptCache** with trie-based prefix matching and LRU eviction for reusing KV state across requests with shared prefixes
- **Batch-aware RoPE** via the new `applyRotaryPosition` helper — all 45 modified `MLXLLM` model files with RoPE-based attention paths now use batch-aware position handling that works in both single and batched modes
- **Automatic fallback** for incompatible requests (VLMs, hybrid/SSM models using `MambaCache` or `CacheList`, and quantized KV-cache requests) — routed to the single-request path with no caller changes needed

## Design

This PR adds batching support for text-generation LLMs. VLM requests are not batched in this PR, and embedding models are out of scope for this batching path.

Most LLM model types are supported. Batched generation is currently limited to models whose cache stack uses `KVCacheSimple` and/or `RotatingKVCache`, and is unavailable for models that use `MambaCache` or `CacheList`. Requests using those models automatically fall back to the single inference path. Support for `MambaCache` and `CacheList` will be added in a separate PR.

`QuantizedKVCache` is also not supported for batching. Requests with `kvBits != nil`, or requests that already carry quantized KV cache state, automatically fall back to the single inference path.

The batching system is opt-in via `ModelContainer.scheduler`:

```swift
let container = ModelContainer(context: context)
container.scheduler = InferenceScheduler()
// Existing generate() calls now batch transparently when the request is compatible
```

Incompatible models that use `MambaCache` and/or `CacheList` are:
- falcon_h1
- baichuan_m1
- lfm2
- lfm2_moe
- granitemoehybrid
- qwen3_next
- qwen3_5, qwen3_5_text, qwen3_5_moe
- nemotron_h
- jamba_3b

## Checklist

### DIFF
--- a/Libraries/MLXLMCommon/ModelFactory.swift
+++ b/Libraries/MLXLMCommon/ModelFactory.swift
@@ -367,11 +367,11 @@ final public class ModelFactoryRegistry: @unchecked Sendable {
     private init() {
         self.trampolines = [
             {
-                (NSClassFromString("MLXLLM.TrampolineModelFactory") as? ModelFactoryTrampoline.Type)?
+                (NSClassFromString("MLXVLM.TrampolineModelFactory") as? ModelFactoryTrampoline.Type)?
                     .modelFactory()
             },
             {
-                (NSClassFromString("MLXVLM.TrampolineModelFactory") as? ModelFactoryTrampoline.Type)?
+                (NSClassFromString("MLXLLM.TrampolineModelFactory") as? ModelFactoryTrampoline.Type)?
                     .modelFactory()
             },
         ]


davidkoski:
Haha, yes, that might be a good idea -- I think file-wise there are a ton of changes that can happen in the models independent of the batching, e.g.

```
        queries = applyRotaryPosition(rope, to: queries, cache: cache)
        keys = applyRotaryPosition(rope, to: keys, cache: cache)
```

that might cut down the size, but whatever you think is best.

ronaldmannak:
Great. Let me start with that and I'll close this PR

Issue #123, opened 2026-02-27
[BUG] KVCache can be reloaded but not used with ChatSession


davidkoski:
**Describe the bug**

See #45 and #121.  The first item added a way to restore a chat session from messages and we purposefully left off restore from KVCache.

But perhaps we should revisit that since we do have a way to save and restore the `KVCache`:

```swift
public func loadPromptCache(
    url: URL
) throws -> ([KVCache], [String: String]?) {
```

You can use it with the lower level generation APIs but not with ChatSession.

PR #151, opened 2026-03-18
Add KV cache initializers and cache access to ChatSession

alankessler:
## Summary

Addresses #123 — `loadPromptCache()` had no path back into `ChatSession`.

- Adds two new initializers that accept a pre-built `[KVCache]` (one for `ModelContainer`, one for `ModelContext`), enabling prefix caching across sessions
- Cache parameter is `consuming` to prevent the caller from reusing shared mutable cache references
- Parameter order matches the `history` initializers: `instructions` before `cache`
- Adds `currentCache() async -> [KVCache]?` to read the live cache after generation
- Adds `saveCache(to url: URL) async throws` to persist the cache to disk as `.safetensors`
- Adds `ChatSessionError: LocalizedError` with `noCacheAvailable` case
- Doc comments warn against passing `instructions` when the cache already encodes a system prompt

The `.kvcache` branch in `streamMap` already handles an injected cache correctly — no changes to generation logic required.

## Use case

Build a KV cache from a long shared context (system prompt + document) once, save it to disk, and restore it across multiple sessions to skip re-prefilling the same tokens each time.

```swift
// Build and save
let session = ChatSession(modelContainer)
_ = try await session.respond(to: systemPrompt)
try await session.saveCache(to: cacheURL)

// Restore in a future session
let (cache, _) = try loadPromptCache(url: cacheURL)
let session = ChatSession(modelContainer, cache: cache)
// generation starts from the cached prefix
```

## Test plan

- `testCurrentCacheNilBeforeGeneration` — nil before any generation
- `testCurrentCacheAfterGeneration` — non-nil after generation
- `testCurrentCacheNilForHistorySessionBeforeGeneration` — nil for history session before generation
- `testCurrentCacheNonNilForHistorySessionAfterGeneration` — non-nil after generation from history session
- `testCurrentCacheNilAfterClear` — nil after `clear()`
- `testInitWithKVCache` — restored session generates correctly
- `testSaveCacheThrowsBeforeGeneration` — throws `noCacheAvailable` before generation
- `testSaveAndRestoreCache` — full save → load → restore round-trip

### DIFF
--- a/Libraries/MLXLMCommon/ChatSession.swift
+++ b/Libraries/MLXLMCommon/ChatSession.swift
@@ -164,11 +164,17 @@ public final class ChatSession {
     /// system prompt and document) once, save it via ``saveCache(to:)``, and restore it
     /// across multiple sessions to avoid re-prefilling the same tokens each time.
     ///
+    /// > Important: If the cache was built from a session that already included system
+    /// > instructions, do not pass the same `instructions` here — they would be
+    /// > re-tokenized on each call to ``respond(to:role:images:videos:)`` without matching
+    /// > KV state, producing incoherent output.
+    ///
     /// - Parameters:
     ///   - model: the ``ModelContainer``
-    ///   - cache: a ``[KVCache]`` previously obtained from ``saveCache(to:)`` or
-    ///     ``currentCache()``
-    ///   - instructions: optional system instructions for the session
+    ///   - instructions: optional system instructions for the session — leave `nil` if the
+    ///     cache already encodes a system prompt
+    ///   - cache: a non-empty ``[KVCache]`` previously obtained from ``saveCache(to:)`` or
+    ///     ``currentCache()``, matching the given model
     ///   - generateParameters: parameters that control generation
     ///   - processing: media processing configuration for images/videos
     ///   - tools: optional tool specifications
@@ -176,8 +182,8 @@ public final class ChatSession {
     ///   - additionalContext: optional model-specific context
     public init(
         _ model: ModelContainer,
-        cache: [KVCache],
         instructions: String? = nil,
+        cache: consuming [KVCache],
         generateParameters: GenerateParameters = .init(),
         processing: UserInput.Processing = .init(resize: CGSize(width: 512, height: 512)),
         additionalContext: [String: any Sendable]? = nil,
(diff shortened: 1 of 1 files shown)


Libraries/MLXLMCommon/ChatSession.swift
```diff
@@ -158,6 +158,78 @@ public final class ChatSession {
...
+    ///     ``currentCache()``
+    ///   - instructions: optional system instructions for the session
+    ///   - generateParameters: parameters that control generation
+    ///   - processing: media processing configuration for images/videos
+    ///   - tools: optional tool specifications
+    ///   - toolDispatch: optional tool dispatch -- required for toolcalls if streaming strings rather than details
+    ///   - additionalContext: optional model-specific context
+    public init(
+        _ model: ModelContainer,
+        cache: [KVCache],
```
davidkoski: One big concern: this is passing a pointer to shared mutable data.  I think this needs to be `consuming`.  That will have follow-on effects on `KVCache` itself -- it will either have to be loaded from disk multiple times OR it needs a way to clone itself (see https://github.com/ml-explore/mlx-swift/pull/367 -- this is not yet in a tag).

alankessler: Good call — fixed. The cache parameter is now `consuming`, and I reordered it after `instructions` to match the `history` initializers.

I also tightened the docs: the `instructions` parameter warns against re-passing a system prompt that's already baked into the cache (since those tokens would be re-tokenized without matching KV state), and `currentCache()` notes that the returned references shouldn't be used concurrently with active generation.

Re: cloning — I have a `KVCache.copy()` implementation ready (protocol requirement + all subclass implementations using `state`/`metaState`, no dependency on a new mlx-swift tag) but kept it out of this PR to stay scoped. Happy to put it up as a follow-on if you'd like.

davidkoski: I think we need the copy to be able to use this correctly -- it might make sense to have that PR go in first.

I am trying to wrap up the mlx-swift side (among other things) so we could potentially use that fix, but if you have something that works, that is ok too.

alankessler: Done — #158 has the copy() implementation. Uses constructor params + state/metaState so it doesn't need a new mlx-swift tag. Once that's in, this PR should be ready to go.

PR #155, opened 2026-03-21
Fix prompt-cache round-trip support for `ArraysCache`, `MambaCache`, and `CacheList`

ronaldmannak:
## Proposed changes

This change keeps the existing public `savePromptCache` / `loadPromptCache` API intact and extends cache reconstruction so composite and array-backed caches can be restored correctly from persisted prompt caches.

## What changed

- `ArraysCache` now serializes enough metadata in `metaState` to preserve:
  - total slot count
  - sparse slot positions
  - `leftPadding`
- `MambaCache` now round-trips through the same array-cache restore path while preserving its concrete type
- `CacheList` now serializes child cache structure in `metaState` and reconstructs children recursively on load
- prompt-cache save/load continues to use the existing safetensors metadata format and Python-compatible cache class names

## Tests

Added round-trip coverage for:

- sparse `ArraysCache`
- `ArraysCache` with `leftPadding`
- `MambaCache` type preservation
- `CacheList` with KV-only children
- `CacheList` with hybrid children (`MambaCache` + `KVCacheSimple`)

## Notes

- No public API changes
- No changes to `LRUPromptCache`, direct-generation integration, or batching behavior in #150 
- This PR is focused on persisted prompt-cache correctness only

## Checklist

### DIFF
--- a/Libraries/MLXLMCommon/KVCache.swift
+++ b/Libraries/MLXLMCommon/KVCache.swift
@@ -1269,7 +1269,7 @@ public class CacheList: BaseKVCache {
 
     public override func copy() -> any KVCache {
         let copiedCaches = caches.map { $0.copy() }
-        let new = CacheList(copiedCaches)
+        let new = CacheList(caches: copiedCaches)
         return new
     }
 
--- a/Libraries/MLXLMCommon/KVCache.swift
+++ b/Libraries/MLXLMCommon/KVCache.swift
@@ -1165,7 +1165,9 @@ public class ArraysCache: BaseKVCache {
             return result
         }
         set {
-            assertionFailure("ArraysCache.metaState should not be set directly. Use restoreFromMetaState() instead")
+            assertionFailure(
+                "ArraysCache.metaState should not be set directly. Use restoreFromMetaState() instead"
+            )
         }
     }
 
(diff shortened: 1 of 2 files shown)


Libraries/MLXLMCommon/KVCache.swift
```diff
@@ -1089,6 +1089,63 @@ public class ArraysCache: BaseKVCache {
...
+                "\(cache.count)",
+                presentSlotIndices.map(String.init).joined(separator: ","),
+            ]
+            if let lp = leftPaddingValues {
+                result.append(lp.map(String.init).joined(separator: ","))
+            }
+            return result
+        }
+        set {
+            // Handled by restoreFromMetaState
```
davidkoski: Is this OK that it is a NOP?  I don't see any calls to set in the code, but other KVCache types do allow set.  Should this be a `fatalError()` instead -- is it a bug to call the setter on CacheList?

davidkoski (approved):
This looks good, thank you.  See what you think of my question -- not a blocker for merge.

Libraries/MLXLMCommon/KVCache.swift
ronaldmannak: @davidkoski The `set` is required to conform to the `KVCache` protocol. The same NOP pattern exists in `CacheList.metaState` (line 1220) as well, btw. I agree a silent NOP could bite us in the future and I've replaced both with `assertionFailure()` so they trigger in debug builds but are stripped in release. Happy to switch to `fatalError()` if you'd prefer a hard crash in all builds.

davidkoski:
GitHub is showing conflicts here -- I think that is the one thing blocking this.

ronaldmannak:
@davidkoski Fixed

PR #156, opened 2026-03-23
Add model-defined pooling fallback for embedding models

sxy-trans-n:
## Proposed changes

This PR fixes embedding pooling for models that do not include sentence-transformers pooling metadata such as `1_Pooling/config.json`.

In `MLXEmbedders`, pooling previously fell back to `.none` when `1_Pooling/config.json` was missing. For Qwen3 embedding models, that meant returning raw hidden states instead of a fixed-size sentence embedding.

This PR adds a model-defined pooling fallback and uses it for Qwen3 embedding models. It also fixes `.last` pooling so it selects the last non-padding token instead of always taking the final sequence position.

Concretely, this change:

- adds `poolingStrategy` to `EmbeddingModel`
- makes `loadPooling(...)` prefer `1_Pooling/config.json`, then fall back to the model-defined pooling strategy
- declares Qwen3 embedding pooling as `.last`
- makes `.last` pooling respect the attention mask

This fixes cases where converted embedding repos are missing `1_Pooling/config.json` and would otherwise produce variable-length embeddings depending on token count.

Validated end-to-end in swama with `mlx-community/Qwen3-Embedding-0.6B-4bit-DWQ`:
- `""` returns 1024 dims
- `"hello world"` returns 1024 dims
- mixed-length batches also return stable 1024-dim embeddings

## Checklist

### DIFF
--- a/Libraries/MLXEmbedders/Pooling.swift
+++ b/Libraries/MLXEmbedders/Pooling.swift
@@ -152,12 +152,14 @@ public class Pooling: Module {
             pooled = inputs.hiddenStates![0..., 0, 0...]
         case .last:
             let hiddenStates = inputs.hiddenStates!
-            let tokenCounts = sum(_mask.asType(.int32), axis: -1).asArray(Int32.self)
-            pooled = stacked(
-                tokenCounts.enumerated().map { batchIndex, count in
-                    let tokenIndex = max(Int(count) - 1, 0)
-                    return hiddenStates[batchIndex, tokenIndex, 0...]
-                })
+            let tokenCounts = sum(_mask.asType(.int32), axis: -1)
+            let tokenIndices = MLX.maximum(
+                tokenCounts - MLXArray(Int32(1)),
+                MLXArray(Int32(0))
+            )
+            let indices = tokenIndices.expandedDimensions(axes: [1, 2])
+            let gathered = MLX.takeAlong(hiddenStates, indices, axis: 1)
+            pooled = MLX.squeezed(gathered, axis: 1)
         case .cls:
             pooled =
                 inputs.pooledOutput


Libraries/MLXEmbedders/Pooling.swift
```diff
@@ -146,7 +151,13 @@ public class Pooling: Module {
...
             pooled = inputs.hiddenStates![0..., 0, 0...]
         case .last:
-            pooled = inputs.hiddenStates![0..., -1, 0...]
+            let hiddenStates = inputs.hiddenStates!
+            let tokenCounts = sum(_mask.asType(.int32), axis: -1).asArray(Int32.self)
+            pooled = stacked(
+                tokenCounts.enumerated().map { batchIndex, count in
+                    let tokenIndex = max(Int(count) - 1, 0)
+                    return hiddenStates[batchIndex, tokenIndex, 0...]
+                })
```
davidkoski: This requires a read back from the GPU (the asArray) which may not be desired here.

Maybe something like this, but check my math:

```swift
// shape: [batch]
let tokenCounts = sum(_mask.asType(.int32), axis: -1)

// shape: [batch]
let tokenIndices = maximum(tokenCounts - 1, 0)

// shape: [batch, 1, 1]
let indices = expandedDimensions(tokenIndices, axes: [1, 2])

// hidden states shape: [batch, sequence, hidden]
// gathered shape: [batch, 1, hidden]
let gathered = takeAlong(hiddenStates, indices, axis: 1)

// squeeze down to [batch, hidden]
pooled = squeezed(gathered, axis: 1)
```

davidkoski (changes requested):
- [ ] don't eval in pooling if we can avoid it

Libraries/MLXEmbedders/Pooling.swift
sxy-trans-n: Good catch. I updated this to use pure MLX tensor ops with `takeAlong` instead of reading indices back with `asArray`, so `.last` pooling now stays on-device and still selects the last non-padding token.

PR #158, opened 2026-03-24
Add copy() to KVCache protocol and all implementations

alankessler:
Follows up on the discussion in #151 — with `consuming` on the cache init, there's no way to reuse a prefix cache across multiple sessions without reloading from disk. This adds `copy()` so you can do `cache.map { $0.copy() }` instead.

Implementation uses constructor params + `state`/`metaState` roundtrip, so there's no dependency on a new mlx-swift tag. Empty caches are handled (the state setters fatalError on empty arrays, so copy() guards against that).

One visibility change: `ArraysCache.leftPadding` goes from `private` to `internal` so `MambaCache.copy()` can preserve it.

**Tests:**
- Parameterized copy independence test across all 6 cache types
- Empty cache copy (no crash)
- CacheList with mixed sub-cache types

### DIFF
--- a/Libraries/MLXLMCommon/KVCache.swift
+++ b/Libraries/MLXLMCommon/KVCache.swift
@@ -431,7 +431,7 @@ public class KVCacheSimple: BaseKVCache, CustomDebugStringConvertible {
         new.step = self.step
         let s = self.state
         if !s.isEmpty {
-            new.state = s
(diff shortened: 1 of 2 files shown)


davidkoski (approved):

davidkoski:
And that is tagged, let's go!

PR #163, opened 2026-03-27
Guard embedder inputs against maxPositionEmbeddings

### LINKED ISSUE #62: [BUG] MLXEmbedders need to respect max token counts
**Describe the bug**

See #60 

**To Reproduce**

In the repro for that issue there was a string that produced 757 tokens and it ran it through a BertModel with:

  "max_position_embeddings": 512,

```
private class BertEmbedding: Module {

    @ModuleInfo(key: "word_embeddings") var wordEmbeddings: Embedding
...
    @ModuleInfo(key: "position_embeddings") var positionEmbeddings: Embedding

    init(_ config: BertConfiguration) {
        _wordEmbeddings.wrappedValue = Embedding(
            embeddingCount: config.vocabularySize, dimensions: config.embedDim)
        _positionEmbeddings.wrappedValue = Embedding(
            embeddingCount: config.maxPositionEmbeddings,
            dimensions: config.embedDim)
    }

    func callAsFunction(
        _ inputIds: MLXArray,
        positionIds: MLXArray? = nil,
        tokenTypeIds: MLXArray? = nil
    ) -> MLXArray {
        let posIds = positionIds ?? broadcast(MLXArray.arange(inputIds.dim(1)), to: inputIds.shape)
        var words = wordEmbeddings(inputIds) + positionEmbeddings(posIds)
```

Either the model should be truncating the text or guarding against it or something in the layer above should do so.

alankessler:
Fixes #62.

BERT models crash when input exceeds `maxPositionEmbeddings` because the position embedding table is fixed-size. Per the discussion between @davidkoski and @anishbasu, this truncates with a warning rather than crashing.

Also exposes `maxPositionEmbeddings` on the `EmbeddingModel` protocol (default `nil`, non-breaking) so callers who want to chunk or split long inputs can check the limit themselves.

Qwen3 and Gemma3 use RoPE — no change needed.

### DIFF
--- a/Libraries/MLXEmbedders/EmbeddingModel.swift
+++ b/Libraries/MLXEmbedders/EmbeddingModel.swift
@@ -14,6 +14,14 @@ public protocol EmbeddingModel: BaseLanguageModel {
     var vocabularySize: Int { get }
     var poolingStrategy: Pooling.Strategy? { get }
 
+    /// The maximum number of position embeddings supported by this model, or `nil`
+    /// if the model uses a position encoding (e.g. RoPE) that handles arbitrary lengths.
+    ///
+    /// Inputs exceeding this length are truncated internally with a warning.
+    /// Callers may pre-truncate for efficiency or to implement custom strategies
+    /// (e.g. chunking with pooling).
+    var maxPositionEmbeddings: Int? { get }
+
     func callAsFunction(
         _ inputs: MLXArray,
         positionIds: MLXArray?,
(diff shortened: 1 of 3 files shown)


JaeminKim-amoz (approved):
**APPROVE**

Prevents a real crash when inputs exceed `maxPositionEmbeddings`. Clean API design with protocol default of `nil` (non-breaking for RoPE models). Correctly truncates all related tensors together.

**Minor notes:**
- [LOW] Uses `print()` for warning instead of `os.Logger` — inconsistent with repo convention
- [LOW] Near-identical truncation logic duplicated between `Bert.swift` and `NomicBert.swift` — could be extracted to a shared helper

davidkoski (approved):
changes look good, thank you!

Issue #157, opened 2026-03-23
[BUG] Qwen 3.5 VLM model crashes on subsequent requests with manual KV cache prefix matching


neon52:
**Describe the bug**
When using VLMModelFactory to load a multimodal model (specifically Qwen 3.5), reusing a ModelContainer for multiple sequential requests fails if manual KV cache prefix matching is employed.

While this pattern (finding a matching KV cache "slot" and initializing a TokenIterator with remainingTokens) works reliably for text-only models, VLM models appear to maintain internal state that is not cleared or updated correctly between perform blocks unless the entire ModelContainer is reloaded via VLMModelFactory.shared.loadContainer().

**To Reproduce**
1. Load a Qwen 3.5 VLM model (for example: https://huggingface.co/mlx-community/Qwen3.5-9B-MLX-4bit) using VLMModelFactory.shared.loadContainer.
2. Perform the first inference request using any KV cache (it will always succeed since it's a fresh ModelContainer)
3. For the 2nd request load a KV cache that is different than what was used for the 1st request and attempt to initialize a TokenIterator for the remaining tokens.  

On step 3 the app will crash with a broadcast_shapes mismatch error.  Note that I can workaround the issue by completely reloading the model before step 3, but this somewhat defeats the purpose of KV cache reuse.

Example code snippet
```swift
// Current workaround: uncommenting the lines below for VLM models only makes everything work
//let freshModelContainer = try await VLMModelFactory.shared.loadContainer(configuration: modelContainer.configuration)
//modelContainer = freshModelContainer

// Find the best matching KV cache or create a new one
let preparedUserInput = try await modelContainer.prepare(input: userInput)
let preparedTokens = preparedUserInput.text.tokens.asArray(Int.self)
let (bestCache, remainingTokens) = findBestCache(preparedTokens)

// Perform inference using the loaded KV cache
await modelContainer.perform { context in
    let tokenIterator = try TokenIterator(
        input: LMInput(tokens: MLXArray(remainingTokens).reshaped([1, -1])), 
        model: context.model,
        cache: bestCache
    )
    generateTokenTask(
        promptTokenCount: preparedTokens.count, // use the full prompt size here instead of remainingTokens.count so the stats are correct
        modelConfiguration: context.configuration,
        tokenizer: context.tokenizer,
        iterator: tokenIterator
    )
}
```

[...]

PR #164, opened 2026-03-27
feat: add ParoQuant (pairwise rotation quantization) support

spokvulcan:
## Summary

Adds support for [ParoQuant](https://arxiv.org/abs/2511.10645) (ICLR 2026) INT4-quantized models. ParoQuant uses learned pairwise Givens rotations to suppress weight outliers before quantization, achieving ~2.4% accuracy improvement over standard AWQ on reasoning tasks.

**What's included:**

- `RotateQuantizedLinear` — applies channel-scaling + pairwise Givens rotation to activations at runtime via Metal kernel, then feeds into standard `quantizedMatmul`
- `ParoQuantLoader` — loads AutoAWQ-format checkpoints, converts weight layout, and patches rotation layers with their learned parameters (theta, pairs, channel_scales)
- Qwen 3.5 fixes — `in_proj_ba` split in sanitize for PARO checkpoint compatibility, hybrid cache handling for mixed attention/Mamba layers
- 7 unit tests (pair packing, AWQ conversion, quantization round-trip)

## Agent benchmark results

We run a 14-scenario tool-calling benchmark (file reads, edits, multi-step instructions, clarification requests) on Apple Silicon. Results are multi-run averages.

| Model | Runs | Passed | Tool Acc | Dup Rate | tok/s | Peak Mem |
|---|---|---|---|---|---|---|
| **Qwen3.5-9B PARO (INT4)** | 4 | **7.2/14** | **85.7%** | **0.9%** | 51 | 9.0 GB |
| Qwen3.5-9B OptiQ (4.5-bit) | 4 | 6.2/14 | 84.3% | 3.1% | 40 | 8.3 GB |
| Qwen3.5-4B PARO (INT4) | 7 | 5.6/14 | 77.5% | 9.7% | 67 | 4.2 GB |
| Qwen3.5-4B (8-bit) | 7 | 4.3/14 | 79.6% | 12.3% | 58 | 5.6 GB |

The 9B PARO outperforms 4.5-bit mixed-precision (OptiQ) on scenarios passed, duplicate rate, and throughput while using comparable memory.

**Key finding:** an earlier version pre-baked rotations into quantized weights for +23% throughput. This degraded structured output quality on the 9B model (5/14 passed, 11% dups). Applying rotation to activations at runtime instead — keeping weights in their original INT4 form — restored accuracy at a small throughput cost.

## Test plan

- [x] Unit tests pass
- [x] End-to-end inference with Qwen3.5-4B-PARO and Qwen3.5-9B-PARO
- [x] Non-PARO Qwen 3.5 models still load correctly (regression check)
- [x] Agent benchmark on 4 model variants (table above)

### DIFF
--- a/Libraries/MLXLLM/Models/Qwen35.swift
+++ b/Libraries/MLXLLM/Models/Qwen35.swift
@@ -174,8 +174,7 @@ final class Qwen35GatedDeltaNet: Module {
     @ModuleInfo(key: "conv1d") var conv1d: Conv1d
     @ModuleInfo(key: "in_proj_qkv") var inProjQKV: Linear
     @ModuleInfo(key: "in_proj_z") var inProjZ: Linear
-    @ModuleInfo(key: "in_proj_b") var inProjB: Linear
-    @ModuleInfo(key: "in_proj_a") var inProjA: Linear
+    @ModuleInfo(key: "in_proj_ba") var inProjBA: Linear
 
     @ParameterInfo(key: "dt_bias") var dtBias: MLXArray
     @ParameterInfo(key: "A_log") var aLog: MLXArray
@@ -212,8 +211,7 @@ final class Qwen35GatedDeltaNet: Module {
 
         _inProjQKV.wrappedValue = Linear(hiddenSize, keyDim * 2 + valueDim, bias: false)
         _inProjZ.wrappedValue = Linear(hiddenSize, valueDim, bias: false)
-        _inProjB.wrappedValue = Linear(hiddenSize, numVHeads, bias: false)
-        _inProjA.wrappedValue = Linear(hiddenSize, numVHeads, bias: false)
+        _inProjBA.wrappedValue = Linear(hiddenSize, numVHeads * 2, bias: false)
 
         _dtBias.wrappedValue = MLXArray.ones([numVHeads])
         let a = MLXRandom.uniform(low: 0, high: 16, [numVHeads])
@@ -235,8 +233,9 @@ final class Qwen35GatedDeltaNet: Module {
 
         var qkv = inProjQKV(inputs)
         let z = inProjZ(inputs).reshaped(B, S, numVHeads, headVDim)
-        let b = inProjB(inputs)
-        let a = inProjA(inputs)
+        let ba = inProjBA(inputs)
+        let b = ba[0..., 0..., ..<numVHeads]
+        let a = ba[0..., 0..., numVHeads...]
 
         let convState: MLXArray
         if let cacheState = cache?[0] {
@@ -624,6 +623,17 @@ public class Qwen35TextModel: Module, LLMModel, KVCacheDimensionProvider {
             {
                 weights[k] = v + MLXArray(1, dtype: v.dtype)
             }
+        }
+
+        // Fuse in_proj_b + in_proj_a → in_proj_ba (eliminates 1 matmul dispatch per layer)
+        for k in Array(weights.keys) where k.hasSuffix(".in_proj_b.weight") {
+            let prefix = String(k.dropLast(".in_proj_b.weight".count))
+            let aKey = "\(prefix).in_proj_a.weight"
+            guard let bWeight = weights.removeValue(forKey: k),
+                let aWeight = weights.removeValue(forKey: aKey)
+            else { continue }
+            weights["\(prefix).in_proj_ba.weight"] = concatenated(
+                [bWeight, aWeight], axis: 0)
         }
 
         return weights
--- a/Libraries/MLXLLM/Models/Qwen35.swift
+++ b/Libraries/MLXLLM/Models/Qwen35.swift
@@ -626,20 +626,6 @@ public class Qwen35TextModel: Module, LLMModel, KVCacheDimensionProvider {
             }
         }
 
-        // Split pre-fused in_proj_ba → in_proj_b + in_proj_a
-        // (PARO checkpoints store these as a single fused key)
-        for k in Array(weights.keys) where k.hasSuffix(".in_proj_ba.weight") {
-            let prefix = String(k.dropLast(".in_proj_ba.weight".count))
-            guard weights["\(prefix).in_proj_b.weight"] == nil else { continue }
-            for suffix in [".weight", ".scales", ".biases", ".bias"] {
-                let baKey = "\(prefix).in_proj_ba\(suffix)"
-                guard let baVal = weights.removeValue(forKey: baKey) else { continue }
-                let half = baVal.dim(0) / 2
-                weights["\(prefix).in_proj_b\(suffix)"] = baVal[0 ..< half]
-                weights["\(prefix).in_proj_a\(suffix)"] = baVal[half...]
-            }
-        }
-
         return weights
     }
 }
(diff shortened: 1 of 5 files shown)


Libraries/MLXLMCommon/ParoQuant/RotateQuantizedLinear.swift
```diff
@@ -0,0 +1,238 @@
...
+    """
+}
+
+// MARK: - Kernel Cache
+
+/// Cached compiled Metal kernels keyed by tile size.
+/// `nonisolated(unsafe)` — kernel creation is idempotent (same tile size always produces
+/// the same compiled kernel), so concurrent stores are benign. Same pattern as
+/// `GatedDeltaKernelManager` in the existing codebase.
+nonisolated(unsafe) private var kernelCache: [Int: MLXFast.MLXFastKernel] = [:]
```
davidkoski, line 91: This has to use locks to protect it -- callers _will_ be multithreaded.

Libraries/MLXLMCommon/ParoQuant/ParoQuantLoader.swift
```diff
@@ -0,0 +1,482 @@
...
+    static let packFactor = 32 / bits  // 8 values per uint32
+    static let mask: Int32 = (1 << bits) - 1
+    static let shifts: [Int32] = (0 ..< 8).map { Int32($0 * bits) }
+    /// Inverse of AutoAWQ reorder [0,2,4,6,1,3,5,7] → [0,4,1,5,2,6,3,7]
+    static let inverseReorder = [0, 4, 1, 5, 2, 6, 3, 7]
+
+    // Pre-computed MLXArrays (created once, reused across calls)
+    nonisolated(unsafe) static let shiftsArray = MLXArray(shifts.map { Int64($0) }).reshaped(
+        1, 1, 8)
+    nonisolated(unsafe) static let reorderIndices = MLXArray(inverseReorder.map { Int32($0) })
```
davidkoski: I think this is doable, but _only_ if the arrays are evaluated.  Two threads can't consume unevaluated arrays safely.

We would want lazy eval, so this may need to be done with some locks and computed properties.

davidkoski: Are these only used at model load time (for unpacking)?  I wonder if we even need to cache these?  Maybe just make a function to return them?

That way they are deallocated when not used and I suspect that these are pretty cheap to create and dispose of during loading.  That would be cleaner all around, unless you have measurements showing that this is needed.

Libraries/MLXLLM/Models/Qwen35.swift
```diff
@@ -626,6 +626,20 @@ public class Qwen35TextModel: Module, LLMModel, KVCacheDimensionProvider {
             }
         }
 
+        // Split pre-fused in_proj_ba → in_proj_b + in_proj_a
+        // (PARO checkpoints store these as a single fused key)
```
davidkoski: Are there other models that might use this technique?  Does this belong as library code in the ParoQuantLoader or ParoQuantizedLinear?

Libraries/MLXLMCommon/ParoQuant/RotateQuantizedLinear.swift
```diff
@@ -0,0 +1,238 @@
...
+/// step before the standard quantized matmul.
+///
+/// Rotation is applied to activations at runtime via a Metal kernel, preserving
+/// the quantization-friendly properties of the original weights.
+nonisolated open class RotateQuantizedLinear: QuantizedLinear {
+
+    // Rotation parameters — discovered by Module reflection for update(parameters:)
+    let theta: MLXArray
+    let pairs: MLXArray
+    let channel_scales: MLXArray  // swiftlint:disable:this identifier_name
```
davidkoski: This can be done using `@ParameterInfo(key: "channel_scales")` and a normal swift variable name.

Libraries/MLXLMCommon/ParoQuant/RotateQuantizedLinear.swift
```diff
@@ -0,0 +1,238 @@
...
+        let packedPairs: MLXArray
+        let scalesFlat: MLXArray
+        let dim: Int
+        let halfGroup: Int
+        let numGroups: Int
+        let krot: Int
+    }
+
+    private var cached: CachedRotation?
+    private var cachedParams: MLXArray?
```
davidkoski: Modules cannot have any eval-time generated state, see #157.

Libraries/MLXLMCommon/ParoQuant/RotateQuantizedLinear.swift
```diff
@@ -0,0 +1,238 @@
...
+    }
+
+    private func rotate(_ x: MLXArray, cache c: CachedRotation) -> MLXArray {
+        let batch = x.dim(0)
+        let tile = batch <= 1 ? 1 : 4
+
+        // Cache params array — batch rarely changes during autoregressive generation
+        if batch != cachedBatch {
+            cachedParams = MLXArray([Int32(batch), Int32(c.dim), Int32(c.krot), Int32(groupSize)])
+            cachedBatch = batch
```
davidkoski: The model will be used in a multi-threaded context, so no mutation is allowed at eval-time.

davidkoski (changes requested):

JaeminKim-amoz (commented):
**COMMENT — concurrency concern**

Solid ParoQuant feature with good benchmarks. Well-structured with clear separation of loader, linear layer, and tests. Thorough tensor shape validation.

**Requested fix:**
- [MEDIUM] `kernelCache` in `RotateQuantizedLinear.swift` uses `nonisolated(unsafe)` with a plain dictionary. Dictionary mutation is not thread-safe — concurrent writes could race. Please add a lock or use an actor-isolated cache.
- [LOW] `FileManager.enumerator` in the loader is not filtered to top directory — nested safetensors in subdirectories would be loaded.

spokvulcan:
@davidkoski  thanks for the detailed review

Verified on 4B PARO and 9B PARO inside [BenchLocal](https://github.com/stevibe/BenchLocal) no perf or memory regression.

image
image

Libraries/MLXLMCommon/ParoQuant/RotateQuantizedLinear.swift
spokvulcan: kernelCache behind NSLock, same pattern as AbstractModelRegistry

Libraries/MLXLMCommon/ParoQuant/ParoQuantLoader.swift
spokvulcan: Dropped the static AWQ arrays, rebuilt inline in unpackAndReorder load-time only

Libraries/MLXLMCommon/ParoQuant/RotateQuantizedLinear.swift
spokvulcan: done

Libraries/MLXLMCommon/ParoQuant/RotateQuantizedLinear.swift
spokvulcan: Fixed no eval-time state, derived arrays are private _-prefixed fields, populated once by a new prepareDerivedRotationState() that the loader runs between update(parameters:) and eval(model)

Libraries/MLXLLM/Models/Qwen35.swift
spokvulcan: moved out of Qwen35.sanitize into splitFusedMambaProjections in ParoQuantLoader

davidkoski (approved):
Changes look good, thank you!

PR #167, opened 2026-03-28
Pass tools schema to ToolCallProcessor for type-aware parsing

### LINKED ISSUE #159: [BUG] Array parameters in tool calls are decoded as strings instead of arrays with Qwen 3.5
**Describe the bug**

I have defined a tool that accepts a list as a parameter. When Qwen 3.5 calls it, the model correctly produces a JSON string representing the array of values, but it can't be converted to the specified native Swift list type. 

Error log:

```
typeMismatch(Swift.Array<Any>, Swift.DecodingError.Context(codingPath: [CodingKeys(stringValue: "param", intValue: nil)], debugDescription: "Expected to decode Array<Any> but found a string instead.", underlyingError: nil))

```

While investigating the issue, I noticed that  `getParameterType`  fails to parse the stringified array because the  `tools`parameter passed to it is unexpectedly  `nil`. This happens despite the tool list being declared and passed correctly during setup.

**To Reproduce**

1.  Define a tool requiring a typed list parameter (see  `Tool Definition`  below).
    
2.  Set up the  `UserInput`  and trigger the generation using the Qwen 3.5 model.
    
3.  The model returns the array as a stringified JSON.
    
4.  Observe the  `typeMismatch`  error because the framework fails to cast/decode the string back into an  `Array<Any>`.
    

**Code Snippets**

_Tool Definition:_

```swift
tool1 = Tool<Tool1Input, Tool1Output>(
    name: "tool1",
    description: "example of a requiring list parameter tool",
    parameters: [
        .required(
            "param",
            type: .array(elementType: .string),
            description: "List of file params"
        ),
    ]
) { input in
    print("invoked")
}

```

_Tool Injection & Generation:_

Even though I declare the tool list properly, the  `tools`  parameter ends up being  `nil`  later in the pipeline:

```swift
let toolList = [
    tool1!.schema,
    tool2!.schema,
    tool3!.schema
]
        
let userInput = UserInput(
    chat: chat,
    tools: toolList
)
        
do {
    let lmInput = try await context.processor.prepare(input: userInput)
            
    let stream = try generate(
        input: lmInput,
        parameters: params!,
        context: context,
    )
    // ...
}

```

_Where it fails internally:_

Because  `tools`  is  `nil`, the fallback logic to check the parameter type cannot execute:

```swift
func getParameterType(
    funcName: String, paramName: String, tools: [[String: any Sendable]]?
) -> String? {
    guard let tools else { return nil } // <--- Fails here, tools is nil
    for tool in tools {
        guard let function = tool["function"] as? [String: any Sendable],
              function["name"] as? String == funcName,
              let parameters = function["parameters"] as? [String: any Sendable],
              let properties = parameters["properties"] as? [String: any Sendable],
              let param = properties[paramName] as? [String: any Sendable],
              let type = param["type"] as? String
        else { continue }
        return type
    }
    return nil
}

```

**Expected behavior**

The framework should correctly pass the  `tools`  context so that the stringified JSON array returned by Qwen 3.5 can be identified and automatically parsed/converted into the appropriate Swift  `Array`  runtime type for the tool implementation.

-   **OS Version:**  macOS 26.3.1
    
-   **Device:**  MacBook Pro 14-inch, M4 Pro (unbinned), 24 GB RAM
    
-   **Library Version:**  latest  `main`  (as of ~5 days ago)
    

**Additional context**

When using the older  **Qwen 3**  model, this problem does not arise (maybe because it uses a different parser?).

alankessler:
Fixes #159.

The tools schema was getting to the chat template fine (so the model knew what to call), but it never reached the ToolCallProcessor that parses the response. Without it, the XML parser for Qwen 3.5 couldn't tell that `["a","b"]` was an array vs a string, so non-string parameter types came back as raw strings.

Threads `tools` through `generateTask` → `TextToolTokenLoopHandler` → `ToolCallProcessor`. All default to nil so existing callers are unaffected.

### DIFF
--- a/Libraries/MLXLMCommon/ChatSession.swift
+++ b/Libraries/MLXLMCommon/ChatSession.swift
@@ -421,7 +421,8 @@ public final class ChatSession {
                             promptTokenCount: input.text.tokens.size,
                             modelConfiguration: modelConfiguration,
                             tokenizer: tokenizer,
-                            iterator: iterator
+                            iterator: iterator,
+                            tools: tools
                         )
 
                         var pendingToolCalls: [ToolCall] = []
--- a/Libraries/MLXLMCommon/Evaluate.swift
+++ b/Libraries/MLXLMCommon/Evaluate.swift
@@ -499,7 +499,7 @@ protocol TokenIteratorProtocol: Sequence, IteratorProtocol where Element == Int 
 
 /// Generator of tokens.
 ///
-/// This is typically used via a call to ``generate(input:cache:parameters:context:wiredMemoryTicket:)`` returning `AsyncStream<Generation>`.
+/// This is typically used via a call to ``generate(input:cache:parameters:context:wiredMemoryTicket:tools:)`` returning `AsyncStream<Generation>`.
 ///
 /// To use it directly:
 ///
@@ -1144,7 +1144,7 @@ private func runSynchronousGenerationLoop(
 
 /// Given prompt tokens generate text using the given model and parameters.
 ///
-/// ``generate(input:cache:parameters:context:wiredMemoryTicket:)`` returning `AsyncStream<Generation>` is the preferred call.
+/// ``generate(input:cache:parameters:context:wiredMemoryTicket:tools:)`` returning `AsyncStream<Generation>` is the preferred call.
 ///
 /// - Parameters:
 ///   - promptTokens: tokenized prompt
@@ -1183,7 +1183,7 @@ public func generate(
 
 /// Generate tokens from an ``LMInput`` and a ``ModelContext``.
 ///
-/// Prefer using ``generate(input:cache:parameters:context:wiredMemoryTicket:)`` returning `AsyncStream<Generation>` instead.
+/// Prefer using ``generate(input:cache:parameters:context:wiredMemoryTicket:tools:)`` returning `AsyncStream<Generation>` instead.
 ///
 /// - Parameters:
 ///   - input: prepared language model input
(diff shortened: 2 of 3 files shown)


davidkoski (approved):
Nice find, thank you for the fix!

Just waiting on CI

denoide1:
I have tested your branch and in some cases the typeMismatch error still applies. I will try to investigate further: do you know where it can happen that it wasn't checked already?

alankessler:
Thanks for testing this! Do you know which model and format is triggering it? I'm wondering if it's the model outputting wrong JSON types (like `"5"` instead of `5`) vs an actual parser gap. If you can share the tool schema and the raw model output I can take a look.

denoide1:
> Thanks for testing this! Do you know which model and format is triggering it? I'm wondering if it's the model outputting wrong JSON types (like `"5"` instead of `5`) vs an actual parser gap. If you can share the tool schema and the raw model output I can take a look.

Thank you for the work! Actually, further testing, I discovered there was an issue with my code where I didn't pass the tool schema, so the type parsing couldn't be done. After passing the tool list schema, everything worked as expected

so the solution is working. Sorry for the false alarm.

denoide1:
In my opinion, as a small tip, if I already pass the list in the UserInput that is passed into the context, it’s redundant to pass it again in the generate method.

I’ve attached a small code example for reference:

```swift
let userInput = UserInput(
      chat: chat,
      tools: toolList
)
        
do {
    let lmInput = try await context.processor.prepare(input: userInput)
    
    let stream = try generate(
        input: lmInput,
        parameters: params!,
        context: context,
        tools: toolList
    )
....
....
```

davidkoski:
> In my opinion, as a small tip, if I already pass the list in the UserInput that is passed into the context, it’s redundant to pass it again in the generate method.

That would be true, but I don't think there is any place the tools (schema) get passed to the generate loop.  Also, FWIW, there are several cases where "tools" are used:

- the tool schema, which is passed to the tokenizer (UserInput)
- the tool processor, which is used in the layer _above_ the generate loop (e.g. ChatSession)
- the tool parser, which is used in the generate loop to detect tool calls (ModelContext)

JaeminKim-amoz (approved):
**APPROVE**

Simple, non-breaking plumbing fix (+12/-7). All new parameters default to `nil`, preserving backward compatibility. Clean threading of tools schema through the generation pipeline.

PR #170, opened 2026-03-29
fix: flatten prompt in TokenRing.loadPrompt to handle 2D inputs

### LINKED ISSUE #168: [BUG] Potential crashing regression in #147
**Describe the bug**
Program crashes during LM inference with `Fatal error: [broadcast_shapes] Shapes (20) and (32) cannot be broadcast` at `Evaluate.swift` in `MLXCommon`

Image

The issue most likely revolves around the highlighted line, although I am not an expert in MLX. `dim(0)` will always return `1` for any prompt in this context (which is a 1 x n matrix), and cause the `buffer` to grow beyond the capacity of the ring buffer (because the padding calculation is subsequently faulty), hence the error thrown by the package.

The workaround I have right now is to change `.dim(0)` to `.dim(1)` and that _seems_ to fix the issue, although that may be a naive/wrong solution, hence I'm preferring to report the issue and not a PR.

**To Reproduce**
Using latest main branch - any LLM inference should do - I'm testing out Qwen 3.5 but the issue is not model-specific.

**Expected behavior**
Working inference.

 - OS Version: macOS 26.4
 - Device: Mac Studio M2 Ultra
 - Version: Latest main branch, commit 2a296f1 right now

spokvulcan:
## Summary

Fixes #168.

- `TokenRing.loadPrompt` used `prompt.dim(0)` to count tokens, which returns `1` for VLM models that pass `[1, n]`-shaped prompts (via `.expandedDimensions(axis: 0)`). This caused the ring buffer to be incorrectly sized, leading to a broadcast shape crash on the next `append` call during token generation.
- Flatten the prompt to 1D upfront with `.flattened()` so the token count and all downstream slicing work correctly regardless of input shape.
- Added regression tests that exercise the full prompt → process → didSample → process cycle with 2D prompts, covering the exact crash path.

## Test plan

- [x] `swift test --filter SampleTests` — all penalty processor tests pass, including the new 2D-prompt variants
- [x] Run inference with a VLM model (e.g. Qwen2VL) using repetition/presence/frequency penalty to verify the crash no longer occurs

### DIFF
--- a/Libraries/MLXLMCommon/Evaluate.swift
+++ b/Libraries/MLXLMCommon/Evaluate.swift
@@ -334,19 +334,19 @@ struct TokenRing {
 
     /// Bulk-load from a prompt. Keeps the last `capacity` tokens.
     mutating func loadPrompt(_ prompt: MLXArray) {
-        let n = prompt.dim(0)
-        let promptTokens = prompt.asType(.int32)
+        let promptTokens = prompt.asType(.int32).flattened()
+        let n = promptTokens.count
         if n <= capacity {
             if n < capacity {
                 let padding = MLXArray.zeros([capacity - n], type: Int32.self)
-                buffer = concatenated([promptTokens.reshaped(-1), padding])
+                buffer = concatenated([promptTokens, padding])
             } else {
-                buffer = promptTokens.reshaped(-1)
+                buffer = promptTokens
             }
             count = n
             writeIndex = n % capacity
         } else {
-            buffer = promptTokens[(-capacity)...].reshaped(-1)
+            buffer = promptTokens[(-capacity)...]
             count = capacity
             writeIndex = 0
         }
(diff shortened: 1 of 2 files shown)


Satchitananda:
Waiting for this one to be reviewed and merged 🙏🏻

JaeminKim-amoz (approved):
**APPROVE**

Clean, well-scoped bug fix. `.flattened()` upfront correctly handles both 1D and 2D prompt shapes. `.flattened()` on an already-1D array is a no-op, so backward compatibility is preserved. Tests are thorough — cover all three penalty processor types with 2D prompts.

ptsochantaris (approved):

john-rocky:
Note for triage: this also fixes #220 (same root cause, different reporter). I had a duplicate fix in #251 with a slightly less clean implementation (`reshaped(-1)` + `dim(0)` instead of `.flattened().count`) and narrower test coverage; closed in favor of this one.

davidkoski (approved):
Looks good, thank you!

PR #172, opened 2026-03-30
Handle stringified JSON tool call arguments

### LINKED ISSUE #169: [BUG] JSONToolCallParser fails to parse stringified arguments (used by Granite 4 and others)
**Describe the bug**
The JSONToolCallParser fails to parse tool calls from models that output the arguments field as a stringified JSON string rather than a nested JSON object. This is the default behavior for several models, including the IBM Granite 4 family.

When a model outputs:
{"name": "func", "arguments": "{\"key\": \"value\"}"}
The current JSONDecoder implementation fails to decode the arguments field because it expects an object ({) but encounters a string (").

**To Reproduce**
Run any Granite 4 model using mlx-swift-lm with tools enabled.  I used this model for testing: https://huggingface.co/mlx-community/granite-4.0-h-micro-4bit

**Expected behavior**
The parser should be resilient enough to handle both standard nested JSON objects and stringified JSON objects in the arguments field, matching the flexible behavior of the mlx-lm Python implementation.

I believe the fix is fairly simple.  Update this old code for JSONToolCallParser.swift [line 30](https://github.com/ml-explore/mlx-swift-lm/blob/2a296f145c3129fea4290bb6e4a0a5fb458efa06/Libraries/MLXLMCommon/Tool/Parsers/JSONToolCallParser.swift#L30):
```swift
        let jsonStr = text.trimmingCharacters(in: .whitespacesAndNewlines)

        guard let data = jsonStr.data(using: .utf8),
            let function = try? JSONDecoder().decode(ToolCall.Function.self, from: data)
        else { return nil }

        return ToolCall(function: function)
```

To this new version which allows both nested objects or stringified arguments.
```swift
        // parse the object and make sure the name field is present
        guard let jsonObj = try? JSONSerialization.jsonObject(with: data) as? [String: Any],
              let name = jsonObj["name"] as? String else {
            return nil
        }
        
        // parse the arguments which may be a nested object or stringified
        var arguments: [String: Any] = [:]
        if let argsDict = jsonObj["arguments"] as? [String: Any] { // nested object
            arguments = argsDict
        } else if let argsString = jsonObj["arguments"] as? String { // stringified (used by Granite 4 models)
            if let argsData = argsString.data(using: .utf8),
               let nestedDict = try? JSONSerialization.jsonObject(with: argsData) as? [String: Any] {
                arguments = nestedDict
            } else { // the string wasn't valid JSON
                return nil
            }
        } else { // arguments property is missing
            return nil
        }
        return ToolCall(function: ToolCall.Function(name: name, arguments: arguments))
```

I tested this fix locally and it allowed the Granite 4 tool calling to work successfully, including calls that required no arguments (Granite 4 includes "arguments": "{}" in that case which is still handled correctly above).

 - OS Version: MacOS 15
 - Device: M4 MacBook Pro
 - mlx-swift-lm Version: latest main from 1 week ago (https://github.com/ml-explore/mlx-swift-lm/tree/edd42fcd947eea0b19665248acf2975a28ddf58b)

kuosuko:
Fixes #169.

## Problem

`JSONToolCallParser` assumes `arguments` is always a JSON object, but some models emit it as a JSON-encoded string instead.

That means payloads like this fail to parse today:

```json
{"name":"get_weather","arguments":"{\"location\":\"Paris\"}"}
```

As a result, Swift drops the tool call entirely.

## Fix

Normalize the outer JSON payload before decoding `ToolCall.Function`:

- decode the top-level object with `JSONSerialization`
- if `arguments` is a string, parse that string as JSON
- replace `arguments` with the parsed value
- decode the normalized payload using the existing `JSONDecoder`

This preserves the current behavior for object-shaped arguments while adding compatibility for stringified arguments.

## Testing

- `swift test --filter ToolTests`
- added regression tests for stringified object arguments
- added regression tests for stringified empty arguments
- added regression tests for stringified array arguments

I also verified locally with Granite tool calling that the original parser emitted no parsed tool call while the patched parser successfully emitted `get_weather(location: "Tokyo", unit: "celsius")`.

### DIFF
--- a/Libraries/MLXLMCommon/Tool/Parsers/JSONToolCallParser.swift
+++ b/Libraries/MLXLMCommon/Tool/Parsers/JSONToolCallParser.swift
@@ -29,10 +29,29 @@ public struct JSONToolCallParser: ToolCallParser, Sendable {
 
         let jsonStr = text.trimmingCharacters(in: .whitespacesAndNewlines)
 
-        guard let data = jsonStr.data(using: .utf8),
-            let function = try? JSONDecoder().decode(ToolCall.Function.self, from: data)
+        guard
+            let data = jsonStr.data(using: .utf8),
+            let normalizedData = normalizedToolCallData(from: data),
+            let function = try? JSONDecoder().decode(ToolCall.Function.self, from: normalizedData)
         else { return nil }
 
         return ToolCall(function: function)
+    }
+
+    private func normalizedToolCallData(from data: Data) -> Data? {
+        guard var jsonObject = try? JSONSerialization.jsonObject(with: data) as? [String: Any] else {
+            return nil
+        }
+
+        if let stringifiedArguments = jsonObject["arguments"] as? String {
+            guard
+                let argumentsData = stringifiedArguments.data(using: .utf8),
+                let argumentsObject = try? JSONSerialization.jsonObject(with: argumentsData)
+                    as? [String: Any]
+            else { return nil }
+            jsonObject["arguments"] = argumentsObject
+        }
+
+        return try? JSONSerialization.data(withJSONObject: jsonObject)
     }
 }
--- a/Libraries/MLXLMCommon/Tool/Parsers/JSONToolCallParser.swift
+++ b/Libraries/MLXLMCommon/Tool/Parsers/JSONToolCallParser.swift
@@ -39,7 +39,8 @@ public struct JSONToolCallParser: ToolCallParser, Sendable {
     }
 
     private func normalizedToolCallData(from data: Data) -> Data? {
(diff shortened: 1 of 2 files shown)


JaeminKim-amoz (approved):
**APPROVE**

Good defensive coding for a real compatibility issue. The normalization approach preserves backward compatibility (object arguments pass through unchanged). 3 regression tests cover stringified objects, empty objects, and arrays.

**Minor notes:**
- [MEDIUM] Bare top-level array arguments would fail silently (returns nil), but this matches `ToolCall.Function` expectations for named arguments
- [LOW] CI swift-format label suggests formatting issues

DePasqualeOrg:

davidkoski (approved):
Looks good, thank you!

PR #173, opened 2026-03-31
Add speculative decoding

petrukha-ivan:
## Proposed changes

Add speculative decoding support, based on the approach in [mlx-lm](https://github.com/ml-explore/mlx-lm/blob/6ddfdda1acd01077bd463d785d88f0ad2a6236d3/mlx_lm/generate.py#L469).

A new `SpeculativeTokenIterator` uses a smaller draft model to propose tokens that are verified in batch by the main model, yielding accepted tokens one at a time. Both `generate()` and `generateTokens()` get overloads accepting a draft model. This also extracts a `TokenIteratorProtocol` so the shared generation loop works with both iterator types.

**Bug fix:** `trimPromptCache` previously only trimmed the first cache layer. This PR fixes it to trim all layers, which is required for speculative decoding to correctly rewind rejected tokens.

## Limitations

Speculative decoding requires trimmable KV caches to rewind rejected tokens. Models that use `MambaCache` are not supported and will throw an error. A possible solution would be to snapshot the cache state before drafting and restore it on rejection, but this adds memory overhead and complexity that felt out of scope for this PR.

## Benchmarks

Benchmarked on M3 Max using a short text translation prompt (~150 tokens) generating ~130 tokens, with 2 draft tokens for speculative generation. Using more than 2 draft tokens makes results worse in most cases. The draft model generates further ahead and divergence from the main model lowers the acceptance rate, so the drafting overhead ends up making generation slower.

| Main model | Draft model | Prompt tokens/s | Generation tokens/s | Memory peak |
| - | - | - | - | - |
| Qwen3-4B-4bit | - | 788 | 99 | 2564M |
| Qwen3-4B-4bit | Qwen3-0.6B-4bit | 708 | 128 (+29%) | 2915M |
| Qwen3-14B-4bit | - | 258 | 32 | 8237M |
| Qwen3-14B-4bit | Qwen3-0.6B-4bit | 250 | 53 (+66%) | 8351M |
| Qwen3-32B-4bit | - | 71 | 14 | 17916M |
| Qwen3-32B-4bit | Qwen3-0.6B-4bit | 73 | 25 (+79%) | 18267M |

The larger the gap between the main and draft model, the greater the benefit. The 32B model sees the biggest speedup (+79%) since its baseline generation is slow and there is more room for the draft model to help. Prompt processing speed stays largely unaffected, and the memory overhead from the 0.6B draft model is minimal in all cases (under 400MB).

> **Note:** The benchmarks above use a translation task, which is relatively deterministic and tends to produce high acceptance rates. Results will vary depending on the draft model selection and the prompt. More creative tasks may see lower acceptance rates, and in some cases speculative decoding can even slow down generation.

## Checklist


Libraries/MLXLMCommon/Evaluate.swift
```diff
@@ -699,6 +706,261 @@ public struct TokenIterator: Sequence, IteratorProtocol {
...
+        case .tokens(let tokens):
+            y = tokens
+        case .logits(let result):
+            var logits = result.logits[0..., -1, 0...]
+            logits = processor?.process(logits: logits) ?? logits
+            let token = sampler.sample(logits: logits)
+            processor?.didSample(token: token)
+            y = .init(tokens: token)
+            mainState = result.state
+            asyncEval(y.tokens)
```
davidkoski: I think the python code doesn't do this.  It isn't incorrect, but it might collide with the prefill of the draft (below).

Libraries/MLXLMCommon/Evaluate.swift
```diff
@@ -1094,6 +1356,81 @@ public func generate(
...
+///   - numDraftTokens: Number of tokens the draft model proposes per round (default: 2).
+///   - wiredMemoryTicket: Optional wired memory ticket for policy-based coordination.
+/// - Returns: An `AsyncStream` that emits `Generation` values.
+/// - Throws: An error if the iterator initialization fails.
+public func generate(
+    input: LMInput,
+    parameters: GenerateParameters,
+    context: ModelContext,
+    draftModel: any LanguageModel,
+    draftCache: [KVCache]? = nil,
```
davidkoski, line 1424: Does this need the main KVCache?  or is there no need to keep that around?

Libraries/MLXLMCommon/KVCache.swift
```diff
@@ -1519,6 +1519,7 @@ public func canTrimPromptCache(_ cache: [KVCache]) -> Bool {
 @discardableResult
 public func trimPromptCache(_ cache: [KVCache], numTokens: Int) -> Int {
     guard canTrimPromptCache(cache), !cache.isEmpty else { return 0 }
+    cache.dropFirst().forEach { $0.trim(numTokens) }
```
davidkoski, line 1522: This is curious --  perhaps we were not using cache trimming in the past but this looks important.

Libraries/MLXLMCommon/Evaluate.swift
petrukha-ivan: Oh, it's actually something I missed. Added parameters for the main cache 👍

Libraries/MLXLMCommon/KVCache.swift
petrukha-ivan: Yeah, I think it went unnoticed because this function was never used internally, but since it's public it's definitely important.

Libraries/MLXLMCommon/Evaluate.swift
petrukha-ivan: Good catch! We do batch processing for this token along with the drafted tokens in the verify step, so I think there's no reason to force evaluation here. Removed.

davidkoski (approved):
This looks good.  I filed #181 to track adding support in ChatSession.

PR #174, opened 2026-03-31
Forward tools and additionalContext in GlmOcr and SmolVLM2 processors

alankessler:
Follow-up to #140. GlmOcr was added after that fix and missed the pattern entirely. SmolVLM2's video path was also missed while its other two branches were fixed.

Separately, `ModelContainer.generate()` also doesn't forward `tools` to the underlying `generate()` call — anyone using that convenience method instead of `ChatSession` won't get type-aware tool call parsing. Not sure if that's worth adding a parameter for or if the expectation is that tool-aware callers use `ChatSession` or the lower-level `perform { }` API. Happy to add it if you want.

### DIFF
--- a/Libraries/MLXVLM/Models/GlmOcr.swift
+++ b/Libraries/MLXVLM/Models/GlmOcr.swift
@@ -831,7 +831,8 @@ public struct GlmOcrProcessor: UserInputProcessor {
     public func prepare(input: UserInput) async throws -> LMInput {
         let messages = GlmOcrMessageGenerator().generate(from: input)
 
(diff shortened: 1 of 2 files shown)


JaeminKim-amoz (approved):
**APPROVE**

Minimal, surgical fix (+4/-2 lines). Correctly forwards missing `tools:` and `additionalContext:` parameters in `GlmOcr.swift` and `SmolVLM2.swift`. Matches the pattern established in PR #140.

davidkoski (approved):
Thank you!

PR #175, opened 2026-04-01
Fix inaccuracies in (and possibly remove) "skills"

DePasqualeOrg:
While I was working on #118, I noticed many inaccuracies in the "skills" documents that were added by @ronaldmannak in #92. I corrected some of these in #118 and have corrected more here. However, I think it would be best to remove most of these documents for the following reasons:

- They appear to be mostly AI-generated summaries of source files. It's better for agents to directly read the source files, which should be self-documenting.
- They are liable to get out of sync with the codebase, as happened with the latest refactoring, and are thus detrimental to agents that use them.

A better use of skills is to record knowledge that isn't directly apparent from the codebase. Only the porting guide fits this criterion. I suggest that the other documents be removed.

### DIFF
--- a/skills/mlx-swift-lm/references/concurrency.md
+++ b/skills/mlx-swift-lm/references/concurrency.md
@@ -11,8 +11,6 @@ mlx-swift-lm uses Swift concurrency with specialized utilities to handle the uni
 | Type | Purpose |
 |------|---------|
 | `SerialAccessContainer<T>` | Exclusive async access to wrapped state |
-| `AsyncMutex` | Lock that works with async blocks |
-| `SendableBox<T>` | Transfer non-Sendable values across isolation |
 | `ModelContainer` | Thread-safe model wrapper (uses SerialAccessContainer) |
 | `ChatSession` | NOT thread-safe (single task only) |
 
@@ -23,8 +21,8 @@ Provides exclusive access to state across `async` calls:
 ```swift
 // Unlike actors, this guarantees exclusive access for entire async operation
 final class SerialAccessContainer<T>: @unchecked Sendable {
-    func read<R>(_ body: (T) async throws -> R) async rethrows -> R
-    func update<R>(_ body: (inout T) async throws -> R) async rethrows -> R
+    func read<R>(_ body: @Sendable (T) async throws -> sending R) async rethrows -> sending R
+    func update<R>(_ body: @Sendable (inout T) async throws -> sending R) async rethrows -> sending R
 }
 ```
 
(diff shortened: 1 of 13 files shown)


skills/mlx-swift-lm/references/concurrency.md
```diff
@@ -11,8 +11,6 @@ mlx-swift-lm uses Swift concurrency with specialized utilities to handle the uni
 | Type | Purpose |
 |------|---------|
 | `SerialAccessContainer<T>` | Exclusive async access to wrapped state |
-| `AsyncMutex` | Lock that works with async blocks |
```
davidkoski, line 14: I think removing this makes sense -- it is private to the file and part of the implementation of other pieces.

skills/mlx-swift-lm/references/concurrency.md
```diff
@@ -11,8 +11,6 @@ mlx-swift-lm uses Swift concurrency with specialized utilities to handle the uni
...
-| `SendableBox<T>` | Transfer non-Sendable values across isolation |
```
davidkoski, line 15: but I think this one is useful for documenting how some of the implementation works.

davidkoski: ah, but you added in its own section below.  that seems fine

skills/mlx-swift-lm/references/concurrency.md
```diff
@@ -80,35 +80,14 @@ Task {
...
+// Inside the library, SendableBox wraps the value:
 let box = SendableBox(input)
 Task {
-    let input = box.consume()  // Transfer ownership
+    let input = box.consume()  // Transfer ownership (single use)
     use(input)
 }
 ```
 
-### Pattern: Consuming Parameters
```
davidkoski, line 91: This might be useful, but let me write why:

Functions that take parameters that contain `MLXArray`, such as `LMInput` or `Chat.Message` should have their parameters marked as `consuming`.  This prevents callers from using these values after passing them to the function -- effectively it transfers ownership to the function.

For example `KVCache` contains `MLXArray` and is updated with the LLM context as it runs.  Consider a ChatSession that is initialized with KVCache values to restore a session:

```
struct ChatSession {
    public init(
        cache: consuming [KVCache]
    )
}
```

The `KVCache` is transferred to the ChatSession -- this prevents a user from making multiple ChatSession using the same underlying KVCache.  It wouldn't be thread-safe and further wouldn't be semantically correct as the ChatSessions would leak to each other.  The correct way is to make an independent copy of the KVCache and pass it to each ChatSession.


Something like that.

davidkoski:
After reading through this more carefully, a lot of it isn't even refactoring, it is just forward progress.  I wonder if the skill could contain a pointer to the appropriate area of code -- it would be more like an index.  I think the explanatory text (that doesn't live in the documentation itself) would be very good in the skills.

Potentially we could make tests that 1) compile and 2) demonstrate some of the features and reference that.  It would make sure it still works and still get the benefit of the skills.  I am not 100% sure that works though!

JaeminKim-amoz (approved):
**APPROVE**

Valuable maintenance work. Thorough corrections of outdated API references across 13 skill docs — updates to new macro syntax, fixes method signatures, removes deleted type references. Prevents agents from using stale APIs.

PR #176, opened 2026-04-02
Fix doc comments and verify in CI

DePasqualeOrg:
I fixed many inaccurate doc comments and added a CI step that fails when inaccuracies in the doc comments are found. The CI step runs `swift package generate-documentation --target "$TARGET" --warnings-as-errors` on each target.


DePasqualeOrg:
I'm investigating why the docs check is failing in CI, since the script passes the test on my machine.

PR #178, opened 2026-04-03
Batched LLM inference part 1 - consolidating RoPE calls

ronaldmannak:
## Proposed changes

As discussed in #150, this PR is the first PR continuous batching feature.

- Introduce `applyRotaryPosition(_:to:cache:)` helper in `MLXLMCommon` that replaces the repeated conditional offset pattern (`if let cache { rope(x, offset: cache.offset) } else { rope(x, offset: 0) }`) with a single call site
- Migrate all 45 model attention layers to use the unified helper, removing ~160 lines of boilerplate
- Standardize rope property types to `RoPELayer (OffsetLayer & ArrayOffsetLayer)`, removing ad-hoc wrapper enums and protocols

## Checklist

### DIFF
--- a/Libraries/MLXLLM/Models/Phi3.swift
+++ b/Libraries/MLXLLM/Models/Phi3.swift
@@ -20,21 +20,7 @@ class Phi3Attention: Module {
     @ModuleInfo(key: "qkv_proj") var wqkv: Linear
     @ModuleInfo(key: "o_proj") var wo: Linear
 
-    enum PositionalEncoding {
-        case rope(RoPE)
-        case suScaledRoPE(SuScaledRoPE)
-
-        func applyEncoding(_ x: MLXArray, offset: Int = 0) -> MLXArray {
-            switch self {
-            case .rope(let rope):
-                return rope.callAsFunction(x, offset: offset)
-            case .suScaledRoPE(let suScaledRoPE):
-                return suScaledRoPE(x, offset: offset)
-            }
-        }
-    }
-
-    let rope: PositionalEncoding
+    let rope: RoPELayer
 
     public init(_ args: Phi3Configuration) {
         self.args = args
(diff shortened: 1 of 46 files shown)


Libraries/MLXLLM/Models/Phi3.swift
```diff
@@ -64,19 +50,19 @@ class Phi3Attention: Module {
             ropeScaling.type == "su" || ropeScaling.type == "longrope",
             let shortFactor = ropeScaling.shortFactor, let longFactor = ropeScaling.longFactor
         {
-            self.rope = .suScaledRoPE(
+            self.rope =
```
davidkoski, line 53: This probably predates the new rope factory on the mlx-lm side and thus here too.  I wonder if everything could just use the new rope factory?  Probably best to keep it consistent with mlx-lm, but I wonder if they should?

davidkoski (approved):
Well, the easy part looks good :-)

davidkoski:
@DePasqualeOrg and @ronaldmannak I appreciate both of your contributions.  We are all (except me I guess) volunteers here -- people contributing their time and efforts to help each other.

Yes, the title on this was not ideal.  I didn't even notice that as I was expecting this PR and knew what it was about -- my bad on not catching and correcting that and perhaps on Ronald for submitting it.  I think it is best to assume mistakes rather than intent.

Please use "PR phrasing":  review the code, not the person.  Use softening phrases like "I think" or "maybe we should" for matters of opinion.  If something is _wrong_ then it is fine to use factual and non-emotional sentences:  "this will leak memory if called concurrently".  I try to stick to this but it is a manual effort and I probably don't always succeed.

Anyway, let's keep it professional and constructive.

- https://github.com/ml-explore/mlx-swift-lm/blob/main/CODE_OF_CONDUCT.md

I appreciate both of you and I want everybody to feel safe to contribute.  If there is a problem in the quality of the contributions, that is on me -- give me feedback.  My personal style is more "move it forward".  Everything should build and run:  no regressions and if we find some they should be fixed quickly.

Thank you!

PR #180, opened 2026-04-03
Add gemma 4 model (text, vision, MoE)

adrgrondin:
## Proposed changes

AI assisted port of Gemma 4 (text, vision, MoE) based on @Blaizzy implementation (https://github.com/Blaizzy/mlx-vlm/tree/main/mlx_vlm/models/gemma4)

Model tested text + vision:
- `mlx-community/gemma-4-e2b-it-4bit`
- `mlx-community/gemma-4-e4b-it-4bit`
- `mlx-community/gemma-4-26b-a4b-it-4bit`

- `mlx-community/gemma-4-31b-it-4bit` is too slow on my machine. Would appreciate if someone can try (it loads but can't really test the 4-bit correctly)

https://github.com/ml-explore/mlx-swift-lm/issues/177

## Checklist

- [x] I have read the [CONTRIBUTING](https://github.com/ml-explore/mlx/blob/main/CONTRIBUTING.md) document
- [x] I have run `pre-commit run --all-files` to format my code / installed pre-commit prior to committing changes
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] I have updated the necessary documentation (if needed)

PR #185, opened 2026-04-04
Add Gemma 4 text model support (E2B and E4B)

stefan-geens:
## Proposed changes

Port of `gemma4.py` and `gemma4_text.py` from [mlx-lm](https://github.com/ml-explore/mlx-lm/tree/main/mlx_lm/models). Adds support for Google's Gemma 4 text-only architecture:

- **Gemma4Text.swift** (~750 lines) — Full model implementation including:
  - Per-Layer Embeddings (PLE) with gated residual signals
  - Shared KV cache across later layers (saves memory)
  - Dual RoPE configurations (proportional for full-attention, default for sliding-window)
  - `ProportionalRoPE` — new RoPE variant supporting `partial_rotary_factor` (only 25% of head dims get positional encoding on full-attention layers)
  - Global head dimensions (512-dim heads for full-attention vs 256-dim for sliding)
  - Double-wide MLP for KV-shared layers
  - Logit softcapping (`tanh(x/cap) * cap`)
  - K-eq-V attention support (for future 26B/31B models)
  - Quantized KV cache support via `KVCacheDimensionProvider`

- **Gemma4.swift** (~100 lines) — Thin wrapper handling the `"gemma4"` model_type with nested `text_config` from HuggingFace configs. Includes VLM weight prefix stripping in `sanitize()`.

- **LLMModelFactory.swift** — Registers `gemma4` and `gemma4_text` model types plus `mlx-community/gemma-4-e4b-it-4bit` and `mlx-community/gemma-4-e2b-it-4bit` configurations

Both E2B and E4B use the same code path — they differ only in config values (hidden size, layer count, KV heads).

MoE support (needed only for the 26B model) is not included in this PR.

## Checklist

## Test plan

- [x] `swift build --target MLXLLM` compiles successfully
- [x] `swift-format lint` passes clean (zero warnings)
- [x] Python `mlx-lm` verification: both E2B and E4B produce coherent English and Chinese output (tested on M4 Pro, 24 GB — E2B: 127 tok/s, 2.66 GB peak; E4B: 71 tok/s, 4.31 GB peak)
- [ ] Run `llm-tool --model mlx-community/gemma-4-e2b-it-4bit --prompt "Hello"` to verify Swift text generation
- [ ] Run `llm-tool --model mlx-community/gemma-4-e4b-it-4bit --prompt "Hello"` to verify E4B variant

### DIFF
--- a/Libraries/MLXLLM/Models/Gemma4Text.swift
+++ b/Libraries/MLXLLM/Models/Gemma4Text.swift
@@ -33,7 +33,7 @@ public struct Gemma4TextConfiguration: Codable, Sendable {
     var attentionKeqV: Bool = false
     var finalLogitSoftcapping: Float = 30.0
     var useDoubleWideMlp: Bool = true
-    var layerTypes: [String]?
+    var layerTypes: [String] = []
     var tieWordEmbeddings: Bool = true
 
     // RoPE parameters (nested dict with full_attention/sliding_attention sub-configs)
@@ -110,15 +110,10 @@ public struct Gemma4TextConfiguration: Codable, Sendable {
             try container.decodeIfPresent(Float.self, forKey: .finalLogitSoftcapping) ?? 30.0
         self.useDoubleWideMlp =
             try container.decodeIfPresent(Bool.self, forKey: .useDoubleWideMlp) ?? true
-        self.layerTypes = try container.decodeIfPresent([String].self, forKey: .layerTypes)
-        self.tieWordEmbeddings =
-            try container.decodeIfPresent(Bool.self, forKey: .tieWordEmbeddings) ?? true
-        self.ropeParameters =
-            try container.decodeIfPresent(
-                [String: [String: StringOrNumber]].self, forKey: .ropeParameters)
-
-        // Derive layer types from sliding window pattern if not provided
-        if self.layerTypes == nil {
+        if let decoded = try container.decodeIfPresent([String].self, forKey: .layerTypes) {
+            self.layerTypes = decoded
+        } else {
+            // Derive layer types from sliding window pattern
             var pattern = [String]()
             for i in 0 ..< slidingWindowPattern {
                 pattern.append(
@@ -130,6 +125,11 @@ public struct Gemma4TextConfiguration: Codable, Sendable {
             }
             self.layerTypes = Array(types.prefix(numHiddenLayers))
         }
+        self.tieWordEmbeddings =
+            try container.decodeIfPresent(Bool.self, forKey: .tieWordEmbeddings) ?? true
+        self.ropeParameters =
+            try container.decodeIfPresent(
+                [String: [String: StringOrNumber]].self, forKey: .ropeParameters)
 
         // Extract RoPE parameters from nested config
         if let ropeParams = ropeParameters {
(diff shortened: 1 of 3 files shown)


john-rocky:
Tested PR #185 on iPhone — Gemma 4 E2B 4-bit loads and generates correctly after one fix:

`Gemma4Model.languageModel` needs `@ModuleInfo(key: "language_model")`. Without it, the module key defaults to camelCase `languageModel`, which doesn't match the snake_case `language_model` in the weight files.

```diff
- fileprivate let languageModel: Gemma4TextModel
+ @ModuleInfo(key: "language_model") fileprivate var languageModel: Gemma4TextModel
```

and in init:
```diff
- self.languageModel = Gemma4TextModel(config.textConfig)
+ self._languageModel.wrappedValue = Gemma4TextModel(config.textConfig)
```

Without this, loading fails with:
```
keyNotFound(path: ["languageModel", "model", "per_layer_model_projection", "weight"], ...)
```

Everything else works great. Tested with `mlx-community/gemma-4-e2b-it-4bit` on iPhone 16 Pro, conversation quality is solid.

stefan-geens:
Thanks for testing and the fix @john-rocky Fixed in `68b82ec`.

Libraries/MLXLLM/Models/Gemma4Text.swift
```diff
@@ -0,0 +1,743 @@
...
+
+            currentOffset = cache?.offset ?? 0
+            k = rope(k, offset: currentOffset)
+
+            var v: MLXArray
+            if useKeqV {
+                v = k
+            } else {
+                v = vProj!(x).reshaped(B, L, nKvHeads, effectiveHeadDim)
+            }
```
davidkoski: this would be better as `if let vProj` -- then it doesn't need the force unwrap.  What you have is safe because of the condition where vProj is set, but not an ideal pattern.

Libraries/MLXLLM/Models/Gemma4Text.swift
```diff
@@ -0,0 +1,743 @@
...
+    @ModuleInfo(key: "q_norm") var qNorm: RMSNorm
+    @ModuleInfo(key: "k_norm") var kNorm: RMSNorm
+    @ModuleInfo(key: "v_norm") var vNorm: RMSNormNoScale
+
+    @ModuleInfo var rope: ProportionalRoPE
+
+    init(_ config: Gemma4TextConfiguration, layerIdx: Int) {
+        self.config = config
+        self.layerIdx = layerIdx
+        self.layerType = config.layerTypes![layerIdx]
```
davidkoski: Everywhere that has a force unwrap like this is essentially an "assert layerTypes is not nil".

I wonder if it would be better to make it non-optional in the config?

```
        self.layerTypes = try container.decodeIfPresent([String].self, forKey: .layerTypes) ?? []
```

davidkoski: Or based on this code:

```
        let layerTypes =
            config.layerTypes ?? Array(repeating: "full_attention", count: config.numHiddenLayers)
```

davidkoski (changes requested):
Overall this looks good, just a few comments around force unwraps -- if you are able to fix those quickly, let's do it, otherwise I think we can take this and do a followup PR.

Libraries/MLXLLM/Models/Gemma4Text.swift
```diff
@@ -0,0 +1,743 @@
...
+        model.layers.map { $0.selfAttn }
+    }
+}
+
+// MARK: - ProportionalRoPE
+
+/// RoPE variant where only a fraction of dimensions are rotated.
+/// Unrotated dimensions get infinite-frequency (effectively zero embedding).
+/// Used by Gemma 4 full-attention layers (partial_rotary_factor = 0.25).
+public class ProportionalRoPE: Module, OffsetLayer, ArrayOffsetLayer {
```
davidkoski: This has been added here:

https://github.com/ml-explore/mlx-swift-lm/blob/main/Libraries/MLXLMCommon/RoPEUtils.swift#L90

from #180 .  Let's use that one.

davidkoski (approved):
Looks good, thank you!

PR #165, opened 2026-03-27
switch to swift 6 -- prevent concurrency issues, fix concurrency issues

davidkoski:
## Proposed changes

- some concurrency issues snuck in for test support
- fix and prevent more from coming in by switching Package to swift 6
- depends on https://github.com/ml-explore/mlx-swift/pull/379
    - will update package.swift to pick that up when ready

## Checklist

Put an `x` in the boxes that apply.

- [x] I have read the [CONTRIBUTING](https://github.com/ml-explore/mlx/blob/main/CONTRIBUTING.md) document
- [x] I have run `pre-commit run --all-files` to format my code / installed pre-commit prior to committing changes
- [x] I have added tests that prove my fix is effective or that my feature works
- [ ] I have updated the necessary documentation (if needed)

PR #192, opened 2026-04-07
Add Gemma 4 audio tower support (ASR via Conformer encoder)

antmanler:
## Summary

Add audio support to Gemma 4 models (E2B and E4B), enabling native speech-to-text on Apple Silicon via MLX.

Built on top of PR #180 (Gemma 4 text + vision by @adrgrondin).

### Audio Encoder (`Gemma4.swift`)
- `Gemma4AudioConfiguration` — all encoder parameters from model config
- `Gemma4SubSampleConvProjection` — Conv2d blocks for mel frame downsampling
- `Gemma4ConformerBlock` — FFW → Attention → LightConv1d → FFW → RMSNorm
- `Gemma4AudioAttention` — chunked local attention with sinusoidal relative position embeddings and logit softcapping
- `Gemma4ConformerLightConv1d` — GLU + depthwise causal conv1d
- `Gemma4AudioEncoder` — SSCP + 12 Conformer layers + output projection

### Audio Feature Extractor (`Gemma4AudioFeatureExtractor.swift`)
- Log-mel spectrogram extraction using Accelerate vDSP FFT
- HTK-scale mel filter bank (128 bins, 16kHz)
- Matching Python `Gemma4AudioFeatureExtractor` parameters

### Integration
- `LMInput.ProcessedAudio` — new struct for audio features + mask
- `UserInput.audios` — audio input field (raw PCM float arrays)
- `Gemma4Processor.prepare()` — audio token injection + mel extraction
- `Gemma4.prepare(LMInput)` — routes audio features through encoder
- `getInputEmbeddings()` — scatters audio embeddings at `<|audio|>` positions

### Key Implementation Details
- vDSP FFT output scaled by 0.5 to match numpy rfft normalization
- Audio mask polarity: mel extractor outputs 1=valid, encoder expects True=padding
- `relPosInvTimescales` stored as `[Float]` (not MLXArray) to avoid Module parameter lookup
- Conv2d/Conv1d weights: mlx-community models store in MLX layout, no transpose needed
- Depthwise conv1d uses `groups=hiddenSize` for correct weight shape `[1024, 5, 1]`

## Testing

- Mel spectrogram alignment: shape and values match Python within 0.003 tolerance
- End-to-end transcription: produces correct output matching Python mlx-vlm
- Python reference data in `Tests/MLXVLMTests/Fixtures/`

## Models Tested
- `mlx-community/gemma-4-e2b-it-4bit` (2B, ~3.4GB)
- `mlx-community/gemma-4-e4b-it-4bit` (4B, ~5.2GB)

## Dependencies
- Requires PR #180 for base Gemma 4 text + vision support
- Uses Accelerate framework for FFT (no additional dependencies)


davidkoski:
#180 and #185 are now merged -- can you rebase this on main and integrate with the code from #180?  Thank you!

vahsaechao:
Hi @antmanler, @davidkoski suggested we collaborate on combining our audio encoder work on top of #180. My PR #194 adds Gemma 3n audio support using the same conformer architecture. Happy to help integrate.

antmanler:
Rebased onto current `main` (after #180, #185, #118 tokenizer decoupling, #165 swift 6 migration). The diff is now 5 logical commits on top of `c1ff9f8`:

1. `feat(mlxlmcommon): add LMInput.ProcessedAudio + UserInput.audios`
2. `feat(gemma4): add mel spectrogram feature extractor`
3. `feat(gemma4): add audio tower — Conformer encoder + model wiring`
4. `test(gemma4-audio): unit + alignment tests with Python reference fixtures`
5. `fix(gemma4): use audio encoder output dim for embed_audio projection`

The previous 21-commit branch has been preserved at tag `backup/feat-gemma4-audio-tower-pre-rebase` (commit `668347f`) on this fork in case anyone needs to reference the original development history.

### Integration notes for #180

* Preserved upstream's `Gemma4SharedKVState`, `gemma4AdjustAttentionMask`, MoE `sanitize`, and the existing audio-aware `perLayerInputs` wiring at `getInputEmbeddings:1683-1694`.
* `Gemma4Configuration` extended with optional `audioConfiguration: Gemma4AudioConfiguration?` (decodes `audio_config` block, missing → nil so text/vision-only models still load).
* `Gemma4` model class gains `@ModuleInfo` audio_tower / embed_audio fields, conditionally initialized.
* `Gemma4.sanitize` now only strips `audio_tower` / `embed_audio` weights when `audioConfiguration == nil`.
* `Gemma4Processor.prepare` extracts mel features, computes audio token count from actual subsampling math, injects `<|audio|>` placeholders, and inverts the mel mask to True=padding.
* `Gemma4ProcessorConfiguration.audioTokenId` defaults to `258_881` for models whose `processor_config.json` omits it.
* No changes to `VLMModelFactory._load`; it already exposes the post-#118 signature.

### Correctness fixes preserved

The 8 numerical/correctness fixes from the original branch are all intact in the rebased commits:

1. vDSP FFT ×0.5 scaling (`Gemma4AudioFeatureExtractor.swift:238`)
2. Audio mask polarity inversion in `Gemma4Processor.prepare` (mel extractor outputs 1=valid, encoder expects True=padding)
3. Depthwise `Conv1d(groups: hiddenSize)` in `Gemma4ConformerLightConv1d`
4. `relPosInvTimescalesData: [Float]` (not `MLXArray`) so MLX doesn't treat it as a loadable parameter
5. EOS tokens `[1, 106, 50]` (in caller's generation loop, not in the library)
6. Audio token count from actual subsampling math, not duration estimate
7. `audioTokenId` default 258881 in `Gemma4ProcessorConfiguration`
8. `<|audio|>` placeholder injection into the prompt before the assistant turn

Plus one new fix that I caught while validating against real model weights:

9. `embed_audio.embedding_projection` uses `audioConfig.outputProjDims ?? hiddenSize` (commit 5 above). Without this, gemma-4-e4b-it-4bit fails to load with `mismatchedSize: biases [2560, 16] vs [2560, 24]`. The old branch had this correct; I dropped it during rebase and the standalone E2E test caught it.

### Verification

* `swift test --filter Gemma4` — 6 SPM-runnable tests pass (config decoding, token count, encoder output shape, mel preprocessing, end-to-end integration, Python alignment). The 4 Metal-dependent tests need `default.metallib` colocated with the test binary; they pass under Xcode.
* Standalone E2E inference test on a real 7.27s audio clip with `mlx-community/gemma-4-e4b-it-4bit`:
  ```
  prepare:    0.129s
  generation: 0.539s (20 tokens)
  audioShape: [1, 726, 128]   matches pre-rebase baseline exactly
  transcript: "我现在已经切換到詹姆斯四。這個就是詹姆斯四截屏的短視頻。"
  baseline:   "我现在已经切换到詹姆斯四。这个就是詹姆斯四接情的这段时期。"
  ```
  AudioShape and the first 12 tokens of the transcript are identical to the pre-rebase baseline (the trailing-segment difference is bf16/sampling noise on long ASR contexts; same prompt produces deterministic prefix). This proves the mel + Conformer + scatter + LLM-decode path is numerically equivalent.
* Talk app (the downstream consumer driving this work) verified separately on the pre-rebase commit; loader-API migration to the new `from:using:` shape is queued for a follow-up Talk PR once #192 lands.

Ready for review.

antmanler:
@vahsaechao thanks for reaching out about #194! After diffing both branches I think we can move both PRs in parallel without stepping on each other:

- #192 only touches `Gemma4.swift`, `Gemma4AudioFeatureExtractor.swift`, and the shared `LMInput.ProcessedAudio` / `UserInput.audios` infra.
- #194 only adds `Gemma3nAudio.swift`, `Gemma3nAudioConfig.swift`, `Gemma3nVLM.swift`.

Zero file-level conflicts — whichever lands first, the other rebases trivially.

For the shared Conformer code: the Gemma 4 and Gemma 3n configs differ enough (hidden_size 1024 vs 1536, head_dim 128 vs 192, different chunk sizes) that a generic shared module needs real design work to handle the parameterization cleanly. I'd rather do that in a follow-up PR once both implementations are in and we have two concrete call sites to factor against, instead of blocking either PR on it. Happy to co-author that extraction PR with you when the time comes.

Also worth noting: #192 ships a complete mel spectrogram feature extractor (`Gemma4AudioFeatureExtractor.swift`, vDSP RFFT + HTK mel filter bank) — your stub-preprocessing TODO in Gemma3nAudio is essentially the same pipeline with different mel parameters. Once #194 is in, it should be straightforward to port; let me know if you want me to do that as part of the consolidation PR.

antmanler:
Rebased this branch onto current `main` (`b331db1`) to resolve conflicts with #211 and #212. Details below in case it helps review.

## What changed in this rebase

The 5 commits are the same audio-tower series, replayed on top of the new base. The only non-mechanical change is the merge resolution in `Libraries/MLXVLM/Models/Gemma4.swift` inside `Gemma4Processor.prepare(input:)`:

- `#211` moved the image-token expansion inside the `if !input.images.isEmpty { ... }` block and added a `Gemma4MessageGenerator`.
- Our audio tower commit also touched the tail of `prepare()` to append audio feature extraction, mask building, and a 4-arg `LMInput(...)` return with `audio: processedAudio`.
- Resolution: keep `#211`'s image handling exactly as-is, run audio handling right after the `if` block closes, then return `LMInput(text:, image:, audio:)`. No logic moved between the two features.

While running the rebased test suite I also fixed one pre-existing flaky test in `Gemma4AudioAlignmentTest.melFilterBankShape` — at `n_mels=128, fmin=0, n_fft=512` the lowest HTK triangular filter has upper edge ≈ 27.9 Hz, which is below the 31.25 Hz FFT bin spacing, so that one filter column is legitimately all-zero (matches `librosa.filters.mel(htk=True, norm=None)`). The test now asserts "at most one all-zero filter, and filters from index 1 onward are all non-zero" instead of "every filter has non-zero coefficients." This failure reproduces on pre-rebase `ed51bf6` as well — it is not introduced by the rebase, just caught more reliably now.

## Verification

**Unit tests** — all Gemma 4 audio suites green after rebase:
```
✔ Suite "Gemma4 Audio Tower" — 3 tests
✔ Suite "Gemma4 Audio Integration" — 4 tests
✔ Suite "Gemma4 Audio Python Alignment" — 2 tests
✔ Suite "Gemma4 Audio Alignment" — 2 tests (incl. fixed melFilterBankShape)
11/11 passed
```

**End-to-end** — real-audio transcription against `mlx-community/gemma-4-e4b-it-4bit`, 7.272 s Mandarin clip loaded from local HF cache:
- prepare (mel extraction + token injection): **0.131 s**
- generation (20 tokens): **0.563 s**
- total: **0.694 s** → ~10.5× real-time
- output shape: `audio=[1, 726, 128]`, `tokens=[1, 202]`
- transcript: *我现在已经切換到詹姆斯四。這個就是詹姆斯四截屏的短視頻。*

Comparison against the pre-#211 baseline (`platx-ai/mlx-swift-lm@668347f`) on the same audio clip produced *我现在已经切换到詹姆斯四。这个就是詹姆斯四接情的这段时期。* Two notable differences:
1. Traditional/simplified character choices shifted (切換 vs 切换, 這個 vs 这个, 視頻 vs 视频) — this is the chat template change in #211 altering the prompt token sequence and therefore the sampling distribution.
2. Semantic quality went **up**: "截屏的短视频" (a screen-recorded short video) is a plausible utterance, while the baseline "接情的这段时期" is a hallucinated-sounding substring. So the rebase is a net quality improvement on this clip, not a regression.

Happy to rerun against additional clips or add more assertion coverage if you'd like.

vahsaechao:
OK, parallel merge makes sense. I've rebased #194 on main and clean it up for review.

For the mel extractor, I did a detailed comparison against the HuggingFace Python sources and found Gemma 3n and 4 differ more than expected: 32ms vs 20ms frames, 0.97 vs 0.0 preemphasis, standard vs offset Hann window, 1e-5 vs 1e-3 mel floor. Happy to help contribute to the consolidated extractor PR once both are in.

JaeminKim-amoz (approved):
**APPROVE**

Well-structured audio encoding pipeline. Clean separation of config, feature extractor, encoder, and integration. Good documentation in PR body explaining design decisions (FFT scaling, mask polarity, weight layout).

**Minor notes:**
- [MEDIUM] Large addition (~870 lines in Gemma4.swift) — consider splitting audio components into a separate `Gemma4Audio.swift` file
- [LOW] `expected_text` vs `python_output` mismatch in e2e reference JSON ("更新一版" vs "更新一把") — is this intentional?
- [LOW] `relPosInvTimescalesData` stored as `[Float]` to avoid Module parameter lookup could benefit from a comment

vahsaechao:
> @vahsaechao thanks for reaching out about #194! After diffing both branches I think we can move both PRs in parallel without stepping on each other:
> 
> * [Add Gemma 4 audio tower support (ASR via Conformer encoder) #192](https://github.com/ml-explore/mlx-swift-lm/pull/192) only touches `Gemma4.swift`, `Gemma4AudioFeatureExtractor.swift`, and the shared `LMInput.ProcessedAudio` / `UserInput.audios` infra.
> * [Add Gemma 3n E4B audio encoder (Conformer) support #194](https://github.com/ml-explore/mlx-swift-lm/pull/194) only adds `Gemma3nAudio.swift`, `Gemma3nAudioConfig.swift`, `Gemma3nVLM.swift`.
> 
> Zero file-level conflicts — whichever lands first, the other rebases trivially.
> 
> For the shared Conformer code: the Gemma 4 and Gemma 3n configs differ enough (hidden_size 1024 vs 1536, head_dim 128 vs 192, different chunk sizes) that a generic shared module needs real design work to handle the parameterization cleanly. I'd rather do that in a follow-up PR once both implementations are in and we have two concrete call sites to factor against, instead of blocking either PR on it. Happy to co-author that extraction PR with you when the time comes.
> 
> Also worth noting: #192 ships a complete mel spectrogram feature extractor (`Gemma4AudioFeatureExtractor.swift`, vDSP RFFT + HTK mel filter bank) — your stub-preprocessing TODO in Gemma3nAudio is essentially the same pipeline with different mel parameters. Once #194 is in, it should be straightforward to port; let me know if you want me to do that as part of the consolidation PR.

@antmanler Thanks for the detailed breakdown. 

I agree that the shared conformer extraction is better as a follow-up with both implementations in the tree. Happy to co-author that.

For the mel preprocessing I would appreciate your help porting `Gemma4AudioFeatureExtractor` to the Gemma 3n path. I've replaced the stub in `Gemma3nAudioVLMProcessor.prepare()` with a `preconditionFailure` guard so it's clear where the integration point is. Main difference is 80-bin mel vs 128-bin. Let me know if you want to pick that up in the consolidation PR or if I should take a first pass. Thanks!

Libraries/MLXVLM/Models/Gemma4.swift
```diff
@@ -1641,12 +1704,719 @@ private final class Gemma4MultimodalEmbedder: Module, UnaryLayer {
...
+        h = ffwLayer1(h)
+        h = silu(h)
+        h = ffwLayer2(h)
+        h = clip(h, min: -gradientClipping, max: gradientClipping)
+        h = postLayerNorm(h)
+        return residual + h * residualWeight
+    }
+}
+
+private final class Gemma4AudioRelativePositionEmbedding: Module {
```
davidkoski, line 1917: This class is not used -- should it be (I wonder if it is a bug that it is not called)?  If it isn't needed please remove.

Tests/MLXLMTests/Gemma4AudioTests.swift
```diff
@@ -0,0 +1,157 @@
...
+                    "num_audio_tokens": 750
+                }
+            }
+            """.data(using: .utf8)!
+
+        // This test will compile once AudioConfiguration is added to Gemma4.swift
+        // let config = try JSONDecoder().decode(Gemma4AudioConfig.self, from: json)
+        // #expect(config.numMelBins == 128)
+        // #expect(config.encoderLayers == 32)
+        // #expect(config.numAudioTokens == 750)
```
davidkoski, line 49: All of these assertions are commented out -- please update.

Libraries/MLXVLM/Models/Gemma4.swift
```diff
@@ -1909,6 +2763,8 @@ public struct Gemma4ProcessorConfiguration: Codable, Sendable {
         imageTokenId = try c.decodeIfPresent(Int.self, forKey: CodingKeys.imageTokenId) ?? 258_880
         boiTokenId = try c.decodeIfPresent(Int.self, forKey: CodingKeys.boiTokenId) ?? 255_999
         eoiTokenId = try c.decodeIfPresent(Int.self, forKey: CodingKeys.eoiTokenId) ?? 258_882
+        // Fix #7: default audioTokenId to 258881 when absent from processor_config.json
```
davidkoski, line 2766: Please remove comments like this (seems to refer back to PR tasks?)

Libraries/MLXLMCommon/UserInput.swift
```diff
@@ -177,6 +177,9 @@ public struct UserInput {
     /// collect the videos from the chat messages, otherwise these are the stored videos with the ``UserInput``.
     public var videos = [Video]()
 
+    /// Audio inputs as raw PCM float arrays (16 kHz mono expected by Gemma 4).
+    public var audios = [[Float]]()
```
davidkoski, line 181: Please introduce an `Audio` type modeled after the Image/Video types.

Libraries/MLXVLM/Models/Gemma4.swift
```diff
@@ -1641,12 +1704,719 @@ private final class Gemma4MultimodalEmbedder: Module, UnaryLayer {
...
+    @ModuleInfo(key: "relative_k_proj") var relativeKProj: Linear
+    @ParameterInfo(key: "per_dim_scale") var perDimScale: MLXArray
+    @ModuleInfo(key: "q_proj") var qProj: Gemma4AudioClippableLinear
+    @ModuleInfo(key: "k_proj") var kProj: Gemma4AudioClippableLinear
+    @ModuleInfo(key: "v_proj") var vProj: Gemma4AudioClippableLinear
+    @ModuleInfo(key: "post") var post: Gemma4AudioClippableLinear
+
+    // Relative position embedding (inline)
+    // Note: relPosInvTimescales is NOT a model parameter — it's a computed constant.
+    // Store as [Float] to avoid MLX Module treating it as a loadable weight.
```
davidkoski, line 2033: You can write the property like this:

```swift
private let _relPosInvTimescalesData: MLXArray
```

and initialize the array -- model loading won't touch it because of the leading underscore.

Libraries/MLXVLM/Models/Gemma4AudioFeatureExtractor.swift
```diff
@@ -0,0 +1,263 @@
...
+            samplingRate: samplingRate
+        )
+    }
+
+    /// Extract log-mel spectrogram from raw audio samples.
+    /// - Parameters:
+    ///   - audio: Raw waveform [numSamples] as Float array
+    ///   - maxLength: Maximum number of samples (default 480000 = 30s at 16kHz)
+    /// - Returns: (melSpectrogram: [frames, featureSize], mask: [frames])
+    public func extract(audio: [Float], maxLength: Int = 480_000) -> (MLXArray, MLXArray) {
```
davidkoski, line 135: Ideally the `Audio` type will produce an `MLXArray` and we can consume that here -- I think this code can probably be written to deal with the audio in that form and it will be executed on the GPU.  See what you think.

Libraries/MLXVLM/Models/Gemma4.swift
```diff
@@ -1865,9 +2685,41 @@ public struct Gemma4Processor: UserInputProcessor {
...
+        // Audio handling: extract mel features, inject <|audio|> placeholder tokens, build ProcessedAudio
+        var processedAudio: LMInput.ProcessedAudio? = nil
+        if !input.audios.isEmpty, let audioTokenId = config.audioTokenId {
+            let extractor = Gemma4AudioFeatureExtractor()
+            let (melFeatures, melMask) = extractor.extract(audio: input.audios[0])
+
+            // Fix #6: audio token count from actual subsampling math (two 2x conv blocks)
+            let melFrames = melFeatures.dim(0)
+            let afterConv0 = (melFrames + 2 - 3) / 2 + 1
+            let numAudioTokens = min((afterConv0 + 2 - 3) / 2 + 1, 750)
```
davidkoski, line 2697: What does the 750 represent?

Libraries/MLXVLM/Models/Gemma4.swift
```diff
@@ -1865,9 +2685,41 @@ public struct Gemma4Processor: UserInputProcessor {
...
+
+            // Fix #8: inject audio placeholder tokens into the prompt
+            // Insert before final assistant turn (last newline token 108) or append
+            let audioPlaceholders = Array(repeating: audioTokenId, count: numAudioTokens)
+            if let lastNewlineIdx = promptTokens.lastIndex(of: 108) {
```
davidkoski, line 2702: What does the 108 represent?

Libraries/MLXVLM/Models/Gemma4.swift
```diff
@@ -1865,9 +2685,41 @@ public struct Gemma4Processor: UserInputProcessor {
...
+                promptTokens.insert(contentsOf: audioPlaceholders, at: lastNewlineIdx)
+            } else {
+                promptTokens.append(contentsOf: audioPlaceholders)
+            }
+
+            // Fix #2: mask polarity inversion — extractor outputs 1=valid but encoder expects True=padding
+            let invertedMask = melMask .== 0
```
davidkoski, line 2709: Can the mask be generated with the correct polarity in the first place?  And as a bool MLXArray?

Libraries/MLXVLM/Models/Gemma4.swift
```diff
@@ -1682,41 +2470,65 @@ public final class Gemma4: Module, VLMModel, KVCacheDimensionProvider {
...
-            mask: imageMaskExpanded,
-            source: imageFeatures
-        )
+        // Scatter audio features into <|audio|> placeholder positions
+        if let audioFeatures,
+            let audioTower,
+            let embedAudio,
+            let audioTokenId = config.audioTokenId
+        {
+            // audioFeatures: [1, frames, melBins] ; audioMask: [1, frames] (True=padding)
```
davidkoski, line 2515: I think this path needs an equivalent to `Gemma4Error.imageTokenCountMismatch` -- it  can fatal inside `gemma4MaskedScatter` but I think it should throw an error like it does in the image path.

Libraries/MLXVLM/Models/Gemma4.swift
```diff
@@ -1865,9 +2685,41 @@ public struct Gemma4Processor: UserInputProcessor {
             promptTokens = expandedTokens
         }
 
+        // Audio handling: extract mel features, inject <|audio|> placeholder tokens, build ProcessedAudio
+        var processedAudio: LMInput.ProcessedAudio? = nil
+        if !input.audios.isEmpty, let audioTokenId = config.audioTokenId {
+            let extractor = Gemma4AudioFeatureExtractor()
+            let (melFeatures, melMask) = extractor.extract(audio: input.audios[0])
```
davidkoski, line 2692: Does this handle multiple audio files? Maybe it needs singleVideoAllowed (but for audio).

Libraries/MLXVLM/Models/Gemma4.swift
```diff
@@ -1682,41 +2470,65 @@ public final class Gemma4: Module, VLMModel, KVCacheDimensionProvider {
...
-        inputsEmbeds = gemma4MaskedScatter(
-            inputTensor: inputsEmbeds,
-            mask: imageMaskExpanded,
-            source: imageFeatures
-        )
+        // Scatter audio features into <|audio|> placeholder positions
+        if let audioFeatures,
+            let audioTower,
+            let embedAudio,
+            let audioTokenId = config.audioTokenId
```
davidkoski, line 2513: What if they provide an audio resource but it doesn't have the audioTower etc.?  Should this produce an error instead of silently dropping the resource?

Tests/MLXLMTests/Gemma4AudioAlignmentTest.swift
```diff
@@ -0,0 +1,148 @@
...
+
+@testable import MLXVLM
+
+@Suite("Gemma4 Audio Alignment")
+struct Gemma4AudioAlignmentTest {
+
+    /// Load JSON fixture from Tests/MLXVLMTests/Fixtures/
+    private func loadFixture(_ name: String) throws -> [String: Any] {
+        // #filePath points to source file location
+        let sourceDir = URL(fileURLWithPath: #filePath).deletingLastPathComponent()
```
davidkoski, line 20: Use this pattern:

```swift
    func testVideoFileAsSimpleProcessedSequence() async throws {
        guard let fileURL = Bundle.module.url(forResource: "1080p_30", withExtension: "mov") else {
            XCTFail("Missing file: 1080p_30.mov")
            return
        }
```

Libraries/MLXVLM/Models/Gemma4.swift
```diff
@@ -1641,12 +1704,719 @@ private final class Gemma4MultimodalEmbedder: Module, UnaryLayer {
     }
 }
 
+// MARK: - Audio
+
+private final class Gemma4AudioRMSNorm: Module, UnaryLayer {
```
davidkoski, line 1709: This is identical to Gemma4RMSNormZeroShift -- do we need a separate type?

Libraries/MLXVLM/Models/Gemma4AudioFeatureExtractor.swift
```diff
@@ -0,0 +1,263 @@
...
+        }
+    }
+
+    return MLXArray(filterBank, [numFrequencyBins, numMelFilters])
+}
+
+// MARK: - Feature Extractor
+
+/// Gemma4 audio feature extractor — converts raw waveform to log-mel spectrogram.
+public struct Gemma4AudioFeatureExtractor {
```
davidkoski, line 69: This seems somewhat generic -- do other audio towers have something similar?  I wonder if this is really Gemma4 specific?

And if not, perhaps it belongs more in MediaProcessing (maybe +Audio)?

Libraries/MLXLMCommon/UserInput.swift
davidkoski: Also, this needs to be represented in `ChatSession`.

Tests/MLXLMTests/Fixtures/gemma4_e2e_reference.json
davidkoski, line 1: This file looks like it is not used.  Only gemma4_mel_alignment.json is used.

Libraries/MLXVLM/Models/Gemma4AudioFeatureExtractor.swift
```diff
@@ -0,0 +1,263 @@
...
+            if idx < mask.count {
+                frameMask[f] = mask[idx]
+            }
+        }
+
+        return (melSpec, MLXArray(frameMask))
+    }
+
+    /// Compute mel spectrogram from windowed frames using Accelerate FFT.
+    private func computeMelSpectrogram(frames: [Float], numFrames: Int) -> MLXArray {
```
davidkoski, line 205: Again, if frames were an MLXArray then this could use MLX FFT (GPU).

Libraries/MLXLMCommon/UserInput.swift
davidkoski: The input is specific to gemma4 today, but much like the images it should be up to the input processor to convert to the desired format.

davidkoski:
Once #298 is merged, please rebase this on top of it -- I added generic audio support in all the right places.

antmanler:
Thanks @davidkoski — that's a much cleaner home for the audio representation than carrying it in this PR. I'll rebase `feat/gemma4-audio-tower` on top of #298 once it lands and drop the `UserInput.audios` / `LMInput.ProcessedAudio` bits this PR was adding, switching over to the generic `UserInput+Audio` support. I'll start prepping the rebase against the `audio-support` head now so the turnaround is quick once #298 merges.

davidkoski:
FYI #298 is merged -- please rebase when you get a chance.  Thank you!

davidkoski:

PR #193, opened 2026-04-07
feat: expose speculative decoding in ChatSession (#181)

VDurocher:
Closes #181

## What
Exposes the existing speculative decoding implementation (added in #173) through `ChatSession`'s public API.

Two files are modified:

**`Libraries/MLXLMCommon/Evaluate.swift`** — adds a public `generateTask` overload that accepts `consuming SpeculativeTokenIterator`. This mirrors the existing overload for `TokenIterator` and gives `ChatSession` a `(AsyncStream<Generation>, Task<Void, Never>)` pair it can use for clean early-termination.

**`Libraries/MLXLMCommon/ChatSession.swift`** — adds `SpeculativeDecodingConfig` and wires it in:
- New `public struct SpeculativeDecodingConfig: Sendable` with `draftModel: ModelContainer` and `numDraftTokens: Int`
- New `public let speculativeDecoding: SpeculativeDecodingConfig?` property
- All six `init` overloads gain a `speculativeDecoding: SpeculativeDecodingConfig? = nil` parameter (fully backward-compatible — default is `nil`)
- `streamMap` branches on `speculativeDecoding`: when set it extracts the draft `LanguageModel` from its `ModelContainer` (same pattern as the main model) and calls `SpeculativeTokenIterator` + the new `generateTask` overload; when `nil` the existing code path is taken unchanged

## Why
Speculative decoding provides ~2–3× inference speedup with no quality degradation by using a small draft model to generate candidate token sequences that the main model verifies in a single forward pass. The internal machinery already exists — this PR surfaces it to `ChatSession` users.

## API
```swift
// Load main and draft models
let main  = try await LLMModelFactory.shared.loadContainer(configuration: mainConfig)
let draft = try await LLMModelFactory.shared.loadContainer(configuration: draftConfig)

let session = ChatSession(
    main,
    speculativeDecoding: SpeculativeDecodingConfig(draftModel: draft, numDraftTokens: 5)
)

// Use exactly as before — speculative decoding is transparent
print(try await session.respond(to: "Hello!"))
```

## Notes
- `numDraftTokens` defaults to 5 (good balance speed/accuracy per the literature; the PR #173 benchmarks used 2 tokens — users can tune this)
- Falls back gracefully to standard decoding when `speculativeDecoding` is `nil`
- No public API change for existing `ChatSession` users (fully backward compatible)
- The draft model's KVCache is freshly allocated per-turn (same lifecycle as the main model's cache), so there is no cross-turn state leak
- `SpeculativeDecodingConfig` holds a `ModelContainer` (not a raw `LanguageModel`) to respect the same `Sendable` constraints already used by `ChatSession`


Libraries/MLXLMCommon/ChatSession.swift
```diff
@@ -4,6 +4,39 @@ import CoreGraphics
...
+///     speculativeDecoding: SpeculativeDecodingConfig(draftModel: draft, numDraftTokens: 5)
+/// )
+/// ```
+public struct SpeculativeDecodingConfig: Sendable {
+
+    /// Le modèle léger utilisé pour proposer des tokens candidats.
+    public let draftModel: ModelContainer
+
+    /// Nombre de tokens proposés par le modèle draft à chaque cycle de vérification.
+    /// La valeur par défaut de 5 offre un bon équilibre vitesse/précision.
```
davidkoski: Please translate the comments to English

Libraries/MLXLMCommon/ChatSession.swift
```diff
@@ -412,21 +476,49 @@ public final class ChatSession {
...
-                        )
+                        // Choisir l'itérateur selon la configuration du décodage spéculatif.
+                        let (genStream, genTask): (AsyncStream<Generation>, Task<Void, Never>)
+
+                        if let speculativeDecoding {
+                            // Extraire le modèle draft hors de son container (même pattern que le modèle principal).
+                            let draftModel = await speculativeDecoding.draftModel.perform { context in
+                                SendableBox(context.model)
+                            }.consume()
+                            let draftCache = draftModel.newCache(parameters: generateParameters)
```
davidkoski: I suspect we want the draftCache to be kept as state, just like the normal KVCache.  Otherwise I think the draft model won't have the state for the conversation.

Libraries/MLXLMCommon/Evaluate.swift
```diff
@@ -1500,6 +1500,44 @@ public func generateTask(
...
+///   - iterator: a pre-initialised ``SpeculativeTokenIterator``
+///   - wiredMemoryTicket: Optional wired memory ticket for policy-based coordination.
+/// - Returns: An `AsyncStream` that emits `Generation` values and a `Task`
+public func generateTask(
+    promptTokenCount: Int,
+    modelConfiguration: ModelConfiguration,
+    tokenizer: Tokenizer,
+    iterator: consuming SpeculativeTokenIterator,
+    wiredMemoryTicket: WiredMemoryTicket? = nil
+) -> (AsyncStream<Generation>, Task<Void, Never>) {
```
davidkoski: Rather than make a new entry point, I wonder if the existing entry could take `TokenIteratorProtocol`?  Either:

```swift
public func generateTask<TOKEN: TokenIteratorProtocol>(
    promptTokenCount: Int,
    tokenizer: Tokenizer,
    iterator: consuming TOKEN,
    wiredMemoryTicket: WiredMemoryTicket? = nil
```

or

```swift
public func generateTask(
    promptTokenCount: Int,
    tokenizer: Tokenizer,
    iterator: consuming any TokenIteratorProtocol,
    wiredMemoryTicket: WiredMemoryTicket? = nil
```

Libraries/MLXLMCommon/ChatSession.swift
VDurocher: Done, all French comments should be translated to English now

Libraries/MLXLMCommon/ChatSession.swift
VDurocher: True `draftKVCache` is now stored as an associated
  value on `Cache.kvcache` alongside the main KV cache. It's allocated lazily
  on the first turn and reused across subsequent turns, matching the lifecycle
  of the main model's cache.

JaeminKim-amoz (commented):
**COMMENT — one fix needed**

Clean API design. Fully backward-compatible with default `nil`. Good use of generics on `generateTask<TOKEN: TokenIteratorProtocol>`. Draft KV cache reuse across turns is well-handled.

**Requested fix:**
- [MEDIUM] `draftKVCache!` force-unwrap is fragile even though the `nil` check is 2 lines above. Please use `guard let` instead for safety.
- [LOW] No unit tests — a compile-time API test would be a minimum.
- [LOW] The `Cache` enum growing to `.kvcache([KVCache], draftKVCache: [KVCache]?)` — consider a struct if more fields are added later.

davidkoski (commented):
Thank you for your contribution!

davidkoski (approved):

strikingleon:
Do you know when we might get a new tagged release that includes the ChatSession support for Speculative Decoding that was part of #181 @VDurocher @davidkoski 

I think this will be a stellar help for something I am working on.

PR #196, opened 2026-04-08
Add Gemma 4 VLM support

louis030195:
## Summary
- Adds full Gemma 4 vision-language model support, ported from [mlx-vlm](https://github.com/Blaizzy/mlx-vlm/tree/main/mlx_vlm/models/gemma4) and [mlx-lm](https://github.com/ml-explore/mlx-lm/blob/main/mlx_lm/models/gemma4_text.py) Python implementations
- Creates `Gemma4.swift` with all model components (text, vision, multimodal projector, processor)
- Registers `gemma4` model type and `Gemma4Processor` processor type in `VLMModelFactory.swift`

### Key architectural differences from Gemma 3:
- **layer_types array** instead of `slidingWindowPattern` for per-layer attention type routing
- **rope_parameters dict** with per-layer-type RoPE configs (ProportionalRoPE for full attention with partial rotary factor)
- **global_head_dim** for full attention layers (can differ from sliding attention head_dim)
- **attention_k_eq_v** — K=V weight sharing in full attention layers
- **num_kv_shared_layers** — later layers reuse KV cache from earlier layers
- **use_double_wide_mlp** — 2x intermediate size for KV-shared layers
- **Per-Layer Embeddings (PLE)** — `hidden_size_per_layer_input` with per-layer token embeddings and gated projection
- **RMSNorm with scale_shift=0** — weight used directly (no `1+weight` offset as in Gemma 3)
- **layer_scalar** per decoder layer
- **New vision encoder**: RMSNorm (not LayerNorm), 2D multidimensional RoPE, learned position embedding table, average-pool based spatial downsampling
- **MultimodalEmbedder** — `Linear` projection + `RMSNormNoScale` replaces Gemma 3's avg-pool + einsum projector
- **Aspect-ratio preserving image resize** in processor

## Testing status
- [x] Swift build succeeds (`swift build` passes with zero errors)
- [ ] Text generation — **note:** the upstream Python `mlx-vlm` library also produces empty output with `mlx-community/gemma-4-e2b-it-4bit` (both 4-bit and 8-bit). This appears to be a known issue with very early Gemma 4 quantized checkpoints, not specific to this Swift port. The architecture implementation follows the Python reference faithfully.
- [ ] Vision generation — same upstream issue; Python `mlx-vlm` fails to generate with image input on Gemma 4 as well.
- [ ] Will need re-testing once stable Gemma 4 MLX checkpoints are available

## Test plan
- [ ] Re-test when mlx-community publishes updated Gemma 4 checkpoints
- [ ] Verify text-only generation
- [ ] Verify image+text generation with single image
- [ ] Verify KV cache sharing for models with `num_kv_shared_layers > 0`
- [ ] Verify PLE for E2B/E4B models with `hidden_size_per_layer_input > 0`

### DIFF
--- a/Libraries/MLXVLM/Models/Gemma4.swift
+++ b/Libraries/MLXVLM/Models/Gemma4.swift
@@ -1,0 +1,1756 @@
+// Copyright © 2025 Apple Inc.
+
+import CoreImage
+import MLX
+import MLXLMCommon
+import MLXNN
+
+// Based on https://github.com/Blaizzy/mlx-vlm/tree/main/mlx_vlm/models/gemma4
+
+// MARK: - RoPE Configuration
+
+/// Single layer-type RoPE config (e.g. for "full_attention" or "sliding_attention")
+public struct RoPEParameterEntry: Codable, Sendable {
+    let ropeTheta: Float?
+    let ropeType: String?
+    let partialRotaryFactor: Float?
+
+    enum CodingKeys: String, CodingKey {
+        case ropeTheta = "rope_theta"
+        case ropeType = "rope_type"
+        case partialRotaryFactor = "partial_rotary_factor"
+    }
+}
+
+/// ProportionalRoPE for Gemma 4 full-attention layers.
+///
+/// Frequencies are computed relative to the full head dimension (not just the
+/// rotated portion), and rotation is applied to the first rotated_dims//2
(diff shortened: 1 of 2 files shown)


mikedoise:
I believe this PR is similar to #180.

davidkoski:
Thank you for working on this -- #180 is merged, so closing this one.

PR #205, opened 2026-04-11
Fix sporadic JSON tool-call detection and add parser hardening for mixed text/tool outputs

aleroot:
## Proposed changes

This PR fixes sporadic cases where models output raw JSON tool calls (without wrapper tags) and the stream showed raw JSON instead of emitting a toolCall. It also hardens parser behavior to avoid regressions with normal text/JSON and mixed tagged + untagged content.
ToolCallProcessor primarily relied on tagged boundaries for streaming detection. When models (e.g. Qwen2.5 Instruct) occasionally emitted bare JSON tool calls, the content could bypass tool extraction and appear as raw text.

## Checklist

### DIFF
--- a/Libraries/MLXLMCommon/Evaluate.swift
+++ b/Libraries/MLXLMCommon/Evaluate.swift
@@ -1478,6 +1478,7 @@ public func generate(
 ///   - modelConfiguration: model configuration (for EOS/extra EOS tokens and tool-call format)
 ///   - tokenizer: tokenizer (for EOS id, unknown token id, and detokenization)
 ///   - iterator: token iterator
+///   - tools: Optional tool schemas used to validate emitted tool names.
 ///   - wiredMemoryTicket: Optional wired memory ticket for policy-based coordination.
 /// - Returns: An `AsyncStream` that emits `Generation` values and a `Task`
 public func generateTask(
@@ -1485,6 +1486,7 @@ public func generateTask(
     modelConfiguration: ModelConfiguration,
     tokenizer: Tokenizer,
     iterator: consuming TokenIterator,
+    tools: [[String: any Sendable]]?,
     wiredMemoryTicket: WiredMemoryTicket? = nil
 ) -> (AsyncStream<Generation>, Task<Void, Never>) {
     generateLoopTask(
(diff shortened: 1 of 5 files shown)


JaeminKim-amoz (approved):
**APPROVE**

Solid bug fix with defensive coding. State machine extension with `maxJSONFallbackBufferLength` safety valve is well-designed. 311 lines of new tests provide strong coverage. Backward compatibility preserved via `supportsBareJSONFallback` gating.

**Minor notes:**
- [MEDIUM] `ToolCallProcessor` is growing complex (470+ lines, 4 states) — consider extracting `JSONLeadingObjectScanner` handling into a separate strategy in the future
- [LOW] `maxJSONFallbackBufferLength = 32_768` — a doc comment explaining the sizing rationale would be helpful

DePasqualeOrg:

Libraries/MLXLMCommon/Evaluate.swift
```diff
@@ -1508,6 +1508,27 @@ public func generateTask<TOKEN: TokenIteratorProtocol>(
     )
 }
 
+/// Compatibility overload that preserves the original public signature.
+///
+/// This forwards to ``generateTask(promptTokenCount:modelConfiguration:tokenizer:iterator:tools:wiredMemoryTicket:)``
```
davidkoski: The docs build failed -- it looks like this is named:

```suggestion
/// This forwards to ``generateTask(promptTokenCount:modelConfiguration:tokenizer:iterator:wiredMemoryTicket:tools:)``
```

Libraries/MLXLMCommon/Tool/Parsers/JSONToolCallParser.swift
```diff
@@ -35,6 +35,22 @@ public struct JSONToolCallParser: ToolCallParser, Sendable {
             let function = try? JSONDecoder().decode(ToolCall.Function.self, from: normalizedData)
         else { return nil }
 
+        // If tool schemas are provided, only accept calls to declared tools.
```
davidkoski, line 38: This seems ok to me, but it is a difference from before and seems different than how mlx-lm handles it.  I am not sure how the LLM would propose to call tools that are not registered, so I think this is ok.

Libraries/MLXLMCommon/Evaluate.swift
```diff
@@ -1511,6 +1511,27 @@ public func generateTask<TOKEN: TokenIteratorProtocol>(
...
+///
+/// This forwards to ``generateTask(promptTokenCount:modelConfiguration:tokenizer:iterator:wiredMemoryTicket:tools:)``
+/// with `tools` unset.
+public func generateTask(
+    promptTokenCount: Int,
+    modelConfiguration: ModelConfiguration,
+    tokenizer: Tokenizer,
+    iterator: consuming TokenIterator,
+    wiredMemoryTicket: WiredMemoryTicket? = nil
+) -> (AsyncStream<Generation>, Task<Void, Never>) {
```
davidkoski: I think this overload is not needed -- perhaps it was a moving target, this is an old PR.  We already have:

```swift
public func generateTask<TOKEN: TokenIteratorProtocol>(
    promptTokenCount: Int,
    tokenizer: Tokenizer,
    iterator: consuming TOKEN,
    wiredMemoryTicket: WiredMemoryTicket? = nil,
    tools: [[String: any Sendable]]? = nil
```

which accepts nil tools.

I think this should be removed.

davidkoski (changes requested):
Looks good, but please remove the override in Evaluate -- I think it is not necessary (maybe no longer necessary).

Libraries/MLXLMCommon/Evaluate.swift
aleroot: Done, thanks.

aleroot:
> Looks good, but please remove the override in Evaluate -- I think it is not necessary (maybe no longer necessary).

Updated, let me know if now is good to go. Thanks.

davidkoski (approved):

PR #212, opened 2026-04-14
Prepare Gemma4Text for batched RoPE offsets

ronaldmannak:
## Proposed changes

This PR prepares the LLM-side Gemma 4 path for upcoming batching work without pulling in the rest of the batching implementation.

Changes:
- Add BatchPositionedKVCache as a small shared hook for caches that expose per-sequence position offsets.
- Update `applyRotaryPosition(...)` to use `batchOffset` when available and fall back to the existing scalar offset path otherwise.
- Refactor `Gemma4Text` to snapshot the pre-update RoPE position before cache mutation instead of reading mutable cache state later.
- Thread that captured position through Gemma 4 shared-KV layers so position handling is consistent across the model.

Why:
- Gemma 4 needs stable pre-update RoPE positions for correct decode behavior.
- This keeps the Gemma 4 model changes aligned with the future batching API shape.
- It reduces churn for the follow-up batching PR by landing the shared RoPE seam now.

Notes:
- This applies to the LLM Gemma 4 path routed through Gemma4Text, including model_type: `gemma4` and model_type: `gemma4_text`.

## Checklist

### DIFF
--- a/Libraries/MLXLLM/Models/Gemma4Text.swift
+++ b/Libraries/MLXLLM/Models/Gemma4Text.swift
@@ -175,6 +175,33 @@ private class ScaledLinear: Module {
     }
 }
 
+private enum Gemma4PositionOffset {
+    case scalar(Int)
+    case batch(MLXArray)
+}
+
+private func gemma4CapturePositionOffset(from cache: KVCache?) -> Gemma4PositionOffset {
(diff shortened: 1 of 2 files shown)


davidkoski (approved):
Changes look good, thank you!

PR #213, opened 2026-04-14
Gemma 4 VLM implementation with proportional RoPE and tool calling

BRVWL:
## Summary

- **Gemma 3/4 Tool calling** with dual format support in `GemmaFunctionParser` and `hasPrefix("gemma")` format inference for all gemma variants

## Changes

- `Libraries/MLXLMCommon/Tool/Parsers/GemmaFunctionParser.swift` - dual format support (Gemma 3 & 4)
- `Libraries/MLXLMCommon/Tool/ToolCallFormat.swift` - `hasPrefix("gemma")` for gemma4 variant matching

Note: Gemma4Text.swift and Gemma4.swift use the official upstream implementations from mlx-lm.


davidkoski:
Are these meant to be modifications of the existing Gemma 4 implementations?  For MLXLLM we don't want a port of the VLM text tower (unless that is all we have).  Preferable a port of the one from mlx-lm -- sometimes there are interesting differences and it is good to have those here.

We don't need a rewrite, but if there is something focused that you think needs fixing we can take that.

JaeminKim-amoz (approved):
**APPROVE**

Well-structured Gemma 4 VLM implementation. Good config refactor from `var` defaults to `let` with `decodeIfPresent`. `effectiveLayerTypes` computed property is a nice pattern. VLM tests included.

**Minor notes:**
- [LOW] `Gemma4RMSNormNoScale` / `Gemma4RMSNormZeroShift` are `internal` — intentional for cross-module VLM use?
- [LOW] `ToolCallFormat.swift` uses `hasPrefix("gemma")` which may over-match future model families
- [LOW] `Gemma4Processor.swift` (334 lines) has no processor-specific tests

PR #183, opened 2026-04-04
Adopt GemmaFunctionParser to accomodate Gemma4 tool calls.

viktike:
## Proposed changes

I see #180 by @drgrondin is already looking good, this complements Gemma 4 with tool calling.
It reuses **GemmaFunctionParser**, and initializes it with three parameters:
- startTag
- endTag
- escapeMarker

To distinguish between the different flavors of tool calling.

## Checklist

Put an `x` in the boxes that apply.

- [x] I have read the [CONTRIBUTING](https://github.com/ml-explore/mlx/blob/main/CONTRIBUTING.md) document
- [x] I have run `pre-commit run --all-files` to format my code / installed pre-commit prior to committing changes
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] I have updated the necessary documentation (if needed)

PR #215, opened 2026-04-15
Gemma 3/4 tool calling support

BRVWL:
- Add dual format support in `GemmaFunctionParser` for both Gemma 3 (`<start_function_call>/<end_function_call>`) and Gemma 4 (`<|tool_call>/<tool_call|>`) formats
- Change `ToolCallFormat.infer()` to use `hasPrefix("gemma")` instead of exact match, enabling tool calling for `gemma2`, `gemma3`, `gemma3n`, `gemma4`, `gemma4_text` model variants
- Add comprehensive tests for Gemma 4 format parsing, escaped strings, and auto-detection

### DIFF
--- a/Libraries/MLXLMCommon/Tool/Parsers/GemmaFunctionParser.swift
+++ b/Libraries/MLXLMCommon/Tool/Parsers/GemmaFunctionParser.swift
@@ -3,24 +3,44 @@
 import Foundation
 
 /// Parser for Gemma format: call:name{key:value,k:<escape>str<escape>}
+/// Supports both Gemma 3 (<start_function_call>/<end_function_call>) and
+/// Gemma 4 (<|tool_call>/<tool_call|>) formats.
 /// Reference: https://github.com/ml-explore/mlx-lm/blob/main/mlx_lm/tool_parsers/function_gemma.py
 public struct GemmaFunctionParser: ToolCallParser, Sendable {
-    public let startTag: String? = "<start_function_call>"
-    public let endTag: String? = "<end_function_call>"
+    // Gemma 4 format tags (primary, as Gemma 4 is the current model)
+    public let startTag: String? = "<|tool_call>"
+    public let endTag: String? = "<tool_call|>"
 
+    // Gemma 3 format tags (fallback detection)
+    private static let gemma3StartTag = "<start_function_call>"
+    private static let gemma3EndTag = "<end_function_call>"
+
+    // Gemma 3 escape marker
     private let escapeMarker = "<escape>"
+    // Gemma 4 escape marker
+    private static let gemma4EscapeMarker = "<|\"|>"
 
     public init() {}
 
+    /// Detects which Gemma format is being used based on content
+    private func detectFormat(in text: String) -> (isGemma4: Bool, stripped: String) {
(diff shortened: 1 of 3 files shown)


Libraries/MLXLMCommon/Tool/Parsers/GemmaFunctionParser.swift
```diff
@@ -3,24 +3,44 @@
...
 
 /// Parser for Gemma format: call:name{key:value,k:<escape>str<escape>}
+/// Supports both Gemma 3 (<start_function_call>/<end_function_call>) and
+/// Gemma 4 (<|tool_call>/<tool_call|>) formats.
 /// Reference: https://github.com/ml-explore/mlx-lm/blob/main/mlx_lm/tool_parsers/function_gemma.py
 public struct GemmaFunctionParser: ToolCallParser, Sendable {
-    public let startTag: String? = "<start_function_call>"
-    public let endTag: String? = "<end_function_call>"
+    // Gemma 4 format tags (primary, as Gemma 4 is the current model)
+    public let startTag: String? = "<|tool_call>"
```
aleroot, line 11: Changing `startTag` to only the Gemma 4 marker breaks streamed Gemma 3 tool calls. `ToolCallProcessor` uses `parser.startTag` as the sole trigger, so `<start_function_call>...<end_function_call>` output will now pass through as regular text even though `parse(content:)` can handle it directly. Since this PR also maps `gemma3`/`gemma3n` to `.gemma`, please either keep streaming detection compatible with both start tags or split Gemma 3 and Gemma 4 into distinct parser formats.

Libraries/MLXLMCommon/Tool/Parsers/GemmaFunctionParser.swift
```diff
@@ -3,24 +3,44 @@
...
+            var stripped = text
+            stripped = stripped.replacingOccurrences(of: st, with: "")
+            stripped = stripped.replacingOccurrences(of: et, with: "")
+            return (true, stripped)
         }
+        // Fallback to Gemma 3 tags
+        var stripped = text
+        stripped = stripped.replacingOccurrences(of: Self.gemma3StartTag, with: "")
+        stripped = stripped.replacingOccurrences(of: Self.gemma3EndTag, with: "")
+        return (false, stripped)
```
aleroot, line 38: This fallback mis-detects unterminated Gemma 4 calls during `processEOS()`. The default `ToolCallParser.parseEOS` splits the buffer on `startTag` before calling `parse`, so a buffered value like `<|tool_call>call:search{query:<|"|>hello, world<|"|>}` reaches `detectFormat` without either Gemma 4 wrapper tag. It then falls back to the Gemma 3 escape marker and splits the escaped string at the comma. Please override `parseEOS` for Gemma or otherwise preserve the detected Gemma 4 format after splitting.

aleroot (changes requested):
Thanks for adding this. I’m interested in Gemma 4 tool calling support as well, and the general direction looks good. I think this needs a couple of adjustments before merging. Once those are covered with processor-level regression tests for both formats, this should be merged.

DePasqualeOrg:

davidkoski:
Closing as dup of #183.  See also #360 which is in the same area.

PR #216, opened 2026-04-16
Doc corrections for v3

DePasqualeOrg:
I made a few corrections and updates to the docs.

I also expanded the version range for swift-syntax, since I ran into a conflict with another project that uses that package. I think this should just make this package more compatible with others.

I'm wondering if the usage examples for the integration packages could be confusing to users. It might be better to instead just link to the readmes of the integration packages, each of which shows the usage specific to that package. The issue is that any example has to show specific usage for one package or set or packages.

I don't want to favor my own packages. I just want to avoid confusion for users, and I'm thinking a less-is-more approach could be better.

Even with the current docs, Claude Code was able to set up a new project and get inference working without directions from me, but it stumbled over some of the errors that I fixed here.

### DIFF
--- a/Libraries/MLXLLM/Documentation.docc/evaluation.md
+++ b/Libraries/MLXLLM/Documentation.docc/evaluation.md
@@ -10,8 +10,8 @@ let model = try await loadModel(
     id: "mlx-community/Qwen3-4B-4bit"
 )
 let session = ChatSession(model)
-print(try await session.respond(to: "What are two things to see in San Francisco?")
-print(try await session.respond(to: "How about a great place to eat?")
+print(try await session.respond(to: "What are two things to see in San Francisco?"))
+print(try await session.respond(to: "How about a great place to eat?"))
 ```
 
 The second question actually refers to information (the location) from the first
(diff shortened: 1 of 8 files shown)


davidkoski:
> I'm wondering if the usage examples for the integration packages could be confusing to users. It might be better to instead just link to the readmes of the integration packages, each of which shows the usage specific to that package. The issue is that any example has to show specific usage for one package or set or packages.

I was thinking along these lines too.  I wanted to have a taste of the integration in the README (e.g. it is pretty easy) but you are right, it does have to pick a method of integration. 

Perhaps it can just show a snippet of code (incomplete) and tell users to read the docs (with link).  I think it is still useful to have something short to show it is quite approachable.

Libraries/MLXLLM/Documentation.docc/evaluation.md
```diff
@@ -10,8 +10,8 @@ let model = try await loadModel(
     id: "mlx-community/Qwen3-4B-4bit"
 )
 let session = ChatSession(model)
-print(try await session.respond(to: "What are two things to see in San Francisco?")
-print(try await session.respond(to: "How about a great place to eat?")
+print(try await session.respond(to: "What are two things to see in San Francisco?"))
+print(try await session.respond(to: "How about a great place to eat?"))
```
davidkoski, line 14: I swear this particular set of missing parens has been fixed a dozen times!

davidkoski:
FYI I am going to be on vacation from 18 April - 7 May -- if you are wondering why I am less responsive than usual.

davidkoski (approved):
Changes are good, thank you!

Issue #221, opened 2026-04-17
Upstreaming improvements from fast-moving forks


ronaldmannak:
I wanted to raise something that seems worth discussing.
There are several forks that have been improving quickly and experimenting with exciting features, fixes, and significant performance improvements. At the same time, they also appear to be diverging more and more from this repo, which could make it harder over time to bring useful work back and avoid fragmentation.

Most notable examples:
- [ekryski/mlx-swift-lm](https://github.com/ekryski/mlx-swift-lm) by @ekryski
- [osaurus-ai/mlx-swift-lm](https://github.com/osaurus-ai/mlx-swift-lm) and [osaurus-ai/vmlx-swift-lm](https://github.com/osaurus-ai/vmlx-swift-lm) by @jjang-ai and @tpae

For example, here are benchmarks shared for the Osaurus fork:

| Model | Upstream | This Fork | Gain |
|-------|---------:|----------:|-----:|
| Gemma 4 26B MoE | 25.0 tok/s | **101 tok/s** | **+304%** |
| Qwen 3.5-35B MoE | 42.4 tok/s | **61 tok/s** | **+44%** |
| NemotronH 30B-A3B | ~25 tok/s | **48 tok/s** | **+92%** |
| Qwen 3.5-4B Dense | 123 tok/s | 145 tok/s | +18% |

There’s also a recent demo from @ekryski [here](https://x.com/ekryski/status/2045277902649073826?s=46&t=CSJUc-EX4J03ow_Gi8B9Lw):  

It would be helpful to get maintainer thoughts on a few things:
- whether you’re already tracking the mentioned forks
- which kinds of improvements from those forks would be welcome here
- whether there’s a preferred process for proposing or extracting those changes back into this repo
- whether there are architectural, maintenance, or scope reasons that would make some of those changes a poor fit upstream

A discussion around this could help clarify:

- how contributors should think about upstreaming improvements from forks
- whether maintainers would prefer smaller targeted PRs instead of fork-sized syncs
- whether a roadmap or “areas open to contribution” note would help reduce unnecessary divergence

Even a brief maintainer perspective on if and how these fork improvements can realistically make their way back into this repo would be very useful.

ronaldmannak:
@davidkoski I know you'll be offline for the next two weeks, so I'm happy to continue this conversation after you return.

davidkoski:
[...]

PR #222, opened 2026-04-19
Fix Qwen2.5-VL: 7 bugs + preprocessing for Python mlx-vlm parity

NivDvir:
## Summary

Fixes 7 bugs in the native Swift Qwen2.5-VL implementation that collectively cause the model to produce wrong bounding box coordinates on grounding tasks, plus a PIL-matching image preprocessing path that closes the remaining sub-pixel gap against the Python reference. With everything applied, Swift output matches the Python reference (`mlx-vlm 0.4.3`, `mlx-community/Qwen2.5-VL-7B-Instruct-4bit`) at **≤ 2 px on all bbox edges — 6 of 8 bit-exact** on the LeetCode 2-panel test image.

Raised in the context of #221 ("Upstreaming improvements from fast-moving forks"). The current shape (see "File map" below) is already three topical files plus one 6-line ordering fix, so each file can be reviewed in isolation; happy to split further across commits if maintainers prefer that.

## File map

| File | Size | What it does |
|---|---|---|
| `Libraries/MLXVLM/MediaProcessing.swift` | +300 / -13 | Adds `resamplePILLanczos`, a bit-identical port of Pillow 10.4's `Resample.c` Lanczos kernel, plus an ICC-retag pathway that avoids the Display-P3 → sRGB conversion macOS screenshots otherwise get. Used only by the Qwen2.5-VL preprocessor. |
| `Libraries/MLXVLM/Models/Qwen25VL.swift` | +136 / -43 | The 7 model-side fixes (below). File-level header comment orients a reviewer to the MROPE extensions. |
| `Libraries/MLXVLM/Models/Qwen25VLRopeIndex.swift` | +135 / -0 | Extracted `Qwen25VL.getRopeIndex` — the port of HuggingFace `Qwen2_5_VLForConditionalGeneration.get_rope_index`. Pure compute, no MLX state. Split into its own file so the model diff stays readable. |
| `Libraries/MLXVLM/Models/Qwen2VL.swift` | +6 / -10 | `Qwen2VLMessageGenerator` emits vision tokens before text (bug #6). Used by Qwen2-VL and Qwen2.5-VL alike. |

## Background

These bugs surfaced while building a native Swift macOS screen-reader ([screen-overlay-toolkit][toolkit]) that uses `mlx-swift-lm`'s Qwen2.5-VL as a real-time UI panel detector on Apple Silicon. The symptom was consistent: the model returned JSON with `bbox_2d` fields, but the coordinates were off by 200–800 px from ground truth. The same image through the Python `mlx-vlm` reference returned correct coordinates. That gap drove the investigation.

The [full debugging writeup][writeup] covers the methodology end-to-end, including two additional bugs that are consumer-side (prompt format, `max_tokens`) and therefore **out of scope for this PR**.

### 1. MROPE section selection (split-select vs slice-replace)

Multi-Resolution Rotary Position Embedding assigns different frequency bands to temporal (T), height (H), width (W). The Swift implementation split the frequency tensor with modulo indexing (`i % 3`), which **interleaves** frequencies across dimensions. Python starts with temporal frequencies and **overwrites H/W slices in place**, producing `[T₀..₁₅, H₁₆..₃₉, W₄₀..₆₃]`. Interleaving destroys the positional structure the attention layer relies on.

### 2. `invFreq` registered as a Module weight

`invFreq` is computed from `ropeTheta` and `headDim`, not loaded. Declaring it as an `MLXArray` property on a `Module` subclass exposed it to MLX's weight-loading reflection, which either raised `keyNotFound` or silently overwrote it from the checkpoint. Fix: wrap in a nested non-`Module` class (`InvFreqBox`) to hide from reflection.

(Qwen3VL and Qwen35 solve the same problem by pulling their `RotaryEmbedding` into a standalone non-`Module` class — that's a cleaner idiom but a larger refactor; the box is the minimal form of the same pattern.)

### 3. `rope_deltas` unused during autoregressive generation

After prefill, the code cleared cached position IDs but never applied `rope_deltas` during token generation. Correct: `positionIds = cacheOffset + ropeDeltas + arange(seqLen)`. Without the deltas, position embeddings drifted with each generated token and output degraded progressively.

### 4. MROPE state not reset between successive images

`_positionIds` and `_ropeDeltas` from one inference persisted into the next. Second, third, and subsequent inferences used the prior image's accumulated offset as their starting point. Each run got progressively worse.

### 5. Image resize cap at 1800 px instead of 1280 px

The Swift path capped the longest edge at 1800 px. Python mlx-vlm defaults to 1280 px, matching the `min_pixels` / `max_pixels` the model was trained under. 1800 px produces a noticeably larger visual-token count and pushes token positions outside the training distribution.

### 6. Chat template ordering (text before image)

`Qwen2VLMessageGenerator` emitted the content array as `[text, image]`. The Python reference emits `[image, text]` → `<|vision_start|><|image_pad|>×N<|vision_end|>PROMPT`. Attention patterns are position-dependent — placing text before the image-feature injection point shifts what the early layers attend to. This is the only change in `Qwen2VL.swift`.

### 7. Vision attention mask ignored

The vision encoder's self-attention uses a mask to implement **windowed attention** (each patch attends only to patches in its local window). The Swift code passed `mask: .none` to `scaledDotProductAttention` instead of `mask: .array(floatMask)`. Without the mask, every patch attended globally, destroying the spatial locality the model relies on for coordinate prediction. This is a single-line fix; in isolation it has an outsized visual impact, but empirically each of the 7 fixes is load-bearing — reverting any one of them breaks coherent output.

```swift
// Before:
let attn = scaledDotProductAttention(queries: q, keys: k, values: v, scale: scale, mask: .none)
// After:
let attn = scaledDotProductAttention(queries: q, keys: k, values: v, scale: scale, mask: .array(floatMask))
```

## Preprocessing: PIL-matching Lanczos resampler

Fixing the 7 bugs above gets to ~30 px parity. Closing the rest comes down to a preprocessing disagreement that's independent of the model.

Python mlx-vlm calls `PIL.Image.resize(..., Image.LANCZOS)`, which is Pillow's int32-fixed-point separable Lanczos kernel (a=3, clamp boundary, `PRECISION_BITS=22`). Core Image's `CILanczosScaleTransform` — and `vImage`, and MPS — implement a Lanczos variant that diverges from Pillow on high-frequency content, enough to shift downstream VLM bbox coordinates by tens of pixels.

Rather than switch resamplers, this PR adds `resamplePILLanczos` in `MediaProcessing.swift`: a direct Swift port of Pillow 10.4's `Resample.c` 8bpc fast path (~300 lines). It also retags `CGImage` as sRGB without conversion, because macOS screenshots carry Display-P3 ICC profiles that `ImageIO` would otherwise convert on load — producing different uint8 bytes than `PIL.Image.open` sees (which ignores the ICC tag).

The Qwen25VL preprocess path uses this helper and bypasses the `CIImage → sRGB → bicubic → normalize → MLXArray` downstream chain entirely, which otherwise re-encodes bytes via the working colorspace.

This helper is scoped to the Qwen VLM preprocessing path — callers of the existing `MediaProcessing.resampleLanczos` (Core-Image-based) are unaffected.

## Parity evidence

Against `mlx-community/Qwen2.5-VL-7B-Instruct-4bit` via `mlx-vlm 0.4.3` on a screenshot of the LeetCode "Two Sum" problem (a 2-panel UI — question left, code editor right):

```
python: {"question": [1, 146, 421, 626], "editor": [421, 146, 881, 626]}
swift:  {"question": [1, 146, 421, 628], "editor": [421, 146, 881, 628]}
deltas: [0,0,0,2] and [0,0,0,2] — max 2 px, 6 of 8 edges bit-exact
```

I have a ~130-line bash reproducer that pins this branch, runs Python + Swift on a fixed test image, and prints per-edge deltas. Happy to land it somewhere public (a separate gist, a `scripts/` directory in this repo, or inlined into a PR comment) — tell me which you'd prefer.

**Cross-architecture smoke test.** I also ran `mlx-community/UI-TARS-1.5-7B-mlx-bf16` through the same patched Swift path and got coherent output — UI-TARS shares the `Qwen2_5_VLForConditionalGeneration` architecture, so the fixes are architecture-level rather than Qwen2.5-VL-specific. Happy to share that report if useful (it's not public yet).

## Notes for reviewers

- **`getRopeIndex` extracted to its own file** — it's a standalone compute, and extracting it keeps the model diff focused on the stateful parts. If you'd prefer it inline, happy to revert that split.
- **Stored state on `Qwen25LanguageModel`** (`_positionIds`, `_ropeDeltas`) — architectural smell. MROPE state needs to persist across prefill → decode steps, and the `KVCache` protocol doesn't carry it. Pulling this into the cache protocol (or threading it through the call chain) would touch upstream contracts; opted not to do that here, but happy to if you'd prefer. See the file-level header comment in `Qwen25VL.swift` for the full rationale.
- **No unit tests** — I don't have a clean way to exercise the full vision encoder + language model against a reference output within the existing test harness. Pointers welcome; happy to add targeted tests.

## Related

- Full debugging writeup: [_Building a Real-Time Screen Reader on macOS That Actually Works_][writeup]
- Consumer app demonstrating these fixes: [screen-overlay-toolkit][toolkit]
- Upstream issue that prompted this PR: #221

[writeup]: https://dev.to/nivdvir/building-a-real-time-screen-reader-on-macos-that-actually-works-471
[toolkit]: https://github.com/NivDvir/screen-overlay-toolkit

### DIFF
--- a/Libraries/MLXVLM/MediaProcessing.swift
+++ b/Libraries/MLXVLM/MediaProcessing.swift
@@ -74,27 +74,327 @@ public enum MediaProcessing {
 
     /// Resample the image using Lanczos interpolation.
     static public func resampleLanczos(_ image: CIImage, to size: CGSize) -> CIImage {
-        // Create a bicubic scale filter
-
-        let yScale = size.height / image.extent.height
-        let xScale = size.width / image.extent.width
+        // clampedToExtent replicates edge pixels, matching PIL's clamp boundary mode.
+        let extent = image.extent
+        let yScale = size.height / extent.height
+        let xScale = size.width / extent.width
 
         let filter = CIFilter.lanczosScaleTransform()
-        filter.inputImage = image
+        filter.inputImage = image.clampedToExtent()
         filter.scale = Float(yScale)
         filter.aspectRatio = Float(xScale / yScale)
         let scaledImage = filter.outputImage!
 
-        // Create a rect with the exact dimensions we want
-        let exactRect = CGRect(
-            x: 0,
-            y: 0,
-            width: size.width,
-            height: size.height
-        )
+        return scaledImage.cropped(to: CGRect(x: 0, y: 0, width: size.width, height: size.height))
+    }
 
-        // Crop to ensure exact dimensions
-        return scaledImage.cropped(to: exactRect)
+    /// PIL-matching Lanczos resampler, bit-identical to `PIL.Image.LANCZOS` (a=3,
+    /// separable, clamp boundary, int32 fixed-point with PRECISION_BITS=22).
+    ///
+    /// Apple's Lanczos family (Core Image / MPS / vImage) all share the same
+    /// underlying implementation that diverges from PIL on high-frequency
+    /// content — enough to shift VLM grounding bboxes by tens of pixels. This
+    /// function is a direct port of Pillow 10.4.0's `Resample.c` 8bpc fast path.
+    ///
+    /// Also bypasses Core Graphics' automatic ICC-profile conversion: macOS
+    /// screenshots are tagged Display P3, and ImageIO would otherwise convert
+    /// P3→sRGB on load, producing different uint8 bytes than `PIL.Image.open`
+    /// sees (which ignores the ICC tag). We retag without converting.
+    static public func resamplePILLanczos(_ image: CIImage, to size: CGSize) -> CIImage {
+        let tw = Int(size.width)
+        let th = Int(size.height)
+        let srgb = CGColorSpace(name: CGColorSpace.sRGB)!
+
+        // Read raw uint8 bytes WITHOUT ICC conversion. If the source CIImage
+        // was built from a CGImage with a tagged colorspace (Display P3, etc.),
+        // we retag as sRGB so CoreGraphics treats the bytes as already-sRGB.
+        let iw: Int
+        let ih: Int
+        var src: [UInt8]
+        if let cg = image.cgImage, let provider = cg.dataProvider {
+            iw = cg.width
+            ih = cg.height
+            let retagged = CGImage(
+                width: iw, height: ih,
+                bitsPerComponent: cg.bitsPerComponent,
+                bitsPerPixel: cg.bitsPerPixel,
+                bytesPerRow: cg.bytesPerRow,
+                space: srgb, bitmapInfo: cg.bitmapInfo,
+                provider: provider, decode: nil,
+                shouldInterpolate: false, intent: .defaultIntent) ?? cg
+            src = [UInt8](repeating: 0, count: ih * iw * 4)
+            let cgctx = CGContext(
+                data: &src, width: iw, height: ih,
+                bitsPerComponent: 8, bytesPerRow: iw * 4,
+                space: srgb,
+                bitmapInfo: CGImageAlphaInfo.premultipliedLast.rawValue
+                          | CGBitmapInfo.byteOrder32Big.rawValue)!
+            cgctx.draw(retagged, in: CGRect(x: 0, y: 0, width: iw, height: ih))
+        } else {
+            iw = Int(image.extent.width)
+            ih = Int(image.extent.height)
+            src = [UInt8](repeating: 0, count: ih * iw * 4)
+            let cictx = CIContext(options: [.workingColorSpace: srgb])
+            cictx.render(image, toBitmap: &src, rowBytes: iw * 4,
+                         bounds: image.extent, format: CIFormat.RGBA8,
+                         colorSpace: srgb)
+        }
+
+        // --- Pillow 10.4.0 Resample.c port ---
+
+        let PRECISION_BITS: Int32 = 22
+        let ROUND_BIAS: Int32 = 1 << (PRECISION_BITS - 1)
+        let ONE = Double(Int32(1) << PRECISION_BITS)
+        let a = 3.0
+
+        @inline(__always) func sinc(_ x: Double) -> Double {
+            if x == 0 { return 1 }
+            let p = Double.pi * x
+            return Foundation.sin(p) / p
+        }
+        @inline(__always) func kernel(_ x: Double) -> Double {
+            if abs(x) >= a { return 0 }
+            return sinc(x) * sinc(x / a)
+        }
+        func precompute(_ inSize: Int, _ outSize: Int) -> ([Int], [Int], [[Int32]]) {
+            let scale = Double(inSize) / Double(outSize)
+            let fs = max(1.0, scale)
+            let support = a * fs
+            var lo = [Int](); var hi = [Int](); var qq = [[Int32]]()
+            lo.reserveCapacity(outSize); hi.reserveCapacity(outSize); qq.reserveCapacity(outSize)
+            for i in 0 ..< outSize {
+                let c = (Double(i) + 0.5) * scale
+                let l = max(Int(floor(c - support + 0.5)), 0)
+                let r = min(Int(floor(c + support + 0.5)), inSize)
+                var w = [Double](repeating: 0, count: r - l)
+                var sum = 0.0
+                for k in 0 ..< (r - l) {
+                    let x = (Double(l + k) + 0.5 - c) / fs
+                    let v = kernel(x); w[k] = v; sum += v
+                }
+                if sum != 0 { for k in 0 ..< w.count { w[k] /= sum } }
+                // Quantize to int32 with 22-bit precision. C's (int) truncates toward zero.
+                let q = w.map { weight -> Int32 in
+                    let s = weight * ONE
+                    let b = s < 0 ? -0.5 + s : 0.5 + s
+                    return Int32(b.rounded(.towardZero))
+                }
+                lo.append(l); hi.append(r); qq.append(q)
+            }
+            return (lo, hi, qq)
+        }
+
+        @inline(__always) func clip8(_ ss: Int32) -> UInt8 {
+            let s = ss >> PRECISION_BITS
+            if s < 0 { return 0 }
+            if s > 255 { return 255 }
+            return UInt8(s)
+        }
+
+        let (hlo, hhi, hw) = precompute(iw, tw)
+        let (vlo, vhi, vw) = precompute(ih, th)
+
+        // Horizontal pass: uint8 → uint8 at (ih, tw).
+        var horiz = [UInt8](repeating: 0, count: ih * tw * 4)
+        for y in 0 ..< ih {
+            for x in 0 ..< tw {
+                let l = hlo[x], r = hhi[x], w = hw[x]
+                var s0: Int32 = ROUND_BIAS, s1: Int32 = ROUND_BIAS
+                var s2: Int32 = ROUND_BIAS, s3: Int32 = ROUND_BIAS
+                for k in 0 ..< (r - l) {
+                    let p = y * iw * 4 + (l + k) * 4
+                    let kw = w[k]
+                    s0 &+= Int32(src[p + 0]) &* kw
+                    s1 &+= Int32(src[p + 1]) &* kw
+                    s2 &+= Int32(src[p + 2]) &* kw
+                    s3 &+= Int32(src[p + 3]) &* kw
+                }
+                let o = (y * tw + x) * 4
+                horiz[o + 0] = clip8(s0); horiz[o + 1] = clip8(s1)
+                horiz[o + 2] = clip8(s2); horiz[o + 3] = clip8(s3)
+            }
+        }
+
+        // Vertical pass: uint8 → uint8 at (tw, th).
+        var dst = [UInt8](repeating: 0, count: th * tw * 4)
+        for y in 0 ..< th {
+            let l = vlo[y], r = vhi[y], w = vw[y]
+            for x in 0 ..< tw {
+                var s0: Int32 = ROUND_BIAS, s1: Int32 = ROUND_BIAS
+                var s2: Int32 = ROUND_BIAS, s3: Int32 = ROUND_BIAS
+                for k in 0 ..< (r - l) {
+                    let p = ((l + k) * tw + x) * 4
+                    let kw = w[k]
+                    s0 &+= Int32(horiz[p + 0]) &* kw
+                    s1 &+= Int32(horiz[p + 1]) &* kw
+                    s2 &+= Int32(horiz[p + 2]) &* kw
+                    s3 &+= Int32(horiz[p + 3]) &* kw
+                }
+                let o = (y * tw + x) * 4
+                dst[o + 0] = clip8(s0); dst[o + 1] = clip8(s1)
+                dst[o + 2] = clip8(s2); dst[o + 3] = clip8(s3)
+            }
+        }
+
+        return CIImage(bitmapData: Data(dst), bytesPerRow: tw * 4,
+                       size: CGSize(width: tw, height: th),
+                       format: .RGBA8, colorSpace: srgb)
+    }
+
+    /// PIL-matching Lanczos that returns raw uint8 RGBA bytes (no CIImage
+    /// round-trip). Internal helper used by both `resamplePILLanczos` (which
+    /// re-wraps as CIImage) and `resamplePILLanczosToArray` (which normalizes
+    /// directly to MLXArray). Splitting this out avoids a round-trip through
+    /// CIImage that would re-encode the bytes via the working colorspace.
+    internal static func _pilLanczosCore(
+        _ image: CIImage, to size: CGSize
+    ) -> (bytes: [UInt8], tw: Int, th: Int) {
+        let tw = Int(size.width), th = Int(size.height)
+        let srgb = CGColorSpace(name: CGColorSpace.sRGB)!
+
+        let iw: Int, ih: Int
+        var src: [UInt8]
+        if let cg = image.cgImage, let provider = cg.dataProvider {
+            iw = cg.width
+            ih = cg.height
+            let retagged = CGImage(
+                width: iw, height: ih,
+                bitsPerComponent: cg.bitsPerComponent,
+                bitsPerPixel: cg.bitsPerPixel,
+                bytesPerRow: cg.bytesPerRow,
+                space: srgb, bitmapInfo: cg.bitmapInfo,
+                provider: provider, decode: nil,
+                shouldInterpolate: false, intent: .defaultIntent) ?? cg
+            src = [UInt8](repeating: 0, count: ih * iw * 4)
+            let cgctx = CGContext(
+                data: &src, width: iw, height: ih,
+                bitsPerComponent: 8, bytesPerRow: iw * 4,
+                space: srgb,
+                bitmapInfo: CGImageAlphaInfo.premultipliedLast.rawValue
+                          | CGBitmapInfo.byteOrder32Big.rawValue)!
+            cgctx.draw(retagged, in: CGRect(x: 0, y: 0, width: iw, height: ih))
+        } else {
+            iw = Int(image.extent.width)
+            ih = Int(image.extent.height)
+            src = [UInt8](repeating: 0, count: ih * iw * 4)
+            CIContext(options: [.workingColorSpace: srgb]).render(
+                image, toBitmap: &src, rowBytes: iw * 4,
+                bounds: image.extent, format: CIFormat.RGBA8, colorSpace: srgb)
+        }
+
+        // PIL Lanczos int32 fixed-point (Pillow 10.4.0 Resample.c port).
+        let PRECISION_BITS: Int32 = 22
+        let ROUND_BIAS: Int32 = 1 << (PRECISION_BITS - 1)
+        let ONE = Double(Int32(1) << PRECISION_BITS)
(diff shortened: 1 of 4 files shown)


NivDvir:
Quick note on broader applicability for visibility:

Spot-checking the other MLXVLM models, several of the patterns fixed here appear in analogous form in related VLMs. Most notably:

- `Qwen2VL.swift` passes `mask: .none` to `scaledDotProductAttention` in the vision encoder — the same shape as **bug #7** in this PR.
- `GlmOcr.swift` uses `mask: .none` in its chunked attention path — same pattern.
- All MROPE-based VLMs (`Qwen2VL`, `Qwen25VL`, `Qwen3VL`, `Qwen35`, `GlmOcr`) share the `mropeSection` / `invFreq` / `rope_deltas` plumbing that bugs **#1–#4** address.

I'll validate each of those models against the Python reference and, where the bugs manifest, file small follow-up PRs that reference this one as the pattern source — so the fixes can be reviewed per-model in digestible chunks rather than as one sweeping change. Will hold off on those until this PR has landed (or had a first maintainer pass) so the pattern is settled before generalizing.

Happy to adjust scope or timing based on maintainer preference.

NivDvir:
**Cross-architecture validation: the combined fix also makes UI-TARS-1.5-7B work.**

UI-TARS-1.5-7B (ByteDance) uses the same `Qwen2_5_VLForConditionalGeneration` architecture, same `hidden_size` (3584), same `image_token_id` (151655), same BOS/EOS — so if the MROPE fixes here are architecture-wide rather than Qwen2.5-VL-specific, UI-TARS should behave the same way on a patched vs unpatched `mlx-swift-lm`.

Ran a deterministic A/B on `mlx-community/UI-TARS-1.5-7B-4bit` — same fixed input image, same model, same prompt, only the source of `Qwen25VL.swift` differs:

**With this PR applied** (1467 lines, `mask: .array(floatMask)`, 7 MROPE fixes):
```json
[
  {"bbox_2d": [146, 566], "label": "question"},
  {"bbox_2d": [146, 608], "label": "question"},
  {"bbox_2d": [146, 640], "label": "question"},
  {"bbox_2d": [146, 672], "label": "question"},
  {"bbox_2d": [146, 714], "label": "question"},
  {"bbox_2d": [146, 746], "label": "question"},
  {"bbox_2d": [146, 778], "label": "question"},
  {"bbox_2d": [146, 810], "label": "question"},
  …
]
```
200 tokens, 9+ distinct coordinates tracking actual content positions.

**Without this PR** (upstream 1098 lines, `mask: .none`):
```json
[
  {"bbox_2d": [141, 141], "label": "question"},
  {"bbox_2d": [141, 141], "label": "editor"}
]
```
52 tokens, output collapses to two entries at identical coordinates — classic symptom of broken position encoding.

Same pattern observed on Qwen2.5-VL-7B-Instruct-4bit in earlier testing (this PR's original target). Two independent Qwen2.5-VL-family models, same result: this combined patch is required for functional VLM grounding on Apple Silicon via mlx-swift-lm.

Happy to provide the minimal reproducer (ffmpeg-captured deterministic image + script that builds `build-app.sh`-style with the PR applied vs reverted) if useful.

vade:
Might i politely suggest submitting 7 smaller prs isolated to each bug so they can be discussed in isolation?

NivDvir:
yeah, will split. MROPE (section + rope_deltas + invFreq) needs to stay bundled, parity needs all three. other four split fine.

attention mask PR in a day or two.

NivDvir:
yeah, will split. MROPE (section + rope_deltas + invFreq) needs to stay bundled, parity needs all three. other four split fine.

attention mask PR in a day or two.

NivDvir:
well it trained on PIL which ignores the ICC tag, so as I see it
color-correct = off-distribution. you think I missed something here?

vade:
Hey, sorry i deleted my comment about the color correction, my thought is generally 

a) the existing path should be able to handle disabling color correction, core image can do it via null CGColorSpaceRef if memory serves

b)  honestly it seems like introducing complexity for something the model should handle, if its been trained on sRGB images and DisplayP3 (and other color gamuts) without normalizing for an output space, then the media pipeline staying in distribution (ie outputting sRGB)  should be fine. I'll grant you if you want bit for bit pipeline accuracy in an attempt to get the predictions exact across platforms, thats likely needed (as are the rescaling). 

Have you run ablations on these changes to see which contributes most to precision differences in the inference? 

Asking out of genuine curiosity !

NivDvir:
So, i spent way too long hunting for a silver bullet to make half this code go away. But after your questions (the exact fair ones I should've asked myself first) I went what amounted to brute-force on the ablations.

I hit that point 400 lines in where the quiet voice in the back of your head starts saying "you don't actually understand this bug yet." Every shortcut I tried just confirmed the complexity was earned though. When you're feeding a real-time overlay, a few pixels of drift is basically just a bug with a better personality.

The update:

Filed splits as #238, #239, #242, and #243 (mask, MROPE, template, prep) to keep the reviews sane.

On colorSpace: nil - agreed it'd be cleaner. our custom retag actually lives in our own consumer code, not in any mlx-swift-lm path these PRs touch, so out of scope here.

Ablations: biggest single fix turned out to be a 10th bug I'd missed - Qwen2VLMessageGenerator was emitting text-then-image while HF's chat template emits image-first. Killing that alone removed a deterministic 9px shift (#242).

On bicubic vs Lanczos: this turned out to be the bit-parity case you flagged in your second comment. Tested bicubic with everything else applied - parity regresses to 28px on one image and 850px on another. mlx-swift-lm doesn't currently have a PIL-matching Lanczos (Core Image's kernel produces different bytes), and since the model was trained on PIL-resized inputs, bit-exact PIL Lanczos turned out to be the requirement. That's what the ~150-line Resample.c port in #243 is doing.

Strict ≤2 px gate: https://github.com/NivDvir/screen-overlay-toolkit/tree/main/scripts/parity

Yet, if anyone can find the silver bullet and make half this code vanish, I’d love to see it. I'm fine with my pride being deleted along with the lines.

Libraries/MLXVLM/Models/Qwen25VL.swift
```diff
@@ -44,6 +65,10 @@ private enum Language {
         let headDim: Int
         let scale: Float
         let mropeSection: [Int]
+        let mropeSectionRaw: [Int]
+        // Box hides invFreq from Module's weight-loader reflection.
+        private class InvFreqBox { let value: MLXArray; init(_ v: MLXArray) { value = v } }
+        private let _invFreqBox: InvFreqBox
```
davidkoski, line 74: This isn't required:  simply name the variable with a leading `_`:

- https://swiftpackageindex.com/ml-explore/mlx-swift/main/documentation/mlxnn/custom-layers#Parameters
- https://swiftpackageindex.com/ml-explore/mlx-swift/main/documentation/mlxnn/module/parameterisvalid(_:)

See also: 

```swift
private let _routerInputScale: MLXArray
```

Libraries/MLXVLM/MediaProcessing.swift
davidkoski, line 1: This might be better as a separate PR -- it seems independent of the ROPE changes.

Libraries/MLXVLM/MediaProcessing.swift
```diff
@@ -74,27 +74,327 @@ public enum MediaProcessing {
...
+    ///
+    /// Apple's Lanczos family (Core Image / MPS / vImage) all share the same
+    /// underlying implementation that diverges from PIL on high-frequency
+    /// content — enough to shift VLM grounding bboxes by tens of pixels. This
+    /// function is a direct port of Pillow 10.4.0's `Resample.c` 8bpc fast path.
+    ///
+    /// Also bypasses Core Graphics' automatic ICC-profile conversion: macOS
+    /// screenshots are tagged Display P3, and ImageIO would otherwise convert
+    /// P3→sRGB on load, producing different uint8 bytes than `PIL.Image.open`
+    /// sees (which ignores the ICC tag). We retag without converting.
```
davidkoski, line 102: See also:

```swift
    /// VLM media processing is normally done without regard to the colorspace. Many,
    /// though not all, images are stored in sRGB and this will be the implicit colorspace
    /// used. This converts to a colorspace with an sRGB tone curve, though not necessarily
    /// sRGB primaries, etc.
    ///
    /// See ``inLinearToneCurveSpace(_:)``
    public static func inSRGBToneCurveSpace(_ image: CIImage) -> CIImage {
```

Perhaps we should do this part more like that -- as an option rather than a requirement.

IMHO PIL is loading these incorrectly -- Display P3 images should _not_ be treated as sRGB images (which the models are likely trained on, though perhaps not!).  Converting to sRGB seems correct, even though it will produce different output than the python code.

Anyway, simply retagging the CIImage would probably do the trick _if_ callers want to defeat correct color handling.

Libraries/MLXVLM/MediaProcessing.swift
```diff
@@ -74,27 +74,327 @@ public enum MediaProcessing {
...
+            src = [UInt8](repeating: 0, count: ih * iw * 4)
+            let cgctx = CGContext(
+                data: &src, width: iw, height: ih,
+                bitsPerComponent: 8, bytesPerRow: iw * 4,
+                space: srgb,
+                bitmapInfo: CGImageAlphaInfo.premultipliedLast.rawValue
+                          | CGBitmapInfo.byteOrder32Big.rawValue)!
+            cgctx.draw(retagged, in: CGRect(x: 0, y: 0, width: iw, height: ih))
+        } else {
+            iw = Int(image.extent.width)
```
davidkoski, line 135: I would just keep this path.  CI likely has something similar to the other path internally (there are other fast paths to consider as well).  I don't think specializing the CG path is worth the extra code and confusion.

Libraries/MLXVLM/MediaProcessing.swift
```diff
@@ -74,27 +74,327 @@ public enum MediaProcessing {
...
-            width: size.width,
-            height: size.height
-        )
+        return scaledImage.cropped(to: CGRect(x: 0, y: 0, width: size.width, height: size.height))
+    }
 
-        // Crop to ensure exact dimensions
-        return scaledImage.cropped(to: exactRect)
+    /// PIL-matching Lanczos resampler, bit-identical to `PIL.Image.LANCZOS` (a=3,
+    /// separable, clamp boundary, int32 fixed-point with PRECISION_BITS=22).
```
davidkoski, line 92: This seems reasonable if you want to match the python output.  Perhaps the model was trained with PIL resized images?

Libraries/MLXVLM/MediaProcessing.swift
```diff
@@ -74,27 +74,327 @@ public enum MediaProcessing {
...
+    ///
+    /// Apple's Lanczos family (Core Image / MPS / vImage) all share the same
+    /// underlying implementation that diverges from PIL on high-frequency
```
davidkoski, line 95: I don't think this is strictly true -- CoreImage uses the GPU for the resampling vs CPU for vImage.  They all have different implementations (and afaik not _exactly_ the same).

My guess is that the CoreImage implementation is probably high quality FWIW, but I do agree it may be different than PIL.  In particular the CI version does the resampling in linear.

Libraries/MLXVLM/MediaProcessing.swift
```diff
@@ -74,27 +74,327 @@ public enum MediaProcessing {
...
+                       size: CGSize(width: tw, height: th),
+                       format: .RGBA8, colorSpace: srgb)
+    }
+
+    /// PIL-matching Lanczos that returns raw uint8 RGBA bytes (no CIImage
+    /// round-trip). Internal helper used by both `resamplePILLanczos` (which
+    /// re-wraps as CIImage) and `resamplePILLanczosToArray` (which normalizes
+    /// directly to MLXArray). Splitting this out avoids a round-trip through
+    /// CIImage that would re-encode the bytes via the working colorspace.
+    internal static func _pilLanczosCore(
```
davidkoski, line 272: This seems to duplicate most of the code above.  Should the CIImage variant call this?

Libraries/MLXVLM/Models/Qwen25VL.swift
```diff
@@ -678,40 +777,32 @@ public struct Qwen25VLProcessor: UserInputProcessor {
...
-        let size = images[0].extent.size
-        let (resizedHeight, resizedWidth) = try QwenVL.targetSize(
-            height: Int(size.height), width: Int(size.width),
-            factor: config.patchSize * config.mergeSize,
-            minPixels: config.size.minPixels, maxPixels: config.size.maxPixels)
+        // Cap longest side at 1280 and snap to multiples of (patchSize *
+        // mergeSize), matching mlx-vlm's default min_pixels / max_pixels.
+        let orig = images[0].extent.size
+        let factor = config.patchSize * config.mergeSize
+        let maxLong: CGFloat = 1280.0
```
davidkoski, line 794: Should this use Qwen25VLProcessorConfiguration.maxPixels?

Libraries/MLXVLM/Models/Qwen25VL.swift
```diff
@@ -678,40 +777,32 @@ public struct Qwen25VLProcessor: UserInputProcessor {
...
+        let ratio = min(maxLong / max(orig.width, orig.height), 1.0)
+        let resizedWidth = max(Int(orig.width * ratio) / factor * factor, factor)
+        let resizedHeight = max(Int(orig.height * ratio) / factor * factor, factor)
```
davidkoski, line 797: I wonder if this rect computation should be in MediaProcessing?

Libraries/MLXVLM/Models/Qwen25VL.swift
NivDvir: Done in #239 — switched to `private let _invFreq: MLXArray` (leading-underscore convention via `parameterIsValid`). Thanks for the pointer.

Libraries/MLXVLM/MediaProcessing.swift
NivDvir: Done — MediaProcessing changes split into #243. #222 will be closed once the splits are reviewed.

Libraries/MLXVLM/MediaProcessing.swift
NivDvir: Will follow up on #243: refactor to a public `retainSourcePixelBytes(_:)` opt-in; default path stays Apple-correct (sRGB via CIContext working colorspace). Reticle's screenshot pipeline opts in because PIL ignores ICC during training, so retag-without-conversion is what stays in-distribution.

Libraries/MLXVLM/MediaProcessing.swift
NivDvir: Agreed — will follow up on #243, collapse `_pilLanczosCore` to a single CIImage path once the retag becomes an explicit opt-in.

Libraries/MLXVLM/MediaProcessing.swift
NivDvir: Yes — HF processor calls `PIL.Image.resize`, and bit-exact PIL Lanczos is what holds the strict ≤2 px parity gate. CI Lanczos regresses by 28+ px on one canonical image.

Libraries/MLXVLM/MediaProcessing.swift
NivDvir: Comment was tightened on #243 — no longer claims a shared kernel family; just notes CI does linear-space resampling vs PIL's gamma-space. I'll re-check #243 and trim further if anything still overstates.

Libraries/MLXVLM/MediaProcessing.swift
NivDvir: Done on #243 — both `resamplePILLanczos` and `resamplePILLanczosToArray` now share `_pilLanczosCore`.

Libraries/MLXVLM/Models/Qwen25VL.swift
NivDvir: This is a long-edge cap (1280) snapped to /28. The existing `Qwen25VLProcessorConfiguration.maxPixels` field is a total-pixels cap (12,845,056) — different semantic. Want me to add a `maxLongEdge` config field, or change `maxPixels`'s meaning? Latter would be a config-API change.

Libraries/MLXVLM/Models/Qwen25VL.swift
NivDvir: Agreed — will follow up on #243, extract as `MediaProcessing.smartResizeTarget(from:longEdge:factor:)`.

Libraries/MLXVLM/Models/Qwen25VL.swift
davidkoski: No, I think this is fine if there is no current matching config.

NivDvir:
FYI: preprocessing was simplified in #243 — dropped `_pilLanczosCore`, switched to the existing CIImage Bicubic chain (+15/-5 in one file). This PR's body still references the old approach; I'll update it once #243 lands.

NivDvir:
Closing — superseded by the split PRs. The MROPE bundle landed in #239 and the Qwen2-VL chat-template fix in #242 (both merged); the remaining preprocessing change is in #243. Nothing unique left here.

PR #224, opened 2026-04-20
fix gated delta state precision -- fp32 state to match Python

TheTom:
## Proposed changes

State was typed as `q.dtype` (bf16), losing precision across T-step recurrence.
Python mlx-lm keeps state in fp32. This aligns the Swift implementation.

Three coupled fixes:
- kernel source: write `o_state` as `StT` not `InT`
- dispatch: add `StT` template param, use `state.dtype` for output
- `gatedDeltaUpdate`: create state in `.float32`, upcast if needed

Before: kernel showed ~0.25 max diff vs ops fallback at T>1.
After: matches ops path.

Tested on M5 Max 128GB, Qwen3.6-35B-A3B-4bit.

## Checklist

### DIFF
--- a/Tests/MLXLMTests/GatedDeltaTests.swift
+++ b/Tests/MLXLMTests/GatedDeltaTests.swift
@@ -1,0 +1,84 @@
+// Copyright © 2026 Apple Inc.
+
+import Foundation
+import MLX
+import MLXLLM
+import MLXLMCommon
+import XCTest
+
+public class GatedDeltaTests: XCTestCase {
+
+    /// Decode a minimal Qwen35 config from JSON -- small dims for fast testing.
+    /// Exercises the GatedDelta kernel at T>1 during prefill.
+    private func makeTestConfig() throws -> Qwen35TextConfiguration {
+        let json = """
+            {
+                "model_type": "qwen3_5",
+                "hidden_size": 64,
+                "num_hidden_layers": 2,
+                "intermediate_size": 128,
+                "num_attention_heads": 4,
+                "num_key_value_heads": 2,
+                "linear_num_value_heads": 4,
+                "linear_num_key_heads": 2,
+                "linear_key_head_dim": 32,
+                "linear_value_head_dim": 16,
+                "linear_conv_kernel_dim": 4,
+                "rms_norm_eps": 1e-6,
+                "vocab_size": 100,
+                "full_attention_interval": 4
+            }
+            """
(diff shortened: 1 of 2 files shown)


davidkoski (approved):
Looks good, thank you

Tests/MLXLMTests/GatedDeltaTests.swift
```diff
@@ -0,0 +1,84 @@
...
+                "num_attention_heads": 4,
+                "num_key_value_heads": 2,
+                "linear_num_value_heads": 4,
+                "linear_num_key_heads": 2,
+                "linear_key_head_dim": 32,
+                "linear_value_head_dim": 16,
+                "linear_conv_kernel_dim": 4,
+                "rms_norm_eps": 1e-6,
+                "vocab_size": 100,
+                "full_attention_interval": 4
```
davidkoski: This causes a crash inside the model because there are only two layers.  It should probably be set to 2.

Tests/MLXLMTests/GatedDeltaTests.swift
```diff
@@ -0,0 +1,84 @@
...
+            }
+            """
+        return try JSONDecoder().decode(
+            Qwen35TextConfiguration.self, from: json.data(using: .utf8)!)
+    }
+
+    /// Test that the GatedDelta kernel produces finite, non-zero output at T>1.
+    /// This catches the precision bug where bf16 state accumulation produced
+    /// divergent results across recurrence steps.
```
davidkoski: I found that both of these tests passed whether the GatedDelta change was applied or not.

davidkoski (changes requested):
I think this PR should either:

1) omit the tests

2) fix the config issue + make tests that fail when GatedDelta is _not_ applied

I think either is fine -- I agree that the change to GatedDelta matches the python side (though it has moved even further forward).

TheTom:
Thanks for the careful read on this. You were right that the previous tests passed either way, so I rewrote them as a direct cross-chunk precision check: same T=16 inputs through `gatedDeltaUpdate` as one shot vs as two T=8 chunks passing state across the boundary. Pre-fix the second chunk drifts by >10 max abs; post-fix it matches within bf16 input noise. Verified locally by reverting the fix to confirm the test fails, then restoring. Skipped the model-construction path entirely so the config-layer-count issue you flagged is moot.

davidkoski (approved):
Thank you!

PR #225, opened 2026-04-20
pipeline prefill chunks with asyncEval -- 10x on GDN models

TheTom:
## Proposed changes

`eval(cache)` between prefill chunks was a blocking sync -- CPU idle while
GPU worked, then built next chunk serially. Python mlx-lm avoids this because
eval is deferred until a value is read.

`asyncEval` on cache state per chunk lets the CPU build ahead while the GPU
pipelines through. One terminal `eval(cache)` after the loop.

Prefill throughput (tok/s), Qwen3.6-35B-A3B-4bit, M5 Max 128GB:

| ctx  | before | after  | speedup |
|------|--------|--------|---------|
| 128  | 260    | 696    | 2.7x    |
| 512  | 235    | 2201   | 9.4x    |
| 1k   | 244    | 3130   | 12.8x   |
| 2k   | 270    | 3937   | 14.6x   |

No regression on dense models (Gemma 4 E2B, Llama 3.2 3B).
Decode unchanged -- different code path.

## Checklist

### DIFF
--- a/Libraries/MLXLLM/LLMModel.swift
+++ b/Libraries/MLXLLM/LLMModel.swift
@@ -24,13 +24,24 @@ extension LLMModel {
         let prefillStepSize = windowSize ?? 512
         var y = input.text
 
-        // Prepare the prompt in chunks if larger than the prefill size
+        // Prepare the prompt in chunks if larger than the prefill size.
+        // asyncEval lets the CPU build chunk N+1's graph while the GPU evaluates
+        // chunk N. Python mlx-lm gets this pipelining for free because its bindings
+        // defer eval until a value is read. The previous `eval(cache)` call was a
+        // blocking sync that drained the GPU pipeline between chunks.
         while y.tokens.size > prefillStepSize {
             let input = y[.newAxis, ..<prefillStepSize]
             _ = self(input, cache: cache.isEmpty ? nil : cache, state: nil)
-            eval(cache)
+
+            var cacheArrays: [MLXArray] = []
+            for c in cache { cacheArrays.append(contentsOf: c.innerState()) }
+            asyncEval(cacheArrays)
+
             y = y[prefillStepSize...]
         }
+
+        // Single sync after the loop — flush any remaining async work.
+        eval(cache)
 
         return .tokens(y)
     }
--- a/Libraries/MLXLLM/LLMModel.swift
+++ b/Libraries/MLXLLM/LLMModel.swift
@@ -26,21 +26,15 @@ extension LLMModel {
 
         // Prepare the prompt in chunks if larger than the prefill size.
         // asyncEval lets the CPU build chunk N+1's graph while the GPU evaluates
-        // chunk N. Python mlx-lm gets this pipelining for free because its bindings
-        // defer eval until a value is read. The previous `eval(cache)` call was a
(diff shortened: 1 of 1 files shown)


TheTom:
M2 Mini 16GB numbers (Qwen3.5-2B-4bit, summarization bench):

| ctx  | before (tok/s) | after (tok/s) | speedup |
|------|-------|------|---------|
| 128  | 535   | 648  | 1.2x    |
| 1k   | 665   | 884  | 1.3x    |
| 4k   | 762   | 874  | 1.15x   |
| 32k  | 661   | 666  | neutral |

smaller GPU saturates faster so the pipeline stall is shorter to begin with -- gains scale with GPU compute time per chunk. M5 Max sees 10x because each chunk takes long enough to fully hide the next chunk's graph-building.

Libraries/MLXLLM/LLMModel.swift
```diff
@@ -24,14 +24,25 @@ extension LLMModel {
         let prefillStepSize = windowSize ?? 512
         var y = input.text
 
-        // Prepare the prompt in chunks if larger than the prefill size
+        // Prepare the prompt in chunks if larger than the prefill size.
+        // asyncEval lets the CPU build chunk N+1's graph while the GPU evaluates
+        // chunk N. Python mlx-lm gets this pipelining for free because its bindings
+        // defer eval until a value is read. The previous `eval(cache)` call was a
+        // blocking sync that drained the GPU pipeline between chunks.
```
davidkoski: I am not sure the part about mlx-lm gets this right -- the semantics of eval and when things are triggered is exactly the same between mlx and mlx-swift.  I think trim this to:


```suggestion
        // Prepare the prompt in chunks if larger than the prefill size.
        // asyncEval lets the CPU build chunk N+1's graph while the GPU evaluates
        // chunk N.
```

Libraries/MLXLLM/LLMModel.swift
```diff
@@ -24,14 +24,25 @@ extension LLMModel {
...
         while y.tokens.size > prefillStepSize {
             let input = y[.newAxis, ..<prefillStepSize]
             _ = self(input, cache: cache.isEmpty ? nil : cache, state: nil)
-            eval(cache)
+
+            var cacheArrays: [MLXArray] = []
+            for c in cache { cacheArrays.append(contentsOf: c.innerState()) }
+            asyncEval(cacheArrays)
+
```
davidkoski: This should be doable as `asyncEval(cache)`:

```swift
public func asyncEval(_ values: Any...) {
    var arrays = [MLXArray]()

    for item in values {
        collect(item, into: &arrays)

private func collect(_ item: Any, into arrays: inout [MLXArray]) {
    case let v as Evaluatable:
        arrays.append(contentsOf: v.innerState())
```

```suggestion
            asyncEval(cache)
```

davidkoski (changes requested):

TheTom:
Thanks for the careful read, both calls are spot on. Trimmed the comment as suggested and swapped to `asyncEval(cache)` directly since `KVCache: Evaluatable` makes the manual `innerState()` loop redundant. Build is green. Appreciate the review.

davidkoski (approved):
Thank you!

PR #226, opened 2026-04-21
fix: disable CIContext intermediate caching to prevent IOSurface exha…

Satchitananda:
…ustion

CIContext() default options cache IOSurface-backed GPU textures for intermediate filter results. Each frame in the VLM image pipeline (tone-curve → bicubic resample → color-matrix normalize) allocates multiple IOSurfaces that are held in the context cache across calls.

Processing a large library (hundreds of videos, each yielding several frames) accumulates thousands of cached surfaces and hits the macOS per-process IOSurface kernel limit of 16384, causing render failures:

  IOSurface creation failed: e00002be (likely per client IOSurface limit of
  16384 reached)
  -[CIContext _startTaskToRender:...] Render failed because of failure to
  allocate intermediate.

Fix: pass `.cacheIntermediates: false` when constructing the shared context. Batch inference never re-renders the same CIImage twice, so the intermediate cache provides no benefit. The change eliminates the surface accumulation while leaving all functional behavior intact.

## Proposed changes

## Checklist

### DIFF
--- a/Libraries/MLXVLM/MediaProcessing.swift
+++ b/Libraries/MLXVLM/MediaProcessing.swift
@@ -13,7 +13,12 @@ public struct ProcessedFrames {
     public let totalDuration: CMTime
 }
 
-private let context = CIContext()
+// `.cacheIntermediates: false` prevents CoreImage from holding IOSurface-backed
+// GPU textures between frames. With the default (caching) context, a large-library
+// scan accumulates thousands of cached intermediate surfaces and hits the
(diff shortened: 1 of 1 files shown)


davidkoski (approved):
Good idea!

davidkoski:
This is still marked as draft -- should it be ready for review?

Satchitananda:
@davidkoski sorry, just forgot about this PR. Thank you for merging!

PR #227, opened 2026-04-21
add FusedGateUpSwitchGLU -- single fused gate_up_proj for MoE models

TheTom:
## Proposed changes

SwitchGLU variant for models that ship a single fused `gate_up_proj` weight
of shape `[numExperts, 2*hiddenDims, inputDims]` instead of separate
`gate_proj` / `up_proj`.

One `gatherMM` dispatch for the combined projection, split, then activate.
Same `gatherSort` optimization as `SwitchGLU`.

## Checklist

### DIFF
--- a/Libraries/MLXLMCommon/SwitchLayers.swift
+++ b/Libraries/MLXLMCommon/SwitchLayers.swift
@@ -86,6 +86,69 @@ public class SwitchGLU: Module {
     }
 }
 
+// MARK: - FusedGateUpSwitchGLU
+
+/// SwitchGLU variant for models that ship a single fused `gate_up_proj` weight
+/// of shape `[numExperts, 2*hiddenDims, inputDims]` instead of separate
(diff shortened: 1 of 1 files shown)


davidkoski (approved):

davidkoski:
The change looks good but there are no uses of this -- are there follow-on PRs that will adopt?

TheTom:
> The change looks good but there are no uses of this -- are there follow-on PRs that will adopt?

Hey David, this PR preps for future upstream PRs adopting it on Qwen35MoE, Gemma 4 MoE, and Qwen3Next, which currently split fused HF `gate_up_proj` at load time. `FusedGateUpSwitchGLU` eliminates four costs: 2x peak load memory, an extra slice at load, two `gatherMM` dispatches per MoE forward, and two expert-weight reads per token. Happy to stack the Qwen35MoE adoption alongside #227 if you'd prefer to see the consumer before merging.

davidkoski:
No, this is fine -- easy to review, I just wanted to make sure it would be used :-)

PR #228, opened 2026-04-21
fix Gemma 4 MoE router -- softmax order + fuse norm dispatches

TheTom:
## Proposed changes

Two improvements to `Gemma4TextRouter`:

1. **Softmax order**: softmax was applied to all 128 expert scores before top-k
   selection, so selected weights came from the full distribution rather than just
   the top-k. Moved softmax to after selection so it operates only on the chosen
   experts. Also drops the renormalization step (divide by sum) since softmax on
   top-k already produces a valid distribution.

2. **Fused norm**: norm + scale used 3 separate dispatches (`norm(x)`,
   `x * rootSize`, `x * scale`). Fused into one `MLXFast.rmsNorm` call
   with `weight = scale * rootSize`.

### Results

gemma-4-26b-a4b-it-4bit on M5 Max 128GB:

**Decode speed:** 32.7 → 35.9 tok/s (+9.8% from fused norm)

**Raw completion** (no chat template, greedy):

| | output |
|---|---|
| before | `\n\nsدle- l_e_e_e_e_e` |
| after | `\n\nin the ownle of kind-approachter-approachter` |

Garbled characters → recognizable English. Both degrade to repetition without
a chat template, but routing fix produces coherent tokens.

**Long generation** (500 tokens, chat template, cryptography essay):

Both produce coherent essays. The fixed router covers more historical ground
in the same token budget (includes Scytale, Al-Kindi, frequency analysis
attribution that the original version omits).

## Checklist

### DIFF
--- a/Libraries/MLXVLM/Models/Gemma4.swift
+++ b/Libraries/MLXVLM/Models/Gemma4.swift
@@ -499,9 +499,9 @@ private final class Gemma4TextMLP: Module, UnaryLayer {
 
 private final class Gemma4TextRouter: Module {
     let topKExperts: Int
+    let config: Gemma4TextConfiguration
     private let rootSize: Float
 
-    @ModuleInfo(key: "norm") var norm: Gemma4RMSNormNoScale
     @ModuleInfo(key: "proj") var proj: Linear
     @ParameterInfo(key: "scale") var scale: MLXArray
     @ParameterInfo(key: "per_expert_scale") var perExpertScale: MLXArray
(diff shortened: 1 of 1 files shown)


TheTom:
before/after on gemma-4-26b-a4b-it-4bit, greedy decode (no chat template, no temperature):

prompt: "Explain the difference between TCP and UDP in one sentence:"

**before** (softmax all 128 experts → top-k):
```
\n\nsدle- l_e_e_e_e_e
```

**after** (top-k → softmax selected):
```
\n\nin the ownle of kind-approachter-approachter
```

both degrade to repetition without a chat template, but the routing fix produces recognizable English instead of garbled characters.

davidkoski (approved):
changes look good, thank you!

Issue #233, opened 2026-04-22
[BUG] Gemma 3n is broken after 3.31.3 upgrade


nusov:
**Describe the bug**
Hello, it seems that Gemma 3n 4bit models have issues with tokenizer

**To Reproduce**
- Load LLMRegistry.gemma3n_E2B_it_lm_4bit
- Try to get a response

Currently the model returns random tokens, backticks or asterisks

```
 are are are

**

**

**
``` 

Include code snippet
```swift
    func test() async {
        let modelConfiguration = LLMRegistry.gemma3n_E2B_it_lm_4bit
        
        // customize this line per the integration package
        let model = try! await #huggingFaceLoadModelContainer(
            configuration: modelConfiguration
        )
        
        let session = ChatSession(model)
        print(try! await session.respond(to: "What are two things to see in San Francisco?"))
    }
```

**Expected behavior**

The MLX runtime produces correct response.

**Desktop (please complete the following information):**
 - OS Version: 26.3.1 (a)
 - Device: iPhone 17
 - Version: 3.31.3

**Additional context**

Both E2B and E4B variants are not functional. 

I tested other models including LFM2.5 and Gemma4 and they work without issues.

DePasqualeOrg:
This is due to the RoPE changes in #178 (cc @ronaldmannak).

It appears to affect only Gemma 3n, but rather than patching just that model, I suggest addressing the root cause, which is the hidden temporal coupling in the new [`applyRotaryPosition`](https://github.com/ml-explore/mlx-swift-lm/blob/7e2b7107be52ffbfe488f3c7987d3f52c1858b4b/Libraries/MLXLMCommon/RoPEApplication.swift#L35-L43) helper. Instead of reading `cache.offset` inside the helper, the offset should be passed as an explicit argument (see [usage](https://github.com/ml-explore/mlx-lm/blob/ed1fca4cef15a824c5f1702c80f70b4cffc8e4dd/mlx_lm/models/gemma3n.py#L152) in mlx-lm).

This is the fix I applied in https://github.com/ml-explore/mlx-swift-lm/pull/234.

I also added an integration test suite in https://github.com/ml-explore/mlx-swift-lm/pull/235 that can be run to catch issues like this in the future.

davidkoski:
See #280

PR #234, opened 2026-04-22
Remove temporal coupling in `applyRotaryPosition`

DePasqualeOrg:
`applyRotaryPosition(_:to:cache:)`, which was introduced in [#178](https://github.com/ml-explore/mlx-swift-lm/pull/178) (cc [@ronaldmannak](https://github.com/ronaldmannak)), reads `cache.offset` implicitly. Because `cache.update(...)` advances the offset, the helper's result depends on call ordering relative to cache updates, and that dependency is invisible at the call site.

44 of the 45 attention layers using this helper call it for queries and keys back to back before any `cache.update`, so both reads return the same value and the coupling stays hidden. Gemma 3n is the exception: its KV-sharing branch calls `cache.update` between the key and query RoPE calls, so queries get rotated at `cache.offset + L` while keys use `cache.offset`. That mismatch produces garbled output (see [#233](https://github.com/ml-explore/mlx-swift-lm/issues/233)).

Gemma 4 had the same interleaving and worked around it with file-local helpers that corresponded to this PR's API, which is now uniformly applied.

MLX LM in Python avoids this by making `offset` an explicit argument to `nn.RoPE`. Callers that interleave `cache.update_and_fetch` snapshot `offset = cache.offset` first, making the timing visible.

## Solution

Make the offset explicit in Swift too. Preserve the batch polymorphism the helper was introduced for, but move the cache read out of the helper and into a one-line snapshot at the top of each `callAsFunction`.

### New API (additive)

In `Libraries/MLXLMCommon/RoPEApplication.swift`, add a `RoPEOffset` enum (`.scalar(Int)` / `.array(MLXArray)`), a `KVCache.ropeOffset` snapshot property that defensively copies the batched variant, and a new `applyRotaryPosition(_:to:offset:)` overload.

### Deprecated shim

Mark the existing `applyRotaryPosition(_:to:cache:)` `@available(*, deprecated)` and forward it to the new overload via `cache?.ropeOffset ?? .scalar(0)`, so downstream callers get a warning, not a compile error, during migration.

### Call site example

Before:
```swift
queries = applyRotaryPosition(rope, to: queries, cache: cache)
keys    = applyRotaryPosition(rope, to: keys,    cache: cache)
```

After:
```swift
let ropeOffset = cache?.ropeOffset ?? .scalar(0)
...
queries = applyRotaryPosition(rope, to: queries, offset: ropeOffset)
keys    = applyRotaryPosition(rope, to: keys,    offset: ropeOffset)
```

The snapshot goes at the top of `callAsFunction`, or at the narrowest scope containing the RoPE calls when branches or conditionals require it (Gemma 3n, GPTOSS, and AfMoE).

### DIFF
--- a/Libraries/MLXLMCommon/RoPEApplication.swift
+++ b/Libraries/MLXLMCommon/RoPEApplication.swift
@@ -15,29 +15,79 @@ public protocol BatchPositionedKVCache: KVCache {
     var batchOffset: MLXArray { get }
 }
 
+// MARK: - RoPEOffset
+
+/// Positional offset for RoPE. Polymorphic so single-sequence and batched
+/// inference share the same call site.
+///
+/// Snapshot once per forward pass via ``KVCache/ropeOffset`` before any
+/// `cache.update` call, then reuse the value for every RoPE application in
+/// the layer. Reading `cache.offset` at each RoPE call is unsafe because
+/// `cache.update` advances the offset mid-pass.
+public enum RoPEOffset {
+    /// Single-sequence offset.
+    case scalar(Int)
+    /// Per-sequence offsets with shape `[B]`.
+    case array(MLXArray)
+}
+
+extension KVCache {
+    /// Snapshot of the current RoPE offset suitable for passing to
+    /// ``applyRotaryPosition(_:to:offset:)``.
+    ///
+    /// Read once at the top of `callAsFunction` and reuse for every RoPE
+    /// call in that layer. Do not re-read across a `cache.update` — the
+    /// offset advances and a later read will not match an earlier one.
+    ///
+    /// The batched form defensively copies `batchOffset` (via `+ 0`) so the
+    /// returned value is decoupled from any in-place mutation that
+    /// `cache.update` may perform on the cache's stored offset array.
+    public var ropeOffset: RoPEOffset {
+        if let batched = self as? BatchPositionedKVCache {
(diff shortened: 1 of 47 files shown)


ronaldmannak:
Thanks @DePasqualeOrg 
This was a real bug and this PR simplifies Gemma4Text as well, which is great.

davidkoski (approved):
Very interesting -- thank you for the fix!

Libraries/MLXLMCommon/RoPEApplication.swift
```diff
@@ -15,29 +15,79 @@ public protocol BatchPositionedKVCache: KVCache {
...
+    ///
+    /// The batched form defensively copies `batchOffset` (via `+ 0`) so the
+    /// returned value is decoupled from any in-place mutation that
+    /// `cache.update` may perform on the cache's stored offset array.
+    public var ropeOffset: RoPEOffset {
+        if let batched = self as? BatchPositionedKVCache {
+            return .array(batched.batchOffset + 0)
+        } else {
+            return .scalar(offset)
+        }
```
davidkoski, line 50: I don't know if it matters, but see also:

- https://github.com/ml-explore/mlx-swift/pull/367
- https://github.com/ml-explore/mlx-swift/issues/364

I don't know if `batchOffset[.ellipsis]` is more idiomatic or cheaper.  The comment explains what it is doing well, enough.

PR #235, opened 2026-04-23
Add coherence integration tests

DePasqualeOrg:
I've added an integration test suite that checks for coherent responses from a representative sample of models. The suite can be run whenever wide-ranging changes such as #178 are made to catch regressions such as the one noted in #233.

The test suite uses smaller models (≤9B) so that they can run on machines with 16 GB of RAM.

The new suite already caught a latent bug: `LLMTypeRegistry` had Jamba registered under the key `"jamba_3b"`, but actual Jamba configs specify `model_type: "jamba"`. I fixed that here as well.

It also surfaced a second issue: loading OLMoE (or any model using `GPTNeoXTokenizer`) fails because Swift Transformers doesn't register that class. My [Swift Tokenizers](https://github.com/DePasqualeOrg/swift-tokenizers) package already works correctly, because the class is registered there.

There are also two failing tool-calling integration tests that are outside the scope of this PR.

## Changes

- Added new `CoherenceIntegrationTests` suite
- Refactored `IntegrationTestModels` around generic `llmContainer(for:)` / `vlmContainer(for:)` accessors and reduced boilerplate code
- Added `LLMRegistry.qwen3_5_2b_4bit` (used by the new suite) and `qwen3_6_27b_4bit`
- Renamed `LLMRegistry.jamba_3b` to `jamba_3b_4bit` and pointed it at the 4-bit variant (1.7 GB vs 6.1 GB bf16)
- Fixed `LLMTypeRegistry` Jamba creator key: `"jamba_3b"` → `"jamba"`

### DIFF
--- a/Libraries/MLXLLM/LLMModelFactory.swift
+++ b/Libraries/MLXLLM/LLMModelFactory.swift
@@ -73,7 +73,7 @@ public enum LLMTypeRegistry {
         "nanochat": create(NanoChatConfiguration.self, NanoChatModel.init),
         "nemotron_h": create(NemotronHConfiguration.self, NemotronHModel.init),
         "afmoe": create(AfMoEConfiguration.self, AfMoEModel.init),
-        "jamba_3b": create(JambaConfiguration.self, JambaModel.init),
+        "jamba": create(JambaConfiguration.self, JambaModel.init),
         "mistral3": create(Mistral3TextConfiguration.self, Mistral3TextModel.init),
         "apertus": create(ApertusConfiguration.self, ApertusModel.init),
     ])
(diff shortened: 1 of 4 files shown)


Libraries/MLXLLM/LLMModelFactory.swift
```diff
@@ -73,7 +73,7 @@ public enum LLMTypeRegistry {
         "nanochat": create(NanoChatConfiguration.self, NanoChatModel.init),
         "nemotron_h": create(NemotronHConfiguration.self, NemotronHModel.init),
         "afmoe": create(AfMoEConfiguration.self, AfMoEModel.init),
-        "jamba_3b": create(JambaConfiguration.self, JambaModel.init),
+        "jamba": create(JambaConfiguration.self, JambaModel.init),
```
davidkoski, line 76: I wonder if we need both?  Perhaps not:

- https://huggingface.co/mlx-community/AI21-Jamba-Reasoning-3B-bf16/blob/main/config.json

That is from #8 where it was added and it uses `jamba` as you have.

davidkoski (approved):
Changes look good and very useful.  Thank you!

PR #237, opened 2026-04-24
Fix Gemma4Text quantized KV cache attention

aleroot:
## Proposed changes

Fix run of gemma-4-e4b-it-4bit which was causing: MLXLMCommon/KVCache.swift:894: Fatal error: update was called on QuantizedKVCache. Use updateQuantized instead.

## Checklist

### DIFF
--- a/Libraries/MLXLLM/Models/Gemma4Text.swift
+++ b/Libraries/MLXLLM/Models/Gemma4Text.swift
@@ -175,6 +175,53 @@ private class ScaledLinear: Module {
     }
 }
 
+private enum Gemma4PositionOffset {
+    case scalar(Int)
+    case batch(MLXArray)
+}
+
+private enum Gemma4TextKVState {
+    case regular(keys: MLXArray, values: MLXArray)
+    case quantized(
+        keys: (MLXArray, MLXArray, MLXArray?),
+        values: (MLXArray, MLXArray, MLXArray?),
+        groupSize: Int,
+        bits: Int,
+        mode: QuantizationMode
+    )
+
+    var sequenceLength: Int {
+        switch self {
+        case .regular(let keys, _):
+            keys.dim(2)
+        case .quantized(let keys, _, _, _, _):
+            keys.0.dim(-2)
+        }
+    }
+}
+
+private func gemma4CapturePositionOffset(from cache: KVCache?) -> Gemma4PositionOffset {
+    if let batchCache = cache as? BatchPositionedKVCache {
+        // Snapshot the per-sequence offsets before cache.update(...) advances them.
+        .batch(batchCache.batchOffset + 0)
+    } else {
+        .scalar(cache?.offset ?? 0)
+    }
+}
+
+private func gemma4ApplyRotaryPosition<R: RoPELayer>(
+    _ rope: R,
+    to x: MLXArray,
+    offset: Gemma4PositionOffset
+) -> MLXArray {
+    switch offset {
+    case .scalar(let value):
+        rope(x, offset: value)
+    case .batch(let values):
+        rope(x, offset: values)
+    }
+}
+
 // MARK: - Attention
 
 private class Gemma4Attention: Module {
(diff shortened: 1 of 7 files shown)


aleroot:
@davidkoski I have rebased this pull request on top of main to resolve the conflicts, so if you want you can try to run the actions again. Thanks.

Libraries/MLXLLM/Models/Gemma4Text.swift
```diff
@@ -175,6 +175,53 @@ private class ScaledLinear: Module {
     }
 }
 
+private enum Gemma4PositionOffset {
+    case scalar(Int)
+    case batch(MLXArray)
+}
```
davidkoski: This can use:

```swift
public enum RoPEOffset {
    case scalar(Int)
    case batch(MLXArray)
}
```

This was added after you wrote this but it was picked up with the rebase.

Libraries/MLXLLM/Models/Gemma4Text.swift
```diff
@@ -175,6 +175,53 @@ private class ScaledLinear: Module {
...
+    to x: MLXArray,
+    offset: Gemma4PositionOffset
+) -> MLXArray {
+    switch offset {
+    case .scalar(let value):
+        rope(x, offset: value)
+    case .batch(let values):
+        rope(x, offset: values)
+    }
+}
```
davidkoski: This can be replaced with:

```swift
public func applyRotaryPosition<R: RoPELayer>(_ rope: R, to x: MLXArray, offset: RoPEOffset?)
    -> MLXArray
```

and `KVCache.ropeOffset`.

Libraries/MLXLMCommon/KVCache.swift
```diff
@@ -422,24 +422,40 @@ public class KVCacheSimple: BaseKVCache, CustomDebugStringConvertible {
...
+                let effectiveGroupSize = resolvedKVQuantizationGroupSize(
+                    requested: groupSize,
+                    keyHeadDim: currentKeys.dim(3),
+                    valueHeadDim: currentValues.dim(3)
+                )
+            else {
+                fatalError(
+                    "KV cache quantization requires head dimensions divisible by one of the supported group sizes (32, 64, 128). Requested group size: \(groupSize). Key head dim: \(currentKeys.dim(3)). Value head dim: \(currentValues.dim(3))."
+                )
+            }
```
davidkoski, line 439: Something to consider:  the group size and bits are computed here but if saved and restored, this will not restore the groupSize or bits:

```swift
    public override var metaState: [String] {
        get { [String(step), String(offset), String(groupSize), String(bits)] }
        set {
            guard newValue.count == 4 else {
                fatalError("QuantizedKVCache metaState must have exactly 4 values")
            }

            self.offset = Int(newValue[1]) ?? 0
        }
    }
```

That was already the case but this makes it likely to hit a bug since these values will change from the defaults.

Libraries/MLXLLM/Models/Gemma4Text.swift
```diff
@@ -175,6 +175,53 @@ private class ScaledLinear: Module {
     }
 }
 
+private enum Gemma4PositionOffset {
+    case scalar(Int)
+    case batch(MLXArray)
+}
+
+private enum Gemma4TextKVState {
```
davidkoski: This is duplicated from `Gemma4SharedKVState` in Gemma4.swift.  Can they be shared?

aleroot: I have tried moving the shared state into MLXLMCommon/Models/Gemma4.swift

Libraries/MLXLMCommon/KVCache.swift
aleroot: metaState now restores offset, groupSize, and bits. I also added KVCacheTests coverage for restore, copy, and empty-cache quantization metadata.

strikingleon:
Looks like we ran into this one as well with Gemma 4 E2B Instruct OptiQ.

davidkoski:
I wonder if there is some overlap with #247 here?  When you get a chance can you please rebase?  Thanks!

aleroot:
> I wonder if there is some overlap with #247 here? When you get a chance can you please rebase? Thanks!

Just rebased on top of the updated upstream master

Tests/MLXLMTests/KVCacheTests.swift
```diff
@@ -65,6 +65,67 @@ func testCacheSerialization(creator: (() -> any KVCache)) async throws {
     }
 }
 
+@Test func testQuantizedKVCacheRestoresNonDefaultQuantizationMetadata() throws {
```
davidkoski, line 68: I am seeing crashes locally per https://github.com/ml-explore/mlx-swift/pull/410.

I cut a tag for it (0.31.4) and updated Package.swift to pick it up

davidkoski: Doh, too many PRs.  That was for 237.

davidkoski (approved):
Looks good, thank you.

PR #238, opened 2026-04-24
Fix vision-encoder attention mask being ignored in Qwen2.5-VL

NivDvir:
Vision.Attention.callAsFunction receives attentionMask but the SDPA call was hardcoded to mask: .none, so the mask is silently discarded.

The mask is [1, seqLen, seqLen] boolean; SDPA's .array fast path wants an additive Float16 mask, so the change folds in the conversion.

+9/-1, one function. Split out from #222 at @vade's suggestion.

### DIFF
--- a/Libraries/MLXVLM/Models/Qwen25VL.swift
+++ b/Libraries/MLXVLM/Models/Qwen25VL.swift
@@ -316,12 +316,20 @@ private enum Vision {
             k = k.reshaped(1, sequenceLength, numHeads, -1).transposed(0, 2, 1, 3)
             v = v.reshaped(1, sequenceLength, numHeads, -1).transposed(0, 2, 1, 3)
 
+            // attentionMask is [1, seqLen, seqLen] boolean -- convert to an additive
+            // Float16 mask for SDPA's .array fast path on the vision fp16 code path.
+            let boolMask = attentionMask[.newAxis, 0..., 0..., 0...]
+            let floatMask = MLX.where(
+                boolMask,
+                MLXArray(Float16(0)),
+                MLXArray(Float16(-10000)))
+
             let output = MLXFast.scaledDotProductAttention(
                 queries: q,
                 keys: k,
                 values: v,
                 scale: scale,
-                mask: .none
+                mask: .array(floatMask)
             )
             .transposed(0, 2, 1, 3)
             .reshaped(sequenceLength, -1)
--- a/Libraries/MLXVLM/Models/Qwen25VL.swift
+++ b/Libraries/MLXVLM/Models/Qwen25VL.swift
@@ -316,20 +316,17 @@ private enum Vision {
             k = k.reshaped(1, sequenceLength, numHeads, -1).transposed(0, 2, 1, 3)
             v = v.reshaped(1, sequenceLength, numHeads, -1).transposed(0, 2, 1, 3)
 
-            // attentionMask is [1, seqLen, seqLen] boolean -- convert to an additive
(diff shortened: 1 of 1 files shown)


davidkoski:

Libraries/MLXVLM/Models/Qwen25VL.swift
```diff
@@ -316,12 +316,20 @@ private enum Vision {
             k = k.reshaped(1, sequenceLength, numHeads, -1).transposed(0, 2, 1, 3)
             v = v.reshaped(1, sequenceLength, numHeads, -1).transposed(0, 2, 1, 3)
 
+            // attentionMask is [1, seqLen, seqLen] boolean -- convert to an additive
+            // Float16 mask for SDPA's .array fast path on the vision fp16 code path.
+            let boolMask = attentionMask[.newAxis, 0..., 0..., 0...]
+            let floatMask = MLX.where(
+                boolMask,
+                MLXArray(Float16(0)),
+                MLXArray(Float16(-10000)))
```
davidkoski: Although the weight might be float16 right now, this should probably use the dtype, see also 3a7503d


```suggestion
                MLXArray(0, dtype: q.dtype),
                MLXArray(-10000, dtype: q.dtype))
```

Libraries/MLXVLM/Models/Qwen25VL.swift
```diff
@@ -316,12 +316,20 @@ private enum Vision {
             k = k.reshaped(1, sequenceLength, numHeads, -1).transposed(0, 2, 1, 3)
             v = v.reshaped(1, sequenceLength, numHeads, -1).transposed(0, 2, 1, 3)
 
+            // attentionMask is [1, seqLen, seqLen] boolean -- convert to an additive
+            // Float16 mask for SDPA's .array fast path on the vision fp16 code path.
+            let boolMask = attentionMask[.newAxis, 0..., 0..., 0...]
+            let floatMask = MLX.where(
```
davidkoski: This work on the attention mask is done per layer -- it would be better to hoist it into the VisionModel so that it can be used by all of the many `blocks: [Qwen25VLVisionBlock]`

davidkoski (changes requested):

Libraries/MLXVLM/Models/Qwen25VL.swift
```diff
@@ -583,9 +588,29 @@ private enum Vision {
...
+            // Build boolean masks once, then convert to additive float masks
+            // matching the q-tensor dtype (carried by hiddenStates as it
+            // flows through the encoder). Done ONCE here instead of inside
+            // each of the 32 encoder blocks — per davidkoski #238 review:
+            //   • dtype: parity with the precedent in commit 3a7503d
+            //     (Bert/NomicBert attention mask dtype fix)
+            //   • hoist: avoid 32× redundant graph construction per layer
+            // The mask is constant across all blocks for a given input,
+            // so building it inside the per-block attention was pure
+            // duplication.
```
davidkoski: A comment is fine, but this maybe more commentary than is needed -- people can always go back to the PR the lines came from to see the discussion.


```suggestion
            // Build boolean masks once, then convert to additive float masks
            // matching the q-tensor dtype (carried by hiddenStates as it
            // flows through the encoder). Done ONCE here instead of inside
            // each of the 32 encoder blocks 
```

davidkoski:
Looks great, but I think the comment needs to be trimmed down a bit, then it is ready to merge!

davidkoski (approved):
Great, thank you!

PR #239, opened 2026-04-24
Fix MROPE, rope_deltas, and invFreq loading in Qwen2.5-VL

NivDvir:
Three coupled fixes for Qwen2.5-VL to match Python mlx-vlm: MROPE section slices are slice-replaced not modulo-interleaved; rope_deltas is reconstructed during autoregressive decode; invFreq is wrapped to hide it from MLX's weight loader.

Split out from #222. Coupled because parity requires all three together.

### DIFF
--- a/Libraries/MLXVLM/Models/Qwen25VL.swift
+++ b/Libraries/MLXVLM/Models/Qwen25VL.swift
@@ -43,7 +43,11 @@ private enum Language {
         let kvHeads: Int
         let headDim: Int
         let scale: Float
-        let mropeSection: [Int]
+        let mropeSection: [Int]      // cumulative section indices (for half-dim split)
+        let mropeSectionRaw: [Int]   // raw section sizes [16, 24, 24] (for full-dim split)
+        // Wrapped in a class to hide from Module's weight loading (not a trained weight)
+        private class InvFreqBox { let value: MLXArray; init(_ v: MLXArray) { value = v } }
+        private let _invFreqBox: InvFreqBox
 
         @ModuleInfo(key: "q_proj") var wq: Linear
         @ModuleInfo(key: "k_proj") var wk: Linear
@@ -65,6 +69,8 @@ private enum Language {
             self._wo.wrappedValue = Linear(heads * headDim, dim, bias: false)
 
             if let v = args.ropeScaling?["mrope_section"], let array = v.asInts() {
+                // Raw sections e.g. [16, 24, 24] — used for splitting full-dim cos/sin
+                self.mropeSectionRaw = array
                 // mrope_section = np.cumsum(mrope_section * 2)[:-1].tolist()
                 self.mropeSection = sequence(state: (0, array.makeIterator())) { state in
                     if let v = state.1.next() {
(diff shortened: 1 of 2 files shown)


davidkoski:
I am unclear what the intent is here -- #222 did seem like a couple of things put together (rope + image handling).  Should #222 be closed and we continue with this and other PRs that you opened?

NivDvir:
Yes — please close #222, or I'll close it after these are reviewed.

Splits:
- #238 — vision attention mask (was bug 7 in #222)
- #239 (this) — MROPE section + rope_deltas + invFreq (kept bundled, parity needs all three together)
- #242 — chat-template content order (the 10th bug, ~9 px deterministic shift)
- #243 — preprocessing (PIL Lanczos + smart-resize + ICC handling — the part you flagged as 'better as a separate PR')

Each is independently buildable; review order is up to you.

davidkoski:
OK, I will port my comments from #222 to the various new PRs and close it when done.  Thanks!

Oh, it looks like you already saw those.

NivDvir:
Thanks — watching the splits for the ported comments.

davidkoski:

Libraries/MLXVLM/Models/Qwen25VL.swift
```diff
@@ -209,6 +272,10 @@ private enum Language {
 
         var kvHeads: [Int]
 
+        // MROPE state: stored position IDs from prefill, reused during autoregressive generation
+        var _positionIds: MLXArray?
+        var _ropeDeltas: MLXArray?
```
davidkoski: The model can't (well, it _shouldn't_) mutate state during inference -- the one model instance can be used by many chat sessions.

See #283 for the pattern to cache state like this.

Libraries/MLXVLM/Models/Qwen25VL.swift
```diff
@@ -316,12 +415,20 @@ private enum Vision {
             k = k.reshaped(1, sequenceLength, numHeads, -1).transposed(0, 2, 1, 3)
             v = v.reshaped(1, sequenceLength, numHeads, -1).transposed(0, 2, 1, 3)
 
+            // Apply attention mask (window or full attention)
+            // attentionMask is [1, seqLen, seqLen] boolean — convert to float16 mask for SDPA
+            let boolMask = attentionMask[.newAxis, 0..., 0..., 0...]  // [1, 1, seqLen, seqLen]
+            let floatMask = MLX.where(
+                boolMask,
+                MLXArray(Float16(0)),
+                MLXArray(Float16(-10000)))
```
davidkoski: See my comment in #238 regarding dtypes.

davidkoski: And in the same PR the comment about hoisting this computation up into where the attentionMask is created. 

Actually I think 238 and this PR may overlap on these lines, right?  Maybe fix it in 238, we get that merged and then rebase this one?

Libraries/MLXVLM/Models/Qwen25VL.swift
```diff
@@ -678,40 +785,36 @@ public struct Qwen25VLProcessor: UserInputProcessor {
...
-        let resizedSize = CGSize(width: resizedWidth, height: resizedHeight)
-
-        // Process images
+        // Compute target size from the original image before any processing, matching Python's
+        // single-step resize: ratio = max_size / max(w,h); new_w = int(w*ratio)//28*28
+        let origSize = images[0].extent.size
+        let factor = config.patchSize * config.mergeSize
+        let targetSize: CGSize
+        if let resize = processing?.resize {
+            targetSize = resize
```
davidkoski: Previously all paths went through `QwenVL.targetSize` -- I think this may need to do so as well.  There are constraints on the size and that function makes sure they are enforced (otherwise I think we get reshape errors later on).

Libraries/MLXVLM/Models/Qwen25VL.swift
```diff
@@ -678,40 +785,36 @@ public struct Qwen25VLProcessor: UserInputProcessor {
     func preprocess(image: CIImage, resizedSize: CGSize) -> CIImage {
```
davidkoski: This is now used for the video case, perhaps rename to indicate that?  Maybe it was already this way?  Anyway, I think naming would help understanding.

Libraries/MLXVLM/Models/Qwen25VL.swift
```diff
@@ -678,40 +785,36 @@ public struct Qwen25VLProcessor: UserInputProcessor {
...
         image
             .toSRGB()
-            .resampled(to: resizedSize, method: .bicubic)
+            .resampled(to: resizedSize, method: .lanczos)
```
davidkoski: Do we really want videos to use lanczos resampling?  Why not bicubic like it used to be (and we use in the image page)?

Libraries/MLXVLM/Models/Qwen25VL.swift
```diff
@@ -79,12 +86,54 @@ private enum Language {
...
+            let base = MLXArray(args.ropeTheta)
+            self._invFreq = 1.0 / pow(base, freqIndices / Float(headDim))
+
             self._rotaryEmbedding.wrappedValue = RoPE(
                 dimensions: headDim, traditional: args.ropeTraditional, base: args.ropeTheta)
         }
 
+        /// Compute cos/sin for MROPE from 3D position IDs
+        /// Matches Python apply_mrope: start with temporal, overwrite H/W ranges
+        private func mropeCosSin(positionIds: MLXArray) -> (MLXArray, MLXArray) {
```
davidkoski, line 108: Where does this come from?  I am looking at Qwen2RotaryEmbedding.apply_mrope in mlx-vlm qwen2_5_vl and this looks similar, but more complex.  I wonder if mlx-vlm has moved forward since this was ported?

Right now it is just a few lines:

```python
    def apply_mrope(self, freqs, mrope_section):
        """Apply MRoPE to 3D rotary embeddings.
        Reorganizes frequency layout to chunked [TTT...HHH...WWW]
        args:
            freqs: (3, bs, seq_len, head_dim // 2)
            mrope_section: (3,)
        returns:
            x_t: (bs, seq_len, head_dim // 2)
        """
        freqs_t = freqs[0]  # just overwrite the first dimension T
        offset = mrope_section[0]
        for dim, length in enumerate(mrope_section[1:], start=1):  # H, W
            idx = slice(offset, offset + length)
            freqs_t[..., idx] = freqs[dim, ..., idx]
            offset += length
        return freqs_t
```

davidkoski (changes requested):

NivDvir:
Rebased on main and addressed three of the review items: video resampler reverted to bicubic, apply_mrope ported to the slice-assignment form from mlx-vlm, and per-call MROPE state plumbed through `LMOutput.State` per the #283 pattern. Targeted-size routing was already in the current head, so that one's a no-op now.

davidkoski:
Rebased on main again (fixed conflict).  The attention mask (floats) was already computed in #238 so I removed that here -- please double check this.

Libraries/MLXVLM/Models/Qwen25VL.swift
davidkoski: @NivDvir what do you think about this one?  This is the one I worry might break things.

NivDvir: Oh, you're right — that slipped in while I was matching the sizing to the Python reference. Reverted to the upstream form.

NivDvir:
Checked it against main — the vision attention path matches #238 now (boolean mask straight to SDPA), so dropping the float16 conversion is right. Thanks for handling the rebase.

NivDvir:
All addressed — targetSize back to the upstream form, boolMask confirmed against main, and the video helper renamed. When you get a chance, could you approve the workflow run (so CI picks up the new commits) and take another look? Thanks again for the rebase and the 0.31.4 fix.

davidkoski (approved):
Looks great, thank you!

PR #241, opened 2026-04-25
Fix: tolerate 1D token inputs in Gemma 4 VLM forward pass

### LINKED ISSUE #240: [BUG] Gemma 4 VLM: crash on continuation request with manual KV cache reuse (1D input shape)
# [BUG] Gemma 4 VLM (`gemma-4-e4b-it-4bit`): crash on continuation request with manual KV cache reuse

## Summary

When loading `mlx-community/gemma-4-e4b-it-4bit` via `VLMModelFactory` and driving generation with a manually-managed `[KVCache]` for prefix caching across requests, the **second** request crashes inside `Gemma4TextBackbone.callAsFunction` with an MLX-level `_mlx_error` raised from `MLXArray.subscript.getter` → `mlx_array_dim`.

Same overall pattern as #157 (Qwen 3.5 VLM with manual KV cache prefix matching) but a **different root cause**: not per-evaluation mutable state — instead, a 1D-token-input shape mismatch that's only exposed when the iterator runs the model on a short tail input + populated cache, where the model's internal `perLayerInputs` indexing `[0..., 0..., idx, 0...]` (4D) silently expects the input to have arrived through `Gemma4Processor.prepare(input:)` (which expands to 2D) rather than being constructed directly via `LMInput(tokens: MLXArray(rawTokens))` (which stays 1D).

The first request happens to succeed because the iterator's `step` adds `.newAxis` for autoregressive single-token decode, so subsequent token feeds are 2D — but the *initial prefill pass* of the second request, with its short prompt and populated cache, hits the model with a 1D input shape that the per-layer-inputs pathway doesn't tolerate.

## Reproduction

```swift
import MLX
import MLXLMCommon
import MLXVLM
import Hub

let modelPath = URL(fileURLWithPath: "/path/to/mlx-community/gemma-4-e4b-it-4bit")
let container = try await VLMModelFactory.shared.loadContainer(
    from: modelPath,
    using: #huggingFaceTokenizerLoader()
)

let processor = await container.processor
let userInput = UserInput(chat: [.user("Reply with just the word ok.")])
let lmInput = try await processor.prepare(input: userInput)
let allTokens = lmInput.text.tokens.asArray(Int.self)

// First request — fresh cache, full prompt as input. Works.
let kvCache = await container.perform { ctx in ctx.model.newCache(parameters: nil) }
var generated: [Int] = []

let parameters = GenerateParameters(maxTokens: 4)
let firstStream = try await container.perform { ctx in
    try MLXLMCommon.generateTokens(
        input: LMInput(tokens: MLXArray(allTokens)),  // 1D — works
        cache: kvCache,
        parameters: parameters,
        context: ctx
    )
}
for await event in firstStream {
    if case .token(let id) = event { generated.append(id) }
}
// At this point kvCache has ~allTokens.count + generated.count entries.

// Second request — same cache, continue with the assistant's reply as
// part of the new prefix. Compute a "tail" containing only the new
// tokens (e.g. by appending a follow-up user turn).
let userInput2 = UserInput(chat: [
    .user("Reply with just the word ok."),
    .assistant("ok"),
    .user("Reply with just the word yes."),
])
let lmInput2 = try await processor.prepare(input: userInput2)
let allTokens2 = lmInput2.text.tokens.asArray(Int.self)
let cachedPrefixLen = allTokens.count + generated.count
let tailTokens = Array(allTokens2.suffix(allTokens2.count - cachedPrefixLen))

let secondStream = try await container.perform { ctx in
    try MLXLMCommon.generateTokens(
        input: LMInput(tokens: MLXArray(tailTokens)),  // 1D — CRASH
        cache: kvCache,
        parameters: parameters,
        context: ctx
    )
}
for await _ in secondStream { /* never reached */ }
```

Crashes with `EXC_BREAKPOINT` (`SIGTRAP`) inside `Gemma4TextBackbone.callAsFunction(_:inputsEmbeds:mask:cache:perLayerInputs:)`.

## Crash trace (relevant frames)

```
_mlx_error
mlx_array_dim
MLXArray.dim(_:)
getItemND(src:operations:stream:)
MLXArray.subscript.getter
MLXArray.subscript.getter
Gemma4TextBackbone.callAsFunction(_:inputsEmbeds:mask:cache:perLayerInputs:)
Gemma4TextLanguageModel.callAsFunction(_:cache:inputsEmbeds:perLayerInputs:mask:)
Gemma4.prepare(_:cache:windowSize:)
TokenIterator.prepare(input:windowSize:)
TokenIterator.init(input:model:cache:parameters:)
generateTokens(input:cache:parameters:context:includeStopToken:wiredMemoryTicket:)
```

## Root cause

In `MLXVLM/Models/Gemma4.swift:1018-1053` (`Gemma4TextBackbone.callAsFunction`):

```swift
let layerInput: MLXArray? =
    if let finalPerLayerInputs {
        finalPerLayerInputs[0..., 0..., idx, 0...]   // 4D subscript
    } else {
        nil
    }
```

`finalPerLayerInputs` is built by `getPerLayerInputs(_ inputIds:)` and `projectPerLayerInputs(...)`. Both use `Array(inputIds.shape) + [hiddenLayers, hiddenSizePerLayerInput]` and `Array(inputsEmbeds.shape.dropLast()) + [...]` respectively to compute the result shape.

When the input `MLXArray` is **1D** `(L,)` instead of **2D** `(B, L)`:
- `inputsEmbeds = embedTokens(inputs)` produces shape `(L, D)` (2D)
- `inputsEmbeds.shape.dropLast()` is `[L]`
- `processedPerLayerInputs.shape = [L, hiddenLayers, hiddenSizePerLayerInput]` (3D)
- Subscript `[0..., 0..., idx, 0...]` requires 4D → MLX raises an internal dimension error.

`Gemma4Processor.prepare(input:)` (line 1868) prevents this for the standard pipeline because it always expands the prompt to 2D before returning `LMInput`:

```swift
let promptArray = MLXArray(promptTokens).expandedDimensions(axis: 0)
```

But when callers construct `LMInput` directly (e.g. for manual prefix caching where you slice `processor`-prepared tokens to a `[Int]` and re-wrap them into a smaller `MLXArray`), the leading batch dim is dropped and the model crashes.

### Option A — make `Gemma4TextBackbone` tolerate 1D inputs

At the top of `callAsFunction`, expand `inputs` to 2D if the caller passed 1D:

```swift
func callAsFunction(
    _ inputs: MLXArray? = nil,
    inputsEmbeds: MLXArray? = nil,
    ...
) -> MLXArray {
    let inputs = inputs.map { $0.ndim == 1 ? $0.expandedDimensions(axis: 0) : $0 }
    let inputsEmbeds = inputsEmbeds.map { $0.ndim == 2 ? $0.expandedDimensions(axis: 0) : $0 }
    // ... existing body
}
```

This matches what `LLMModel.prepare` does implicitly via `y[.newAxis, ..<prefillStepSize]` for chunked prefill. The cost is a single zero-copy reshape; output behaves identically for 2D inputs.

### Option B — make `MLXLMCommon.generateTokens(input:cache:...)` document that callers pass 2D `LMInput`

Add a `precondition(input.text.tokens.ndim >= 2, ...)` at the public API entry point so misuse is loud rather than crashing inside the model. Document the requirement on `LMInput.init(tokens:)`.

I'd prefer Option A — it removes a sharp edge that's not obvious from the public API, and matches Python `mlx-vlm`'s tolerance for 1D inputs in similar paths.

## Workaround

Until this is fixed, callers doing manual KV cache prefix caching with VLM models can wrap their tail tokens themselves:

```swift
LMInput(tokens: MLXArray(tailTokens).expandedDimensions(axis: 0))
```

Verified locally — the fix unblocks `gemma-4-e4b-it-4bit` continuation requests; cache reuse cuts TTFT from ~14 s to ~1 s on a 3000-token system prompt.

## Environment

- mlx-swift-lm: 3.31.3 (pinned exact)
- mlx-swift: 0.31.3
- macOS: 14
- Apple Silicon: M-series
- Swift: 6.0
- Model: `mlx-community/gemma-4-e4b-it-4bit`

## Related issues

- #157 — same pattern (manual cache prefix matching crashes on continuation), Qwen 3.5 VLM, different root cause (per-evaluation mutable state)
- #149 — Qwen 3.5 VLM 1D input crash, similar shape mismatch but in a different file
- #237 — Gemma4Text quantized KV cache fix (PR), addresses a different Gemma 4 bug

I'm happy to open a PR with Option A — the diff is small and contained to `MLXVLM/Models/Gemma4.swift`. Let me know if there's a preference for the API-level precondition (Option B) instead.

brdcastro:
Fixes #240.

## Summary

`Gemma4TextBackbone.callAsFunction` indexes `finalPerLayerInputs` with `[0..., 0..., idx, 0...]` (4D), which silently requires the input to have arrived through `Gemma4Processor.prepare(input:)` — that path always expands to 2D `(1, L)` via `MLXArray(promptTokens).expandedDimensions(axis: 0)`. Callers that construct `LMInput` directly with raw token IDs (e.g. for manual KV-cache prefix reuse across requests) end up handing the model a 1D `(L,)` shape; `embedTokens` produces 2D `(L, D)`; `processedPerLayerInputs` becomes 3D; and the 4D subscript crashes inside `MLXArray.subscript.getter` → `mlx_array_dim` → `_mlx_error`.

The first request happens to work because the iterator's autoregressive `step` adds `.newAxis` to each generated token, so the 2D path is hit. The crash shows up on the *initial prefill* of a continuation request when prefix-cache reuse leaves only a short tail to feed.

## Fix

At the top of `callAsFunction`, expand 1D `inputs` and 2D `inputsEmbeds` to add a leading batch dim. Zero-copy reshape; identical behavior when the caller already passed the canonical 2D shape:

```swift
let inputs = inputs.map { $0.ndim == 1 ? $0.expandedDimensions(axis: 0) : $0 }
let inputsEmbeds = inputsEmbeds.map {
    $0.ndim == 2 ? $0.expandedDimensions(axis: 0) : $0
}
```

This matches Python `mlx-vlm`'s tolerance for 1D inputs in similar paths, and matches what `LLMModel.prepare` already does implicitly via `y[.newAxis, ..<prefillStepSize]` for chunked prefill.

## Validation

End-to-end via a standalone CLI reproducer that loads `mlx-community/gemma-4-e4b-it-4bit` and runs a 2-turn manual KV-cache flow:
- **Without this patch**: turn 1 generates "ok" then turn 2 prefill crashes inside `Gemma4TextBackbone.callAsFunction` with `_mlx_error` from `mlx_array_dim`.
- **With this patch**: turn 1 generates "ok", turn 2 generates "yes", cache offset advances `0 → 19 → 38` correctly. Cache reuse cuts TTFT from ~14 s to ~1 s on a 3000-token system prompt in real usage (Lloyd app, M-series).

The full reproduction code is inline in #240.

## Notes

The alternative considered (Option B in the issue) was a `precondition(input.text.tokens.ndim >= 2, ...)` at the public `MLXLMCommon.generateTokens` entry point, but that turns a sharp edge into a louder sharp edge rather than removing it. Option A here removes the trap entirely with a single zero-copy reshape per forward pass.

### DIFF
--- a/Libraries/MLXVLM/Models/Gemma4.swift
+++ b/Libraries/MLXVLM/Models/Gemma4.swift
@@ -970,6 +970,18 @@ private final class Gemma4TextBackbone: Module {
         cache: [KVCache?]? = nil,
         perLayerInputs: MLXArray? = nil
     ) -> MLXArray {
+        // Tolerate callers that hand us a 1D `(L,)` token array instead
+        // of the canonical 2D `(B, L)` produced by `Gemma4Processor.prepare`.
+        // The downstream `perLayerInputs` indexing path (`finalPerLayerInputs[
+        // 0..., 0..., idx, 0...]`) requires 4D shapes; with 1D inputs the
+        // model otherwise crashes inside `MLXArray.subscript.getter`
+        // → `mlx_array_dim` → `_mlx_error`. This expansion is zero-copy
+        // and behaves identically when the caller already passed 2D.
+        let inputs = inputs.map { $0.ndim == 1 ? $0.expandedDimensions(axis: 0) : $0 }
+        let inputsEmbeds = inputsEmbeds.map {
+            $0.ndim == 2 ? $0.expandedDimensions(axis: 0) : $0
+        }
+
         let h0: MLXArray
         if let inputsEmbeds {
             h0 = inputsEmbeds
--- a/Libraries/MLXVLM/Models/Gemma4.swift
+++ b/Libraries/MLXVLM/Models/Gemma4.swift
@@ -968,6 +968,18 @@ private final class Gemma4TextBackbone: Module {
         cache: [KVCache?]? = nil,
         perLayerInputs: MLXArray? = nil
     ) -> MLXArray {
+        // Tolerate callers that hand us a 1D `(L,)` token array instead
+        // of the canonical 2D `(B, L)` produced by `Gemma4Processor.prepare`.
+        // The downstream `perLayerInputs` indexing path (`finalPerLayerInputs[
+        // 0..., 0..., idx, 0...]`) requires 4D shapes; with 1D inputs the
+        // model otherwise crashes inside `MLXArray.subscript.getter`
+        // → `mlx_array_dim` → `_mlx_error`. This expansion is zero-copy
+        // and behaves identically when the caller already passed 2D.
+        let inputs = inputs.map { $0.ndim == 1 ? $0.expandedDimensions(axis: 0) : $0 }
+        let inputsEmbeds = inputsEmbeds.map {
+            $0.ndim == 2 ? $0.expandedDimensions(axis: 0) : $0
+        }
+
         let h0: MLXArray
         if let inputsEmbeds {
             h0 = inputsEmbeds


davidkoski (approved):
Thank you for the fix!

davidkoski:
Waiting on CI -- infrastructure issue.

davidkoski:
Rebased on main to pick up CI fix

PR #243, opened 2026-04-25
Qwen2/2.5-VL: default to the model card's recommended image budget

NivDvir:
Qwen2.5-VL's `preprocessor_config.json` ships `max_pixels = 12,845,056` — about 12× the `1280*28*28` budget its model card recommends for image tasks. Grounding is noticeably sharper at the documented size, so this defaults there. Callers can still set any budget via `UserInput.Processing.minPixels` / `maxPixels`. ([#1175](https://github.com/Blaizzy/mlx-vlm/issues/1175) tracked the same finding for the Python side.)

### DIFF
--- a/Libraries/MLXLMCommon/UserInput.swift
+++ b/Libraries/MLXLMCommon/UserInput.swift
@@ -151,8 +151,18 @@ public struct UserInput {
 
         public var audio = AudioProcessing()
 
-        public init(resize: CGSize? = nil) {
+        /// Optional per-call overrides for the image resize budget. When set,
+        /// they replace the model's configured `min_pixels` / `max_pixels` for
+        /// this request; when `nil` the model configuration is used. This lets
+        /// a caller request the resolution a model was tuned for without
+        /// hard-coding pixel counts in the processor.
+        public var minPixels: Int?
+        public var maxPixels: Int?
+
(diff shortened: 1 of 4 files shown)


davidkoski:
Rebase on main to pick up CI fixes

NivDvir:
Rebased on main.

davidkoski (approved):
Looks good, thank you!

There may be follow-on work:

- should other models adopt this?
- should this live in MediaProcessing so that the capability applies across the board?

NivDvir:
Thanks David!

Both follow-ons fall out of one identity: `max_pixels = max_tokens · (patch·merge)²`.

- **Other models:** the recommended *token* budget is the same across the Qwen-VL line (256–1,280) — the only per-model term is `factor = patch·merge`, which is 28 for Qwen2/2.5 and 32 for Qwen3, and it's already in each config. So `1280 · factor²` gives the right default for the whole family with no per-model lookup. (Non-Qwen tokenizers are a different mechanism — out of scope.)
- **MediaProcessing:** since `factor` is the only model-specific piece, the budget + the `min`/`max` override are model-agnostic and belong there — each model just hands in its `factor`. The per-request override stays the safety valve for detail/OCR work that genuinely wants more tokens.

Happy to take a pass at the MediaProcessing move if useful.

PR #248, opened 2026-04-30
Add LLM generation benchmark helper

john-rocky:
## Summary

Adds `benchmarkLLMGeneration` and `LLMGenerationStats` to `BenchmarkHelpers`, covering prefill (prompt processing) and decode (generation) throughput in a single helper. Mirrors the surface of the existing loading and tokenization helpers (warm-up + multi-run timing, returns `BenchmarkStats`).

## Why

`BenchmarkHelpers` already covers loading, tokenization, and download. Generation throughput (tok/s) is the missing slot — useful for evaluating model-level performance changes (fusion patches, custom kernels, KV cache tweaks) without having to wire up a separate CLI tool.

## How

- Internally relies on the `GenerateCompletionInfo` emitted by the existing generation stream, so prefill/decode times are measured the same way `llm-tool` and `ChatSession` surface them.
- Prompt defaults to text from `BenchmarkDefaults.textSource`; pass an explicit `prompt:` to override.
- `temperature` is 0 so trials are deterministic and tps comparisons are reproducible across runs.
- A warm-up run is performed before the timed trials.

## Test plan

- [x] `swift build --target BenchmarkHelpers` passes (Swift 6.2.1 / macOS 26 / arm64).
- [ ] Reviewers can run the helper against any model in `LLMRegistry`.

Existing `BenchmarkHelpers` users (loading/tokenization/download) are unchanged.

### DIFF
--- a/Libraries/BenchmarkHelpers/BenchmarkHelpers.swift
+++ b/Libraries/BenchmarkHelpers/BenchmarkHelpers.swift
@@ -462,3 +462,144 @@ public func benchmarkDownloadCacheHit(
 
     return BenchmarkStats(times: times)
 }
+
+// MARK: - LLM Generation Benchmarks
+
+/// Combined prefill and decode timing for an LLM generation run.
+public struct LLMGenerationStats: Sendable {
+    public let promptTokenCount: Int
+    public let generationTokenCount: Int
+    /// Prefill latency per run, in milliseconds.
+    public let promptTimeStats: BenchmarkStats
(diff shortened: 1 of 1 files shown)


davidkoski (approved):
Looks good, thank you!

davidkoski:
Waiting on CI -- currently having infrastructure issues.

davidkoski:
rebase to pick up CI fix

PR #249, opened 2026-04-30
Optimize Gemma 4 text decode: compile-fused RMSNorm-add and gelu-mul (+23.8%)

john-rocky:
## Summary

Fuse three small graph fragments in `Gemma4DecoderLayer` into `compile()` helpers so each compiled graph spans `add+norm` or `gelu+mul` instead of emitting them as separate ops:

- Post-attention residual: `residual + RMSNorm(attnOut)`
- Post-MLP residual: `residual + RMSNorm(out)`
- PLE gating: `gelu_approx(g) * perLayerInput`, `residual + RMSNorm(g)`

The eps is hardcoded to `1e-6` to keep the compiled graph shapeless and single-instance. Every shipping Gemma 4 weight uses this value (it is the `Gemma4TextConfiguration.rmsNormEps` default).

## Numerical equivalence

`compile()` preserves graph semantics; fusion is a scheduling-level optimization, no algebraic rewrite. Output values should be bit-identical to the previous code path. Greedy output parity verified by visual inspection across all benchmark trials.

## Benchmark

Apple M4 Max (128 GB), `mlx-community/gemma-4-e2b-it-4bit`, prompt 183 tok, decode 64 tok, temperature 0, seed 42. Eight in-process trials per branch using `llm-tool` with a local `--repeat 8` flag. Trial 1 dropped as warm-up (compile lazy phase / Metal pipeline build).

Decode throughput, trials 2-8 (tok/s):

| | trials 2-8 (sorted) | median |
|---|---|---:|
| `main` | 48.2, 48.7, 50.5, 51.2, 55.4, 58.0, 59.1 | **51.2** |
| this PR | 51.9, 60.9, 61.7, 63.4, 63.8, 65.0, 66.4 | **63.4** |

→ **+23.8% decode throughput**

Prefill is unchanged (~1300-1500 tok/s in both branches), as expected — the fusions live on the per-token decode path.

The Python equivalent of these fusions on `mlx-lm` measures +2.4% on the same hardware/model. The Swift gain is larger because Swift's per-op MLX dispatch has more overhead than Python's, so consolidating ops via `compile()` recovers more of that overhead.

## Test plan

- [x] `swift build --target MLXLLM` passes (358/358, no warnings).
- [x] Decode tps measured on M4 Max as above.
- [x] Greedy output parity verified.
- [ ] Reviewers running `gemma-4-e2b-it-4bit` or `gemma-4-e4b-it-4bit` should observe similar relative speedups.

### DIFF
--- a/Libraries/MLXLLM/Models/Gemma4Text.swift
+++ b/Libraries/MLXLLM/Models/Gemma4Text.swift
@@ -9,6 +9,33 @@ import MLX
 import MLXLMCommon
 import MLXNN
 
+// MARK: - Compiled fusion fragments
+//
+// Gemma 4 uses a single rms_norm_eps (1e-6) for every RMSNorm in the model
+// (see Gemma4TextConfiguration.rmsNormEps default — all upstream Gemma 4
+// weights ship with this value). Hardcoding lets one compiled graph serve
+// every layer without per-layer specialization.
+//
+// Mirrors the upstream mlx-lm Python optimization
+// (https://github.com/ml-explore/mlx-lm/blob/main/mlx_lm/models/gemma4_text.py)
+// which fuses (residual + RMSNorm(x) * weight) and gelu(g) * other into a
+// single compiled graph. Measured ~+2.4% decode tps on M4 Max for
+// gemma-4-e2b-it-4bit at batch=1.
+
+private let kRMSEps: Float = 1e-6
+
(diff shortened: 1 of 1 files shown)


davidkoski:

davidkoski (approved):
Change looks good, thank you!

PR #247, opened 2026-04-29
Fix Gemma4 attention crash when vProj is nil (attentionKeqV)

jfranknichols:
## Bug

`Gemma4Attention.callAsFunction` crashes in `broadcast_shapes` when running models that share K and V (`attentionKeqV: true`, `vProj` nil). Reproduces reliably with [`mlx-community/gemma-4-31b-it-4bit`](https://huggingface.co/mlx-community/gemma-4-31b-it-4bit).

## Cause

`k` is transposed to `(B, nKvHeads, L, head_dim)` earlier in the method. In the `vProj == nil` branch the code did:

```swift
v = k
// ...
v = vNorm(v)
v = v.transposed(0, 2, 1, 3)
```

That second transpose reversed `k`'s layout back to `(B, L, nKvHeads, head_dim)`, which doesn't match the shape the attention call expects, producing the `broadcast_shapes` fatal error.

## Fix

Move `vNorm` + transpose into the `vProj` branch (where `v` actually starts as `(B, L, nKvHeads, effectiveHeadDim)` and needs both), and in the nil-`vProj` branch apply `vNorm` directly to the already-transposed `k`.

The `vProj` branch is behaviorally unchanged.

## Verification

- Tested locally with `gemma-4-31b-it-4bit` — generation runs without crashing.
- Models that have a separate `v_proj` (e.g. `gemma-4-e2b/e4b`) hit the unchanged branch and continue to work.

PR #250, opened 2026-04-30
Fix Gemma 4 text K-eq-V attention: don't double-process v from already-transposed k (#231)

### LINKED ISSUE #231: Gemma4 (MLXLLM): K-eq-V attention path double-transposes V, causes broadcast_shapes crash
## Summary

The text-only Gemma 4 implementation in `MLXLLM/Models/Gemma4Text.swift` crashes at the first attention step when `attention_k_eq_v: true`, because `V` is computed from the already-transposed `K` and then transposed a second time. This produces a key/value shape mismatch that fails inside `scaledDotProductAttention`.

**Affected models:** any Gemma 4 variant with `attention_k_eq_v: true` in the text config — notably:
- `mlx-community/gemma-4-26b-a4b-it-4bit`
- `mlx-community/gemma-4-31b-it-4bit`

The dense E2B / E4B models have `attention_k_eq_v: false` and work fine, which is why this hasn't surfaced yet. The 26B/31B happen to also be available through `MLXVLM`, where the attention code path is correct.

## Reproduction

```swift
import MLXLLM
import MLXLMCommon

let container = try await LLMModelFactory.shared.loadContainer(
    configuration: ModelConfiguration(id: "mlx-community/gemma-4-31b-it-4bit")
)

try await container.perform { context in
    let input = try await context.processor.prepare(
        input: UserInput(chat: [.user("Hello")])
    )
    let stream = try generateTokensTask(
        input: input,
        parameters: GenerateParameters(maxTokens: 8),
        context: context
    )
    for await _ in stream.0 {}
}
```

Crashes with:

```
Fatal error: [broadcast_shapes] Shapes (1,512,4,512) and (1,4,512,512) cannot be broadcast.
```

## Root cause

`MLXLLM/Models/Gemma4Text.swift` (around lines 302–323), inside `Gemma4Attention.callAsFunction`, the K-eq-V branch:

```swift
var k = kProj(x).reshaped(B, L, nKvHeads, effectiveHeadDim)
k = kNorm(k)
k = k.transposed(0, 2, 1, 3)                              // (B, H, L, D)
k = gemma4ApplyRotaryPosition(rope, to: k, offset: activePositionOffset)

var v: MLXArray
if let vProj {
    v = vProj(x).reshaped(B, L, nKvHeads, effectiveHeadDim)
} else {
    v = k                                                  // already (B, H, L, D), normalized, RoPE'd
}
v = vNorm(v)                                               // applied over wrong axis ordering
v = v.transposed(0, 2, 1, 3)                               // SECOND transpose -> (B, L, H, D)
```

Step by step when `vProj == nil`:

1. `k` is reshaped to `(B, L, nKvHeads, hdim)`
2. `k = kNorm(k)`
3. `k = k.transposed(0, 2, 1, 3)` → `(B, nKvHeads, L, hdim)`
4. `k` gets RoPE applied
5. `v = k` copies the already-transposed-and-RoPE'd tensor
6. `v = vNorm(v)` is applied over the wrong axes
7. `v = v.transposed(0, 2, 1, 3)` transposes a second time → `(B, L, nKvHeads, hdim)`

Result: `keys` is `(B, H, L, D)` but `values` is `(B, L, H, D)`, and `scaledDotProductAttention` cannot broadcast them.

For comparison, the equivalent path in `MLXVLM/Models/Gemma4.swift` (around lines 676–704) does the right thing — it copies the **pre-transpose, pre-norm** projection into `V`:

```swift
var keys = kProj(x).reshaped(batch, length, numKVHeads, headDim)
var values = useKEqV
    ? keys
    : vProj!(x).reshaped(batch, length, numKVHeads, headDim)
keys = kNorm(keys).transposed(0, 2, 1, 3)
values = vNorm(values).transposed(0, 2, 1, 3)
keys = rope(keys, offset: currentOffset)
```

## Proposed fix

In `Gemma4Text.swift`, capture the raw `kProj` output before it's normalized/transposed, and use that for `V` when `vProj == nil`:

```swift
let kRaw = kProj(x).reshaped(B, L, nKvHeads, effectiveHeadDim)
var k = kNorm(kRaw)
k = k.transposed(0, 2, 1, 3)
k = gemma4ApplyRotaryPosition(rope, to: k, offset: activePositionOffset)

var v: MLXArray
if let vProj {
    v = vProj(x).reshaped(B, L, nKvHeads, effectiveHeadDim)
} else {
    v = kRaw
}
v = vNorm(v)
v = v.transposed(0, 2, 1, 3)
```

This mirrors the MLXVLM implementation and resolves the shape mismatch.

## Environment

- mlx-swift-lm: 3.31.3 (resolved from `from: "3.0.0"`)
- macOS: 14.4+
- Swift: 5.9+

Happy to open a PR if helpful.

john-rocky:
## Summary

Fixes #231.

When \`attention_k_eq_v: true\` (used by \`mlx-community/gemma-4-26b-a4b-it-4bit\` and \`mlx-community/gemma-4-31b-it-4bit\`), \`v_proj\` is omitted and values are sourced from the raw \`k_proj\` output. The previous code took \`v = k\` *after* \`k_norm\`, the \`(B, H, L, D)\` transpose, and RoPE had already been applied to \`k\`, then re-applied \`v_norm\` and another transpose. The result was that \`keys\` ended up as \`(B, H, L, D)\` while \`values\` ended up as \`(B, L, H, D)\`, which \`scaledDotProductAttention\` cannot broadcast:

\`\`\`
Fatal error: [broadcast_shapes] Shapes (B, H, L, D) and (B, L, H, D) cannot be broadcast.
\`\`\`

The fix moves \`v = k\` to before \`k_norm\` / transpose / RoPE so v's subsequent \`vNorm\` + transpose lands it in \`(B, H, L, D)\` to match keys, matching the upstream Python (\`mlx_lm/models/gemma4_text.py\`).

## Verification (M4 Max)

\`Tests/MLXLMTests/Gemma4KeqVTests.swift\` builds a tiny Gemma 4 Text model (1 layer, hidden_size=64, head_dim=16) with \`attention_k_eq_v: true\` and runs a forward pass at both T=1 (decode) and T=8 (prefill) — the two paths the broadcast crash fired in.

\`\`\`
Test Case 'testForwardWithKeqV' passed (0.021 seconds)
Test Case 'testPrefillWithKeqV' passed (0.002 seconds)
\`\`\`

Pre-fix, both tests fatal-error inside the first \`scaledDotProductAttention\` call. The tests use random-init weights (no checkpoint download needed) since the bug is purely a shape mismatch.

Run via:

\`\`\`bash
  -only-testing:MLXLMTests/Gemma4KeqVTests \\
  -skipMacroValidation
\`\`\`

## Test plan

- [x] \`swift build\` clean.
- [x] Two regression tests pin the k / v shape contract under \`attention_k_eq_v=true\` for both decode and prefill.
- [ ] Reviewers running \`mlx-community/gemma-4-26b-a4b-it-4bit\` or \`mlx-community/gemma-4-31b-it-4bit\` should no longer crash with \`broadcast_shapes\` and should produce greedy output matching the Python \`mlx-lm\` reference.

### DIFF
--- a/Libraries/MLXLLM/Models/Gemma4Text.swift
+++ b/Libraries/MLXLLM/Models/Gemma4Text.swift
@@ -300,16 +300,22 @@ private class Gemma4Attention: Module {
             values = sharedV
         } else {
             var k = kProj(x).reshaped(B, L, nKvHeads, effectiveHeadDim)
-            k = kNorm(k)
-            k = k.transposed(0, 2, 1, 3)
-            k = gemma4ApplyRotaryPosition(rope, to: k, offset: activePositionOffset)
 
+            // K-eq-V (attention_k_eq_v=true): values come from the raw k_proj
+            // output, BEFORE k_norm and the (B, H, L, D) transpose / RoPE.
+            // v_norm is then applied to that raw k_proj output, matching the
+            // Python reference (mlx_lm/models/gemma4_text.py).
             var v: MLXArray
             if let vProj {
                 v = vProj(x).reshaped(B, L, nKvHeads, effectiveHeadDim)
             } else {
                 v = k
             }
+
+            k = kNorm(k)
+            k = k.transposed(0, 2, 1, 3)
+            k = gemma4ApplyRotaryPosition(rope, to: k, offset: activePositionOffset)
+
             v = vNorm(v)
             v = v.transposed(0, 2, 1, 3)
 
--- a/Tests/MLXLMTests/Gemma4KeqVTests.swift
+++ b/Tests/MLXLMTests/Gemma4KeqVTests.swift
@@ -1,0 +1,80 @@
+// Copyright © 2026 Apple Inc.
+//
+// Verifies the #231 fix: Gemma 4 Text attention with `attention_k_eq_v=true`
+// must produce keys and values with matching shapes. Pre-fix the code took
+// `v = k` AFTER `k_norm` + transpose + RoPE had been applied to k, then ran
+// v through its own `vNorm` + transpose, so values ended up in
+// `(B, L, H, D)` while keys were in `(B, H, L, D)` — `scaledDotProductAttention`
+// crashes with `broadcast_shapes (B, H, L, D) vs (B, L, H, D)`.
+
+import Foundation
+import MLX
+@testable import MLXLLM
+import MLXLMCommon
+import XCTest
+
+final class Gemma4KeqVTests: XCTestCase {
+
+    /// Build a tiny Gemma 4 Text config (1 layer, hidden_size=64) with
+    /// `attention_k_eq_v: true`. Defaults handle every other field.
+    private func makeConfig() throws -> Gemma4TextConfiguration {
+        let json = """
+            {
+                "model_type": "gemma4_text",
+                "hidden_size": 64,
+                "num_hidden_layers": 1,
+                "intermediate_size": 128,
+                "num_attention_heads": 4,
+                "head_dim": 16,
+                "global_head_dim": 32,
+                "num_key_value_heads": 2,
+                "vocab_size": 64,
+                "vocab_size_per_layer_input": 64,
+                "hidden_size_per_layer_input": 16,
+                "num_kv_shared_layers": 0,
+                "sliding_window": 16,
+                "sliding_window_pattern": 2,
+                "max_position_embeddings": 128,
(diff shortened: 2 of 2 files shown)


davidkoski:
Duplicate of #247 -- closing this but thank you for the submission!

PR #251, opened 2026-04-30
TokenRing: flatten 2D prompts before measuring length (#220)

### LINKED ISSUE #220: TokenRing.loadPrompt crash: prompt.dim(0) returns batch dim for 2D tensors
Hey! 👋

I ran into a crash when using `repetitionPenalty` with Gemma 3 VLM (QAT 4-bit). The issue is in `TokenRing.loadPrompt()` in `Evaluate.swift`.

## The bug

`loadPrompt` uses `prompt.dim(0)` to get the number of tokens, but when the prompt is 2D `(1, seqLen)`, `dim(0)` returns `1` (the batch dimension) instead of the actual sequence length.

This causes the ring buffer to be incorrectly sized — the prompt gets flattened to `(seqLen,)` via `reshaped(-1)` but the capacity check used `n=1`, so the buffer ends up with shape `(seqLen + capacity - 1,)` instead of `(capacity,)`. Then `append()` tries `MLX.where(positions, token, buffer)` where `positions` has shape `(capacity,)` and `buffer` has shape `(seqLen,)` → broadcast crash.

```
Fatal error: [broadcast_shapes] Shapes (64) and (1009) cannot be broadcast.
```

## Suggested fix

In `TokenRing.loadPrompt`, flatten the prompt before measuring its length:

```swift
mutating func loadPrompt(_ prompt: MLXArray) {
    let promptTokens = prompt.asType(.int32).reshaped(-1)  // flatten first
    let n = promptTokens.dim(0)  // now always gives seq length
    ...
}
```

## Repro

Any model with a prompt longer than `repetitionContextSize` (default 64) and a 2D prompt tensor `(1, seqLen)`:

```swift
let params = GenerateParameters(
    maxTokens: 100,
    temperature: 0.7,
    repetitionPenalty: 1.1,
    repetitionContextSize: 64
)
// → crash during first token generation
```

Introduced in f7a235d. Thanks for the great library! 🙏

john-rocky:
## Summary

Fixes #220.

\`TokenRing.loadPrompt\` measured prompt length with \`prompt.dim(0)\`, which returns the batch dimension for a 2D \`(1, L)\` tensor — the standard shape that comes out of every \`*Processor.prepare\` implementation. As a result the ring buffer was sized to length \`L\` (via \`reshaped(-1)\`) while \`count\` was set to \`1\`, producing internally inconsistent state that crashed the next \`append()\`:

\`\`\`
Fatal error: [broadcast_shapes] Shapes (64) and (1009) cannot be broadcast.
\`\`\`

The fix flattens before measuring length so 2D and 1D prompts are treated identically.

## Verification (M4 Max)

\`Tests/MLXLMTests/TokenRingTests.swift\` exercises the path through the public \`RepetitionContext\` (which wraps \`TokenRing\`):

| test | input | pre-fix | post-fix |
|---|---|---|---|
| \`test2DPromptLongerThanContextDoesNotCrash\` | \`(1, 1009)\`, ctx=64 | \`broadcast_shapes\` fatal | passes |
| \`test2DPromptShorterThanContextDoesNotCrash\` | \`(1, 32)\`, ctx=64 | inconsistent state | passes |
| \`test1DPromptStillWorks\` | \`(200,)\`, ctx=64 | already worked | regression guard |

\`\`\`
Test Case 'test1DPromptStillWorks' passed (0.021 seconds)
Test Case 'test2DPromptLongerThanContextDoesNotCrash' passed (0.001 seconds)
Test Case 'test2DPromptShorterThanContextDoesNotCrash' passed (0.001 seconds)
\`\`\`

Run via:

\`\`\`bash
  -only-testing:MLXLMTests/TokenRingTests \\
  -skipMacroValidation
\`\`\`

## Test plan

- [x] \`swift build\` clean.
- [x] Three regression tests pin the 2D-prompt contract on \`RepetitionContext\` (which is the path #220 reports).
- [ ] Reviewers running a 1000+ token prompt with \`repetitionPenalty: 1.1\`, \`repetitionContextSize: 64\` should no longer crash.

### DIFF
--- a/Libraries/MLXLMCommon/Evaluate.swift
+++ b/Libraries/MLXLMCommon/Evaluate.swift
@@ -334,19 +334,24 @@ struct TokenRing {
 
     /// Bulk-load from a prompt. Keeps the last `capacity` tokens.
     mutating func loadPrompt(_ prompt: MLXArray) {
-        let n = prompt.dim(0)
-        let promptTokens = prompt.asType(.int32)
+        // Flatten first so 2D `(B, L)` prompts are treated as length `L` rather
+        // than `B`. Without this, 2D prompts (which is the standard shape from
+        // `*Processor.prepare`) underflow capacity, leaving `buffer` shaped
+        // `(L,)` instead of `(capacity,)` and breaking `append()`'s
+        // mask broadcast.
+        let promptTokens = prompt.asType(.int32).reshaped(-1)
+        let n = promptTokens.dim(0)
         if n <= capacity {
             if n < capacity {
                 let padding = MLXArray.zeros([capacity - n], type: Int32.self)
-                buffer = concatenated([promptTokens.reshaped(-1), padding])
+                buffer = concatenated([promptTokens, padding])
             } else {
-                buffer = promptTokens.reshaped(-1)
+                buffer = promptTokens
             }
             count = n
             writeIndex = n % capacity
         } else {
-            buffer = promptTokens[(-capacity)...].reshaped(-1)
+            buffer = promptTokens[(-capacity)...]
             count = capacity
             writeIndex = 0
         }
--- a/Tests/MLXLMTests/TokenRingTests.swift
+++ b/Tests/MLXLMTests/TokenRingTests.swift
@@ -1,0 +1,67 @@
+// Copyright © 2026 Apple Inc.
+//
+// Verifies the #220 fix: `TokenRing.loadPrompt` must measure prompt length
+// after flattening. Prompts coming from `*Processor.prepare` are typically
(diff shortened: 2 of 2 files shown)


ptsochantaris:
This looks like https://github.com/ml-explore/mlx-swift-lm/pull/170?

john-rocky:
Thanks for the pointer @ptsochantaris — yes, this is the same fix that #170 already lands. Both PRs flatten the prompt before measuring length, with #170 using the cleaner `.flattened().count` form. #170 also has broader test coverage across the three penalty contexts (Repetition / Presence / Frequency).

Closing this in favor of #170. The issue I closed against (#220) reports the same root cause as #170's #168 — different reporters, same bug.

PR #252, opened 2026-04-30
JSONToolCallParser: accept stringified `arguments` field (#169, Granite 4)

### LINKED ISSUE #169: [BUG] JSONToolCallParser fails to parse stringified arguments (used by Granite 4 and others)

john-rocky:
## Summary

Fixes #169.

The IBM Granite 4 family (and others) emit tool calls with the \`arguments\` field as a JSON-encoded *string* rather than a nested object:

\`\`\`
{"name": "func", "arguments": "{\\"key\\": \\"value\\"}"}
\`\`\`

\`JSONDecoder().decode(ToolCall.Function.self, ...)\` requires \`arguments\` to be a nested \`{...}\`, so the parse silently returned \`nil\` and tool calls never reached the user.

This PR adds a private \`WireFunction\` decoder that accepts either shape:

- nested JSON object → use directly
- JSON-encoded string → re-decode the string contents as \`[String: JSONValue]\`

Models that already emit nested objects (Llama 3, Qwen, etc.) keep the original behavior. Empty \`{}\` arguments are preserved. Mirrors the flexibility of the Python \`mlx-lm\` default tool parser. Credit to @neon52 for the analysis and proposed fix in #169.

## Verification (M4 Max)

\`Tests/MLXLMTests/JSONToolCallParserTests.swift\` covers the five interesting shapes:

| test | input arguments shape | expected |
|---|---|---|
| \`nestedObjectArguments\` | nested \`{...}\` (standard) | parses (regression guard) |
| \`stringifiedArguments\` | JSON-encoded string (#169 / Granite 4) | parses (the fix) |
| \`emptyArguments\` | \`{}\` | preserves empty dict |
| \`emptyStringifiedArguments\` | \`"{}"\` | preserves empty dict |
| \`malformedStringifiedArguments\` | \`"not-json"\` | fails gracefully (returns \`nil\`, no crash) |

\`\`\`
✔ Test "Parses nested object arguments (standard shape)" passed (0.016s)
✔ Test "Parses JSON-encoded string arguments (Granite 4 shape, #169)" passed (0.016s)
✔ Test "Empty object arguments are preserved" passed (0.001s)
✔ Test "Empty stringified arguments are also preserved" passed (0.001s)
✔ Test "Malformed stringified arguments returns nil instead of crashing" passed (0.016s)
\`\`\`

Run via:

\`\`\`bash
  -only-testing:MLXLMTests/JSONToolCallParserTests \\
  -skipMacroValidation
\`\`\`

## Test plan

- [x] \`swift build --target MLXLMCommon\` passes (no warnings).
- [x] Five regression tests pin both shapes plus the malformed input failure mode.
- [ ] Reviewers running \`mlx-community/granite-4.0-h-micro-4bit\` with tools enabled should see tool calls parsed instead of silently dropped.

### DIFF
--- a/Libraries/MLXLMCommon/Tool/Parsers/JSONToolCallParser.swift
+++ b/Libraries/MLXLMCommon/Tool/Parsers/JSONToolCallParser.swift
@@ -2,6 +2,48 @@
 
 import Foundation
 
+/// Wire-side decoder that accepts both shapes the upstream models emit:
+///   * arguments as a nested object: `{"name":"f","arguments":{"k":"v"}}`
+///   * arguments as a JSON-encoded string (Granite 4 etc.):
+///     `{"name":"f","arguments":"{\"k\":\"v\"}"}`
+private struct WireFunction: Decodable {
+    let name: String
+    let arguments: [String: JSONValue]
+
+    private enum CodingKeys: String, CodingKey {
+        case name
+        case arguments
+    }
+
+    init(from decoder: Decoder) throws {
+        let c = try decoder.container(keyedBy: CodingKeys.self)
+        self.name = try c.decode(String.self, forKey: .name)
+
+        if c.contains(.arguments) {
+            if let nested = try? c.decode([String: JSONValue].self, forKey: .arguments) {
+                self.arguments = nested
+            } else if let stringified = try? c.decode(String.self, forKey: .arguments) {
+                guard let data = stringified.data(using: .utf8),
+                    let nested = try? JSONDecoder().decode(
+                        [String: JSONValue].self, from: data)
+                else {
+                    throw DecodingError.dataCorruptedError(
+                        forKey: .arguments, in: c,
+                        debugDescription:
+                            "arguments is a String but does not contain a JSON object")
+                }
+                self.arguments = nested
+            } else {
+                throw DecodingError.dataCorruptedError(
+                    forKey: .arguments, in: c,
+                    debugDescription: "arguments must be an object or a JSON-encoded string")
+            }
+        } else {
+            self.arguments = [:]
+        }
+    }
+}
+
 /// Parser for JSON format: <tag>{"name": "...", "arguments": {...}}</tag>
 /// Reference: https://github.com/ml-explore/mlx-lm/blob/main/mlx_lm/tool_parsers/default.py
 public struct JSONToolCallParser: ToolCallParser, Sendable {
@@ -30,9 +72,10 @@ public struct JSONToolCallParser: ToolCallParser, Sendable {
         let jsonStr = text.trimmingCharacters(in: .whitespacesAndNewlines)
 
         guard let data = jsonStr.data(using: .utf8),
-            let function = try? JSONDecoder().decode(ToolCall.Function.self, from: data)
+            let wire = try? JSONDecoder().decode(WireFunction.self, from: data)
         else { return nil }
 
-        return ToolCall(function: function)
+        return ToolCall(
+            function: ToolCall.Function(name: wire.name, arguments: wire.arguments))
     }
 }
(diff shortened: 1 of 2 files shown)


DePasqualeOrg:

john-rocky:
Thanks @DePasqualeOrg — I read through #260, #475, and the Swift LM Response Parser repo. I'll test it against Granite 4's stringified-arguments case (and a couple of other tool-calling models I'm running locally) and report back, either here or in your repo.

A note on this PR's scope: the change here is narrow — it only handles the specific case where a model emits `arguments` as a JSON-encoded string instead of an object (Granite 4 does this consistently, and it's a documented OpenAI-compat dialect issue). It's a 5-line fix with regression tests, and it makes the existing in-tree parser correct for one more model without pulling in a new dependency.

Whether the in-tree parser eventually gets deprecated in favor of an external library is a call for the maintainers, and the broader question is being handled in #260. Until that's resolved, this PR unblocks Granite 4 users on the existing API surface. Happy to close it if the direction lands on "deprecate in-tree parsing" — just looking for the maintainer signal on #260 first.

john-rocky:
Closing — superseded by @kuosuko's #172, which was merged on 2026-05-11 and uses an equivalent JSONSerialization-based normalization to accept the stringified \`arguments\` shape. Issue #169 was closed by @davidkoski at the same time. Sorry for the noise.

Issue #143, opened 2026-03-11
Qwen35 VLM sanitize logic fails to map bare 'model.*' keys to 'language_model


ajayphilipjoy:
### **Overview**
There is a discrepancy between the sanitize implementation in the Qwen35 LLM and VLM paths. When loading the unified Qwen 3.5 0.8B model via the VLM path, the weights fail to initialize because keys starting with a bare model. prefix  are not correctly remapped to the language_model namespace.

### **The Logic Gap**

In the LLM version, there is a robust catch-all:

**Swift**
```
else if !key.hasPrefix("language_model.") {
    key = "language_model." + key
}
```
This ensures that any key not already specifically handled gets wrapped into the language_model submodule.

In the VLM version, the logic is too restrictive:

**Swift**
```
if key.contains("model") {
    if key.contains("model.language_model") {
        // ... handled
    } else if key.contains("model.visual") {
        // ... handled
    }
    // Bare "model.layers" keys fall through here and are NOT modified
}
```
### **Reproduction & Error**

When loading mlx-community/Qwen3.5-0.8B-MLX-4bit weights:

Expected: model.layers.0.mlp.up_proj.weight → language_model.model.layers.0.mlp.up_proj.weight

Actual: model.layers.0.mlp.up_proj.weight (Unchanged)

This results in a keyNotFound error because the Swift property language_model never receives its required weights.

### **Proposed Fix**

Update the VLM sanitize function to include a fallback for bare model. keys, ensuring parity with the LLM loading path:

**Swift**
```
if key.contains("model") {
    if key.contains("model.language_model") {
        key = key.replacingOccurrences(of: "model.language_model", with: "language_model.model")
    } else if key.contains("model.visual") {
        key = key.replacingOccurrences(of: "model.visual", with: "vision_tower")
    } else if key.hasPrefix("model.") {
        // ADD THIS: Fixes bare model.* keys for unified Qwen 3.5
        key = "language_model." + key
    }
}
```

Issue #124, opened 2026-02-28
MoE models 7.3x slower than Python mlx-lm (per-token sync + evalLock)


dirvine:
## Summary

MoE models (Qwen3.5-35B-A3B) generate at **11.7 T/s** in mlx-swift-lm vs **85 T/s** (MLX-internal) in Python mlx-lm — a **7.3x slowdown**. Dense models (Qwen3-8B) show only a 1.33x gap (52.8 vs 70.2 T/s), confirming the issue is MoE-specific.

## Environment

- **Hardware:** Apple M4 Max, 96 GB unified memory
- **mlx-swift:** 0.30.6
- **mlx-swift-lm:** commit 11968af (latest HEAD)
- **Python mlx-lm:** 0.30.7, mlx 0.31.0
- **Model:** `NexVeridian/Qwen3.5-35B-A3B-4bit` (MoE, 35B total / 3B active per token)

### Swift (11.7 T/s)

```swift
let config = ModelConfiguration(id: "NexVeridian/Qwen3.5-35B-A3B-4bit")
let container = try await LLMModelFactory.shared.loadContainer(configuration: config)
let input = UserInput(chat: [
    .system("/no_think\nYou are a helpful assistant."),
    .user("What is the weather like today?"),
])
let lmInput = try await container.prepare(input: input)
let params = GenerateParameters(maxTokens: 128, temperature: 0.7)
let stream = try await container.generate(input: lmInput, parameters: params)
for await generation in stream {
    // info.tokensPerSecond reports ~11.7 T/s
}
```

### Python (85 T/s)

```python
from mlx_lm import load, generate
model, tokenizer = load("NexVeridian/Qwen3.5-35B-A3B-4bit")
# Verbose output shows: Generation: 128 tokens, 85.094 tokens-per-sec
output = generate(model, tokenizer, prompt=prompt_text, max_tokens=128, verbose=True)
```

## Verified measurements

| Model | Type | Python mlx-lm (internal T/s) | Swift mlx-swift-lm (internal T/s) | Ratio |
|---|---|---:|---:|---|
| Qwen3-8B | Dense | 70.2 | 52.8 | 1.33x |
| Qwen3.5-35B-A3B | MoE | 85.1 | 11.7 | **7.3x** |

Python numbers verified across 3 consistent runs (72 T/s wall-time, 85 T/s MLX-internal).

## Root cause analysis

We traced the bottleneck to three patterns in the Swift generation loop:

[...]

PR #257, opened 2026-05-01
Share the fused gated-delta kernel between MLXLLM and MLXVLM

john-rocky:
## Summary

`Libraries/MLXLLM/Models/GatedDelta.swift` defines a fused `gatedDeltaKernel` (custom Metal kernel, single threadgroup walking the time loop) and a `gatedDeltaOps` expression-graph fallback. `gatedDeltaUpdate` prefers the kernel when MLX Metal is available, which is the only path the LLM-side Qwen 3.5 / Qwen 3 Next models take on Apple Silicon.

`MLXVLM/Models/Qwen35.swift` was a copy/paste of the same code with the kernel branch removed:

```swift
private func gatedDeltaUpdate(...) {
    ...
    return gatedDeltaOps(...)   // ops fallback only
}
```

So every VLM Qwen 3.5 checkpoint with `linear_attn` was running the unfused per-step expression-graph version, generating far more intermediates per token. This is a stand-alone VLM-side missed-branch fix; for context on the broader #124 perf discussion, see [my profiling comment there](https://github.com/ml-explore/mlx-swift-lm/issues/124#issuecomment-4358100904) showing the 35B-A3B-4bit text-only MoE no longer reproduces a gap on current `mlx-swift 0.31.3` (mlx-c 0.31.1) builds — the VLM-side branch removal addressed in this PR is independent of that finding.

## Change

- Move `Libraries/MLXLLM/Models/GatedDelta.swift` → `Libraries/MLXLMCommon/GatedDelta.swift`.
- Make `computeGatedDeltaG`, `gatedDeltaKernel`, `gatedDeltaOps`, `gatedDeltaUpdate` `public` so both LLM and VLM call through `MLXLMCommon`.
- Delete the duplicate `computeGatedDeltaG` / `gatedDeltaStepOps` / `gatedDeltaOps` / `gatedDeltaUpdate` helpers from `MLXVLM/Models/Qwen35.swift`. The single VLM call site now resolves to the shared public function with the kernel-preferred dispatch.

Net (excluding the bench test): +10 / −132. No behavior change for MLXLLM (same code, new home); MLXVLM Qwen 3.5 with `linear_attn` now hits the fused Metal kernel.

## Microbench (M4 Max, 128 GB, bfloat16)

`Tests/MLXLMTests/GatedDeltaBenchTests.swift` constructs realistic Qwen 3.5 linear-attn input shapes (B=1, Hk=16, Hv=64, Dk=192, Dv=128) at four representative T values and times `gatedDeltaKernel` vs `gatedDeltaOps` over 20 iterations after warmup.

| shape | T | `gatedDeltaOps` median (ms) | `gatedDeltaKernel` median (ms) | speedup |
|---|---:|---:|---:|---:|
| decode_T1 | 1 | 0.605 | 0.315 | **1.92×** |
| decode_T8 | 8 | 2.909 | 0.369 | **7.88×** |
| prefill_T64 | 64 | 20.255 | 0.799 | **25.35×** |
| prefill_T256 | 256 | 87.577 | 1.467 | **59.69×** |

Reproduce:

```bash
xcodebuild test -scheme mlx-swift-lm-Package \
  -destination 'platform=macOS,arch=arm64' \
  -only-testing:MLXLMTests/GatedDeltaBenchTests \
  -skipMacroValidation
```

(SPM `swift test` is unsupported because the metal kernel needs the .metallib bundle that only xcodebuild produces.)

## End-to-end (M4 Max, 128 GB, `mlx-community/Qwen3.5-0.8B-MLX-4bit`)

The 0.8B unified VLM checkpoint has 24 transformer blocks; 18 of them are `linear_attention` (gated-delta) per its `text_config.layer_types` and 6 are `full_attention`. With `max-tokens 64`, `temperature 0`, `seed 42`, prompt `"What is the capital of Japan?"` (24 tokens), 6 trials per branch with the first dropped as warmup:

| build | prompt median (tok/s) | generation median (tok/s) |
|---|---:|---:|
| Swift VLM, baseline (origin/main + #143 sanitize fix only, ops fallback) | 1377.42 | 164.67 |
| Swift VLM, this PR (sanitize fix + kernel fix) | **2125.60** | **355.95** |
| Python `mlx_lm` 0.31.3 (kernel path) | 669.80 | 213.59 |
| **Swift speedup over baseline** | **1.54×** | **2.16×** |

Decode improvement (1.54× / 2.16×) is smaller than the microbench (1.92× – 59.69×) because the 0.8B model still spends a meaningful fraction of each forward in non-`linear_attn` work (full-attention blocks, MoE / MLP, RMS norm, embed). The decode-per-token cost on `linear_attn` blocks specifically is what the kernel fuses; end-to-end gain scales with the share of those blocks. Models with a higher `linear_attn` ratio (the larger 35B-A3B / 122B-A10B Qwen 3.5 unified VLMs) should see proportionally larger wins.

Greedy output is identical between the two Swift builds:

> The capital of Japan is **Tokyo** (also written as **東京** in Japanese).
>
> Tokyo is the largest city in the country, located in the central part of the country, …

(Python emits the same text minus the bold markers because it doesn't go through Swift's markdown wrapper.)


Libraries/MLXVLM/Models/Qwen35.swift
```diff
@@ -16,134 +16,11 @@ private enum Qwen35VLError: Error {
...
-    let Dv = v.dim(3)
-
-    let state = state ?? MLXArray.zeros([B, Hv, Dv, Dk], dtype: q.dtype)
-    return gatedDeltaOps(q: q, k: k, v: v, g: g, beta: beta, state: state, mask: mask)
-}
+// Gated-delta helpers (`gatedDeltaUpdate`, `gatedDeltaOps`, `gatedDeltaKernel`,
+// `computeGatedDeltaG`) are shared with MLXLLM via `MLXLMCommon/GatedDelta.swift`,
+// which selects the fused Metal kernel when available and falls back to the ops
+// path otherwise. Keeping a single implementation also keeps both paths in sync
+// when the upstream Python kernel evolves.
```
davidkoski: We don't need the comment for removed code.

Tests/MLXLMTests/GatedDeltaBenchTests.swift
```diff
@@ -0,0 +1,122 @@
...
+//
+//   swift test --filter GatedDeltaBenchTests
+//
+// to print median latency over N iterations on the current device.
+
+import MLX
+import MLXLMCommon
+import XCTest
+
+final class GatedDeltaBenchTests: XCTestCase {
```
davidkoski: I don't think we need benchmark tests, though if you did want to add it, look in IntegrationTesting -- that is a more appropriate place.

davidkoski (changes requested):
See comments.  This needs a rebase to pull the updated GatedDelta code.

Thanks!

Libraries/MLXVLM/Models/Qwen35.swift
john-rocky: Done — the duplicate helpers are just deleted now, nothing left in their place.

Tests/MLXLMTests/GatedDeltaBenchTests.swift
john-rocky: Dropped the benchmark test file. The microbench and end-to-end numbers stay in the PR description as the supporting evidence.

john-rocky:
Thanks for the review! Addressed all three points:

- **Rebased onto current `main`** (`5b7e543`) — this picks up #224 (`fix gated delta state precision -- fp32 state to match Python`), so the moved `GatedDelta.swift` carries the fp32-state code rather than the pre-#224 version.
- **Removed the comment for the deleted code** in `MLXVLM/Models/Qwen35.swift` — the duplicate helpers are just gone now.
- **Dropped the benchmark test** (`GatedDeltaBenchTests.swift`). The microbench / end-to-end numbers stay in the PR description as the supporting evidence.

Two things worth flagging since they're new in this diff:

- Only `gatedDeltaUpdate` is `public` now — its helpers (`computeGatedDeltaG`, `gatedDeltaKernel`, `gatedDeltaOps`) stay internal, since nothing outside `GatedDelta.swift` calls them. The earlier revision over-exposed all four.
- `Tests/MLXLMTests/GatedDeltaTests.swift` (added in #224) switches from `@testable import MLXLLM` to `import MLXLMCommon`, following the file's move — it only references the now-public `gatedDeltaUpdate`.

Net is +2 / −132 across 3 files (`GatedDelta.swift` is a 99%-similarity rename). `swift build` and `swift build --build-tests` both pass clean on `main`@`5b7e543`.

davidkoski:
Rebased on main to pick up CI fix + fix merge conflict.

davidkoski (approved):
Looks good, thank you!

PR #262, opened 2026-05-02
Add continuous batching and Gemma 4 text support

Gajesh2007:
## Summary
- Add a continuous-batching generation path with batched KV caches, per-row samplers, and stop-sequence handling.
- Add Gemma 4 text model support, including shared-KV handling and MoE projection sanitization.
- Document continuous batching and add upstream-local tests for batching caches, row samplers, stop matching, and request admission.

## Test plan
- [x] `uvx pre-commit run --all-files`
- [x] `xcodebuild test -scheme mlx-swift-lm-Package -destination 'platform=macOS'`
- [x] `scripts/verify-docs.sh`

Made with [Cursor](https://cursor.com)


Gajesh2007:
Closing for now; we'll reopen when ready.

PR #263, opened 2026-05-02
Add continuous batching and Gemma 4 text support

Gajesh2007:
## Summary
- Add a continuous-batching generation path with batched KV caches, cancellation, per-row samplers, and stop-sequence handling.
- Add Gemma 4 text model support, including shared-KV handling and MoE projection sanitization.
- Document continuous batching and add upstream-local tests for cache behavior, cancellation, row sampling, stop matching, and request admission.

## Performance
Gemma 4 26B-A4B-it-8bit, decode-only throughput on the same machine:
- Python `mlx_lm` BatchGenerator: B=1 77.2 tok/s, B=2 129.5 tok/s, B=4 180.8 tok/s
- Swift BatchGenerator/provider scheduler: B=1 ~79.6 tok/s, B=2 ~132.2 tok/s, B=4 ~179.8 tok/s

## Test plan
- [x] `uvx pre-commit run --all-files`
- [x] `xcodebuild test -scheme mlx-swift-lm-Package -destination 'platform=macOS'`
- [x] `scripts/verify-docs.sh`

### DIFF
--- a/Libraries/MLXLLM/Models/Gemma4Text.swift
+++ b/Libraries/MLXLLM/Models/Gemma4Text.swift
@@ -36,6 +36,12 @@ public struct Gemma4TextConfiguration: Codable, Sendable {
     var layerTypes: [String] = []
     var tieWordEmbeddings: Bool = true
 
+    // MoE (only set on the 26B-A4B variant; 2B/4B/31B are dense)
+    var enableMoeBlock: Bool = false
+    var numExperts: Int?
+    var topKExperts: Int?
+    var moeIntermediateSize: Int?
+
     // RoPE parameters (nested dict with full_attention/sliding_attention sub-configs)
     var ropeParameters: [String: [String: StringOrNumber]]?
 
@@ -69,6 +75,10 @@ public struct Gemma4TextConfiguration: Codable, Sendable {
         case layerTypes = "layer_types"
         case tieWordEmbeddings = "tie_word_embeddings"
         case ropeParameters = "rope_parameters"
+        case enableMoeBlock = "enable_moe_block"
+        case numExperts = "num_experts"
+        case topKExperts = "top_k_experts"
+        case moeIntermediateSize = "moe_intermediate_size"
     }
 
     public init(from decoder: Decoder) throws {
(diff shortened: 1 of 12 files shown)


ronaldmannak:
This is great! I was working on a parallel batching branch but yours is much cleaner. Mine adds a container-level batching surface and an optional upgrade from single-stream to batched (to avoid double-batching overhead). Happy to contribute those on top once your PR lands.
I did notice a few things re: cache and concurrency that will matter as soon as container APIs sit on top. Most notably that CacheList silently fall through to the generic BatchKVCache, so hybrid models end up with the wrong cache topology rather than an error. Since I already have the fixes for these in my branch, I'd like to open two PRs against your fork: one with those fixes, and a second adding the integration hooks needed for a container batching surface to land cleanly later. Does that work for you?

ronaldmannak:
I opened a [small stack of PRs](https://github.com/Layr-Labs/mlx-swift-lm/pulls) against the Layr repo to keep the review pieces separate:

Suggested flow, if this works for you:

1. Review/merge the Layr PR stack first.
2. Use that cleaned-up continuous batching branch as the basis for the upstream `ml-explore` PR.
3. After the upstream batching foundation is accepted, I can follow up with separate upstream PRs for:
   - LRU prompt cache
   - Container/scheduler-level batching APIs

Briefly, the PR stack covers:

- `GenerationBatchResponse` cleanup: removes public KV cache payloads from responses so the response value can be normally `Sendable`.
- Mutable batching state Sendability cleanup: removes `@unchecked Sendable` from `BatchGenerator`, `PromptProcessingBatch`, and `GenerationBatch`, and documents them as single-driver engine types. The idea is that a future scheduler/container actor should own these objects rather than exposing them as cross-task safe.
- Cache topology fixes: makes batched cache construction reject unsupported cache shapes explicitly and preserve supported composite cache topology.
- SSM/ragged prompt padding fixes: fixes the Arrays/Mamba padding path for ragged prompts.
- Input validation and small cleanup PRs: turn a few crash/hang cases into explicit validation and remove some dead/fragile batching state.
- Wired Memory boundary: adds an opt-in `drain(wiredMemoryTicket:onResponse:)` scope for bounded continuous-batching runs.

The only architectural suggestion I’d make is to keep `BatchGenerator` as a low-level, non-Sendable, single-driver primitive. For server/client-app usage with concurrent requests, I think the cleaner follow-up is a scheduler or `ModelContainer`-owned batching facade that serializes access internally and routes per-request streams.

Of course, up to Layr and the mlx maintainers whether this split and ordering makes sense. I’m happy to adjust the stack if you’d prefer a different review shape.

PR #267, opened 2026-05-06
Add Gemma 4 Multi-Token Prediction (MTP) drafter for speculative decoding

john-rocky:
## Summary

Port of Google's official Gemma 4 Multi-Token Prediction (MTP) drafter for speculative decoding (announced 2026-05-05; the announcement explicitly mentions MLX as a supported runtime). The drafter is tightly coupled to a `Gemma4Model` target: every drafter layer is kv-shared and reads K/V from the target's last full-attention and last sliding-attention layers (drafter has no KV cache of its own); the drafter's only recurrent state is the target's last hidden, projected through `post_projection`.

Drafter input each step is `concat([target_embed(token), last_hidden_state], axis=-1)`, projected to drafter hidden by `pre_projection`. RoPE positions are held constant at the bonus token's absolute position across all draft steps within a block. For E2B / E4B variants, `use_ordered_embeddings=true` and the LM head is a centroid-routed sparse softmax (`MaskedEmbedder`) that scores 2048 token clusters, materialises top-K (default 32) clusters' tokens and scatters logits back into the full vocab. 26B-A4B / 31B variants use a tied dense head.

`MTPSpeculativeTokenIterator` drives the round loop: drafter generates `block_size - 1` candidates from `(bonusToken, lastHidden)`; target verifies `[bonus, c_0, ..., c_{K-1}]` in a single forward; greedy compare → accept up to first mismatch → emit correction; trim cache for rejected verify tokens.

## Recent commits (2026-05-11)

Pipelining + Python-reference parity pass; full audit of `mlx_vlm.speculative.drafters.gemma4_assistant` against this Swift port. Three perf commits + one test commit:

- **`perf: pipeline MTP drafter loop with single CPU sync per round`** — drafter now returns a lazy `[1, blockSize-1]` MLXArray (`draftBlockArray`) instead of `[Int]`, the iterator chains it straight into the verify forward, mask construction is hoisted out of the per-step loop, and a greedy short-circuit in `MaskedEmbedder.argMaxToken` skips the `[B, L, 262 144]` vocab scatter for the (sole MVP) greedy path. The old debug `eval(vLogits, vHidden)` is gone — `argMax` now chains off the verify forward and can fuse with the lm_head matmul.
- **`perf: compile-fuse Gemma4 MLP geglu to match Python reference`** — Python's `geglu` is wrapped in `@partial(mx.compile, shapeless=True)`; the Swift port only had `geluApproximate` compiled, leaving the elementwise multiply outside compile. Wrapping the whole `gelu_approx(gate) * x` in `compile(shapeless: true)` keeps the gelu output in fp32 registers across the multiply, eliminating a layer-by-layer bf16 round-trip drift across the 35 decoder layers.
- **`test: add MLX-Swift L-invariance + drafter dtype microbench`** — 12 microbenchmarks pinning every isolated op (matmul fp32/bf16, quantizedMatmul, RoPE, RMSNorm, fast SDPA, KVCache.update) as row-0-invariant for L ∈ {2..16}, plus mid-/last-position SDPA drift, full `Gemma4TextModel` forward drift, and a MaskedEmbedder dtype check (sanitize keeps int32 `token_ordering`, argPartition keeps uint32 indices for vocab > 32 767, large token ids survive gather).

## Files

**Added**
- `Libraries/MLXLLM/Models/Gemma4Assistant.swift` (~620 lines) — `Gemma4AssistantConfiguration`, `MaskedEmbedder` sparse LM head + greedy `argMaxToken` short-circuit, bidirectional full / SWA mask helpers, `Gemma4AssistantDraftModel` (`bind`, `draftBlockArray` / `draftBlock`, `sanitize`, `load(from:)`).
- `Libraries/MLXLMCommon/MTPSpeculativeTokenIterator.swift` (~290 lines) — `MTPTargetModel` + `MTPDrafterModel` protocols (including the new lazy `draftBlockArray` hook), single-batch greedy verify-trim-advance round loop with one CPU sync per round.
- `Tests/MLXLMTests/Gemma4AssistantTests.swift` (139 lines) — config decode + sanitize tests (4 cases).
- `Tests/MLXLMTests/MatmulInvarianceTests.swift` (580 lines) — 12-case L-invariance + drafter dtype microbench.

**Modified additively (no behaviour change on existing target paths)**
- `Libraries/MLXLLM/Models/Gemma4Text.swift` — public MTP hooks (`forwardWithHidden`, `inputEmbeddings`, `inputEmbedScale`, `layerTypes`, `slidingWindow`, `backboneHiddenSize`, `lastFullAttentionLayerIndex`, `lastSlidingAttentionLayerIndex`); opt-in `kvSharedOnly` mode on `Gemma4Attention` / `Gemma4DecoderLayer` (default `false`) so drafter layers can skip K/V projections + norms; `Gemma4Attention` / `Gemma4MLP` / `Gemma4DecoderLayer` / `RMSNormNoScale` / `Gemma4PositionOffset` and the rope helpers demoted from `private` to default visibility so the drafter (in the same module) can reuse them; `compiledGemma4Geglu` introduced to match Python's `@partial(mx.compile)` fusion of `gelu_approx(gate) * x`.
- `Libraries/MLXLLM/Models/Gemma4.swift` — forwards MTP hooks from `Gemma4Model` to its inner `Gemma4TextModel`; `MTPTargetModel` conformance.

## Tests

```
xcodebuild test -scheme mlx-swift-lm-Package \
  -destination 'platform=macOS,arch=arm64' \
  -only-testing:MLXLMTests/Gemma4AssistantTests \
  -only-testing:MLXLMTests/MatmulInvarianceTests \
  -skipMacroValidation
```

16 / 16 cases pass in ~0.16 s on M4 Max — 4 `Gemma4AssistantTests` (config decode + sanitize) + 12 `MatmulInvarianceTests` (per-op L-invariance + drafter dtype). `swift build --target MLXLLM` clean (0 warnings).

## Bench (M4 Max 96 GB, this PR head as of 2026-05-11)

Single-batch, temperature 0, 1 warmup + 3 measured trials. `MTPSpeculativeTokenIterator` driven via a local `mtp-bench` subcommand on `llm-tool` (not part of this PR). Drafter is `mlx-community/gemma-4-E2B-it-assistant-bf16` (158 MB on disk).

### Target = `google/gemma-4-E2B-it` (HF native bf16, 9.6 GB)

| prompt class | block=4 | block=8 | block=16 (Repeat sweep) |
|---|---|---|---|
| `yes` × 30 (repetitive) | 1.66× ✅ byte-id | **2.97×** ✅ byte-id | **5.88×** ✅ byte-id |
| Code (Fibonacci) | 0.98× ✅ byte-id | 1.10× ✅ byte-id | — |
| Translation | 1.29× drift | 1.61× drift | — |
| JSON gen | 0.83× drift | 0.90× drift | — |
| Essay (free-form) | 0.71× drift | 0.73× drift | — |

Greedy is **byte-identical to the no-drafter baseline at temperature 0** on bf16 targets for repetitive / code prompts at block=4 / 8 / 16. Larger blocks pay off only when per-slot acceptance is high (low-entropy content); above block=16 the verify chunk's numerics on synthetic content begin to drift (Repeat block=20 stops being byte-identical), so block=16 is the practical greedy ceiling here.

### Target = `mlx-community/gemma-4-e2b-it-4bit` (1.5 GB, production iOS deploy candidate)

| prompt class | block=4 | block=8 |
|---|---|---|
| `yes` × 30 (repetitive) | 1.56× | 1.40× |
| **Translation (best)** | **1.62×** | 1.44× |
| JSON gen | 1.07× | 0.76× |
| Code (Fibonacci) | 0.95× | 0.72× |
| Essay (free-form) | 0.82× | 0.50× |

INT4 K cache quantisation noise reduces accept rate (consistent with the
[CoreML-LLM project's findings on the same drafter](https://github.com/anemll/coreml-llm) — chunk-1 K cache cos 0.978 vs fp16 0.993 cuts drafter accept ~half), so `block=4` is the production setup on INT4 (larger blocks let drift compound rather than amortise drafter cost). Byte-identical to the no-drafter baseline is generally lost on INT4 but recovers on Repeat at `block=4`.

### Speedup curve vs `block_size` (target = bf16, `yes` × 30, byte-identical at temp 0)

```
block=4   1.66×
block=6   2.34×
block=8   2.97×
block=10  3.26×
block=12  3.67×
block=16  5.88×   ← greedy ceiling
block=20  drift
```

### Drafter quantisation parity (vs fp32 reference, 5 synthetic inputs)

| drafter dtype | cos sim | max\|Δ\| | top-1 match | top-5 overlap | top-32 overlap |
|---|---|---|---|---|---|
| fp32 | 1.000000 | 0 | 100% | 100% | 100% |
| **bf16** (used here) | **1.000000** | 0 | **100%** | **100%** | **100%** |
| fp16 | 1.000000 | 6.1e-5 | 100% | 100% | 100% |
| int8 (Linears only, embed kept bf16) | 0.960 | 81 | 100% | 88% | 95% |
| int4 (Linears only) | 0.762 | 115 | **40%** | 24% | 26% |

→ bf16 ≡ fp32 byte-identical on the drafter's outputs; quantising the drafter further breaks the centroid path (`int4` top-1 only 40% vs fp32). The drafter is also memory-BW dominated by `embed_tokens` (134 MB of the 156 MB file), so quantising the Linears alone only saves 6% of file size with no measurable bench delta. **bf16 stays the production drafter.**

26B-A4B / 31B targets not benched here — download pending. The MLX VLM Python reference reports up to 3.94× / 2.29× respectively at batch=4. iPhone numbers will follow (a small `Applications/LLMMTPBench` SwiftUI harness exists locally; will land as a separate PR on `mlx-swift-examples` once we have phone data).

## Known limitations

- **B=1 single-batch only**, **greedy (temperature 0) only**. Stochastic sampling not yet supported.
- **Multimodal prefill** runs through the target unchanged; MTP only kicks in on the text-decode tail.
- **Output diverges from the no-drafter baseline on INT4 targets for high-entropy content**. Root cause is target-side INT4 K cache quantisation noise (chunk-1 K cache cos ≈ 0.978 vs fp16 ≈ 0.993), not an iterator bug — bf16 target stays byte-identical to baseline across all benched content types. The Python reference's "byte-identical at temp=0" claim is reproduced on bf16 targets; INT4 deploys keep greedy semantics but lose strict byte-identicalness on free-form prompts.
- **Speedup is content-dependent**. Repetitive / structured / code prompts get the biggest win; high-entropy free-form prompts (essay) can regress to 0.7–0.8× because the drafter can't predict enough tokens to amortise its overhead on the M4 Max GPU's already fast 78–167 tok/s baseline. This matches both the Google blog's "up to 2×" framing and the per-prompt curve [reported by mlx-vlm's HF ceiling bench](https://github.com/Blaizzy/mlx-vlm).

## References

- Google announcement: https://ai.google.dev/gemma/docs/mtp/overview
- Python port (full code audit done as part of this PR): [Blaizzy/mlx-vlm — `mlx_vlm/speculative/drafters/gemma4_assistant`](https://github.com/Blaizzy/mlx-vlm/tree/main/mlx_vlm/speculative/drafters/gemma4_assistant)
- mlx-community drafters: `gemma-4-{E2B,E4B,26B-A4B,31B}-it-assistant-bf16`

### DIFF
--- a/Libraries/MLXLLM/Models/Gemma4.swift
+++ b/Libraries/MLXLLM/Models/Gemma4.swift
@@ -105,3 +105,21 @@ extension Gemma4Model: LoRAModel {
         languageModel.loraLayers
     }
 }
+
+// MARK: - MTP drafter support
+
+extension Gemma4Model {
+    public var backboneHiddenSize: Int { languageModel.backboneHiddenSize }
+    public var inputEmbeddings: Embedding { languageModel.inputEmbeddings }
+    public var inputEmbedScale: Float { languageModel.inputEmbedScale }
+    public var layerTypes: [String] { languageModel.layerTypes }
+    public var slidingWindow: Int { languageModel.slidingWindow }
+    public var lastFullAttentionLayerIndex: Int? { languageModel.lastFullAttentionLayerIndex }
+    public var lastSlidingAttentionLayerIndex: Int? { languageModel.lastSlidingAttentionLayerIndex }
+
+    public func forwardWithHidden(_ inputs: MLXArray, cache: [KVCache]?) -> (
+        logits: MLXArray, hidden: MLXArray
+    ) {
+        languageModel.forwardWithHidden(inputs, cache: cache)
+    }
+}
--- a/Libraries/MLXLLM/Models/Gemma4Assistant.swift
+++ b/Libraries/MLXLLM/Models/Gemma4Assistant.swift
@@ -1,0 +1,507 @@
+//
+//  Gemma4Assistant.swift
+//  mlx-swift-lm
+//
+//  Multi-Token Prediction (MTP) drafter for Gemma 4 speculative decoding.
+//
+//  Reference: https://ai.google.dev/gemma/docs/mtp/overview
+//  Python port: Blaizzy/mlx-vlm/mlx_vlm/speculative/drafters/gemma4_assistant
+//
+
+import Foundation
+import MLX
+import MLXLMCommon
+import MLXNN
+
+// MARK: - Configuration
+
+/// Drafter configuration for Gemma 4 Multi-Token Prediction (assistant) models.
+///
+/// The drafter is a small (4-layer for E2B/E4B/26B-A4B/31B variants) model that
+/// shares the target's input embeddings and last-layer hidden state, and reads
+/// K/V from the target's last full-attention and last sliding-attention layers.
+public struct Gemma4AssistantConfiguration: Codable, Sendable {
+    public var modelType: String = "gemma4_assistant"
+    public var backboneHiddenSize: Int = 1536
+    public var useOrderedEmbeddings: Bool = false
+    public var numCentroids: Int = 2048
+    public var centroidIntermediateTopK: Int = 32
+    public var tieWordEmbeddings: Bool = true
+    public var blockSize: Int = 4
+    public var textConfig: Gemma4TextConfiguration
+
+    enum CodingKeys: String, CodingKey {
+        case modelType = "model_type"
+        case backboneHiddenSize = "backbone_hidden_size"
+        case useOrderedEmbeddings = "use_ordered_embeddings"
+        case numCentroids = "num_centroids"
+        case centroidIntermediateTopK = "centroid_intermediate_top_k"
+        case tieWordEmbeddings = "tie_word_embeddings"
+        case blockSize = "block_size"
+        case textConfig = "text_config"
+    }
+
+    public init(from decoder: Decoder) throws {
+        let container = try decoder.container(keyedBy: CodingKeys.self)
+        self.modelType =
+            try container.decodeIfPresent(String.self, forKey: .modelType) ?? "gemma4_assistant"
+        self.backboneHiddenSize =
+            try container.decodeIfPresent(Int.self, forKey: .backboneHiddenSize) ?? 1536
+        self.useOrderedEmbeddings =
+            try container.decodeIfPresent(Bool.self, forKey: .useOrderedEmbeddings) ?? false
+        self.numCentroids =
+            try container.decodeIfPresent(Int.self, forKey: .numCentroids) ?? 2048
+        self.centroidIntermediateTopK =
+            try container.decodeIfPresent(Int.self, forKey: .centroidIntermediateTopK) ?? 32
+        self.tieWordEmbeddings =
+            try container.decodeIfPresent(Bool.self, forKey: .tieWordEmbeddings) ?? true
+        self.blockSize =
+            try container.decodeIfPresent(Int.self, forKey: .blockSize) ?? 4
+
+        var textCfg: Gemma4TextConfiguration
+        if let decoded = try container.decodeIfPresent(
+            Gemma4TextConfiguration.self, forKey: .textConfig)
+        {
+            textCfg = decoded
+        } else {
+            textCfg = try Gemma4TextConfiguration(from: decoder)
+        }
+        // HF Gemma4AssistantConfig.__post_init__: when num_kv_shared_layers is 0
+        // the assistant shares K/V across all layers.
+        if textCfg.numKvSharedLayers == 0 {
+            textCfg.numKvSharedLayers = textCfg.numHiddenLayers
+        }
+        self.textConfig = textCfg
+    }
+}
+
+// MARK: - Bidirectional masks for the drafter forward
+
+/// Bi-directional full-attention mask.
+///
+/// Returns nil when no masking is needed (B=1 unbatched case ⇒ no-op).
+private func bidirectionalFullMask(
+    queryLen: Int,
+    kvLen: Int,
+    kvValidLen: Int?,
+    dtype: DType
+) -> MLXArray? {
+    guard let valid = kvValidLen, valid < kvLen else { return nil }
+    let kIdx = MLXArray(0 ..< kvLen)
+    let inside = kIdx .< valid
+    let bias = MLX.where(
+        inside,
+        MLXArray(0.0).asType(dtype),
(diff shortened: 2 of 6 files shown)


john-rocky:
Converting to draft. Two follow-ups before requesting review:

1. **Larger-target bench** — current numbers are on 4-bit E2B / E4B where drafter overhead dominates (1.0–1.2× at best). Will validate against `gemma-4-26B-A4B-it-bf16` and `gemma-4-31B-it-bf16` on M4 Max to replicate the Python reference's batch=1 speedups.

2. **Numerical drift in long generations** — the multi-position verify forward goes through `MLXFast.scaledDotProductAttention` with `.causal` mask, while the no-drafter baseline uses `.none` for single-token forwards. The two paths produce ε-level differences in bf16 accumulators that flip argmax on tokens with narrow top-1/top-2 logit margins (visible on the 4-bit E2B/E4B benched here, expected to be invisible on 26B/31B). Investigating whether unifying the mask path (`returnArray: true`) or casting verify SDPA to fp32 eliminates this on small targets.

Will mark ready when both land.

davidkoski:
FYI see https://github.com/ml-explore/mlx-swift-lm/pull/308#discussion_r3290355015

john-rocky:
Thanks for the pointer, @davidkoski — and noted on keeping model-weight-dependent tests in IntegrationTesting rather than committing fixtures.

I also see #308 now covers this same Gemma 4 MTP work with more complete integration coverage. I'll leave #267 as a draft and defer to whichever direction you'd prefer — happy to close in favor of #308, or fold in anything useful from here.

davidkoski:
Closing in favor of #308 -- follow-on PRs are welcome!

PR #268, opened 2026-05-07
fix: avoid shared CIContext in media processing

Meapri:
## Summary

- Removes the shared file-level `CIContext` from `MediaProcessing`.
- Creates a local `CIContext` inside `asMLXArray(...)` before rendering.
- Avoids Swift 6.1 strict concurrency failures caused by a non-Sendable global `CIContext` while preserving existing rendering behavior.

## Test Plan

- Verified downstream SwiftPM build from `Meapri/mlx-vlm` against this fork branch.
- The previous strict-concurrency error was:
  - `let context is not concurrency-safe because non-Sendable type CIContext may have shared mutable state`.

### DIFF
--- a/Libraries/MLXVLM/MediaProcessing.swift
+++ b/Libraries/MLXVLM/MediaProcessing.swift
@@ -1,7 +1,7 @@
 // Copyright © 2024 Apple Inc.
 
 import AVFoundation
-import CoreImage.CIFilterBuiltins
+@preconcurrency import CoreImage.CIFilterBuiltins
 import MLX
 import MLXLMCommon
 
@@ -12,6 +12,8 @@ public struct ProcessedFrames {
     public let timestamps: [CMTime]
     public let totalDuration: CMTime
 }
+
+private let context = CIContext()
 
 /// Collection of methods for processing media (images, video, etc.).
 ///
@@ -167,12 +169,11 @@ public enum MediaProcessing {
         let bytesPerRow = w * bytesPerPixel
 
         var data = Data(count: w * h * bytesPerPixel)
-        let renderContext = CIContext()
         data.withUnsafeMutableBytes { ptr in
-            renderContext.render(
+            context.render(
                 image, toBitmap: ptr.baseAddress!, rowBytes: bytesPerRow, bounds: image.extent,
                 format: format, colorSpace: colorSpace)
-            renderContext.clearCaches()
+            context.clearCaches()
         }
 
         var array = MLXArray(data, [h, w, 4], type: Float32.self)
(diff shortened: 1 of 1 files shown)


davidkoski:
I am curious about this one:  `CIContext` _is_ `Sendable`:

- https://developer.apple.com/documentation/coreimage/cicontext#conforms-to

What issue are you seeing here?  This is already building using swift 6.1 and I haven't seen an error.

Meapri:
Thanks for checking this.

You're right that the current Apple documentation lists `CIContext` as conforming to `Sendable`.

The reason I opened this was a downstream SwiftPM build that depends on `mlx-swift-lm`. In that environment, the file-scope shared context triggered Swift's strict-concurrency checking with a diagnostic like:

```text
MediaProcessing.swift:16:13: error: let 'context' is not concurrency-safe because non-'Sendable' type 'CIContext' may have shared mutable state
```

That downstream workflow runs on a `macos-15` GitHub Actions runner and reports:

```text
Apple Swift version 6.1.2 (swiftlang-6.1.2.1.2 clang-1700.0.13.5)
```

So this may be SDK/toolchain-version dependent, or specific to how the package is being built as a dependency. I agree the docs make this less straightforward than I initially thought. The intent here was just to avoid the file-scope shared value that triggered the diagnostic downstream, without changing the public API.

If this does not reproduce in this repo's supported build setup, I'm happy to close or rework the PR.

davidkoski:
Ah, that makes sense -- the older SDK may not have it marked as Sendable.  Huh, that is an interesting one.  We could mark the import as preconcurrency.  I think that would fix it for the older SDK, but I wonder if that hides issues.

Our runners are newer SDKs, so I can't verify it there.  I think I might have a mac with an older macOS where I could try it.

Meapri:
That sounds reasonable. I can try the `@preconcurrency import CoreImage` approach as an alternative if you prefer that over moving the `CIContext` into function scope.

My only concern is similar to yours: `@preconcurrency` would silence concurrency checking for the CoreImage import more broadly, while the current patch keeps the workaround local to the shared file-scope value that triggered the downstream diagnostic.

If you're able to test on an older SDK, that would be very helpful. I'm also happy to update this PR to whichever approach you think fits the project better.

davidkoski:
It would be great if there were something like `#if sdk(> 26)` but there isn't.  I think the best we can do is:

```swift
@preconcurrency import CoreImage.CIFilterBuiltins
```

I verified that it worked with Xcode 16.4 and current Xcode seems fine with it too.

Meapri:
Thanks for checking this on Xcode 16.4 and the current Xcode.

That approach makes sense to me. I updated the PR to use:

```swift
@preconcurrency import CoreImage.CIFilterBuiltins
```

and restored the shared `CIContext`, so the change now targets the SDK compatibility issue directly rather than changing the render context lifetime.

I could not run a local Swift build in this environment because `swift` is not installed here, but the patch is intentionally limited to the import/workaround you suggested.

davidkoski (approved):
Thank you!

PR #271, opened 2026-05-07
Propagate errors through tokenizer and generation APIs

DePasqualeOrg:
> Note: 735 of the 1163 new lines in this PR are tests.

The `Tokenizer` protocol's `encode` and `decode` methods are currently non-throwing, which forces conformers to choose between crashing and silently corrupting the output – neither of which is ideal. Swift Transformers' `PreTrainedTokenizer` does both: [`encode` force-unwraps vocab lookups](https://github.com/huggingface/swift-transformers/blob/349a7ce54ccb8ebe4bc10c2022e9806f01adb7c6/Sources/Tokenizers/Tokenizer.swift#L643), so an out-of-vocabulary token kills the process, and [`decode` `compactMap`s them](https://github.com/huggingface/swift-transformers/blob/349a7ce54ccb8ebe4bc10c2022e9806f01adb7c6/Sources/Tokenizers/Tokenizer.swift#L670), so an unknown ID disappears with no signal to the caller. [Swift Tokenizers](https://github.com/DePasqualeOrg/swift-tokenizers), which wraps the Rust `tokenizers` library, already handles this correctly, but errors can't currently propagate across the MLX bridge.

The non-throwing behavior is at odds with how other tokenizer libraries handle these operations. Rust `tokenizers` surfaces failures from [`encode`](https://github.com/huggingface/tokenizers/blob/bcdd25b97fcd78549903082ecf3ddd87d42c456b/tokenizers/src/tokenizer/mod.rs#L837) and [`decode`](https://github.com/huggingface/tokenizers/blob/bcdd25b97fcd78549903082ecf3ddd87d42c456b/tokenizers/src/tokenizer/mod.rs#L901) as `Result`s, and [the Python binding](https://github.com/huggingface/tokenizers/blob/bcdd25b97fcd78549903082ecf3ddd87d42c456b/bindings/python/src/tokenizer.rs#L1473) preserves that shape via PyO3's `PyResult`, so failures cross into Python as exceptions.

This PR makes `encode` and `decode` throwing, bringing their behavior into line with the broader ecosystem.

## Related changes in streaming detokenizer

`NaiveStreamingDetokenizer.next()` currently calls `tokenizer.decode` and conforms to `IteratorProtocol`, whose `next()` cannot throw. Once `decode` can throw, the only ways to keep the iterator shape are `try?` (which re-introduces the silent-drop pattern this PR removes) or `try!` (which crashes). The current detokenizer isn't really an iterator anyway: `next()` returns `nil` to mean "feed me more", not "end of sequence", so a `for-in` loop would exit at the first multi-byte UTF-8 fragment, before the actual end of the stream. Rust's [`DecodeStream`](https://github.com/huggingface/tokenizers/blob/bcdd25b97fcd78549903082ecf3ddd87d42c456b/tokenizers/src/tokenizer/mod.rs#L1027) skips `Iterator` for the same reason.

To rectify this, `NaiveStreamingDetokenizer` is replaced with a `StreamingDetokenizer` class whose `consume(_:)` method is throwing. The replacement uses the algorithm Rust `tokenizers` exposes as [`step_decode_stream`](https://github.com/huggingface/tokenizers/blob/bcdd25b97fcd78549903082ecf3ddd87d42c456b/tokenizers/src/tokenizer/mod.rs#L1094). The replacement differs from the naive version in two important ways:

- The new version's memory stays roughly constant during generation. The naive version cleared its token buffer only when a chunk ended with a newline, so memory grew across long stretches of newline-free output.
- If the decoder produces inconsistent output (a chunk that doesn't follow from text already emitted), the new version detects this and throws. The naive version silently emitted the corrupted result.

The tool-call handler in `Evaluate.swift` recovers from the new error class by rebuilding the detokenizer and retrying once – [the same recovery shape vLLM uses](https://github.com/vllm-project/vllm/blob/66d1cc0c77628e36df8e5e245c116e6aac3045fb/vllm/v1/engine/detokenizer.py#L220) for the same error.

Some tokenizers rewrite their `decode` output (whitespace fixups, contractions), which can retroactively change earlier text and break streaming. Tokenizers that conform to the new `StreamingDecodeTokenizer` protocol expose a cleanup-free `rawDecode`, which is used by `StreamingDetokenizer` when available.

Using this separate protocol would allow for logging a warning (currently not implemented) when the behavior degrades with libraries such as Swift Transformers, which don't offer `rawDecode`. Swift Tokenizers already behaves correctly.

## Seeding the streaming detokenizer at the prompt boundary

SentencePiece-style tokenizers strip a leading space when decoding a sequence that begins with a metaspace marker. A streaming detokenizer with no prior context decodes the first generated token as sequence-start, so the space at the prompt/generated boundary is silently dropped (`"...isParis"` instead of `"...is Paris"`). Python mlx-lm has the same gap; its [`_maybe_trim_space`](https://github.com/ml-explore/mlx-lm/blob/df1d3f3c9a7aae402dcbb8f41d4c36bcc13a50ae/mlx_lm/tokenizer_utils.py#L195) papers over the symptom with a per-model strip-leading-space flag.

`StreamingDetokenizer` now takes an `initialTokenIds` parameter, and the engine seeds it with the last 7 prompt tokens, [matching vLLM](https://github.com/vllm-project/vllm/blob/66d1cc0c77628e36df8e5e245c116e6aac3045fb/vllm/tokenizers/detokenizer_utils.py#L73). On `invalidStreamingPrefix` recovery the seed is dropped, [matching vLLM](https://github.com/vllm-project/vllm/blob/66d1cc0c77628e36df8e5e245c116e6aac3045fb/vllm/v1/engine/detokenizer.py#L240): once text has been emitted, the boundary is no longer adjacent. `generateTask` and `generateTokenTask` now take `promptTokenIds: [Int]` in place of `promptTokenCount: Int`; high-level `generate(input:)` callers see no change.

## Throwing generation stream

`generate(...)` and the related `generateTask`, `generateTokens`, and `generateTokenTask` overloads previously returned `AsyncStream<Generation>` (or `AsyncStream<TokenGeneration>`). With the throwing tokenizer API and a throwing `StreamingDetokenizer`, errors from deeper in the pipeline could not propagate to the consumer through a non-throwing stream. Those return types have been converted to `AsyncThrowingStream<Generation, Error>` (and the token-ID equivalent), so any error thrown during generation finishes the stream with that error, and the consumer can handle it.

vLLM's [`AsyncLLM.generate`](https://github.com/vllm-project/vllm/blob/66d1cc0c77628e36df8e5e245c116e6aac3045fb/vllm/v1/engine/async_llm.py#L524) takes the same approach: errors from the engine are pushed onto the [output queue](https://github.com/vllm-project/vllm/blob/66d1cc0c77628e36df8e5e245c116e6aac3045fb/vllm/v1/engine/output_processor.py#L45) and raised when the consumer reads them.

## Migration

The migration is mechanical: add `try` to `tokenizer.encode`/`decode` call sites, and `for try await` in place of `for await` on generation streams. Users who implement a lower-level token iteration loop replace `NaiveStreamingDetokenizer`, `append(token:)`, and `next()` with `tokenizer.streamingDetokenizer()` and `consume(_:)`. Callers of `generateTask` and `generateTokenTask` rename `promptTokenCount: input.text.tokens.size` to `promptTokenIds: input.text.tokens.asArray(Int.self)`. Callers that want the old swallow-and-continue behavior can wrap in `try?`, but this is an explicit choice now.

These are breaking changes that will require a major version bump, but the migration for users is simple and contained: in most cases they'll just need to add `try`.

## Testing

`StreamingDetokenizerTests` exercises the new algorithm against a mock tokenizer, with cases modeled on equivalents in [`huggingface/tokenizers`](https://github.com/huggingface/tokenizers/blob/bcdd25b97fcd78549903082ecf3ddd87d42c456b/bindings/python/tests/bindings/test_tokenizer.py#L418), [vLLM](https://github.com/vllm-project/vllm/blob/66d1cc0c77628e36df8e5e245c116e6aac3045fb/tests/v1/engine/test_fast_incdec_prefix_err.py), and [`transformers`](https://github.com/huggingface/transformers/blob/main/tests/generation/test_streamers.py).

### DIFF
--- a/Libraries/MLXLMCommon/Evaluate.swift
+++ b/Libraries/MLXLMCommon/Evaluate.swift
@@ -1317,7 +1317,7 @@ public func generate(
 /// * Important: if the stream is terminated early (e.g. break from the loop) computation will continue
 /// using the model, parameters, KVCache, etc. for some time (typically a few ms).  This is typically OK for
 /// one-shot calls, but for "chat session" type calls consider using
-/// ``generateTask(promptTokenCount:modelConfiguration:tokenizer:iterator:wiredMemoryTicket:)``
+/// ``generateTask(promptTokenIds:modelConfiguration:tokenizer:iterator:wiredMemoryTicket:)``
 /// so that the end of the generation task can be observed.
 ///
 /// - Parameters:
@@ -1363,7 +1363,7 @@ public func generate(
     let iterator = try TokenIterator(
         input: input, model: context.model, cache: cache, parameters: parameters)
     let (stream, _) = generateTask(
-        promptTokenCount: input.text.tokens.size,
+        promptTokenIds: input.text.tokens.asArray(Int.self),
         modelConfiguration: context.configuration,
         tokenizer: context.tokenizer,
         iterator: iterator,
@@ -1443,7 +1443,8 @@ public func generate(
         wiredMemoryTicket: wiredMemoryTicket,
         handler: TextToolTokenLoopHandler(
             tokenizer: context.tokenizer,
-            format: context.configuration.toolCallFormat ?? .json
+            format: context.configuration.toolCallFormat ?? .json,
+            promptTokenIds: input.text.tokens.asArray(Int.self)
         )
     )
     return stream
@@ -1459,7 +1460,7 @@ public func generate(
     wiredMemoryTicket: WiredMemoryTicket? = nil
 ) -> AsyncThrowingStream<Generation, Error> {
     let (stream, _) = generateTask(
-        promptTokenCount: input.text.tokens.size,
+        promptTokenIds: input.text.tokens.asArray(Int.self),
         modelConfiguration: context.configuration,
         tokenizer: context.tokenizer,
         iterator: iterator,
@@ -1475,28 +1476,31 @@ public func generate(
 /// the `task` to observe when the use of the parameters is complete.
 ///
 /// - Parameters:
-///   - promptTokenCount: number of tokens in the prompt
+///   - promptTokenIds: prompt token IDs, used for stats and to seed the
+///     streaming detokenizer at the prompt/generated boundary so the first
+///     generated chunk decodes correctly for SentencePiece/metaspace tokenizers.
 ///   - modelConfiguration: model configuration (for EOS/extra EOS tokens and tool-call format)
 ///   - tokenizer: tokenizer (for EOS id, unknown token id, and detokenization)
 ///   - iterator: token iterator
 ///   - wiredMemoryTicket: Optional wired memory ticket for policy-based coordination.
 /// - Returns: An `AsyncThrowingStream` that emits `Generation` values and a `Task`
 public func generateTask(
-    promptTokenCount: Int,
+    promptTokenIds: [Int],
     modelConfiguration: ModelConfiguration,
     tokenizer: Tokenizer,
     iterator: consuming TokenIterator,
     wiredMemoryTicket: WiredMemoryTicket? = nil
 ) -> (AsyncThrowingStream<Generation, Error>, Task<Void, Never>) {
     generateLoopTask(
-        promptTokenCount: promptTokenCount,
+        promptTokenCount: promptTokenIds.count,
         modelConfiguration: modelConfiguration,
         tokenizer: tokenizer,
         iterator: iterator,
         wiredMemoryTicket: wiredMemoryTicket,
         handler: TextToolTokenLoopHandler(
             tokenizer: tokenizer,
-            format: modelConfiguration.toolCallFormat ?? .json
+            format: modelConfiguration.toolCallFormat ?? .json,
+            promptTokenIds: promptTokenIds
         )
     )
 }
@@ -1527,7 +1531,7 @@ public func generateTokens(
     let iterator = try TokenIterator(
         input: input, model: context.model, cache: cache, parameters: parameters)
     let (stream, _) = generateTokenTask(
-        promptTokenCount: input.text.tokens.size,
+        promptTokenIds: input.text.tokens.asArray(Int.self),
         modelConfiguration: context.configuration,
         tokenizer: context.tokenizer,
         iterator: iterator,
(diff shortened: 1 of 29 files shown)


Libraries/MLXLMCommon/Evaluate.swift
```diff
@@ -1928,46 +1940,78 @@ public enum TokenGeneration: Sendable {
...
     mutating func onGenerationEnd(
-        emit: (sending Output) -> AsyncStream<Output>.Continuation.YieldResult
+        emit: (sending Output) -> AsyncThrowingStream<Output, Error>.Continuation.YieldResult
     )
 
     func infoEvent(_ info: GenerateCompletionInfo) -> Output
 }
 
-private struct TextToolTokenLoopHandler: TokenLoopHandler, @unchecked Sendable {
+internal struct TextToolTokenLoopHandler: TokenLoopHandler, @unchecked Sendable {
```
davidkoski: See #272 -- we can eliminate the `@unchecked Sendable` (different PR -- I just noticed the issue here)

Libraries/MLXLMCommon/Tokenizer.swift
```diff
@@ -53,62 +53,32 @@ extension Tokenizer {
...
+            "Streaming detokenizer prefix invariant violated for token \(tokenId): expected prefix \(expectedPrefix.debugDescription), got \(actualString.debugDescription)."
         }
     }
 }
-
-public protocol StreamingDetokenizer: IteratorProtocol<String> {
-    mutating func append(token: Int)
-}
-
-public struct NaiveStreamingDetokenizer: StreamingDetokenizer {
```
davidkoski, line 71: This is a breaking change -- I wonder if we should leave a deprecated stub behind?

There are _some_ used of it:

- https://github.com/search?q=NaiveStreamingDetokenizer+language%3Aswift&type=code

I don't think it is enough for a major version bump but I think we can probably make something with the same name that is backward compatible.  Eh, though we don't have the protocol any more and the new one is a final class.

I wonder if we can just typalias it (deprecated) and add a deprecated method to the `StreamingDetokenizer`?

e.g. to handle https://github.com/scouzi1966/maclocal-api/blob/4cbbe063bd80fb576e72a96dfffcf7b078fcc474/Sources/MacLocalAPI/Models/BatchScheduler.swift#L1291

Most of the hits are on forks of mlx-swift-examples / mlx-swift-lm, but there are a few with actual calls

DePasqualeOrg: I marked it as unavailable so that users get a message that helps them migrate if they try to compile code that uses it. There are actually a lot of deprecations that should probably be removed in the next major-version release.

Libraries/MLXLMCommon/Evaluate.swift
DePasqualeOrg: Done.

PR #280, opened 2026-05-10
fix ropeOffset use in Gemma3NText

davidkoski:
## Proposed changes

See #234 -- match python pattern and pass offset into applyRotaryPosition.  This is critical for Gemma3NText which has a pattern of apply/update/apply where the second apply uses a _different_ offset.

FYI @ronaldmannak 

## Checklist

### DIFF
--- a/Libraries/MLXLLM/Models/Gemma3nText.swift
+++ b/Libraries/MLXLLM/Models/Gemma3nText.swift
@@ -14,52 +14,11 @@ import MLXNN
 
 // MARK: - Configuration
 
-/// A type that can be decoded as either a single Int or an array of Ints.
-/// This is needed because some models (like Gemma 3n) specify intermediate_size
-/// as a per-layer array, while others use a single value.
-public struct IntOrArray: Codable {
-    public let values: [Int]
-
-    public init(from decoder: Decoder) throws {
-        let container = try decoder.singleValueContainer()
-        if let array = try? container.decode([Int].self) {
-            self.values = array
-        } else if let single = try? container.decode(Int.self) {
-            self.values = [single]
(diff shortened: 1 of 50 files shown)


Libraries/MLXLLM/Models/Gemma3nText.swift
```diff
@@ -14,52 +14,11 @@ import MLXNN
 
 // MARK: - Configuration
 
-/// A type that can be decoded as either a single Int or an array of Ints.
-/// This is needed because some models (like Gemma 3n) specify intermediate_size
-/// as a per-layer array, while others use a single value.
-public struct IntOrArray: Codable {
```
davidkoski, line 20: Not related to the base issue, just noticed duplicated code when I was looking at the config.

Libraries/MLXLLM/Models/Gemma3nText.swift
```diff
@@ -263,6 +255,7 @@ class Gemma3nAttention: Module {
         queries = queries.reshaped(B, L, -1, headDim)
         queries = qNorm(queries)
 
+        let offset = cache?.ropeOffset
```
davidkoski, line 258: This is the key fix

Libraries/MLXLLM/Models/Gemma4Text.swift
```diff
@@ -175,33 +175,6 @@ private class ScaledLinear: Module {
     }
 }
 
-private enum Gemma4PositionOffset {
```
davidkoski, line 178: This is subsumed by RoPEOffset

angeloskath (approved):
👍

PR #281, opened 2026-05-11
fix build

davidkoski:
## Proposed changes

Fix build problem from #164 -- that was based on an older mlx-swift-lm.  CI passed and there were no conflicts, but main no longer built.  `ModelTypeRegistry` is now `ModelTypeRegistry<T>` for use in loading embedding models.

## Checklist

### DIFF
--- a/Libraries/MLXLMCommon/ParoQuant/RotateQuantizedLinear.swift
+++ b/Libraries/MLXLMCommon/ParoQuant/RotateQuantizedLinear.swift
@@ -6,7 +6,7 @@ import MLXNN
 
 /// Pairwise Givens rotation kernel for Metal (Apple Silicon).
 /// Template parameters are substituted at compile time.
-nonisolated private func metalSource(
+private func metalSource(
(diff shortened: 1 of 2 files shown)


Libraries/MLXLMCommon/ParoQuant/RotateQuantizedLinear.swift
```diff
@@ -137,7 +137,7 @@ nonisolated private func packPairs(_ pairs: MLXArray, groupSize: Int) -> MLXArra
 ///
 /// Rotation is applied to activations at runtime via a Metal kernel, preserving
 /// the quantization-friendly properties of the original weights.
-nonisolated open class RotateQuantizedLinear: QuantizedLinear {
+open class RotateQuantizedLinear: QuantizedLinear {
```
davidkoski, line 140: I don't understand how this would have made it through CI ... oh maybe it was an earlier Package.swift (e.g. older swift version).

angeloskath (approved):
👍

Issue #279, opened 2026-05-09
Register gemma4_assistant model type so Gemma 4 MTP drafters load through LLMModelFactory


brdcastro:
## Context

Google's Gemma 4 MTP drafters (released 2026-05-05, [blog post](https://blog.google/innovation-and-ai/technology/developers-tools/multi-token-prediction-gemma-4/)) ship as standalone Hugging Face checkpoints that work with the `generate(..., draftModel:, numDraftTokens:)` speculative-decoding API in `MLXLMCommon/Evaluate.swift`. The drafters are published in MLX format under [`mlx-community/gemma-4-assistant-mtp`](https://huggingface.co/collections/mlx-community/gemma-4-assistant-mtp), one per Gemma 4 size:

- `mlx-community/gemma-4-E2B-it-assistant-bf16` (78 MB)
- `mlx-community/gemma-4-E4B-it-assistant-bf16` (78.8 MB)
- `mlx-community/gemma-4-26B-A4B-it-assistant-bf16` (~400 MB)
- `mlx-community/gemma-4-31B-it-assistant-bf16` (~500 MB)

These are the recommended drafters for the Gemma 4 family — pairing them with the matching target model is what unlocks the 1.7×–3× decode speedup the blog demonstrates.

## The gap

Each drafter checkpoint declares:

```json
{
  "architectures": ["Gemma4AssistantForCausalLM"],
  "model_type": "gemma4_assistant",
  "backbone_hidden_size": 2560,
  "num_centroids": 2048,
  "centroid_intermediate_top_k": 32,
  "use_ordered_embeddings": true,
  "text_config": { "hidden_size": 256, "num_hidden_layers": 4, ... }
}
```

`LLMTypeRegistry.shared` in `Libraries/MLXLLM/LLMModelFactory.swift` registers `gemma4` and `gemma4_text` but not `gemma4_assistant`. Loading any of the four drafters above through `LLMModelFactory.loadContainer` therefore fails — even though the rest of the speculative-decoding plumbing (`SpeculativeTokenIterator`, `generate(..., draftModel:, numDraftTokens:)`) is already in place.

The Hugging Face transformers reference implementation lives in [`src/transformers/models/gemma4_assistant/`](https://github.com/huggingface/transformers/tree/main/src/transformers/models/gemma4_assistant) (`configuration_gemma4_assistant.py`, `modeling_gemma4_assistant.py`).

## Asks

[...]

PR #283, opened 2026-05-12
models should not mutate state during eval

### LINKED ISSUE #157: [BUG] Qwen 3.5 VLM model crashes on subsequent requests with manual KV cache prefix matching

davidkoski:
## Proposed changes

- fixes #157
- the LMOutputState (currently unused) is updated to hold key/value pairs
- that allows models to produce cached values for later use

## Checklist

### DIFF
--- a/Libraries/MLXLLM/Models/GPTOSS.swift
+++ b/Libraries/MLXLLM/Models/GPTOSS.swift
@@ -161,6 +161,7 @@ class AttentionBlock: Module {
     let numKeyValueHeads: Int
     let numKeyValueGroups: Int
     let smScale: Float
+    var sinksActive: Bool
 
     @ParameterInfo(key: "sinks") var sinks: MLXArray
     @ModuleInfo(key: "q_proj") var qProj: Linear
@@ -169,7 +170,6 @@ class AttentionBlock: Module {
     @ModuleInfo(key: "o_proj") var oProj: Linear
 
     let rope: YarnRoPE
-    private var cachedSinksActive: Bool?
 
     public init(_ config: GPTOSSConfiguration) {
         self.headDim = config.headDim
@@ -178,6 +178,8 @@ class AttentionBlock: Module {
         self.numKeyValueGroups = config.attentionHeads / config.kvHeads
 
         _sinks.wrappedValue = zeros([config.attentionHeads])
+        self.sinksActive = false
+
         _qProj.wrappedValue = Linear(
             config.hiddenSize, config.attentionHeads * config.headDim, bias: true)
         _kProj.wrappedValue = Linear(config.hiddenSize, config.kvHeads * config.headDim, bias: true)
@@ -205,6 +207,20 @@ class AttentionBlock: Module {
         }
     }
 
+    open override func update(
+        parameters: ModuleParameters, verify: VerifyUpdate, path: [String] = [],
+        modulePath: [String] = []
+    ) throws -> Self {
+        try super.update(
+            parameters: parameters, verify: verify, path: path, modulePath: modulePath)
+
+        if parameters["sinks"] != nil {
+            self.sinksActive = (sinks * sinks).max().item(Float.self) > 0
+        }
+
+        return self
+    }
+
     public func callAsFunction(
         _ x: MLXArray,
         mask: MLXFast.ScaledDotProductAttentionMaskMode,
@@ -216,13 +232,6 @@ class AttentionBlock: Module {
         var q = qProj(x).reshaped(B, L, -1, D).swappedAxes(1, 2)
         var k = kProj(x).reshaped(B, L, -1, D).swappedAxes(1, 2)
         var v = vProj(x).reshaped(B, L, -1, D).swappedAxes(1, 2)
-        let sinksActive =
-            cachedSinksActive
-            ?? {
-                let active = (sinks * sinks).max().item(Float.self) > 0
-                cachedSinksActive = active
-                return active
-            }()
 
         // Quantized cache path
         if let qcache = cache as? QuantizedKVCacheProtocol {
--- a/Libraries/MLXLLM/Models/GPTOSS.swift
+++ b/Libraries/MLXLLM/Models/GPTOSS.swift
@@ -161,7 +161,6 @@ class AttentionBlock: Module {
     let numKeyValueHeads: Int
     let numKeyValueGroups: Int
     let smScale: Float
-    var sinksActive: Bool
 
     @ParameterInfo(key: "sinks") var sinks: MLXArray
     @ModuleInfo(key: "q_proj") var qProj: Linear
@@ -178,7 +177,6 @@ class AttentionBlock: Module {
         self.numKeyValueGroups = config.attentionHeads / config.kvHeads
 
         _sinks.wrappedValue = zeros([config.attentionHeads])
-        self.sinksActive = false
 
         _qProj.wrappedValue = Linear(
             config.hiddenSize, config.attentionHeads * config.headDim, bias: true)
@@ -207,20 +205,6 @@ class AttentionBlock: Module {
         }
     }
 
-    open override func update(
-        parameters: ModuleParameters, verify: VerifyUpdate, path: [String] = [],
-        modulePath: [String] = []
-    ) throws -> Self {
-        try super.update(
-            parameters: parameters, verify: verify, path: path, modulePath: modulePath)
-
-        if parameters["sinks"] != nil {
-            self.sinksActive = (sinks * sinks).max().item(Float.self) > 0
-        }
-
-        return self
-    }
-
     public func callAsFunction(
         _ x: MLXArray,
         mask: MLXFast.ScaledDotProductAttentionMaskMode,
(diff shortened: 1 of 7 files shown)


davidkoski:
@angelsbrood FYI -- I haven't tested it yet, but something like this

angelsbrood:
Thanks for the FYI David. I've been digging into gemma4_assistant for #279, and this looks like the mechanism the MTP drafter needs: it would let the target share its last-layer hidden states and per-layer-type K/V with the drafter through the iterator. Two questions before I go further:

1. My usage would be cross-model— target writes state keys on its forward, drafter reads them on its own forward. GlmOcr and Qwen35 in #283 look intra-model (each reads keys it wrote itself). The mechanism handles cross-model fine, but I wanted to flag it in case you'd want a different API surface for that, or bake constraints into #283 before it lands.
2. Reading the canonical Python (HF modeling_gemma4_assistant.py and Blaizzy's mlx-vlm port), the drafter is structurally not a LanguageModel— forward takes (inputs_embeds, shared_kv_states, position_ids), it loads via a separate load_drafter(kind="mtp"), and it exposes a K-step draft_block the iterator calls directly. My instinct is a parallel MTPDrafterModel protocol and DrafterTypeRegistry rather than fitting it into LanguageModel. Does that sound right, or would you steer it differently?

Holding off until I hear back. Thanks again!

davidkoski:
> 1. My usage would be cross-model— target writes state keys on its forward, drafter reads them on its own forward. GlmOcr and Qwen35 in [models should not mutate state during eval #283](https://github.com/ml-explore/mlx-swift-lm/pull/283) look intra-model (each reads keys it wrote itself). The mechanism handles cross-model fine, but I wanted to flag it in case you'd want a different API surface for that, or bake constraints into [models should not mutate state during eval #283](https://github.com/ml-explore/mlx-swift-lm/pull/283) before it lands.

I think that would be fine -- the keys would just ned to be more visible than `private`.  These states that I was adding were internal to the model so private makes sense here.  I think returning out-of-band data this way (via public or at least package scoped) keys makes sense.

> 2. Reading the canonical Python (HF modeling_gemma4_assistant.py and Blaizzy's mlx-vlm port), the drafter is structurally not a LanguageModel— forward takes (inputs_embeds, shared_kv_states, position_ids), it loads via a separate load_drafter(kind="mtp"), and it exposes a K-step draft_block the iterator calls directly. My instinct is a parallel MTPDrafterModel protocol and DrafterTypeRegistry rather than fitting it into LanguageModel. Does that sound right, or would you steer it differently?

Possibly -- I know the other draft models were done as normal language models (but I think they were from the mlx-lm side).  If it does need to be a new type of model, I think that `BaseLanguageModel` provides the required loading support (see how it is used by `EmbeddingModel`).  So this is a possibility.

angelsbrood:
Read through `BaseLanguageModel` and how `EmbeddingModel` uses it— yes, this works better than the parallel-registry shape I was tentatively imagining. The drafter would conform to `BaseLanguageModel` and gain its own task protocol the way `EmbeddingModel` does:
```
public protocol MTPDrafterModel: BaseLanguageModel {
    func bind(target: any LanguageModel)
    func draftBlock(
        lastToken: MLXArray, hidden: MLXArray,
        sharedKV: [String: (MLXArray, MLXArray)],
        positionIds: MLXArray,
        blockSize: Int, sampler: Sampler
    ) -> MLXArray
}
```
This composes with #283: the target writes `gemma4Assistant.lastHiddenStates` and `gemma4Assistant.sharedKVStates` keys (public per your "out-of-band" guidance), the MTP iterator reads them off `LMOutput.state` and threads them as method args into `draftBlock` per round. The drafter doesn't carry transient round-state as ivars— only bind-time setup (target's embed table reference, scale) which is set once and never mutated during eval. Aligns with #283's invariant.

One question: where would you want `MTPDrafterModel` to live? Three options I see: in `MLXLMCommon` alongside `LanguageModel`, in `MLXLLM`, or in a new `MLXDrafters` library parallel to `MLXEmbedders`. The last is the most literal parallel to the `EmbeddingModel` precedent, but it does add a library for what's currently a single model type. Happy to go where you'd prefer.

Will draft the PR with this shape once you weigh in on placement (and once #283 lands, since the keys depend on it).

davidkoski:
I think MLXLMCommon (the protocol) and then the implementations in MLXLLM and MLXVLM.

angelsbrood:
@davidkoski Heads-up that I've opened a draft PR at #293 consuming this work: three public `LMOutput.Key` declarations (emit flag, last hidden state, shared K/V) threading state across the target → iterator → drafter boundary for Gemma 4 MTP speculative decoding.

One observation in case it's useful for the review: the fixed `LanguageModel.callAsFunction(_:cache:state:)` signature made `LMOutput.State` the natural carrier for the emit-flag control signal as well as the tensor payloads. What would have been two mechanisms (state-bag for tensors, function parameter for the flag) ended up as one. The typed `Key<T>` subscript pattern carries both cleanly.

Will leave the draft alone until #283 merges and we rebase onto `main`. Happy to share what surfaced during use if it's useful to your review push. Thanks for getting the foundation up.

Libraries/MLXLLM/Models/GPTOSS.swift
```diff
@@ -216,13 +232,6 @@ class AttentionBlock: Module {
...
-        let sinksActive =
-            cachedSinksActive
-            ?? {
-                let active = (sinks * sinks).max().item(Float.self) > 0
-                cachedSinksActive = active
-                return active
-            }()
 
         // Quantized cache path
         if let qcache = cache as? QuantizedKVCacheProtocol {
```
angeloskath, line 221: This is probably a bigger change than this was meant to be but the following code should probably be unified to a call to a `scaledDotProductAttention` that checks internally if the cache is of `QuantizedKVCacheProtocol` rather than duplicating the logic below.

Regarding active sinks, is there a GPT OSS that doesn't have active sinks? In python we ignore that possibility 🤷‍♂️ We could also just add sinks to the `quantizedScaledDotProductAttention` it's straightforward.

Libraries/MLXLLM/Models/GPTOSS.swift
```diff
@@ -205,6 +207,20 @@ class AttentionBlock: Module {
...
     }
 
+    open override func update(
+        parameters: ModuleParameters, verify: VerifyUpdate, path: [String] = [],
+        modulePath: [String] = []
+    ) throws -> Self {
+        try super.update(
+            parameters: parameters, verify: verify, path: path, modulePath: modulePath)
+
+        if parameters["sinks"] != nil {
```
angeloskath: As I mention below we kind of ignore that possibility in python, not sure how important it is.

angeloskath (approved):
👍

Libraries/MLXLLM/Models/GPTOSS.swift
davidkoski: Yeah, good question -- maybe I should sync up this implementation with mlx-lm.  Either way it needs a change.

davidkoski: Yeah, simpler to make it just match mlx-lm

davidkoski:
@angelsbrood FYI merged.  Finally have CI back in working order so I am going to start going through these again.

PR #285, opened 2026-05-13
run the doc linter on linux

davidkoski:
## Proposed changes

run the doc linter on linux -- the macOS runner has limited inventory and I don't think it is needed

## Checklist


davidkoski:
Huh, the doc lint requires a full build, which is still iffy on linux.

PR #288, opened 2026-05-15
feat: add pre-quantized weight loading and minicpmv4_6 model support

aooyoo:
- Load.swift: detect pre-quantized weights (.scales keys), create QuantizedLinear directly with MXFP8 weight/scales/biases, strip VLM-only keys
- LLMModelFactory.swift: register minicpmv4_6 → Qwen35Model

## Proposed changes

## Checklist

### DIFF
--- a/Libraries/MLXLLM/LLMModelFactory.swift
+++ b/Libraries/MLXLLM/LLMModelFactory.swift
@@ -564,6 +564,7 @@ public final class LLMModelFactory: GenericModelFactory {
 
         try loadWeights(
             modelDirectory: modelDirectory, model: model,
+            quantization: baseConfig.quantization,
             perLayerQuantization: baseConfig.perLayerQuantization)
 
         let tokenizer = try await tokenizerTask
--- a/Libraries/MLXLLM/LLMModelFactory.swift
+++ b/Libraries/MLXLLM/LLMModelFactory.swift
@@ -41,6 +41,7 @@ public enum LLMTypeRegistry {
         "qwen3_5_moe": create(Qwen35Configuration.self, Qwen35MoEModel.init),
         "qwen3_5_text": create(Qwen35TextConfiguration.self, Qwen35TextModel.init),
         "minicpm": create(MiniCPMConfiguration.self, MiniCPMModel.init),
+        "minicpmv4_6": create(Qwen35Configuration.self, Qwen35Model.init),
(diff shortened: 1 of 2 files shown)


Libraries/MLXLMCommon/Load.swift
```diff
@@ -36,8 +36,12 @@ public func loadWeights(
...
     weights = model.sanitize(weights: weights, metadata: metadata)
 
-    // quantize if needed
-    if quantization != nil || perLayerQuantization != nil {
+    // handle pre-quantized weights (models where safetensors already contain
+    // quantized weight + scales + biases) vs. float-weight models that need
+    // load-time quantization
+    if hasPreQuantizedWeights(weights) {
+        try loadPreQuantizedWeights(model: model, weights: &weights, quantization: quantization)
+    } else if quantization != nil || perLayerQuantization != nil {
```
davidkoski, line 44: I think the two changes in LLMModelFactory make sense, but I am not sure what problem this is fixing -- the existing code will walk the model and configure it per the quantization configuration.

The added code (`loadPreQuantizedWeights()`) doesn't handle per-layer quantization and duplicates what the code already does:

```swift
        quantize(model: model) { path, module in
            if weights["\(path).scales"] != nil {
                if let perLayerQuantization {
                    return perLayerQuantization.quantization(layer: path)?.asTuple
                } else {
                    return quantization?.asTuple
                }
            } else {
                return nil
            }
        }
```

What error were you seeing?

davidkoski (changes requested):
The changes in LLMModelFactory make sense but please explain the error that the changes in Load.swift are addressing -- that should already be handled with existing code.

Libraries/MLXLMCommon/Load.swift
davidkoski: Perhaps Qwen35Model.sanitize needs to be augmented?  Or a new model for `minicpmv4_6` that handles something in the weights?

Libraries/MLXLLM/LLMModelFactory.swift
```diff
@@ -563,6 +564,7 @@ public final class LLMModelFactory: GenericModelFactory {
 
         try loadWeights(
             modelDirectory: modelDirectory, model: model,
+            quantization: baseConfig.quantization,
```
davidkoski, line 567: Thinking more about it, this isn't needed -- the parameter was there for backward compatibility.  `PerLayerQuantization` already encapsulates the top level quantization configuration.

PR #290, opened 2026-05-15
Fix Gemma 4 vision pooler kernel derivation

### LINKED ISSUE #289: Gemma 4 vision pooler derives kernel from unpadded patch count
First time poster, long time admirer of this project and all contributors.

**Summary:**
The Gemma4VisionPooler kernel-derivation formula uses the wrong patch count and ends up dropping roughly half of every image — specifically the lower portion of the patch grid — before the language model ever sees it. In a downstream Swift app, this surfaced as the model consistently missing content in the bottom half of images.

**Details:**
`Gemma4VisionPooler.callAsFunction` in `Libraries/MLXVLM/Models/Gemma4.swift` derives the pooling kernel from the real (un-padded) patch count rather than the padded sequence length. At the default 280 soft-token budget, this can recover `kernel = 2` for normal processed image sizes. HuggingFace's reference implementation and the model's own `pooling_kernel_size` config field both expect `3`. Real-patch positions whose flat-kernel bucket index lands beyond `pooledLength` are then mapped to all-zero one-hot rows by `gemma4OneHot` and contribute nothing to the einsum.

**Invariant:**
Per HuggingFace's image processor:
[image_processing_gemma4.py](https://github.com/huggingface/transformers/blob/main/src/transformers/models/gemma4/image_processing_gemma4.py)

```
max_patches = max_soft_tokens * pooling_kernel_size**2
```

The padded patch tensor fed into the pooler has length `max_patches`. Recovering `pooling_kernel_size` from this padded length requires:

```
kernel = floor(sqrt(max_patches / max_soft_tokens)) = pooling_kernel_size
```

HuggingFace's `modeling_gemma4.py` does exactly this — it derives `k` from `int((input_seq_len // length) ** 0.5)` where `input_seq_len` is the padded hidden-state length.

The mlx-swift-lm pooler instead uses `validCount / length`, where `validCount` is the real (un-padded) patch count from the resized image. For aspect-ratio-preserving resizes where `validCount < max_patches`, the recovered kernel is too small.

**Reproducer:**
A `768×768` image is the canonical aligned default for Gemma 4 (`patch_size × pooling_kernel_size = 16 × 3 = 48`, and 768 is divisible by 48). For that image:

- `validCount = (768 / 16)² = 2304`
- `paddedPatchCount = 280 × 3² = 2520`

Current code at:
[Gemma4.swift](https://github.com/ml-explore/mlx-swift-lm/blob/main/Libraries/MLXVLM/Models/Gemma4.swift)

```swift
let kernel = Int(sqrt(Double(max(1, validCount / max(length, 1)))))
//         = Int(sqrt(2304 / 280)) = Int(2.87) = 2
```

Expected (matching the HF formula):

```swift
let kernel = Int(sqrt(Double(max(1, pooledHiddenStates.dim(1) / max(length, 1)))))
//         = Int(sqrt(2520 / 280)) = Int(3.0) = 3
```

With `kernel = 2`, `flatKernel = x' + (maxX / kernel) * y'` spans a 24×24 coarse grid (576 cells). Because `gemma4OneHot` is called with `numClasses = pooledLength = 280`, bucket indices 280…575 become all-zero rows and contribute nothing to the einsum. Those buckets correspond to real patches from the image, not padding (the einsum input is `pooledHiddenStates[0..., ..<validCount, 0...]`, i.e., the real-patch prefix only).

**Proposed fix:**
A one-line change to derive the kernel from the padded sequence length:

```diff
- let kernel = Int(sqrt(Double(max(1, validCount / max(length, 1)))))
+ let kernel = Int(sqrt(Double(max(1, pooledHiddenStates.dim(1) / max(length, 1)))))
```

I can open a focused PR with the one-line fix plus a small unit test if this framing looks right. The branch already extracts the formula into a helper (`gemma4VisionPoolingKernel(paddedPatchCount:outputLength:)`) and adds a `Gemma4VisionPoolingTests` suite asserting `kernel = 3` across the documented soft-token budgets `{70, 140, 280, 560, 1120}`.

To be clear, that test is a unit-level invariant check on the kernel-derivation formula; my end-to-end runtime testing on `mlx-community/gemma-4-e4b-it-4bit` was done at the default budget of 280. Non-default budgets exercise additional code paths (resize sizing, prompt expansion) and are out of scope.

**Verification:**
Ran `xcodebuild test -scheme mlx-swift-lm-Package -destination 'platform=macOS'` with the fix applied: **TEST SUCCEEDED**, all 64 tests across 5 suites pass, including the new `Gemma4VisionPoolingTests` suite.

**Observed qualitative impact:**
On `mlx-community/gemma-4-e4b-it-4bit` at default budget, comparing the same engine binary with only this one-line change:
- A bar chart with 8 bars: pre-fix model counted 3, post-fix counted all 8.
- A vertical iOS-app screenshot with two rows of poster cards: pre-fix missed the entire bottom row and misread a visible title as a different similar-looking word; post-fix read every visible title correctly.
- A wide iPad screenshot of a scheduling app: pre-fix named 3 of the visible staff entries, post-fix named 7.

joshuaejackson:
Gemma4VisionPooler.callAsFunction derived the kernel from the real (un-padded) patch count rather than the padded sequence length. At the default 280 soft-token budget, this can recover kernel=2 for typical image sizes, while HuggingFace's reference implementation and the model's pooling_kernel_size config both expect 3. Real-patch positions whose flat-kernel bucket index exceeds pooledLength then map to all-zero one-hot rows by gemma4OneHot and contribute nothing to the einsum output.

Extract the formula into gemma4VisionPoolingKernel(paddedPatchCount: outputLength:) and derive the kernel from pooledHiddenStates.dim(1) (= outputLength * pool^2) so it returns pool=3 for every supported budget.

Add Gemma4VisionPoolingTests asserting kernel=3 for all five documented budgets {70, 140, 280, 560, 1120}, with a regression test documenting the prior formula's behavior.

## Proposed changes

Fixes #289. Aligns the Gemma 4 vision pooler's kernel derivation with HuggingFace's reference image processor so the recovered kernel matches the model's `pooling_kernel_size = 3` at every supported soft-token budget.

## Checklist

### DIFF
--- a/Libraries/MLXVLM/Models/Gemma4.swift
+++ b/Libraries/MLXVLM/Models/Gemma4.swift
@@ -82,6 +82,24 @@ private func gemma4OneHot(_ indices: MLXArray, numClasses: Int) -> MLXArray {
     expandedDimensions(indices, axis: -1) .== MLXArray(0 ..< numClasses)
 }
 
+/// Average-pool kernel for Gemma 4's vision pooler.
+///
+/// The padded patch tensor has length
+/// `paddedPatchCount = outputLength × pool²` where `pool` is the
+/// model's `pooling_kernel_size`. Recovering `pool` from these
+/// two values yields `floor(sqrt(paddedPatchCount / outputLength))`.
+///
+/// Matches HuggingFace's reference image processor (see
+/// `image_processing_gemma4.py`: `max_patches = max_soft_tokens *
+/// pooling_kernel_size**2`).
+internal func gemma4VisionPoolingKernel(
+    paddedPatchCount: Int, outputLength: Int
+) -> Int {
+    let safeLength = max(outputLength, 1)
+    let ratio = max(1, paddedPatchCount / safeLength)
+    return Int(sqrt(Double(ratio)))
+}
+
 private func gemma4RotateHalf(_ x: MLXArray) -> MLXArray {
     let half = x.shape[x.shape.count - 1] / 2
     let x1 = x[.ellipsis, ..<half]
@@ -1485,7 +1503,8 @@ private final class Gemma4VisionPooler: Module {
 
         let actualPositions = patchPositions[0, ..<validCount]
         let maxX = Int(actualPositions[0..., 0].max().item(Int32.self)) + 1
-        let kernel = Int(sqrt(Double(max(1, validCount / max(length, 1)))))
+        let kernel = gemma4VisionPoolingKernel(
+            paddedPatchCount: pooledHiddenStates.dim(1), outputLength: length)
         let divisor = max(kernel * kernel, 1)
         let pooledLength = max(length, 1)
 
--- a/Tests/MLXLMTests/Gemma4PoolingTests.swift
+++ b/Tests/MLXLMTests/Gemma4PoolingTests.swift
@@ -1,0 +1,84 @@
+// Copyright © 2026 Apple Inc.
+
+import Foundation
+import Testing
+
+@testable import MLXVLM
+
+/// Unit tests for `gemma4VisionPoolingKernel`. Pins the recovered
+/// kernel to Gemma 4's `pooling_kernel_size` (3) for every
+/// supported soft-token budget.
+struct Gemma4VisionPoolingTests {
+
+    /// `pooling_kernel_size` for Gemma 4 (per HF + the model
+    /// config's `pooling_kernel_size` field).
+    private static let expectedPoolingKernel = 3
+
+    @Test(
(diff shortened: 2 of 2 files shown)


davidkoski (approved):
Nice find and fix!

davidkoski:
Waiting on CI -- infrastructure issue

joshuaejackson:
Thanks! First PR here. Wide open to feedback on format.

davidkoski:
rebase to pick up CI fix

PR #291, opened 2026-05-16
Fix/isolated mlx test streams (fixes unit tests on main)

christophsturm:
## Proposed changes

Swift Testing can run suites concurrently. The previous .serialized trait only serialized tests within RoPEApplicationTests, so it did not prevent gemma3nAttentionTest from running alongside KV cache and speculative decoding tests.

Those tests all enqueue and evaluate MLX work through StreamOrDevice.default. Without a task-local override, that resolves to the shared default GPU stream, so separate tests can interact through the same underlying command-buffer lifecycle and trip Metal's 'Completed handler provided after commit call' assertion.

Run the RoPE test under Stream.withNewDefaultStream and synchronize that scoped stream before returning. This isolates the MLX stream state for the test while preserving Swift Testing parallelism instead of serializing unrelated tests.

## Checklist

- [X] I have read the [CONTRIBUTING](https://github.com/ml-explore/mlx/blob/main/CONTRIBUTING.md) document

### DIFF
--- a/Tests/MLXLMTests/RoPEApplicationTests.swift
+++ b/Tests/MLXLMTests/RoPEApplicationTests.swift
@@ -51,9 +51,6 @@ struct RoPEApplicationTests {
             // different offsets for keys/queries.
             let match = allClose(outputBatch, outputSeq, atol: 1e-4)
             eval(match)
-            print(outputBatch)
-            print(outputSeq)
-            print(abs(outputSeq - outputBatch))
             #expect(match.item(Bool.self))
         }
     }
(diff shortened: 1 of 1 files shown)


davidkoski:
Can you give more info on the failure?  Separate streams should not be required, even if tests are running concurrently.

I _was_ able to trigger a failure, but this is in mlx-swift: https://github.com/ml-explore/mlx-swift/pull/410

christophsturm:
you are right! it seems the prints in the test are the problem so the real fix is your https://github.com/ml-explore/mlx-swift/pull/410 (and in the meantime removing the prints would also fix it)

PR #269, opened 2026-05-07
Add Gemma 4 MTP assistant model and native MTP speculative decoding

joelnishanth:
## Summary

Adds native Multi-Token Prediction (MTP) speculative decoding support for Gemma 4 models. The MTP drafter uses the model's own assistant heads (~78M params for E4B) that share KV cache and embeddings with the backbone, enabling 2-3x inference speedup with zero quality degradation.

**Key changes:**

- **`Gemma4Assistant.swift`** — Full `Gemma4AssistantForCausalLM` model definition: 4-layer transformer drafter with sliding+full attention, pre/post projection, and centroid-masked embedder for efficient vocabulary narrowing (262K vocab to ~4K candidates during drafting)
- **`MTPGeneration.swift`** — MTP infrastructure: `MTPBackboneModel` / `MTPDraftModel` protocols, `MTPSpeculativeTokenIterator` (follows `SpeculativeTokenIterator` pattern), `MTPVerifier` with Leviathan-Chen rejection sampling (fp32 ratios to avoid BF16 underflow), and `generate()` overloads
- **`LLMModelFactory.swift`** — Registers `"gemma4_assistant"` model type in the type registry
- **`Gemma4Text.swift`** — Adds `MTPBackboneModel` conformance to `Gemma4TextModel`, exposing pre-logit hidden states via `forwardMTP()`
- **`Gemma4.swift`** — Forwards `MTPBackboneModel` through the `Gemma4Model` wrapper
- **`Evaluate.swift`** — Widens access on `generateLoopTask` and `TokenLoopHandler` from `private` to `internal` so `MTPGeneration.swift` can reuse the shared generation infrastructure

## Supported Model Pairs

| Backbone | MTP Assistant | Params | Target |
|----------|--------------|--------|--------|
| gemma-4-e2b-it (4-bit) | gemma-4-E2B-it-assistant-bf16 | 78M | 8GB constrained |
| gemma-4-e4b-it (4-bit) | gemma-4-E4B-it-assistant-bf16 | 78.8M | 16GB+ standard |
| gemma-4-26B-A4B-it (4-bit) | gemma-4-26B-A4B-it-assistant-bf16 | 0.4B | 32GB+ |
| gemma-4-31B-it (4-bit) | gemma-4-31B-it-assistant-bf16 | 0.5B | 64GB+ |

## Architecture

[...]

PR #293, opened 2026-05-17
Add Gemma 4 MTP speculative decoding

angelsbrood:
## Proposed changes

Adds Gemma 4 Multi-Token Prediction (MTP) speculative decoding. Introduces an `MTPDrafterModel` protocol in MLXLMCommon refining `BaseLanguageModel`, registers `gemma4_assistant` as the first concrete drafter implementation, threads cross-model state through three public `LMOutput.Key` declarations, and adds an `MTPSpeculativeTokenIterator` parallel to the existing `SpeculativeTokenIterator`. Tracks #279.

Opening as a draft against `lmoutput-state` because the implementation depends on #283's typed-key `LMOutput.State` API. Will rebase to `main` once #283 merges; review requests deferred until then.

## Architecture

The drafter does not fit `LanguageModel`. Its forward signature differs in fundamental ways: it has no own KV cache; it cross-attends to the target's last full-attention and last sliding-attention layers via a `[String: (MLXArray, MLXArray)]` shared K/V dict; its position IDs are constant within a drafting round; and it requires a bind-time reference to the target to borrow input embeddings. The new `MTPDrafterModel` protocol in `Libraries/MLXLMCommon/MTPDrafterModel.swift` refines `BaseLanguageModel` (so weight loading and sanitize hooks are inherited) and defines its own forward signature. Drafters are loaded via `MTPDrafterModelFactory`, registered into `MTPDrafterTypeRegistry`, and driven by `MTPSpeculativeTokenIterator`.

Cross-model state passthrough uses three public `LMOutput.Key` declarations in MLXLMCommon: `mtpEmitFlagKey: Bool`, `mtpLastHiddenStatesKey: MLXArray`, and `mtpSharedKVStatesKey: [String: (MLXArray, MLXArray)]`. The iterator pre-populates the emit-flag key on every main-model call to signal that the target should write drafter state; the target's top-level `Gemma4.callAsFunction(_:cache:state:)` override reads the flag and threads it through to the backbone, which writes the last-layer hidden state and per-`layer_type` shared K/V into the state on the way out. The emit flag is itself a `LMOutput.Key<Bool>` rather than a function parameter because `LanguageModel.callAsFunction(_:cache:state:)` has a fixed protocol signature; threading the flag through `LMOutput.State` unifies the mechanism — the state-bag carries both tensor payloads and the control signal that triggers them.

`MTPSpeculativeTokenIterator` includes a sticky passthrough fallback. When the target's emit hook returns `nil` shared K/V — which happens once quantization onset converts the regular K/V cache to `QuantizedKVCache`, or any other condition that breaks the regular-K/V invariant — the iterator transparently switches to single-token main-model generation for the remainder of the stream. The drafter is never invoked again on that iterator instance. MTP with mid-stream quantization onset degrades gracefully rather than crashing or producing incorrect outputs; bf16-drafter-against-quantized-target streams give up the speedup but preserve correctness.

On the target side, the drafter reuses existing `Gemma4TextAttention`, `Gemma4TextDecoderLayer`, and norm building blocks rather than duplicating them. Visibility on those types and a `kvSharedOnly: Bool` flag on `Gemma4TextAttention.init` (which skips construction of `kProj`/`vProj`/`kNorm`/`vNorm`, declared Optional `@ModuleInfo`) let the drafter compose its own decoder layers from existing target-side modules. `Gemma4.swift` diff: +53/−26. The non-`kvSharedOnly` regression suite is bit-identical before and after; existing Gemma 4 generation paths are not perturbed.

## Relationship to existing work

#269 targets the same scope — Gemma 4 MTP speculative decoding — and predates this PR by ~9 days. The two implementations diverge on three substantive axes worth surfacing for reviewers:

1. **Foundation.** #269 widens `generateLoopTask` and `TokenLoopHandler` from `private` to `internal` to reuse the existing generation infrastructure, and exposes pre-logit hidden states via a new `forwardMTP()` method on `Gemma4TextModel`. This PR consumes the `LMOutput.State` API from #283 instead, threading drafter state through the dictionary-backed typed-key state-bag (three `LMOutput.Key` declarations, no per-call ivar mutation on the model). The "models should not mutate state during eval" invariant of #283 is honored throughout.
2. **Verification.** This PR is fixture-anchored against `mlx-vlm@d49d428` with a four-rung ladder culminating in byte-identical token-ID parity at temperature=0. #269 lists a test plan as an unchecked checklist.
3. **Quantization onset.** This PR includes the sticky passthrough fallback described above, which preserves correctness when the main model's cache transitions to quantized mid-stream. #269 does not address this case.

The masked-embedder path is the dimension on which #269 has more implemented than this PR; conversely this PR carries the empty-checkpoint `fatalError` stub honestly rather than shipping centroid logic without a verification fixture.

The overlap is real and acknowledged here so reviewers don't have to discover it independently.

## Verification

Fixture-anchored against [`mlx-vlm@d49d428`](https://github.com/Blaizzy/mlx-vlm/tree/d49d428). Reference dataset published at [`angelsbrood/gemma4-mtp-fixtures`](https://huggingface.co/datasets/angelsbrood/gemma4-mtp-fixtures) — 43 files / ~20 MB, generated against target=`mlx-community/gemma-4-31b-it-8bit` and drafter=`mlx-community/gemma-4-31B-it-assistant-bf16`. The same 43-file fixture set is vendored in-repo at `tools/fixtures/` so the Rung 1–4 tests run offline against a known-good frozen reference; schema (`FIXTURE-SCHEMA.md`), manifest (`FIXTURE-MANIFEST.json`), and the generation tooling sit alongside. Regeneration is deterministic (verified across runs).

Verification proceeds in four rungs, each gating the next:

- **Rung 1 — Load / shape.** Drafter checkpoint loads with no missing or unexpected weight keys after sanitize. Weight-key inventory in `RESEARCH-NOTES.md` matches the checkpoint verbatim across both 31B and 26B-A4B drafter siblings (48 keys each, identical layout, shapes differ).
- **Rung 2 — Per-layer activations.** Drafter forward with synthetic inputs produces activations matching Python fixtures at `atol=1e-3, rtol=1e-3`.
- **Rung 3 — Logit parity.** Greedy `draftBlock` logits match Python fixtures at the same tolerance.
- **Rung 4 — Token-for-token cross-runtime parity.** Greedy `draftBlock` at temperature=0 produces byte-identical token-ID sequences to the Python reference across three block sizes (2, 4, 6).

All four rungs pass in Xcode against the fixture set. `swift test` CLI is not a working verification surface for this repository — `CONTRIBUTING.md` notes "swift test does not work yet — use xcodebuild instead," and the metallib initialization failure that affects CLI runs predates this branch and is not introduced here. Full-target acceptance-rate and quantization-onset integration tests are placeholdered in `MLXLMTests` with `Issue.record` + early-return because they require `MLXHuggingFace`'s `#huggingFaceTokenizerLoader()` which is not in `MLXLMTests`' dependency closure; the full end-to-end measurement will live in a downstream integration test.

## Files

**New:**

- `Libraries/MLXLMCommon/MTPDrafterModel.swift` — `MTPDrafterModel` protocol, `MTPDrafterContext`, `MTPDrafterContainer`, three public `LMOutput.Key` declarations
- `Libraries/MLXLMCommon/MTPSpeculativeTokenIterator.swift` — iterator with sticky passthrough fallback
- `Libraries/MLXLMCommon/MTPDrafterModelFactory.swift` — `MTPDrafterTypeRegistry`, `MTPDrafterRegistry`, `MTPDrafterModelFactory`
- `Libraries/MLXLMCommon/BidirectionalMasks.swift` — `createBidirectionalMask` and `createBidirectionalSlidingWindowMask` helpers
- `Libraries/MLXVLM/Models/Gemma4Assistant.swift` — `Gemma4AssistantConfiguration`, `Gemma4AssistantMaskedEmbedder`, `Gemma4AssistantDraftInner`, `Gemma4AssistantDraftModel`
- `Libraries/MLXVLM/Gemma4AssistantRegistration.swift` — `registerModelType` call site for `"gemma4_assistant"`
- `Tests/MLXLMTests/BidirectionalMasksTests.swift`, `MTPDrafterModelTests.swift`, `Gemma4AssistantDraftModelTests.swift`, `Gemma4EmitDrafterStateTests.swift`, `MTPDrafterModelFactoryTests.swift`, `MTPSpeculativeTokenIteratorTests.swift`, `MTPGenerateAPITests.swift`, `MTPRung4TokenParityTests.swift`, `MTPAcceptanceRateTests.swift`, `MTPQuantizationOnsetTests.swift`
- `RESEARCH-NOTES.md` — architectural Q&A with file:line citations against canonical Python (HF Transformers `gemma4_assistant`, mlx-vlm drafter), plus full 48-key weight inventories for both reference drafter checkpoints
- `tools/inspect_drafter_layout.py`, `tools/generate_mtp_fixtures.py`
- `tools/fixtures/` — 43 fixture data files (~20 MB) plus `FIXTURE-SCHEMA.md` and `FIXTURE-MANIFEST.json`. Vendored in-repo so the Rung 1–4 tests run offline against a frozen reference; content is byte-identical to the [`angelsbrood/gemma4-mtp-fixtures`](https://huggingface.co/datasets/angelsbrood/gemma4-mtp-fixtures) HF dataset for cross-verification.

**Modified:**

- `Libraries/MLXLMCommon/Evaluate.swift` — two new `generate(...)` overloads (one returning `AsyncStream<Generation>`, one returning `AsyncStream<TokenGeneration>`), taking an `mtpDrafter` parameter, `blockSize: Int = 4` default
- `Libraries/MLXVLM/Models/Gemma4.swift` — visibility bumps on internal building blocks (the cascade reached 16 symbols once `Gemma4TextLanguageModel`, `Gemma4TextBackbone`, the `languageModel` accessor, and four MLP/router types were pulled along by what they reference), `kvSharedOnly: Bool` flag on `Gemma4TextAttention.init`, emit hooks in `Gemma4TextBackbone.callAsFunction` and `Gemma4TextLanguageModel.callAsFunction`, new top-level `Gemma4.callAsFunction(_:cache:state:)` override (+53/−26)

## Supported model pairs

| Target | Drafter | Status in this PR |
| --- | --- | --- |
| `mlx-community/gemma-4-31b-it-8bit` | `mlx-community/gemma-4-31B-it-assistant-bf16` | Primary; fixture-anchored |
| `mlx-community/gemma-4-26B-A4B-it-bf16` | `mlx-community/gemma-4-26B-A4B-it-assistant-bf16` | Architecturally supported; not fixture-anchored |
| `mlx-community/gemma-4-31B-it-bf16` | `mlx-community/gemma-4-31B-it-assistant-bf16` | Architecturally supported; would re-tighten Rung 2/3 tolerances |

E-series targets (`gemma-4-E2B-it`, `gemma-4-E4B-it`) are architecturally supported but not exercised by the fixture set in this PR; both have published `-assistant-bf16` siblings on the `mlx-community` org.

## Known limitations

- **`Gemma4AssistantMaskedEmbedder.callAsFunction` is a `fatalError` stub.** Both reference checkpoints have `use_ordered_embeddings=false` and ship with zero `masked_embedding.*` weight keys, so the centroid-narrowing path is never exercised by the fixture set or by either currently-published reference checkpoint. The module is declared correctly for weight loading (so a future checkpoint with the flag enabled would load with no key mismatches); the forward implementation can land in a follow-up when a verifiable checkpoint surfaces. Choosing `fatalError` over an unverified implementation matches the verification posture of the rest of the PR.
- **Quantized drafter weights** deferred — bf16 only initially. Drafters are small enough (~78M params for E-series, ~0.4B–0.5B for the larger pairs) that this is acceptable for a first PR.
- **Batched MTP** deferred — single-batch only. mlx-vlm's `_speculative_walk_batch` is out of scope.
- **Audio path** unaffected — the target's audio components remain filtered in sanitize today; this drafter is text-only.
- **Quantized cache during a drafting stream** is covered by the sticky passthrough fallback described above. The drafter is bypassed when fallback activates; correctness is preserved, the MTP speedup is not.
- **Acceptance-rate floor and full-target quantization-onset integration tests** are placeholdered in `MLXLMTests` pending `MLXHuggingFace` in the test target's dependency closure. The passthrough fallback logic itself is unit-tested with synthetic state.

## Checklist

- [x] CONTRIBUTING document read
- [x] `pre-commit run --all-files` passes
- [x] Tests added (ten new test files; Rungs 1–4 covered; full-target integration tests placeholdered with documented gating)
- [ ] Documentation updates — happy to add a README section or wherever convention dictates if reviewers want one

## Acknowledgments

Thanks to @davidkoski for #283 — the typed-key `LMOutput.State` API is the foundation this PR builds on; without it, the cross-model state passthrough would have required something much uglier.

## References

- Tracking issue: #279 (@brdcastro)
- Foundation PR: #283
- Parallel PR: #269 (see "Relationship to existing work" above)
- Fixture dataset: [`angelsbrood/gemma4-mtp-fixtures`](https://huggingface.co/datasets/angelsbrood/gemma4-mtp-fixtures)
- Canonical Python references: HF Transformers `gemma4_assistant`, mlx-vlm `speculative/drafters/gemma4_assistant` at SHA `d49d428`
- Cross-runtime: vLLM PR [#41745](https://github.com/vllm-project/vllm/pull/41745)
- Google announcement: [blog post](https://blog.google/innovation-and-ai/technology/developers-tools/multi-token-prediction-gemma-4/)


angelsbrood:
For future readers: this PR was auto-closed when `lmoutput-state` was deleted on merge of #283. Continued at #308.

PR #194, opened 2026-04-07
Add Gemma 3n E4B audio encoder (Conformer) support

vahsaechao:
## Proposed changes

Adds the audio encoder for Gemma 3n E4B multimodal models. 12-layer USM-style Conformer that converts mel spectrograms into embeddings for the language model.

This is the first audio modality support in mlx-swift-lm. The encoder was ported from the [HuggingFace Python implementation](https://huggingface.co/google/gemma-3n-E4B-it) and tested on iphone 17 pro max.

**Architecture:**
```
mel [1, T, 128] → SubSampleConvProjection (4x reduction)
  → 12 Conformer blocks (chunked local attention, depthwise conv1d, cumulative group norm)
  → Temporal reduction (4x)
  → AudioEmbedder (1536 → 2048)
  → LM token stream
```

**New files:**
- `Gemma3nAudio.swift` - Full Conformer encoder: chunked local attention, depthwise conv1d, cumulative group norm, temporal reduction, sub sample conv projection
- `Gemma3nAudioConfig.swift` - 28 audio encoder configuration parameters with Codable
- `Gemma3nVLM.swift` - Top level VLM wrapper with audio embedding injection and multimodal embedder
- `Gemma3nAudioTests.swift` - Configuration decoding and encoder shape tests
- Factory registration in `VLMModelFactory.swift`

**Updated Files:**
- `Gemma3nAudio.swift` - split `attention`, `norm`, `conv`, and `encoder` classes into separate files
- `Gemma3nAudioAttention.swift`: relative position embeddings, chunked attention
- `Gemma3nAudioNorm.swift`: cumulative group normalization
- `Gemma3nAudioConv.swift`: SSCP conv blocks, subsampling projection
- `Gemma3nAudio.swift`: conformer blocks, top-level encoder
- `Gemma3nVLM.swift`: preconditionFailure on unimplemented processor stub

**Tests:**
```python
Test Suite 'Selected tests' started at 2026-04-15 11:32:48.538.
Test Suite 'mlx-swift-lmPackageTests.xctest' started at 2026-04-15 11:32:48.538.
Test Suite 'Gemma3nAudioTests' started at 2026-04-15 11:32:48.538.
Test Case '-[MLXLMTests.Gemma3nAudioTests testAudioConfigDecoding]' started.
Test Case '-[MLXLMTests.Gemma3nAudioTests testAudioConfigDecoding]' passed (0.001 seconds).
[...]

PR #298, opened 2026-05-18
support for audio resources

davidkoski:
## Proposed changes

- see #192, #194
- this adds representation of audio in UserInput and LMInput

This adds the lower level support so that these other PRs can use it.

## Checklist

### DIFF
--- a/Libraries/MLXLMCommon/Chat.swift
+++ b/Libraries/MLXLMCommon/Chat.swift
@@ -14,14 +14,20 @@ public enum Chat {
         /// Array of video data associated with the message.
         public var videos: [UserInput.Video]
 
+        /// Array of audio data associated with the message.
+        public var audios: [UserInput.Audio]
+
         public init(
-            role: Role, content: String, images: [UserInput.Image] = [],
-            videos: [UserInput.Video] = []
+            role: Role, content: String,
+            images: [UserInput.Image] = [],
+            videos: [UserInput.Video] = [],
+            audios: [UserInput.Audio] = []
         ) {
             self.role = role
             self.content = content
             self.images = images
             self.videos = videos
+            self.audios = audios
         }
 
         public static func system(
@@ -37,9 +43,12 @@ public enum Chat {
         }
 
         public static func user(
-            _ content: String, images: [UserInput.Image] = [], videos: [UserInput.Video] = []
+            _ content: String,
+            images: [UserInput.Image] = [],
+            videos: [UserInput.Video] = [],
+            audios: [UserInput.Audio] = []
         ) -> Self {
-            Self(role: .user, content: content, images: images, videos: videos)
+            Self(role: .user, content: content, images: images, videos: videos, audios: audios)
         }
 
         public static func tool(_ content: String) -> Self {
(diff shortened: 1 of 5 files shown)


Libraries/MLXLMCommon/Chat.swift
```diff
@@ -14,14 +14,20 @@ public enum Chat {
         /// Array of video data associated with the message.
         public var videos: [UserInput.Video]
 
+        /// Array of audio data associated with the message.
+        public var audios: [UserInput.Audio]
```
davidkoski, line 18: Thread audio throughout

Libraries/MLXLMCommon/ChatSession.swift
```diff
@@ -73,6 +73,7 @@ public final class ChatSession {
     public var instructions: String?
     private let cache: SerialAccessContainer<Cache>
     public var processing: UserInput.Processing
+    public var audioProcessing: UserInput.AudioProcessing
```
davidkoski: And processing parameters

Libraries/MLXLMCommon/LanguageModel.swift
```diff
@@ -120,17 +121,32 @@ public struct LMInput {
         }
     }
 
+    /// Representation of prepared input audio(s).
+    public struct ProcessedAudio {
+
+        public let samples: MLXArray
```
davidkoski, line 127: images/video have THW -- we may need something like that here, but we can start with this

Libraries/MLXLMCommon/ChatSession.swift
aleroot: What about using processing and creating section inside the single UserInput.Processing ? 
`processing.audio.channels` 


Cleaner ?

_A single processing configuration seems easier to evolve as we add modalities and avoids growing every ChatSession/UserInput initializer with one top-level parameter per modality. If we keep `audioProcessing` separate, what semantic distinction should users infer between `processing` and `audioProcessing`?_

davidkoski: Yeah, i think that is a good idea.

davidkoski: Done!

Re the initializers I wonder if we need to go to a structure for the parameters for the general case, e.g.

```swift
var chatParameters = ChatParameters()
chatParameters.processing = ...
chatParameters.tools = ...

ChatSession(model, parameters)
```

maybe with a few overrides for common cases?

Libraries/MLXLMCommon/ChatSession.swift
```diff
@@ -337,18 +339,21 @@ public final class ChatSession {
...
     ///   - video: optional video (for use with VLMs)
+    ///   - audio: optional audio (for use with VLMs)
     /// - Returns: the model's response
     public func respond(
         to prompt: String,
         role: Chat.Message.Role = .user,
         image: UserInput.Image? = nil,
-        video: UserInput.Video? = nil
+        video: UserInput.Video? = nil,
+        audio: consuming UserInput.Audio? = nil
```
angeloskath, line 349: That's probably a typo :-)

angeloskath (approved):
Looks great except 1 typo.

Libraries/MLXLMCommon/ChatSession.swift
davidkoski: Yes, they should all be `consuming`

PR #306, opened 2026-05-21
Remove SDK build version command from workflow

madrob:
## Proposed changes

Removed the command to show SDK build version during the build process. This was informative, but ultimately causing problems with installations and not actually guarding anything.

## Checklist


davidkoski (approved):
Thanks for debugging that!

PR #308, opened 2026-05-22
Add Gemma 4 MTP speculative decoding

angelsbrood:
**Status:** New PR against `main` following the merge of [#283](https://github.com/ml-explore/mlx-swift-lm/pull/283). The original draft at [#293](https://github.com/ml-explore/mlx-swift-lm/pull/293) was auto-closed when `lmoutput-state` was deleted on merge; this is the same set of twelve commits rebased onto current `main`, plus eight follow-up commits (two repair pre-existing test-setup gaps; the remaining six address davidkoski's review feedback on this PR — DocC link resolution + migration of fixture-anchored tests into the IntegrationTesting Xcode project — and are detailed in Verification below). cc @davidkoski per your note about going through these again now that CI is back in working order.

## Proposed changes

Adds Gemma 4 Multi-Token Prediction (MTP) speculative decoding. Introduces an `MTPDrafterModel` protocol in MLXLMCommon refining `BaseLanguageModel`, registers `gemma4_assistant` as the first concrete drafter implementation, threads cross-model state through three public `LMOutput.Key` declarations, and adds an `MTPSpeculativeTokenIterator` parallel to the existing `SpeculativeTokenIterator`. Tracks #279.

## Architecture

The drafter does not fit `LanguageModel`. Its forward signature differs in fundamental ways: it has no own KV cache; it cross-attends to the target's last full-attention and last sliding-attention layers via a `[String: (MLXArray, MLXArray)]` shared K/V dict; its position IDs are constant within a drafting round; and it requires a bind-time reference to the target to borrow input embeddings. The new `MTPDrafterModel` protocol in `Libraries/MLXLMCommon/MTPDrafterModel.swift` refines `BaseLanguageModel` (so weight loading and sanitize hooks are inherited) and defines its own forward signature. Drafters are loaded via `MTPDrafterModelFactory`, registered into `MTPDrafterTypeRegistry`, and driven by `MTPSpeculativeTokenIterator`.

Cross-model state passthrough uses three public `LMOutput.Key` declarations in MLXLMCommon: `mtpEmitFlagKey: Bool`, `mtpLastHiddenStatesKey: MLXArray`, and `mtpSharedKVStatesKey: [String: (MLXArray, MLXArray)]`. The iterator pre-populates the emit-flag key on every main-model call to signal that the target should write drafter state; the target's top-level `Gemma4.callAsFunction(_:cache:state:)` override reads the flag and threads it through to the backbone, which writes the last-layer hidden state and per-`layer_type` shared K/V into the state on the way out. The emit flag is itself a `LMOutput.Key<Bool>` rather than a function parameter because `LanguageModel.callAsFunction(_:cache:state:)` has a fixed protocol signature; threading the flag through `LMOutput.State` unifies the mechanism — the state-bag carries both tensor payloads and the control signal that triggers them.

`MTPSpeculativeTokenIterator` includes a sticky passthrough fallback. When the target's emit hook returns `nil` shared K/V — which happens once quantization onset converts the regular K/V cache to `QuantizedKVCache`, or any other condition that breaks the regular-K/V invariant — the iterator transparently switches to single-token main-model generation for the remainder of the stream. The drafter is never invoked again on that iterator instance. MTP with mid-stream quantization onset degrades gracefully rather than crashing or producing incorrect outputs; bf16-drafter-against-quantized-target streams give up the speedup but preserve correctness. Empirically verified end-to-end by `testMTP31BQuantizationOnsetEngagesPassthrough`: with the cache configured to quantize mid-stream, speculation runs at ~78% acceptance pre-quantization, transitions to single-token passthrough mode after quantization onset (with `passthroughReason` populated via the Phase 0 plumbing), and completes the token budget with coherent output.

On the target side, the drafter reuses existing `Gemma4TextAttention`, `Gemma4TextDecoderLayer`, and norm building blocks rather than duplicating them. Visibility on those types and a `kvSharedOnly: Bool` flag on `Gemma4TextAttention.init` (which skips construction of `kProj`/`vProj`/`kNorm`/`vNorm`, declared Optional `@ModuleInfo`) let the drafter compose its own decoder layers from existing target-side modules. `Gemma4.swift` diff: +53/−26. The non-`kvSharedOnly` regression suite is bit-identical before and after; existing Gemma 4 generation paths are not perturbed.

## Relationship to existing work

#269 targets the same scope — Gemma 4 MTP speculative decoding — and predates this PR by ~9 days. The two implementations diverge on three substantive axes worth surfacing for reviewers:

1. **Foundation.** #269 widens `generateLoopTask` and `TokenLoopHandler` from `private` to `internal` to reuse the existing generation infrastructure, and exposes pre-logit hidden states via a new `forwardMTP()` method on `Gemma4TextModel`. This PR consumes the `LMOutput.State` API from #283 instead, threading drafter state through the dictionary-backed typed-key state-bag (three `LMOutput.Key` declarations, no per-call ivar mutation on the model). The "models should not mutate state during eval" invariant of #283 is honored throughout.
2. **Verification.** This PR is fixture-anchored against `mlx-vlm@d49d428` with a four-rung ladder culminating in byte-identical token-ID parity at temperature=0. #269 lists a test plan as an unchecked checklist.
3. **Quantization onset.** This PR includes the sticky passthrough fallback described above, which preserves correctness when the main model's cache transitions to quantized mid-stream. #269 does not address this case.

The masked-embedder path is the dimension on which #269 has more implemented than this PR; conversely this PR carries the empty-checkpoint `fatalError` stub honestly rather than shipping centroid logic without a verification fixture.

The overlap is real and acknowledged here so reviewers don't have to discover it independently.

## Verification

Fixture-anchored against [`mlx-vlm@d49d428`](https://github.com/Blaizzy/mlx-vlm/tree/d49d428). The fixture set lives at the [`angelsbrood/gemma4-mtp-fixtures`](https://huggingface.co/datasets/angelsbrood/gemma4-mtp-fixtures) HF dataset (43 files / ~20 MB, generated against target=`mlx-community/gemma-4-31b-it-8bit` and drafter=`mlx-community/gemma-4-31B-it-assistant-bf16`), pinned to a specific commit SHA at the test call sites. Fixture-anchored tests live in the IntegrationTesting Xcode project per davidkoski's review feedback, and fetch the dataset on demand via `IntegrationTestHelpers`. The schema (`FIXTURE-SCHEMA.md`), per-file manifest (`FIXTURE-MANIFEST.json`), and generation tooling (`tools/generate_mtp_fixtures.py`) remain in-repo at `tools/fixtures/` for reproducibility; the data files themselves are not vendored. Regeneration is deterministic (verified across runs).

Verification proceeds in four rungs, each gating the next:

- **Rung 1 — Load / shape.** Drafter checkpoint loads with no missing or unexpected weight keys after sanitize. Weight-key inventory in `RESEARCH-NOTES.md` matches the checkpoint verbatim across both 31B and 26B-A4B drafter siblings (48 keys each, identical layout, shapes differ).
- **Rung 2 — Per-layer activations.** Drafter forward with synthetic inputs produces activations matching Python fixtures at `atol=1e-3, rtol=1e-3`.
- **Rung 3 — Logit parity.** Greedy `draftBlock` logits match Python fixtures at the same tolerance.
- **Rung 4 — Token-for-token cross-runtime parity.** Greedy `draftBlock` at temperature=0 produces byte-identical token-ID sequences to the Python reference across three block sizes (2, 4, 6).

**End-to-end iterator verification (added 2026-05-23):** Above the four-rung fixture ladder, which exercises drafter forward components, the following IntegrationTesting suite exercises the full iterator loop against the 31B target+drafter pair through the high-level `generate(...)` API:

- `testMTP31BPairProducesAcceptedDrafts` — speculation active, counters reach `GenerateCompletionInfo`, no passthrough engagement
- `testMTP31BMatchesBaselineByteIdentical` — byte-identical to baseline at maxTokens=32 on the sky-blue prompt
- `testMTP31BMatchesBaselineRainbowsPrompt`, `testMTP31BMatchesBaselineSwiftPrompt` — byte-identical to baseline at maxTokens=32 on two additional prompts (verifies the bonus-yield fix's content-independence)
- `testMTP31BLongerStreamProducesValidContinuation` — at maxTokens=128, MTP runs with healthy acceptance and the first 64 tokens are byte-identical to baseline. See "Known limitations" for the numerical-substrate boundary that bounds this assertion
- `testBaselineSelfDeterminism128` — trust anchor; confirms baseline non-speculative generation is itself deterministic at maxTokens=128, a prerequisite for any MTP-vs-baseline comparison to be meaningful
- `testMTP31BQuantizationOnsetEngagesPassthrough` — sticky-passthrough fallback empirically verified under mid-stream KV cache quantization (see Architecture for the numbers)
- `testMTPE2BPairFailureMode`, `testMTPE4BPairFailureMode` (env-gated `TEST_E2B_PAIR` / `TEST_E4B_PAIR`) — E-series drafter failure-mode characterization. Both trap at the documented `Gemma4AssistantMaskedEmbedder` `fatalError` stub at `Gemma4Assistant.swift:95`; the iterator and surrounding plumbing work as designed. The centroid embedder is the only missing piece for E-series support.

Plus CI-scoped (no model loads required): `MTPGenerateLoopTaskContractTests` in `MLXLMTests`, which drives a mock `MTPStatsCollecting` iterator through a minimal `generateLoopTask` reproduction to catch value-type fork regressions at PR-CI scope.

All four rungs pass in Xcode against the fixture set. `swift test` CLI is not a working verification surface for this repository — `CONTRIBUTING.md` notes "swift test does not work yet — use xcodebuild instead," and the metallib initialization failure that affects CLI runs predates this branch and is not introduced here. Full-target acceptance-rate and quantization-onset integration tests are placeholdered in `MLXLMTests` with `Issue.record` + early-return because they require `MLXHuggingFace`'s `#huggingFaceTokenizerLoader()` which is not in `MLXLMTests`' dependency closure; the full end-to-end measurement will live in a downstream integration test.

**Test setup corrections (2026-05-21):** Two pre-existing test-setup gaps were repaired alongside the rebase, as two trailing commits. (1) The Rung 4 token-parity test (`MTPRung4TokenParityTests`) was strengthened to load the full target (`mlx-community/gemma-4-31b-it-8bit`) and bind the drafter to it via `bind(target:)` before invoking `draftBlock`, matching the production iterator path — the bind step had been missing from the test since its introduction. (2) Three synthetic-mock iterator tests in `MTPSpeculativeTokenIteratorTests` used `99` as an inert placeholder for unused prefill positions, but the mock's vocab is `20`, so the placeholder caused an out-of-bounds write inside the mock's `makeLogits`. In both cases the runtime path was correct in production (`MTPSpeculativeTokenIterator.bind` wires bind correctly; the iterator's own logic is sound); only the test scaffolding was incomplete. All four Rungs now pass against the fixture set as claimed above, as do the iterator's synthetic smoke tests.

**Migration to IntegrationTesting (2026-05-22):** Per davidkoski's review feedback on this PR, fixture-anchored and checkpoint-dependent tests have been moved out of `MLXLMTests` and into the IntegrationTesting Xcode project, with fixtures fetched on demand from the published HF dataset rather than vendored in-repo. The unit tests that remain in `MLXLMTests` (synthetic-mock, compile-time, key-roundtrip) run on CI as before. Two test files mixed synthetic and checkpoint/fixture-gated cases and were split rather than moved wholesale: `MTPDrafterModelFactoryTests.swift` (4 synthetic methods stay; 1 checkpoint-gated load test moves to `MTPDrafterModelFactoryIntegrationTests.swift`) and `Gemma4AssistantDraftModelTests.swift` (4 synthetic methods stay; 3 checkpoint/fixture-gated methods move to `Gemma4AssistantDraftModelIntegrationTests.swift`). The dataset-fetch path is a new public `downloadDataset(repo:revision:)` helper in `IntegrationTestHelpers`; fixture revisions are pinned to a specific HF commit SHA at each test call site for byte-exact parity.

**DocC link resolution fix (2026-05-22):** CI's `Verify documentation` step (run #26266924101) surfaced eleven cross-module and internal-visibility `` ``Symbol`` `` references in MTP doc comments that DocC strict mode could not resolve (each module's documentation archive is built independently, so bare-name symbol links in MLXVLM cannot resolve to MLXLMCommon symbols, and links from public-symbol doc comments to internal-visibility same-module symbols are disallowed). Fixed by demoting unresolvable links to inline code formatting and expanding one signature-mismatched same-module link to its full parameter form. Full audit covers all `` ``...`` `` references across touched files, not just those CI named.

**Diagnostic investigation and end-to-end verification (2026-05-23):** Following @joelnishanth's 2026-05-22 benchmark surfacing `proposedDraftTokens=0` on the E4B pair, a 9-commit investigation resolved his report into a real observability gap plus two pre-existing latent bugs in the iterator's surrounding plumbing.

The observability gap: iterator counters (`proposedCount`, `acceptedCount`) existed but weren't plumbed through to `GenerateCompletionInfo`, so consumers using the high-level `generate(...)` API couldn't read them. Phase 0 (`0140ab2`) plumbed three Optional fields — `proposedDraftTokens`, `acceptedDraftTokens`, `passthroughReason` — through the `.info` event and introduced an `MTPStatsCollecting` protocol that drafter iterators conform to. Source-compatible: existing call sites that don't supply the new init parameters continue to compile unchanged.

The two latent bugs:

- **For-in value-type copy in `generateLoopTask`** (`Evaluate.swift:1787`, fixed in `25faa79`). `for token in iterator` over a value-typed iterator expands to `var __it = iterator.makeIterator()` which forks a copy via Sequence's default `makeIterator() -> Self`; mutations on the copy don't propagate to the outer binding observed at any later downcast. This was a long-standing issue affecting both MTP and non-MTP iterators, surfaced only when Phase 0's downcast at the info-construction site started reading iterator state. Secondary fix as a side effect: the same change makes `iterator.tokenCount` at the stopReason check truthful, so iterators hitting their `maxTokens` budget now correctly report `stopReason=.length` rather than falling through to `.cancelled`.

- **Prepare-time bonus-token yield omission in `MTPSpeculativeTokenIterator.prepare(input:)`** (fixed in `ee86bff`). The iterator sampled bonus token(s) during prepare and stored them in `y` as input to the first speculation round, but never appended them to `pendingTokens`. The consumer of the iterator never saw the bonus tokens; the output stream started 1-2 positions ahead of baseline non-speculative generation depending on `PrepareResult` shape. Long-standing; not caught by the fixture-based Rung 1–4 tests because they exercise drafter forward components rather than the full iterator-prepare-to-next loop. Surfaced when the post-Phase-4 byte-identity-to-baseline test (`testMTP31BMatchesBaselineByteIdentical` in `6b320c1`) compared the iterator's full output stream against autoregressive baseline at temp=0.

End-to-end verification added across 8 active test methods (2 additional env-gated) in `MTPIteratorEndToEndDiagnosticTests` (IntegrationTesting), plus a CI-scoped synthetic-loop test in `MLXLMTests` for value-type fork regression coverage without weight loads. Final 31B numbers post-fix: proposed=35, accepted=19, acceptance rate 54.29%, passthrough=none, 13.03 tok/s MTP vs. 8.22 tok/s baseline (1.586× speedup), byte-identical to baseline at maxTokens=32 across three prompts, byte-identical to baseline through the first 64 tokens of maxTokens=128. Full commit list: `0140ab2`, `01fab00`, `25faa79`, `ee86bff`, `6b320c1`, `75c80d0`, `7c57408`, `42d6ad1`, `61777c8`.

## Files

**New:**

- `Libraries/MLXLMCommon/MTPDrafterModel.swift` — `MTPDrafterModel` protocol, `MTPDrafterContext`, `MTPDrafterContainer`, three public `LMOutput.Key` declarations
- `Libraries/MLXLMCommon/MTPSpeculativeTokenIterator.swift` — iterator with sticky passthrough fallback
- `Libraries/MLXLMCommon/MTPDrafterModelFactory.swift` — `MTPDrafterTypeRegistry`, `MTPDrafterRegistry`, `MTPDrafterModelFactory`
- `Libraries/MLXLMCommon/BidirectionalMasks.swift` — `createBidirectionalMask` and `createBidirectionalSlidingWindowMask` helpers
- `Libraries/MLXVLM/Models/Gemma4Assistant.swift` — `Gemma4AssistantConfiguration`, `Gemma4AssistantMaskedEmbedder`, `Gemma4AssistantDraftInner`, `Gemma4AssistantDraftModel`
- `Libraries/MLXVLM/Gemma4AssistantRegistration.swift` — `registerModelType` call site for `"gemma4_assistant"`
- `Tests/MLXLMTests/MTPDrafterModelTests.swift`, `Gemma4AssistantDraftModelTests.swift` (synthetic-mock half of the split — config-decode, sanitize, instantiation), `Gemma4EmitDrafterStateTests.swift`, `MTPDrafterModelFactoryTests.swift` (synthetic-mock half of the split — registration, registry, factory), `MTPSpeculativeTokenIteratorTests.swift`, `MTPGenerateAPITests.swift`
- `IntegrationTesting/IntegrationTestingTests/BidirectionalMasksTests.swift`, `MTPRung4TokenParityTests.swift`, `MTPAcceptanceRateTests.swift` (placeholdered), `MTPQuantizationOnsetTests.swift` (placeholdered), `Gemma4AssistantDraftModelIntegrationTests.swift` (the fixture/checkpoint-gated half of the split `Gemma4AssistantDraftModelTests.swift`), `MTPDrafterModelFactoryIntegrationTests.swift` (the checkpoint-gated half of the split `MTPDrafterModelFactoryTests.swift`)
- `RESEARCH-NOTES.md` — architectural Q&A with file:line citations against canonical Python (HF Transformers `gemma4_assistant`, mlx-vlm drafter), plus full 48-key weight inventories for both reference drafter checkpoints
- `tools/inspect_drafter_layout.py`, `tools/generate_mtp_fixtures.py`
- `tools/fixtures/` — schema (`FIXTURE-SCHEMA.md`), manifest (`FIXTURE-MANIFEST.json`), regeneration tooling, and `README.md` pointing to the published HF dataset. Data files themselves are at [`angelsbrood/gemma4-mtp-fixtures`](https://huggingface.co/datasets/angelsbrood/gemma4-mtp-fixtures).
- `Tests/MLXLMTests/MTPStatsCollectingTests.swift` — synthetic-mock unit tests for the `MTPStatsCollecting` protocol conformance and `GenerateCompletionInfo` field threading
- `Tests/MLXLMTests/MTPGenerateLoopTaskContractTests.swift` — CI-scoped regression test for iterator copy semantics in `generateLoopTask`, no model loads required
- `IntegrationTesting/IntegrationTestingTests/MTPIteratorEndToEndDiagnosticTests.swift` — full end-to-end iterator test surface (8 active methods, 2 env-gated)

**Modified:**

- `Libraries/MLXLMCommon/Evaluate.swift` — two new `generate(...)` overloads (one returning `AsyncStream<Generation>`, one returning `AsyncStream<TokenGeneration>`), taking an `mtpDrafter` parameter, `blockSize: Int = 4` default
- `Libraries/MLXVLM/Models/Gemma4.swift` — visibility bumps on internal building blocks (the cascade reached 16 symbols once `Gemma4TextLanguageModel`, `Gemma4TextBackbone`, the `languageModel` accessor, and four MLP/router types were pulled along by what they reference), `kvSharedOnly: Bool` flag on `Gemma4TextAttention.init`, emit hooks in `Gemma4TextBackbone.callAsFunction` and `Gemma4TextLanguageModel.callAsFunction`, new top-level `Gemma4.callAsFunction(_:cache:state:)` override (+53/−26)
- `Libraries/IntegrationTestHelpers/IntegrationTestHelpers.swift` — public top-level `downloadDataset(repo:revision:matching:)` helper that fetches a public HF dataset snapshot to a per-revision local cache (`~/.cache/huggingface/integration-test-datasets/{repo}/{revision}/`); consumed by the migrated fixture-anchored tests
- `Libraries/MLXLMCommon/MTPDrafterModel.swift` — added `MTPStatsCollecting` protocol
- `Libraries/MLXLMCommon/MTPSpeculativeTokenIterator.swift` — conformance to `MTPStatsCollecting`, stored `passthroughReason` for observability, fix for prepare-time bonus-token yield
- `Libraries/MLXLMCommon/Evaluate.swift` — three Optional fields added to `GenerateCompletionInfo` (Optional preserves source compatibility), downcast in `generateLoopTask` info construction, for-in copy-semantics fix at the iterator-consuming loop

## Supported model pairs

| Target | Drafter | Status in this PR |
| --- | --- | --- |
| `mlx-community/gemma-4-31b-it-8bit` | `mlx-community/gemma-4-31B-it-assistant-bf16` | Primary; fixture-anchored |
| `mlx-community/gemma-4-26B-A4B-it-bf16` | `mlx-community/gemma-4-26B-A4B-it-assistant-bf16` | Architecturally supported; not fixture-anchored |
| `mlx-community/gemma-4-31B-it-bf16` | `mlx-community/gemma-4-31B-it-assistant-bf16` | Architecturally supported; would re-tighten Rung 2/3 tolerances |

E-series targets (`gemma-4-E2B-it`, `gemma-4-E4B-it`) are architecturally supported but not exercised by the fixture set in this PR; both have published `-assistant-bf16` siblings on the `mlx-community` org.

## Known limitations

- **`Gemma4AssistantMaskedEmbedder.callAsFunction` is a `fatalError` stub.** Both reference checkpoints have `use_ordered_embeddings=false` and ship with zero `masked_embedding.*` weight keys, so the centroid-narrowing path is never exercised by the fixture set or by either currently-published reference checkpoint. The module is declared correctly for weight loading (so a future checkpoint with the flag enabled would load with no key mismatches); the forward implementation can land in a follow-up when a verifiable checkpoint surfaces. Choosing `fatalError` over an unverified implementation matches the verification posture of the rest of the PR.
- **Quantized drafter weights** deferred — bf16 only initially. Drafters are small enough (~78M params for E-series, ~0.4B–0.5B for the larger pairs) that this is acceptable for a first PR.
- **Batched MTP** deferred — single-batch only. mlx-vlm's `_speculative_walk_batch` is out of scope.
- **Audio path** unaffected — the target's audio components remain filtered in sanitize today; this drafter is text-only.
- **Quantized cache during a drafting stream** is covered by the sticky passthrough fallback described above. The drafter is bypassed when fallback activates; correctness is preserved, the MTP speedup is not.
- **Acceptance-rate floor and full-target quantization-onset integration tests** are placeholdered in `MLXLMTests` pending `MLXHuggingFace` in the test target's dependency closure. The passthrough fallback logic itself is unit-tested with synthetic state.
- **Byte-identity to autoregressive at temp=0 is verified for short streams; longer streams may diverge.** Empirically on the 31B target+drafter pair, MTP and baseline non-speculative output are byte-identical through 64 tokens and may diverge between 64 and 96 tokens. The cause is MLX's fused SDPA kernel producing shape-dependent floating-point reduction results: MTP's multi-token verify forward (positions `[bonus, d_1, ..., d_K]`) follows a different numerical path than the K+1 single-position autoregressive forwards baseline uses. At temp=0 with argmax sampling, this drift eventually flips a tied or near-tied logit choice and the streams branch. The MTP algorithm's speculation-correctness guarantee — at each emitted position N, the iterator returns what the target would have returned at position N given the iterator's accumulated history through N-1 — is preserved throughout; only independent-baseline byte-identity is bounded by this. Both `testMTP31BMatchesBaselineByteIdentical` (32 tokens, three prompts) and `testMTP31BLongerStreamProducesValidContinuation` (128 tokens, 64-token prefix) regression-test the empirical boundary on this configuration. Investigation tracked in [upstream MLX issue link TBD post-merge].

## Checklist

- [x] CONTRIBUTING document read
- [x] `pre-commit run --all-files` passes
- [x] Tests added (ten new test files; Rungs 1–4 covered; full-target integration tests placeholdered with documented gating)
- [ ] Documentation updates — happy to add a README section or wherever convention dictates if reviewers want one

## Acknowledgments

Thanks to @davidkoski for #283 — the typed-key `LMOutput.State` API is the foundation this PR builds on; without it, the cross-model state passthrough would have required something much uglier.

## References

- Tracking issue: #279 (@brdcastro)
- Foundation PR: #283
- Parallel PR: #269 (see "Relationship to existing work" above)
- Fixture dataset: [`angelsbrood/gemma4-mtp-fixtures`](https://huggingface.co/datasets/angelsbrood/gemma4-mtp-fixtures)
- Canonical Python references: HF Transformers `gemma4_assistant`, mlx-vlm `speculative/drafters/gemma4_assistant` at SHA `d49d428`
- Cross-runtime: vLLM PR [#41745](https://github.com/vllm-project/vllm/pull/41745)
- Google announcement: [blog post](https://blog.google/innovation-and-ai/technology/developers-tools/multi-token-prediction-gemma-4/)


angelsbrood:
@angeloskath — I don't have the org permissions to formally request reviews on this repo, so I'm pinging here. The PR description's Status block also cc's @davidkoski (per a prior thread). Would appreciate your eyes on this when convenient.

tools/fixtures/drafter_block/case_01_block2.safetensors
davidkoski, line 1: I don't think we want all these safetsensors files -- are these something you used during development?

Tests/MLXLMTests/Gemma4AssistantDraftModelTests.swift
```diff
@@ -0,0 +1,298 @@
...
+import MLXNN
+@_spi(Testing) import MLXVLM
+import Testing
+
+/// Locate `tools/fixtures/` relative to this test file by walking up.
+private func toolsFixturesDir(file: String = #filePath) -> URL? {
+    var dir = URL(fileURLWithPath: file).deletingLastPathComponent()
+    let fs = FileManager.default
+    for _ in 0 ..< 10 {
+        let candidate = dir.appendingPathComponent("tools/fixtures")
```
davidkoski: Ah, I see, the safetsensors are for these tests.  We shouldn't have such large files for the tests -- if you want to make integration style tests take a look at IntegrationTesting.  That doesn't run on CI, but it can download models and exercise things.

I think the tests that rely on the safetsensors files probably belong in there.

davidkoski:
FYI @joelnishanth (#269) and @john-rocky (#267) -- these are all efforts and landing gemma4 MTP.  I am inclined to go with #308 as I think not introducing a new type of model evaluation is better.  The MTP and base model are already tightly coupled, so using the State + keys seems like a reasonable approach to manage the out-of-band communication.

We can only land one of these, so if you think there are things missing here please comment or coordinate with @angelsbrood to incorporate.

Thank you!

angelsbrood:
Thanks for the review. Both points addressed in `b81e4c9` and `a2c6c84` (the migration commit pair):

- **Fixture-anchored and checkpoint-dependent tests moved to the IntegrationTesting Xcode project.** `BidirectionalMasksTests`, `MTPRung4TokenParityTests`, `MTPAcceptanceRateTests` (placeholdered), `MTPQuantizationOnsetTests` (placeholdered) moved wholesale. `MTPDrafterModelFactoryTests` and `Gemma4AssistantDraftModelTests` mixed synthetic-mock and checkpoint/fixture-gated cases, so they were split — the synthetic halves stay in `MLXLMTests` (CI-runnable); the checkpoint/fixture-gated halves moved to `MTPDrafterModelFactoryIntegrationTests.swift` and `Gemma4AssistantDraftModelIntegrationTests.swift`.
- **`tools/fixtures/*.safetensors` data files removed.** Tests now fetch the dataset on demand from [`angelsbrood/gemma4-mtp-fixtures`](https://huggingface.co/datasets/angelsbrood/gemma4-mtp-fixtures) via a new public `downloadDataset(repo:revision:)` helper in `IntegrationTestHelpers` (commit `efbdfdc`). Per-revision local cache at `~/.cache/huggingface/integration-test-datasets/{repo}/{revision}/`; the revision is pinned to a specific SHA at each test call site for byte-exact parity. Schema, manifest, generation tooling, and a new `README.md` pointing to the published HF dataset stay at `tools/fixtures/` for reproducibility.
- **Synthetic-mock and compile-time unit tests remain in `Tests/MLXLMTests/`** (no fixture or checkpoint dependency) so they continue to run on CI.

Two other things landed in this push:

- `0d97e61` fixes the eleven DocC symbol-link errors that caused the previous `Verify documentation` CI failure (cross-module and internal-visibility `` ``Symbol`` `` references demoted to inline code; one signature-mismatched same-module link expanded). Full audit covers all `` ``...`` `` references in changed files.
- `7e2f2b8` repairs two test-infrastructure issues surfaced when the moved tests first ran under the IntegrationTesting target's runner: `@Test(.disabled(...))` for the two placeholders (Swift Testing treats `Issue.record` as a failure, not a skip), and a `@Suite(.serialized)` wrapper on the three Gemma4 assistant integration tests so parallel weight loads don't race on MLX runtime state.

Resolving both threads. Let me know if anything looks off.

davidkoski:
> @angeloskath — I don't have the org permissions to formally request reviews on this repo, so I'm pinging here. The PR description's Status block also cc's @davidkoski (per a prior thread). Would appreciate your eyes on this when convenient.

No worries, I will review.

joelnishanth:
@angelsbrood @davidkoski — Thanks for putting this together. The architecture (State + typed keys, sticky passthrough fallback, fixture-anchored verification) is solid and I'm glad David is going with this approach.

A few ways I'd like to help:

**1. Centroid masked embedder**

I have a working `CentroidMaskedEmbedder` implementation in my [mlx-swift-mtp](https://github.com/joelnishanth/mlx-swift-mtp/blob/main/Sources/MLXSwiftMTP/EfficientEmbedder.swift) package — centroid prediction, top-K selection, gather/scatter over ~4K candidates. I understand you stubbed it with `fatalError` because no current checkpoint exercises the path, but happy to contribute the implementation as a follow-up PR (or directly to your branch) whenever a verifiable checkpoint surfaces. The code ports HuggingFace's `Gemma4AssistantMaskedEmbedder` and is tested against the same reference.

**2. Benchmarking**

Once this lands, I can run end-to-end tok/s benchmarks (MTP speculative vs standard autoregressive) across model pairs and Apple Silicon generations using a benchmark CLI I've been maintaining. Happy to produce acceptance-rate and throughput numbers for the PR description or a follow-up.

**3. Downstream validation**

I'm using MTP speculative decoding in a shipping macOS app (on-device inference), so I'll be exercising this in a real integration fairly quickly after merge. Will report any edge cases I hit.

Let me know if there's anything else that would be helpful.

— [Joel Nishanth](https://offlyn.ai) · [offlyn.AI](https://offlyn.ai)

angelsbrood:
Thanks Joel— super appreciate it, and your graciousness on the architecture. All three would be great. 

On the CentroidMaskedEmbedder: yes, please. The fatalError is definitely a placeholder waiting for a verifiable path. When a checkpoint exercises it, your implementation is the natural fill-in. Happy to coordinate on timing, whether that's a follow-up PR from you or porting into this surface together.

Re: benchmarks, would love that. Tok/s and acceptance-rate numbers across model pairs and Apple Silicon generations would round out the verification story considerably. Happy to fold the results into a follow-up to the PR description, or wherever feels right.

On downstream validation: thank you. A shipping integration finding edge cases the fixture tests miss is exactly the kind of surface this needs. Please reach out for anything you hit. Happy to be reachable.

Looking forward to working together on this.

joelnishanth:
## MTP benchmark results (E4B target + E4B assistant)

CLI: [`joelnishanth/mlx-swift-mtp-bench`](https://github.com/joelnishanth/mlx-swift-mtp-bench) against `angelsbrood/gemma4-assistant` + local centroid embedder fix.

**Setup**
- Target: `mlx-community/gemma-4-e4b-it-4bit`
- Drafter: `mlx-community/gemma-4-E4B-it-assistant-bf16`
- Hardware: Apple Silicon, 18 GB RAM, macOS 26.1
- `temperature=0`, `maxTokens=64`, block sizes `{2,4,6}`, 3 measured runs + 1 warmup

### Decode tok/s (median of 3 runs)

| Prompt | Prefill tokens | Baseline | MTP b=2 | MTP b=4 | MTP b=6 |
|--------|---------------:|---------:|--------:|--------:|--------:|
| short  | 33 | **215** | 43 | 43* | 397* |
| medium | 3179 | **534** | 80 | 140 | 93 |

\*Short MTP block 4/6 medians are noisy (early EOS on some runs → sub-30ms decode); block 2 stable ~43 tok/s.

### TTFT (median, ms)

| Prompt | Baseline | MTP b=4 |
|--------|----------|---------|
| short  | 174 | 246 |
| medium | 209 | 478 |

### Observations

1. **`proposedDraftTokens=0` on all MTP runs** in the benchmark JSON — acceptance rate could not be computed. This suggests the iterator never recorded draft proposals (passthrough from missing drafter state, or counter not reached). Worth checking against `MTPAcceptanceRateTests` wiring.
2. **E4B assistant requires `Gemma4AssistantMaskedEmbedder`** (`use_ordered_embeddings=true`). Without it the drafter LM head path is broken; I have a follow-up PR: [angelsbrood/mlx-swift-lm#1](https://github.com/angelsbrood/mlx-swift-lm/pull/1) (Python parity test passes).
3. **Long prompt (~8K tokens) OOM** on 18 GB (`metal::malloc` 12.3 GB allocation during prefill). Medium prompt is the practical limit on this machine.
4. On medium context, baseline is **~3–6× faster** than MTP in this run — likely related to (1) and/or drafter overhead before acceptance is healthy.

Happy to re-run after #308 merges with the centroid embedder landed, and to extend the CLI with `GenerateCompletionInfo`-based timing if helpful.

— [Joel Nishanth](https://offlyn.ai) · [offlyn.AI](https://offlyn.ai)

angelsbrood:
@joelnishanth, thanks much for running that. The `proposedDraftTokens=0` you saw was definitely a real observability gap: counters existed on the iterator but weren't plumbed through to `GenerateCompletionInfo` so the CLI was reading a field that didn't exist and defaulting to zero. Following that thread surfaced two other pre-existing bugs underneath. Counters now reach `info.proposedDraftTokens` and `info.acceptedDraftTokens` through the standard `.info` event. (PR description updated with the full sequence.)

If you get a second to re-run your bench against `gemma4-assistant` head, would love to see the new numbers— your CLI should be able to read the counters directly now. No rush, thanks again.

joelnishanth:
Commenting to check if there has been progress

Libraries/MLXLMCommon/MTPSpeculativeTokenIterator.swift
```diff
@@ -0,0 +1,373 @@
...
+            let state = mainState,
+            let lastHidden = state[mtpLastHiddenStatesKey],
+            let sharedKV = state[mtpSharedKVStatesKey]
+        else {
+            switchToPassthrough(reason: "main model did not emit drafter state")
+            return
+        }
+
+        // Slice last position's hidden -> [B, 1, hidden].
+        let lastPositionHidden = lastHidden[0..., (-1)..., 0...]
```
davidkoski: The mlx-lm code has this:

```python
        # Hidden for next round: pick the slot of the newly accepted bonus.
        hidden = _mtp_draft_hidden(lm, verify.hidden[:, accepted : accepted + 1, :])
        b = new_tokens[-1] if new_tokens else b
```

It notes:

```python
    # Hidden from prefill is full prompt-length; reduce to a single slot.
    # The semantically-correct choice is the *last* prompt token's hidden:
    # the just-sampled bonus is the next-token prediction from that position,
    # so its embedding paired with that hidden is what the drafter expects.
    # (HF's literal ``[:, n_last_matches:n_last_matches+1]`` with ``n_matches=0``
    # on the first round picks position 0, which is BOS — they get away with
    # it because subsequent rounds slice into the per-call verify hidden, but
    # the round-1 acceptance is wasted. We don't replicate that quirk.)
```

I don't know all the nuances here, but in comparing the implementation, this stood out.  I asked claude and it suggested that the acceptance rate is low when blockSize is high.  🤷 

I think we may need to track the accepted state in the iterator.  This might also affect sharedKV.

See what you think.

Libraries/MLXVLM/Models/Gemma4Assistant.swift
```diff
@@ -0,0 +1,301 @@
...
+            self._lmHead.wrappedValue = Linear(
+                textCfg.hiddenSize, textCfg.vocabularySize, bias: false)
+        }
+        if config.useOrderedEmbeddings {
+            self._maskedEmbedding.wrappedValue = Gemma4AssistantMaskedEmbedder(config: config)
+        }
+        super.init()
+    }
+
+    public func bind(target: any LanguageModel) {
```
davidkoski: Called from `MTPSpeculativeTokenIterator` -- this isn't thread safe.  Multiple threads/tasks may be using the same `Gemma4AssistantDraftModel` instance.  It might be that they all pass in the same `LanguageModel` instance, but:

- that kind of mutation without locks leads to crashes
- if they were different instances then it will spill across the different iterators

I suspect this needs to be passed in as a parameter to `draftBlock()` instead.

Libraries/MLXLMCommon/MTPDrafterModel.swift
```diff
@@ -0,0 +1,159 @@
...
+/// target's input embeddings, embed scale, and per-layer type metadata.
+/// Stored references are read-only during eval (consistent with PR #283's
+/// "no mutation during eval" invariant).
+public protocol MTPDrafterModel: BaseLanguageModel {
+    /// One-time setup. Caches references to the target's input embeddings,
+    /// embed scale, and per-layer type metadata. Must be called before any
+    /// `draftBlock(...)` invocation. Stored state is read-only during eval.
+    ///
+    /// - Parameter target: The main language model that this drafter speculates for.
+    func bind(target: any LanguageModel)
```
davidkoski: See https://github.com/ml-explore/mlx-swift-lm/pull/308/changes#r3370592364 -- this is not thread safe.

Libraries/MLXLMCommon/BidirectionalMasks.swift
```diff
@@ -0,0 +1,67 @@
...
+/// Transformers' `create_bidirectional_sliding_window_mask` (after the
+/// kv-axis flip). When `windowSize >= kvLen` the entire kv axis attends
+/// (degenerate case → all zeros).
+///
+/// Note on runtime semantics: mlx-vlm's drafter computes the SWA mask from
+/// a distance formula `|q_idx - k_idx| < windowSize` that accepts a
+/// `query_offset`. For `queryLen == 1, query_offset == 0` (the dominant MTP
+/// case) the two conventions produce identical output. Drafter call sites
+/// that need a non-zero `query_offset` should build the mask inline rather
+/// than calling this helper.
```
davidkoski: Is this referring to variables in the original python code?  It seems to mix python and swift naming -- I think it should probably settle on one.

Libraries/MLXVLM/Models/Gemma4Assistant.swift
```diff
@@ -0,0 +1,301 @@
...
+        // Per-layer-type masks; KV tensor shape is [B, H, S, D] so axis -2 = seq.
+        let fullKvLen = sharedKV["full_attention"].map { $0.0.dim(-2) } ?? 0
+        let slidingKvLen = sharedKV["sliding_attention"].map { $0.0.dim(-2) } ?? 0
+        let fullMask = createBidirectionalMask(
+            queryLen: queryLen, kvLen: fullKvLen, dtype: h.dtype)
+        let slidingMask = createBidirectionalSlidingWindowMask(
+            queryLen: queryLen, kvLen: slidingKvLen,
+            windowSize: textCfg.slidingWindow, dtype: h.dtype)
+
+        let queryOffset = Int(positionIds[0, 0].item(Int32.self))
```
davidkoski: Note: the `createBidirectionalSlidingWindowMask` suggests that the queryOffset should be 0 _or else!_.  Should we verify that here?  Or do we need to handle the case where it is not?

Also, using `item()` during inference will cause a stall.

Is this code actually used outside the tests?  Both this and `createBidirectionalSlidingWindowMask` seem to be special purpose for the tests.  That is ok, but maybe put them in an extension maybe even in a `+Tests.swift` file to keep them apart?

Libraries/MLXLMCommon/MTPSpeculativeTokenIterator.swift
```diff
@@ -0,0 +1,373 @@
...
+
+        // Sample one main-model token per verify position.
+        let mainTokens: MLXArray
+        if var verifyProcessor = processor {
+            var sampled = [MLXArray]()
+            for i in 0 ..< (numDraft + 1) {
+                var logits = mainLogits[0..., verifyStart + i, 0...]
+                logits = verifyProcessor.process(logits: logits)
+                let token = sampler.sample(logits: logits)
+                verifyProcessor.didSample(token: token)
```
davidkoski: I am curious how this works when some of the tokens get rejected -- they may be in the LogitProcessor state (if it has any, e.g. a repetition penalty).

Maybe we can't support stateful logit processors?

Libraries/MLXVLM/Models/Gemma4Assistant.swift
davidkoski: Another option is to have `bind()` return a _new_ `Gemma4AssistantDraftModel` with this set.  The models can share the state.  But I suspect just passing as a parameter is simpler.

Libraries/MLXVLM/Models/Gemma4Assistant.swift
```diff
@@ -0,0 +1,301 @@
...
+    @ModuleInfo(key: "lm_head") public var lmHead: Linear?
+    @ModuleInfo(key: "masked_embedding") public var maskedEmbedding: Gemma4AssistantMaskedEmbedder?
+
+    // Bind-time state. Read-only during eval (consistent with PR #283's
+    // "no mutation during eval" invariant). Not `Sendable` because Embedding
+    // is a reference type; cross-domain access must go through
+    // `MTPDrafterContainer.perform`.
+    private var boundInputEmbed: Embedding?
+    private var boundEmbedScale: Float = 1.0
+    private var boundTargetLayerTypes: [String] = []
```
davidkoski: See thread safety note below.  Also I don't think boundTargetLayerTypes is used.

Libraries/MLXLMCommon/MTPDrafterModel.swift
angelsbrood: Same fix as comment 2 — protocol no longer has `bind()`; `target: any LanguageModel` flows through `draftBlock(...)` per round. Conforming types are now expected to be stateless w.r.t. target. `c6af858`.

Libraries/MLXLMCommon/MTPSpeculativeTokenIterator.swift
angelsbrood: Confirmed — slice was using the last verify position unconditionally regardless of accepted count. Tracked accepted across rounds and applied the `[:, accepted:accepted+1, :]` slice for round 2+ (round 1 still uses `[-1:]` for the last prompt position, matching mlx-lm). SharedKV did not need analogous slicing — its seq dim is cumulative cache size rather than verify-input size, so the existing pass-through is correct. Fixed in `55c55b2`.

Empirical impact at temp=0 sky-blue (31B target + drafter): bs=4 acceptance went from 54.3% to 72.4% (+18.1pp), throughput 13.14 → 15.46 tok/s. bs=6 went from 17.9% to 29.8% (+11.9pp), throughput 7.10 → 9.26 tok/s. Both lift, with bs=6 catching up proportionally more as the bug's mechanism would predict. The effect grows with stream length too: bs=4 at maxTokens=64 lifted from 32.3% to 60.6% (+28.3pp) as per-round losses compound.

Libraries/MLXVLM/Models/Gemma4Assistant.swift
angelsbrood: Took option A — dropped `bind()` from the protocol and pass `target` as a parameter to `draftBlock(...)`. Drafter is now stateless w.r.t. target binding; sharing instances across iterators is safe by construction (no per-iterator mutable state remains). Fixed in `c6af858`.

Libraries/MLXLMCommon/BidirectionalMasks.swift
angelsbrood: Right — the comment was referring to mlx-vlm's Python parameter name `query_offset` rather than the Swift `queryOffset`. Rewrote the docstring to describe the Swift API contract directly. (The full rewrite landed in `bc2f8e3` along with the comment-5 fix, since the docstring also needed updating to document the production invariant that makes the early-exit branch fire.)

Libraries/MLXVLM/Models/Gemma4Assistant.swift
angelsbrood: Three observations here, all good:

1. **`queryOffset == 0` precondition**: turns out the dominant MTP case actually has non-zero queryOffset — it equals the main cache offset, which grows past zero after prefill. But production callers always satisfy `slidingKvLen <= slidingWindow` because the target's sliding-attention cache is capped at `slidingWindow` by `RotatingKVCache(maxSize:keep:)`. That cap keeps the helper on its `windowSize >= kvLen` early-exit branch, which produces an all-zeros mask regardless of `queryOffset`. The non-degenerate branch (which builds an absolute-position mask, NOT a distance-from-`queryOffset` mask) is theoretical and unreachable in production today. Added a `precondition(slidingKvLen <= textCfg.slidingWindow, ...)` in `forwardHidden` to make the invariant explicit — loud crash beats silent wrong-mask if cache policy ever changes. Also rewrote the helper's docstring to describe what it actually computes in the non-degenerate path and document the production invariant the early-exit relies on.

2. **`.item()` during inference**: eliminated. Threaded `queryOffset` as a Swift `Int` from `MTPSpeculativeTokenIterator.speculateRound()` (where `cacheOffset` is already an `Int`) down through `draftBlock(target:queryOffset:...)` → `forwardHidden(queryOffset:...)`. The `Int(positionIds[0, 0].item(Int32.self))` round-trip — which fired K-1 times per round inside the autoregressive draft loop — is gone. The layer's RoPE path takes the `Int` offset directly; no MLXArray of positions was ever needed beyond the round-trip.

3. **Test-only vs production**: the helper IS used in production (the Gemma 4 drafter's `forwardHidden` calls it), it's just that the production invariant keeps it on the trivial early-exit. Kept it in `MLXLMCommon`.

All three fixed in `bc2f8e3`.

Libraries/MLXLMCommon/MTPSpeculativeTokenIterator.swift
angelsbrood: Looked into this carefully. The verify loop uses `var verifyProcessor = processor`, which by Swift's struct value semantics is a local copy — mutations to `verifyProcessor.didSample` don't propagate to `self.processor`. The canonical processor state at `self.processor` only receives `didSample` calls from the accept loop, which runs over the actually-emitted tokens (accepted drafts + bonus). So state pollution from rejected tokens doesn't occur. The pattern is subtle and easy to misread; added a CI-scoped recording-processor regression test in `9b55aa4` that asserts this invariant directly, plus a 31B end-to-end smoke. Renamed the local copy to `verifyProcessorCopy` in the same commit with a comment block above the binding making the value-semantics intent explicit at the read site.

Libraries/MLXVLM/Models/Gemma4Assistant.swift
angelsbrood: Went with option A (parameter-pass). Kept it simpler — `target` is now just a method argument to `draftBlock(...)`, no instance-level binding state involved. Fixed in `c6af858`.

Libraries/MLXVLM/Models/Gemma4Assistant.swift
angelsbrood: Right — confirmed via grep that `boundTargetLayerTypes` was written by `bind()` and never read. Removed it (and the other `bound*` fields) as part of dropping `bind()` per comment 2. `c6af858`.

Libraries/MLXLMCommon/MTPSpeculativeTokenIterator.swift
```diff
@@ -0,0 +1,418 @@
...
+        // in one forward call, emitting state for next round.
+        var verifyState = LMOutput.State()
+        verifyState[mtpEmitFlagKey] = true
+        let verifyTokens = concatenated([bonusToken, flatDraftTokens])
+        let verifyInput = LMInput.Text(tokens: verifyTokens)
+        let verifyStart = verifyInput.tokens.dim(0) - (numDraft + 1)
+        let mainResult = mainModel(
+            verifyInput[text: .newAxis], cache: mainCache, state: verifyState)
+        let mainLogits = mainResult.logits
+        mainState = mainResult.state
```
davidkoski, line 270: This saves the various states as the output of mainModel from verifying the tokens, e.g. `mtpSharedKVStatesKey`

however, see below (309)

Libraries/MLXLMCommon/MTPSpeculativeTokenIterator.swift
```diff
@@ -0,0 +1,418 @@
...
+        let finalToken = mainTokens[accepted ... accepted]
+        processor?.didSample(token: finalToken)
+        pendingTokens.append(mainTokensList[accepted])
+
+        proposedCount += numDraft
+        acceptedCount += accepted
+        lastRoundAccepted = accepted
+
+        // Rewind only the main cache by rejected count. Drafter has no cache.
+        trimPromptCache(mainCache, numTokens: numDraft - accepted)
```
davidkoski: This trims the cache back to the accepted tokens, but the `mtpSharedKVStatesKey` still holds the draft values.  I think that needs very similar logic to trimPromptCache.

RESEARCH-NOTES.md
davidkoski, line 1: I think this is specific to the PR and should be deleted for the merge.

davidkoski (commented):
This is large and pretty complex but I think it is time to merge it and let people use it -- aside from removing the implementation notes, I don't see any issues.

Libraries/MLXLMCommon/MTPSpeculativeTokenIterator.swift
angelsbrood: Confirmed — good catch. The emit happens inside the verify forward, so the snapshot spans the full `[bonus, d_1 ... d_n]` chunk before acceptance is known, and it crossed the trim boundary untouched: any round following a partial acceptance drafted while cross-attending over the rejected rows. mlx-vlm never has this seam because `set_shared_kv()` re-derives from the live cache before every `draft_block`; threading sharedKV as an emitted value (per #283's no-mutation invariant) is where the snapshot acquired a lifetime the cache rewind didn't cover.

Fixed in 34c586c: `trimSharedKVState(_:numTokens:)` rewinds the snapshot at the same site as `trimPromptCache`, fed by its return value so the two stay in lockstep, with a debug assertion at `draftBlock` consumption that per-entry span == cache offset (it held across every 31B integration round). c0f35a6 adds a synthetic regression test that forces a partial acceptance and asserts the span the drafter receives on the following round — it fails with the fix absent (observed span = offset + rejected) and passes at HEAD. Output was never affected (emitted tokens are always the main model's samples against the trimmed cache, and the verify pass gets a fresh state carrying only the emit flag), which is why the byte-identity suites couldn't see it; the observable is acceptance rate.

Re-measured on the hidden-slice validation prompt plus the two cross-prompt byte-identity prompts (maxTokens=64, temperature 0): aggregate acceptance moves 46.4% → 49.1% at blockSize=4 and 23.7% → 36.6% at blockSize=6. Five of six prompt×blockSize cells improve (up to +15.9pp). The exception is the sky-blue prompt at bs=4 (60.6% → 46.8%), which we characterized before posting: the stale rows are the drafter's own rejected tail, sitting at exactly the positions it is about to predict again and fully visible to its bidirectional attention — after a shallow rejection that amounts to a self-lookahead hint, and on that one highly formulaic completion the hint was unusually lucky. A feature-shuffle ablation on the pre-fix tree (stale span kept, phantom content destroyed) reproduces the post-fix trajectory exactly (36/77 at bs=4), so the rejected rows' content was carrying the entire effect, not the span change. Since the reference re-derives shared K/V from the live cache every round, that pre-fix number was inflation from the deviation rather than achievable correct behavior — and conditioning the next draft on tokens the target just rejected isn't semantics we want to keep by accident. If a draft-retry hint is worth having, it should be a deliberate, position-correct design; we're filing a follow-up issue to explore that post-merge.

`lastHidden` doesn't need the analogous trim — the accepted-index slice from the earlier fix selects by position and never reads the rejected rows.

Libraries/MLXLMCommon/MTPSpeculativeTokenIterator.swift
angelsbrood: Addressed together with the thread at the trim site below — https://github.com/ml-explore/mlx-swift-lm/pull/308#discussion_r3392566471.

angelsbrood:
Got it— removed in [a0aea6a](https://github.com/ml-explore/mlx-swift-lm/pull/308/commits/a0aea6a97ad29a3cc7af8dfb7b2881aed14882fa). Thank you for the careful reviews; the sharedKV catch in particular made this better, and the characterization it prompted is interesting enough that I'll file it as a follow-up issue after merge. Looking forward to people using it.

davidkoski (approved):
Thank you for building this!  Let's go!

PR #309, opened 2026-05-22
Make Gemma 4 per_layer_model_projection quantizable

davidkoski:
## Proposed changes

- use a normal Linear layer and multiply by scale:  matches gemma3n and mlx-lm

Replaces #244

## Checklist

### DIFF
--- a/Libraries/MLXLLM/Models/Gemma4Text.swift
+++ b/Libraries/MLXLLM/Models/Gemma4Text.swift
@@ -160,21 +160,6 @@ private class RMSNormNoScale: Module {
     }
 }
 
-private class ScaledLinear: Module {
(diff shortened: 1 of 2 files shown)


Libraries/MLXLLM/Models/Gemma4Text.swift
```diff
@@ -160,21 +160,6 @@ private class RMSNormNoScale: Module {
     }
 }
 
-private class ScaledLinear: Module {
```
davidkoski, line 163: Replace this with a normal Linear * scale

aleroot (approved):
Makes sense

angeloskath (approved):
👍

PR #310, opened 2026-05-22
Add Nemotron Labs Diffusion

smdesai:
## Proposed changes

This PR adds support for NVIDIA's [Nemotron-Labs-Diffusion](https://huggingface.co/nvidia/Nemotron-Labs-Diffusion-3B) model supporting auto-regressive, diffusion and self-speculating modes. 

## Checklist

https://github.com/user-attachments/assets/3706f715-2396-479e-9247-9b6c3d4feda3

### DIFF
--- a/Libraries/MLXLLM/Models/NemotronLabsDiffusion.swift
+++ b/Libraries/MLXLLM/Models/NemotronLabsDiffusion.swift
@@ -30,7 +30,7 @@ import MLXNN
 private func nlsLlama4AttentionScale(
     start: Int, stop: Int, beta: Float, maxPositionEmbeddings: Int
 ) -> MLXArray {
-    let positions = MLXArray(Int32(start) ..< Int32(stop))
+    let positions = arange(start, stop)
     let scaling =
         1 + beta
         * MLX.log(
@@ -486,6 +486,17 @@ extension NemotronLabsDiffusionModel {
     /// Mirrors `ar_generate` in `modeling_nemotron_labs_diffusion.py`. Returns
     /// the prompt concatenated with newly generated tokens, and the number of
     /// forward passes performed.
+    ///
+    /// > Note: This method is provided for **symmetry** with
+    /// > ``diffusionGenerate(promptIds:maxNewTokens:blockLength:threshold:eosTokenId:onStep:)``
+    /// > and ``linearSpecGenerate(promptIds:maxNewTokens:blockLength:threshold:eosTokenId:onAccept:)``
+    /// > so callers can pick a mode by method name. For real applications
+    /// > prefer driving AR generation through the standard pipeline
+    /// > (`ChatSession` / `ModelContainer.generate(...)`), which calls
+    /// > ``callAsFunction(_:cache:)`` token-by-token via `TokenIterator` and
+    /// > adds streaming, temperature / top-p / repetition-penalty sampling,
+    /// > chunked prompt prefill, sliding-window and quantized KV caches, and
+    /// > wired-memory ticket coordination — none of which this method does.
     public func arGenerate(
         promptIds: [Int],
         maxNewTokens: Int = 128,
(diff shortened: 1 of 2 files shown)


davidkoski:
I wonder if the Nemotron and LoRA code could be in two PRs?  They look independent.

Libraries/MLXLLM/Models/NemotronLabsDiffusion.swift
```diff
@@ -0,0 +1,848 @@
...
+// MARK: - Llama-4 Attention Scaling
+//
+// Identical to Mistral3Text.swift's helper but kept private here to keep the
+// file self-contained — the existing helper there is `private` so it cannot be
+// reused.
+
+private func nlsLlama4AttentionScale(
+    start: Int, stop: Int, beta: Float, maxPositionEmbeddings: Int
+) -> MLXArray {
+    let positions = MLXArray(Int32(start) ..< Int32(stop))
```
davidkoski: use `arange(start, stop)` instead -- that will be done on the GPU (may not be a big deal)

Libraries/MLXLLM/Models/NemotronLabsDiffusion.swift
```diff
@@ -0,0 +1,848 @@
...
+    /// Mirrors `generate(...)` in `modeling_nemotron_labs_diffusion.py` with
+    /// `causal_context = False`. The prompt is causally prefilled into a
+    /// shared KV cache once. Each block of `blockLength` tokens is initialized
+    /// to `mask_token_id` and refined in up to `blockLength` denoising steps:
+    /// at each step the block is forwarded bidirectionally, attending against
+    /// the cached prefix without re-running it. After each step the cache is
+    /// trimmed by `blockLength` so the next step writes over the same slots.
+    /// Once a block stabilizes, we run one causal forward over its finalized
+    /// tokens to commit their K/V into the cache for subsequent blocks.
+    public func diffusionGenerate(
```
davidkoski, line 570: Is this a generic technique required for diffusion models?  I wonder if we need a new DiffusionLanguageModel protocol and this becomes generic code?

Same question for arGenerate

What happens if you call the standard generate loop with callAsFunction?  Are you losing functionality?

smdesai: This is the first diffusion model in the repo so hard to say if it's generic. A DiffusionLanguageModel protocol could exist but it would need to be carefully scoped and hard to validate until there's a second or third implementation. 

The scoring/commit logic for Nemotron has some quirks, "fallback to a single most confident token if no candidate beats threshold" is a specific decision. A generic protocol may freeze one policy and possibly limit future ports. I'd hold off on this until we have a second model and re-visit. 

arGenerate was provided as a way of providing symmetry with diffusionGenerate / linearSpecGenerate. It duplicates the work that LLModel and TokenIterator already do. It can be removed and go through ChatSession, this is what the demo app does so same as generate loop with callAsFunction. 

For Diffusion/Self-spec we'd lost almost everything that makes them what they are as it has no idea about bi-directional logits or block-wise denoising. 

Removing arGenerate makes sense unless you'd like to keep it for symmetry purposes, perhaps documented and commented as such.  Happy either way.

Libraries/MLXLLM/Models/NemotronLabsDiffusion.swift
smdesai: Yes this is cleaner and I can make the change.

As for Nemotron and LoRA code as two separate PRs. Yes, they're independent. 
- LoRA changes → standalone, mergeable on their own
- Nemotron port → depends on the LoRA changes

I'm happy if that's cleaner.

davidkoski: A little easier to review smaller pieces and I can get the LoRA change reviewed and merged faster than the nemotron code.

smdesai: PR #316 now contains the LoRA changes. This PR will be updated accordingly once that's merged.

Libraries/MLXLLM/Models/NemotronLabsDiffusion.swift
davidkoski: Yeah, as the first diffusion model I agree we are in uncharted territory.  I think adding all of it with a comment explaining _why_ it is that way will be fine.  If/when we pick up another diffusion model we can revisit.

Libraries/MLXLLM/Models/NemotronLabsDiffusion.swift
smdesai: Using arange(start, stop) now.

Libraries/MLXLLM/Models/NemotronLabsDiffusion.swift
smdesai: arGenerate commented and documented and also added a comment around DiffusionLanguageModel protocol .

smdesai: @davidkoski Changes applied so ready for review.

davidkoski (approved):
Thank you for the contribution -- since it is the first diffusion model I think we should take it as-is (per discussion) and see how things play out.  We can refactor as needed.

PR #313, opened 2026-05-27
Add structured ChatSession continuation

aleroot:
## Proposed changes

Adds structured `ChatSession` continuation APIs so callers can append one or more `Chat.Message` values directly to an existing session. This lets agent and tool-call loops continue a live cached session with `.tool`, `.assistant`, or `.user` messages without rebuilding a new session from the full accumulated transcript each turn.

### Motivation

A `ChatSession` initialized with history has to tokenize and prefill that history once to build the KV cache. Before this change, callers that needed to append structured tool messages often had to rebuild accumulated history externally, which could repeatedly re-apply the chat template, re-tokenize, and prefill the entire prompt across tool turns.

The new overloads expose the natural continuation path: keep using the same `ChatSession`, append structured messages, and preserve the existing KV cache.

## Checklist


davidkoski (approved):
Looks good, thank you!

PR #314, opened 2026-05-27
Add speculative decoding telemetry and memory gating

aleroot:
## Proposed changes

Speculative decoding is workload and hardware-sensitive. A draft model can look good in theory but fail to pay off if acceptance is low or if the extra model pressure is too high.

The change set prepares the project for edge-aware speculative decoding by adding the two missing foundations:

1. observe speculative decoding quality/performance
2. avoid applying auxiliary-model speculation when memory pressure makes it likely to hurt

### Telemetry
It exposes whether speculative decoding is actually helping at runtime. Callers can now inspect acceptance rate, draft tokens, target verification calls, emitted tokens, and average emitted tokens per target call through GenerateCompletionInfo.speculativeDecodingTelemetry.

### Memory-Aware Gating
ChatSession now defaults speculative decoding to a memory policy based on GPU.maxRecommendedWorkingSetBytes(). Before this, enabling speculative decoding meant loading/running the main model plus draft model whenever requested. On memory-constrained systems, that can make things slower or less stable.

## Checklist

### DIFF
--- a/Libraries/MLXLMCommon/Evaluate.swift
+++ b/Libraries/MLXLMCommon/Evaluate.swift
@@ -496,6 +496,13 @@ public protocol TokenIteratorProtocol: Sequence, IteratorProtocol where Element 
     var maxTokens: Int? { get }
     var tokenCount: Int { get }
     var promptPrefillTime: TimeInterval { get }
+    var speculativeDecodingTelemetry: SpeculativeDecodingTelemetry? { get }
+    mutating func discardGeneratedToken()
+}
+
+extension TokenIteratorProtocol {
+    public var speculativeDecodingTelemetry: SpeculativeDecodingTelemetry? { nil }
+    public mutating func discardGeneratedToken() {}
 }
 
 /// Generator of tokens.
@@ -748,7 +755,7 @@ public struct SpeculativeTokenIterator: TokenIteratorProtocol {
     var processor: LogitProcessor?
     let sampler: LogitSampler
 
-    public var tokenCount = 0
+    public var tokenCount: Int { telemetry.emittedTokenCount }
     public let maxTokens: Int?
     let numDraftTokens: Int
 
@@ -758,6 +765,14 @@ public struct SpeculativeTokenIterator: TokenIteratorProtocol {
 
     // Internal metrics
     public var promptPrefillTime: TimeInterval = 0.0
+    private var telemetry = SpeculativeDecodingTelemetry()
+    public var speculativeDecodingTelemetry: SpeculativeDecodingTelemetry? {
+        telemetry.roundCount > 0 ? telemetry : nil
+    }
+
+    public mutating func discardGeneratedToken() {
+        telemetry.discardGeneratedToken()
+    }
 
     /// Initialize a `SpeculativeTokenIterator` with the given input.
     ///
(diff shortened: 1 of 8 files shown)


Libraries/MLXLMCommon/Evaluate.swift
```diff
@@ -756,6 +763,14 @@ public struct SpeculativeTokenIterator: TokenIteratorProtocol {
 
     // Internal metrics
     public var promptPrefillTime: TimeInterval = 0.0
+    private var telemetry = SpeculativeDecodingTelemetry()
+    public var speculativeDecodingTelemetry: SpeculativeDecodingTelemetry? {
+        telemetry.roundCount > 0 ? telemetry : nil
+    }
+
+    public mutating func discardGeneratedToken() {
+        telemetry.discardGeneratedToken()
```
davidkoski, line 774: Should this mutate the iterator's own `tokenCount`?  Should we have a single source of truth, the telemetry?

I think the count is just informative, so not a correctness issue.

Libraries/MLXLMCommon/SpeculativeDecoding.swift
```diff
@@ -0,0 +1,206 @@
...
+
+private func defaultSpeculativeDecodingMemoryLimit() -> Int? {
+    guard let bytes = GPU.maxRecommendedWorkingSetBytes(), bytes > 0 else {
+        return nil
+    }
+    return bytes
+}
+
+/// Runtime counters for a speculative decoding pass.
+public struct SpeculativeDecodingTelemetry: Sendable, Equatable {
```
davidkoski, line 14: Probably worth an issue: MTPSpeculativeTokenIterator should probably have this integrated as well.

davidkoski:
I think this is ready to go but the GitHub web ui claims there are conflicts (SpeculativeDecodingTests.swift), however when I click on "resolve conflicts" it tells me there are zero. 😕

davidkoski:
a rebase _does_ show a conflict, which is strange since you squashed it down to 1 commit -- the rebase and the merge that the editor use should be doing the same thing.

davidkoski:
OK, mystery solved -- the "resolve conflicts" button does this:

```
git merge -X diff-algorithm=patience origin/main
```

which is able to cleanly resolve the merge, while `myers` (which seems to be the default strategy) cannot.  Looking at the change locally and if that is clean I will just do the resolve conflicts via the web editor.

davidkoski:
Two tests are failing with the merge but do not fail on main:

- textOnlyPrefillChunkSizeInvariant
- `Chunked and single-pass prefill agree on logits and cache offsets`

curious because these tests are both calling `model.prepare()` with different window sizes -- a code path that isn't touched in this PR.

aleroot:
Rebased and kept upstream’s stable-logit oracle, kept Gemma3 as smoke coverage, and preserved the telemetry assertions.

davidkoski:
Cool, I think this unit test failure on my side is some environmental issue -- I have two checkouts and one works and the other does not.

CI underway.

davidkoski:
Oh no, this is more TF32 business.  I had the env variable set in one of those checkouts.  OK, not related to this PR -- I will file an issue.

#357

davidkoski (approved):
Thank you for this improvement!

PR #316, opened 2026-05-27
 LoRA: runtime toggle and PEFT adapter loader

smdesai:
## Proposed changes

This PR adds two related, backward-compatible improvements to the LoRA infrastructure in MLXLMCommon.

**Runtime loraEnabled toggle on LoRALayer**
A new loraEnabled: Bool property on the LoRALayer protocol lets callers enable or disable the LoRA term at runtime without unloading the adapter. When false, the layer behaves as the underlying base layer (no LoRA term added).

This is needed for inference patterns that interleave LoRA-on and LoRA-off forward passes against the same model — for example, speculative decoding schemes where a LoRA-tuned drafter feeds an un-tuned verifier with a shared KV cache. Today the only way to "disable" a loaded adapter is to unload and reload it, which is too expensive to do per inference step.

Backward compatibility: Strictly additive. The default value is true (LoRA always applied, matching pre-PR behavior). External LoRALayer conformers compile unchanged because the protocol-extension default satisfies the new requirement; their toggle is silently a no-op until they opt in by adding their own stored property.

**LoRAContainer.fromPEFT(directory:) — load HuggingFace PEFT adapters**
 A new static loader on LoRAContainer reads adapter directories in the standard HuggingFace peft format:
- adapter_config.json (PEFT schema: r, lora_alpha, target_modules, peft_type, …)
- adapter_model.safetensors (with base_model.model.<path>.lora_A.weight / lora_B.weight keys)

These changes are extracted from PR #310 where the model's speculative-decoding mode requires per-phase LoRA toggling and the canonical adapter ships in PEFT format. 

## Checklist


Libraries/MLXLMCommon/Adapters/LoRA/DoRA+Layers.swift
```diff
@@ -109,6 +110,10 @@ public class DoRALinear: Linear, LoRALayer {
 
     public override func callAsFunction(_ x: MLXArray) -> MLXArray {
         let y = matmul(x, weight.T)
+        if !loraEnabled {
+            if let bias { return y + bias }
+            return y
```
davidkoski: I wonder if this should call super?  Linear is simple enough and unlikely to change, but since it is a subtype it might be better to call that way.  It looks like `LoRALinear` does it that way -- I think that is the pattern to follow

Libraries/MLXLMCommon/Adapters/LoRA/PEFTAdapter.swift
```diff
@@ -0,0 +1,161 @@
...
+//     base_model.model.<inner-path>.<target_module>.lora_A.weight  shape [r, in]
+//     base_model.model.<inner-path>.<target_module>.lora_B.weight  shape [out, r]
+//
+// MLX format (LoRALinear / QLoRALinear):
+//   loraA: [in, r],  loraB: [r, out],  scale = lora_alpha / r
+//
+// Conversion:
+//   - Strip the leading `base_model.model.` prefix.
+//   - Rename `.lora_A.weight` -> `.lora_a`,  `.lora_B.weight` -> `.lora_b`.
+//   - Transpose both tensors to match MLX's [in, r] / [r, out] convention.
```
davidkoski: I wonder if this block comment should be on fromPEFT?  As it is you can only see it in the code, not in the built docs.

Libraries/MLXLMCommon/Adapters/LoRA/PEFTAdapter.swift
```diff
@@ -0,0 +1,161 @@
...
+        // Sort for determinism in tests.
+        let keys = Array(seenTargetKeys).sorted()
+        return (renamed, keys)
+    }
+
+    private static func stripLayerPrefix(_ path: String) -> String? {
+        // Match "<encoder|model>.layers.<n>." then return the rest. This
+        // matches the two common backbone layouts the project uses.
+        let parts = path.split(separator: ".", omittingEmptySubsequences: false)
+        for i in 0 ..< (parts.count - 2) {
```
davidkoski: if parts.count is 0 or 1 this will trap -- might want to guard vs that and return nil.

smdesai: Good catch. Yes a guard needs to be in place. It's added.

Libraries/MLXLMCommon/Adapters/LoRA/PEFTAdapter.swift
smdesai: Good catch. It's moved to fromPEFT.

Libraries/MLXLMCommon/Adapters/LoRA/DoRA+Layers.swift
smdesai: Yes you're right, it can and updated accordingly.

davidkoski (commented):
Looks good, thank you!

davidkoski (approved):

Issue #312, opened 2026-05-26
[BUG] maybeQuantizeKVCache silently corrupts context mid-generation


somegeekintn:
_Description created with a little help from Claude since I'm not the best explainer. Hopefully this covers everything_

**Summary**

`maybeQuantizeKVCache` is called on every step inside `TokenIterator`'s generation loop. When the `quantizedKVStart` threshold is crossed mid-generation, it replaces elements in `TokenIterator`'s local copy of the cache array with new `QuantizedKVCache` instances. Because the function takes `cache: inout [KVCache]`, it replaces array elements rather than mutating the cache objects in place. The caller's array (in `ChatSession`) still holds the original `KVCacheSimple` references, which are unaware of any tokens generated after the quantization threshold was crossed. The quantized caches in `TokenIterator` are never surfaced back to the caller and are deallocated at the end of generation.

**Effect**

The model loses all context generated after the quantization threshold. From the model's perspective, that portion of the conversation never happened.

**To Reproduce**

Set `kvBits` to any non-nil value. Ask the model to generate a long response (enough tokens to cross `quantizedKVStart`). In a follow-up turn, ask the model to reference something specific from that response. The model will have no recollection of it.

**Additional notes**

- The `!(firstQuantizable is QuantizedKVCache)` check on line 1806 is dead code — `cache.first(where: { $0 is KVCacheSimple })` can never return a `QuantizedKVCache` instance since `QuantizedKVCache` inherits from `BaseKVCache`, not `KVCacheSimple`.
- `maybeQuantizeKVCache` is called on every generation step, scanning all cache layers each time, even after quantization has already occurred and no `KVCacheSimple` instances remain.
- A minimal workaround is to expose `cache` as `public var` on `TokenIterator` so callers can retrieve the post-generation array. A more complete fix would require rethinking cache ownership so replacements propagate back to the caller.
- Also this was tested with a custom version of `ChatSession` but I believe the same behavior will occur with the "stock" version.

**Desktop (please complete the following information):**
 - OS Version: [e.g. MacOS 26.5]
 - Device: [MacBook Pro M5 Po 24GB]
 - Version: Not present in the last release which was broken in other ways. Last modified in commit 88beb40b.

davidkoski:
See also https://github.com/ml-explore/mlx-swift-examples/issues/440 -- I think that is similar, though perhaps a layer down.

somegeekintn:
[...]

PR #317, opened 2026-05-27
Fix consistent float32 dtype for gated delta SSM operations in Qwen3.5 models

somegeekintn:
In `gatedDeltaUpdate`, which is used by `Qwen35GatedDeltaNet`, `g` and `beta` inherited their dtype from the input tensor (`bfloat16`), while the SSM state is correctly maintained in `float32`. Additionally, `computeGatedDeltaG` computed in `float32` for precision but immediately downcasted back to input dtype, discarding that precision.

This dtype inconsistency caused MLX to compile the `gated_delta_step` Metal kernel with `bfloat16` for `g` and `beta` on the first generation turn. On subsequent turns where the computation graph differed (e.g. when reusing a `QuantizedKVCache` across turns), the correct `float32` dtype was used, causing a kernel source mismatch. MLX's `clear_library` would evict the old compiled pipeline while an `asyncEval` command buffer was still using it, triggering a Metal API validation crash.

**Changes**

Cast `beta` to `float32` explicitly in `gatedDeltaUpdate`, and return `float32` directly from `computeGatedDeltaG` rather than downcasting back to input dtype. I believe this matches the Python mlx-lm reference implementation, which uses `StT` (float32) consistently for all SSM computations.

Also reconciled recent changes from the Python counterpart (`gated_delta.py`): [#997](https://github.com/ml-explore/mlx-lm/pull/997) and [#1072](https://github.com/ml-explore/mlx-lm/pull/1072).

Note: discovered this issue while attempting various solutions for #312 This fix is a pre-requisite.

## Checklist

### DIFF
--- a/Libraries/MLXLMCommon/GatedDelta.swift
+++ b/Libraries/MLXLMCommon/GatedDelta.swift
@@ -13,7 +13,7 @@ import MLXNN
 
 func computeGatedDeltaG(_ aLog: MLXArray, _ a: MLXArray, _ dtBias: MLXArray) -> MLXArray {
     let decay = exp(-exp(aLog.asType(.float32)) * softplus(a + dtBias))
-    return decay.asType(a.dtype)
+    return decay
 }
 
 // MARK: - Metal Kernel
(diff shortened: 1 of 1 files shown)


davidkoski (approved):
Nice find, thank you!

PR #318, opened 2026-05-28
Add Swift model conversion API

### LINKED ISSUE #266: Add convert API for in-Swift model quantization
## Summary

Proposing the addition of a `convert` function to `MLXLMCommon` that downloads, quantizes, and saves a model entirely in Swift. Equivalent to Python's `mlx_lm.convert` CLI.

## Motivation

Today, quantizing a model for use with `mlx-swift-lm` requires running `mlx_lm.convert` in Python, then loading the output in Swift. For Swift-only workflows (CLI tools, on-device pipelines, app-bundled conversion), this means maintaining a Python toolchain just for one step.

The primitives already exist:
- `MLXNN.quantize(model:groupSize:bits:mode:)` for in-place quantization
- `MLX.save(arrays:metadata:url:)` for safetensors output
- `LLMModelFactory` + `Downloader` for fetching MLX-format sources
- Per-model `sanitize(weights:metadata:)` for preprocessing

What's missing is a function that wires them together and handles `config.json` rewriting + tokenizer file passthrough.

## Proposed API

```swift
public func convert(
    from sourceId: String,
    to outputDirectory: URL,
    bits: Int = 4,
    groupSize: Int = 64,
    mode: QuantizationMode = .affine,
    downloader: any Downloader,
    progressHandler: @Sendable @escaping (Progress) -> Void = { _ in }
) async throws
```

Behavior:
1. Download source repo via the provided `Downloader`.
2. Load model + weights using `LLMModelFactory` (so `sanitize` runs).
3. Apply `quantize(model:groupSize:bits:mode:)`.
4. Save quantized weights as `model.safetensors` in `outputDirectory`.
5. Read source `config.json`, inject `quantization: {group_size, bits}` block, write to output.
6. Copy tokenizer files (`tokenizer.json`, `tokenizer_config.json`, `*.jinja`, etc.) to output.

## Scope (v1)

**In:**
- MLX-format source models (already converted from PyTorch)
- Affine quantization, uniform across all layers
- Local output directory

**Out (future PRs):**
- PyTorch `.bin` source conversion
- Per-layer quantization config (`PerLayerQuantization`)
- HF Hub upload helper

## Open questions

1. Should this live in `MLXLMCommon` or a new `MLXLMConvert` target?
2. Naming: `convert` vs `quantizeAndSave` vs other?
3. Should the API also expose a lower-level `quantize(model:bits:groupSize:)` + `save(model:to:)` pair for users who want to control the flow?
4. Is there appetite for this in the repo, or is the Python tool considered the canonical path?

Happy to put up a PR if there's interest. Wanted to align on scope and API shape first.

aleroot:
## Proposed changes

Adds a basic Swift model conversion API for quantizing safetensors-backed LLMs through MLXLMCommon and LLMModelFactory. Fix #266 .

## Checklist

### DIFF
--- a/Libraries/MLXLMCommon/ModelConversion.swift
+++ b/Libraries/MLXLMCommon/ModelConversion.swift
@@ -1,0 +1,845 @@
+// Copyright © 2026 Apple Inc.
+
+import Foundation
+import MLX
+import MLXNN
+
+package let modelConversionSidecarPatterns = [
+    "*.json", "*.jsonl", "*.jinja", "*.txt", "*.model", "*.tiktoken", "*.py",
+]
+package let modelConversionDownloadPatterns = ["*.safetensors"] + modelConversionSidecarPatterns
+
+/// Quantization settings for one conversion pass or layer.
+public struct ModelConversionQuantization: Sendable, Equatable {
+    /// Quantization bit depth.
+    ///
+    /// `nil` lets MLX use the default for the selected quantization mode.
+    public var bits: Int?
+
+    /// Quantization group size.
+    ///
+    /// `nil` lets MLX use the default for the selected quantization mode.
+    public var groupSize: Int?
+
+    /// Quantization mode.
+    public var mode: QuantizationMode
+
+    public init(bits: Int? = nil, groupSize: Int? = nil, mode: QuantizationMode = .affine) {
+        self.bits = bits
+        self.groupSize = groupSize
+        self.mode = mode
+    }
+}
+
+/// Per-layer quantization decision.
+public enum ModelConversionQuantizationDecision: Sendable, Equatable {
+    /// Do not quantize this layer.
+    case skip
+
+    /// Quantize this layer, using the global conversion settings when the associated value is nil.
+    case quantize(ModelConversionQuantization? = nil)
+}
+
+/// Predicate used to customize layer quantization.
+public typealias ModelConversionQuantizationPredicate =
+    @Sendable (
+        _ path: String, _ module: Module
+    ) -> ModelConversionQuantizationDecision
+
+/// Options controlling model conversion.
+public struct ModelConversionOptions: Sendable {
+    /// Default quantization settings.
+    public var quantization: ModelConversionQuantization
+
+    /// Maximum safetensors shard size in bytes.
+    public var maxShardSize: Int64
+
+    /// Remove and recreate the output directory if it already exists.
+    public var overwriteExistingOutput: Bool
+
+    /// Optional per-layer quantization customization.
+    public var quantizationPredicate: ModelConversionQuantizationPredicate?
+
+    public var bits: Int? {
+        get { quantization.bits }
+        set { quantization.bits = newValue }
+    }
+
+    public var groupSize: Int? {
+        get { quantization.groupSize }
+        set { quantization.groupSize = newValue }
+    }
+
+    public var mode: QuantizationMode {
+        get { quantization.mode }
+        set { quantization.mode = newValue }
+    }
+
+    public init(
+        bits: Int? = nil,
+        groupSize: Int? = nil,
+        mode: QuantizationMode = .affine,
+        maxShardSize: Int64 = 5 * 1024 * 1024 * 1024,
+        overwriteExistingOutput: Bool = false,
+        quantizationPredicate: ModelConversionQuantizationPredicate? = nil
+    ) {
+        self.quantization = .init(bits: bits, groupSize: groupSize, mode: mode)
+        self.maxShardSize = maxShardSize
+        self.overwriteExistingOutput = overwriteExistingOutput
+        self.quantizationPredicate = quantizationPredicate
+    }
+}
+
+/// Coarse conversion stage for progress callbacks.
+public enum ModelConversionStage: String, Sendable {
+    case downloading
+    case copyingFiles
+    case loadingWeights
+    case quantizing
+    case savingWeights
+    case updatingConfiguration
+}
+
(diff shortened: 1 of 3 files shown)


Libraries/MLXLMCommon/ModelConversion.swift
```diff
@@ -0,0 +1,713 @@
...
+    /// Quantization bit depth.
+    public var bits: Int
+
+    /// Quantization group size.
+    public var groupSize: Int
+
+    /// Quantization mode.
+    public var mode: QuantizationMode
+
+    public init(bits: Int = 4, groupSize: Int = 64, mode: QuantizationMode = .affine) {
```
davidkoski: I think `bits` and `groupSize` should be nullable and have nil defaults.  The lower level mlx code will use mode appropriate defaults in that case.

Libraries/MLXLMCommon/ModelConversion.swift
```diff
@@ -0,0 +1,713 @@
...
+    bits: Int,
+    mode: QuantizationMode
+) -> Module? {
+    if layer is Quantized {
+        return nil
+    }
+
+    if let linear = layer as? Linear {
+        let (weight, scales, biases) = MLX.quantized(
+            linear.weight, groupSize: groupSize, bits: bits, mode: mode)
```
davidkoski: Not an issue with this PR, but something I noticed when I was looking at the bits/groupSize and defaults -- nvfp4 uses `globalScale`.  The QuantizedLinear and friends don't have this as a property, so using it here wouldn't make sense, but I wonder if this is a gap?  It matches the python implementation, so perhaps not something people actually use?

Libraries/MLXLMCommon/ModelConversion.swift
```diff
@@ -0,0 +1,713 @@
...
+
+    progressHandler(.init(stage: .loadingWeights))
+    try loadWeights(
+        modelDirectory: modelDirectory,
+        model: model,
+        perLayerQuantization: perLayerQuantization)
+
+    progressHandler(.init(stage: .quantizing))
+    let layerQuantization = quantizeForModelConversion(model: model, options: options)
+    eval(model)
```
davidkoski, line 216: claude pointed this out: mlx-lm's version of the conversion does an optional dtype conversion.

```python
MODEL_CONVERSION_DTYPES = ["float16", "bfloat16", "float32"]

    parser.add_argument(
        "--dtype",
        help="Type to save the non-quantized parameters. Defaults to config.json's `torch_dtype` or the current model weights dtype.",
        type=str,
        choices=MODEL_CONVERSION_DTYPES,
        default=None,
    )


    if dtype is None:
        dtype = config.get("torch_dtype", None)
    if dtype is None and (text_config := config.get("text_config", None)):
        dtype = text_config.get("dtype", None)
    if dtype in MODEL_CONVERSION_DTYPES:
        print("[INFO] Using dtype:", dtype)
        dtype = getattr(mx, dtype)
        cast_predicate = getattr(model, "cast_predicate", lambda _: True)

        def set_dtype(k, v):
            if cast_predicate(k) and mx.issubdtype(v.dtype, mx.floating):
                return v.astype(dtype)
            else:
                return v

        model.update(tree_map_with_path(set_dtype, model.parameters()))
```

mlx-swift doesn't have a `issubdtype` so we would need to add that first anyway -- is this a case worth having?  I think we can add it in after in the `ModelConversionOptions` if we need it.

Libraries/MLXLMCommon/ModelConversion.swift
```diff
@@ -0,0 +1,713 @@
...
+        {
+            try FileManager.default.removeItem(at: file)
+        }
+    }
+}
+
+private func quantizeForModelConversion(
+    model: BaseLanguageModel,
+    options: ModelConversionOptions
+) -> [String: ModelConversionQuantization] {
```
davidkoski: I wonder if this needs to be `[String:ModelConversionQuantizationDecision]`?  If the predicate indicates a `skip` then it will omit the layer from the dictionary.  This means "use default", not "skip".

Libraries/MLXLMCommon/ModelConversion.swift
aleroot: Good catch. This points to a real gap. I have the problem even in my app where I have extracted the code from as this is something that needs to be implemented in the mlx-swift library first, I will look into that next few days as I may need to convert in that way. Thank you.

Libraries/MLXLMCommon/ModelConversion.swift
aleroot: This is a valuable comment as well, I need to look into this. So far I had no conversion issues but it is better to implement this as there may be edge cases where the source safetensors are FP32 while the config expects float16 or bfloat16. In that case the converted model would likely still load, but it could keep non-quantized parameters at the wrong dtype.

Libraries/MLXLMCommon/ModelConversion.swift
aleroot: I have added

Libraries/MLXLMCommon/ModelConversion.swift
aleroot: This is probably good to go for the moment, later I can open another PR when the pieced are in place in the mlx-swift new release.

davidkoski (approved):
Looks good, thank you!

Maybe we need directory with an xcodeproj (along the lines of the integration tests) for tools.  A conversion tool and maybe a simple version of llm-tool from the examples repo might be handy.  Being able to run these things in the same repo is convenient.

Or maybe we need a conversion tool in the examples repo?

PR #322, opened 2026-05-30
Add support for Qwen3-VL-MoE models

lucasnewman:
## Proposed changes

This adds support for the Qwen3-VL-MoE model family in addition to the Qwen3-VL models already supported. I also fixed the GELU approximation type for both models, which is designed to be config-driven.

I verified I get valid output in an integration test:

```
Prompt: what do you see?

Response:
Based on the picture the image shows a traditional Japanese painting, likely a hanging scroll or byōdō, depicting a vibrant scene of morning glory flowers (asagao) in full bloom. The key elements are:

- **Flowering Vines:** The central focus is on lush, dark green vines with clusters of deep blue, trumpet-shaped flowers. The flowers are depicted in full bloom, with some facing upwards and others hanging, creating a sense of natural movement.
- **Background:** The entire scene is set against a warm, golden-orange background, which is characteristic of traditional Japanese art. This background is divided into vertical panels, suggesting the painting is on a folding screen or a piece of paper.
- **Composition:** The composition is dynamic, with the vines and flowers flowing diagonally across the frame, creating a sense of life and vitality. The dark blue of the flowers contrasts sharply with the golden background, making them stand out.
- **Style:** The painting is done in a traditional Japanese style, likely with ink and color on paper, emphasizing the natural beauty of the flowers.
```

## Checklist

- [X] I have read the [CONTRIBUTING](https://github.com/ml-explore/mlx/blob/main/CONTRIBUTING.md) document

### DIFF
--- a/Libraries/MLXVLM/Models/Qwen3VL.swift
+++ b/Libraries/MLXVLM/Models/Qwen3VL.swift
@@ -349,7 +349,7 @@ public struct Qwen3VLConfiguration: Codable, Sendable {
         private let _inChannels: Int?
         public var inChannels: Int { _inChannels ?? 3 }
         private let _hiddenAct: String?
-        public var hiddenAct: String { _hiddenAct ?? "gelu" }
+        public var hiddenAct: String { _hiddenAct ?? "gelu_pytorch_tanh" }
         private let _deepstackVisualIndexes: [Int]?
         public var deepstackVisualIndexes: [Int] { _deepstackVisualIndexes ?? [] }
 
@@ -455,6 +455,17 @@ public struct Qwen3VLConfiguration: Codable, Sendable {
 // MARK: - Vision
 
 enum Qwen3VLVision {
+
+    private static func geluApproximation(for hiddenAct: String) -> GELU.Approximation {
+        switch hiddenAct {
+        case "gelu_pytorch_tanh", "gelu_new":
+            .tanh
+        case "gelu_fast":
+            .fast
+        default:
+            .none
+        }
+    }
 
     static func rotateHalf(_ x: MLXArray) -> MLXArray {
         let half = x.dim(-1) / 2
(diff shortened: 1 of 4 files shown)


Libraries/MLXVLM/Models/Qwen3VL.swift
```diff
@@ -570,10 +581,11 @@ enum Qwen3VLVision {
         @ModuleInfo(key: "linear_fc2") var linear2: Linear
         @ModuleInfo(key: "act") var activation: GELU
 
-        init(dim: Int, hiddenDim: Int) {
+        init(dim: Int, hiddenDim: Int, hiddenAct: String) {
             _linear1.wrappedValue = Linear(dim, hiddenDim, bias: true)
             _linear2.wrappedValue = Linear(hiddenDim, dim, bias: true)
-            _activation.wrappedValue = GELU(approximation: .fast)
+            _activation.wrappedValue = GELU(
+                approximation: Qwen3VLVision.geluApproximation(for: hiddenAct))
```
davidkoski, line 695: Curiously it looks like this was swapped in the mlx-vlm MOE code.  In the HF transformers code both types of models use exact/tanh for PatchMerger/VisionMLP.  mlx-vlm MOE seems to swap them (and then this code is out of sync with that, but is actually correct per the HF reference code).

davidkoski:
Rebase to main to hopefully get CI to work.

davidkoski (approved):
Changes look good, thank you!

Libraries/MLXVLM/Models/Qwen3VL.swift
davidkoski: And it fixes a bug in the non MOE code in that it ignored the config :-)

davidkoski:
Just waiting on CI.

PR #323, opened 2026-05-30
Improve Qwen3.5 recurrent cache handling

aleroot:
## Proposed changes

Qwen3.6 models using `model_type: qwen3_5` rely heavily on hybrid linear-attention / GatedDelta layers. Keeping the convolution state contiguous avoids carrying strided slices through decode steps, which should reduce per-token cache overhead and better match upstream MLX Python behavior.

Advancing array-cache metadata and preserving left-padding masks also improves stability for padded or batched generation paths.

## Checklist

### DIFF
--- a/Libraries/MLXLLM/Models/Qwen35.swift
+++ b/Libraries/MLXLLM/Models/Qwen35.swift
@@ -251,7 +251,7 @@ final class Qwen35GatedDeltaNet: Module {
 
         let convInput = concatenated([convState, qkv], axis: 1)
         if let cache {
-            cache[0] = convInput[0..., (-(convKernelSize - 1))...]
+            cache[0] = contiguous(convInput[0..., (-(convKernelSize - 1))..., 0...])
         }
 
         let convOut = silu(conv1d(convInput))
@@ -287,6 +287,7 @@ final class Qwen35GatedDeltaNet: Module {
 
         if let cache {
             cache[1] = state
+            cache.advance(S)
         }
 
         out = norm(out, gate: z)
--- a/Libraries/MLXLMCommon/KVCache.swift
+++ b/Libraries/MLXLMCommon/KVCache.swift
@@ -85,12 +85,46 @@ public protocol KVCache: Evaluatable {
 
     /// Create an independent deep copy of this cache.
     func copy() -> any KVCache
+
+    /// Prepare cache metadata for a batched sequence.
+    func prepare(lengths: [Int]?)
+
+    /// Prepare cache metadata for a batched sequence.
+    func prepare(lengths: MLXArray?)
+
+    /// Clear transient cache metadata after generation.
+    func finalize()
 }
 
 extension KVCache {
     public var ropeOffset: RoPEOffset {
         .scalar(offset)
     }
+
+    public func prepare(lengths: [Int]?) {}
+
+    public func prepare(lengths: MLXArray?) {}
+
+    public func finalize() {}
+}
+
+public func withPreparedCache<Result>(
+    _ cache: [any KVCache],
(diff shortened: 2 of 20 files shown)


Libraries/MLXLMCommon/KVCache.swift
```diff
@@ -1157,9 +1157,17 @@ public class ArraysCache: BaseKVCache {
         leftPadding = nil
     }
 
+    /// Advance sequence-relative metadata after consuming `N` tokens.
+    public func advance(_ N: Int) {
```
davidkoski: This looks ok, but the python code:

```python
    def advance(self, N):
        if self.lengths is not None:
            self.lengths -= N
        if self.left_padding is not None:
            self.left_padding -= N
```

has a `lengths` property and doesn't touch`'offset`.  Perhaps we need an effort to fully sync the python and swift versions of these KVCaches.

Libraries/MLXLLM/Models/Qwen35.swift
```diff
@@ -287,6 +287,7 @@ final class Qwen35GatedDeltaNet: Module {
 
         if let cache {
             cache[1] = state
+            cache.advance(S)
```
davidkoski, line 290: FalconH1 also uses this cache -- does it need the same treatment?

davidkoski: Oh, actually anything that uses MambaCache is also using an ArrraysCache -- do they also need it?

aleroot: Yes, indeed. I have tried to cover them all, have look I hope I have not missed anything

aleroot:
Updated rebasing on latest main , also fixed a MTP tail-budget issue by making speculative rounds budget-aware: the iterator now only drafts when the remaining output budget can hold both the accepted draft prefix and the verifier correction/bonus.

Libraries/MLXLMCommon/KVCache.swift
```diff
@@ -93,6 +93,43 @@ extension KVCache {
...
+        }
+    }
+
+    fileprivate func finalizeArrayMetadata() {
+        if let arrays = self as? ArraysCache {
+            arrays.finalize()
+        } else if let list = self as? CacheList {
+            list.finalize()
+        }
+    }
```
davidkoski: I wonder if this should be done through either methods on the KVCache protocol with default (empty) implementations?  Or add a KVCacheLifecycle (or something) protocol.  I worry that new cache types could get added (we already have quite a few) and this particular check wouldn't be updated.

davidkoski (commented):
Looks good.  See what you think about my question on KVCache finalize.

Libraries/MLXLMCommon/KVCache.swift
aleroot: Agree. Moved this to KVCache lifecycle hooks with no-op defaults, made CacheList forward via the protocol, added coverage, and fixed the format issue. Thanks.

aleroot:
Had to fix the the CI failure which was coming from `SpeculativeDecodingTests`, specifically speculative generation diverging from normal generation.

There were two issues:

1. The parameterized Swift Testing cases were running concurrently, while this suite uses shared MLX/model generation state. I marked the suite `.serialized` so the cases run one at a time.

2. The repetition/presence penalty processors were mutating logits via indexed assignment. In the speculative verification path those logits can be views into a larger multi-token logits tensor, so that could affect later verification positions. I changed those processors to return out-of-place adjusted logits using `takeAlong`/`putAlong`.

davidkoski (approved):
Looks good, thank you!

PR #324, opened 2026-05-31
Moe refactoring -> reduces repeated small graph construction overhead

aleroot:
## Proposed changes

This PR reduces repeated small graph construction overhead in shared MoE paths.

Many MoE models combine routed expert outputs with router scores using the same expression:

```swift
(outputs * weights[..., .newAxis]).sum(axis: -2)
```

This PR centralizes that pattern behind a shared compiled helper, `weightedExpertSum`, and updates MoE model implementations to use it. It also adds a compiled default-SiLU product path for `SwitchGLU` / `FusedGateUpSwitchGLU`, covering the common:

```swift
silu(gate) * up
```

case while preserving the existing generic path for custom activations.

The intent is to reduce Swift/MLX dispatch and graph setup overhead in MoE-heavy models without changing model math, weights, quantization, cache behavior, or dependency versions.

## Checklist

### DIFF
--- a/Libraries/MLXLLM/Models/AfMoE.swift
+++ b/Libraries/MLXLLM/Models/AfMoE.swift
@@ -338,7 +338,7 @@ class AfMoEMoE: Module, UnaryLayer {
 
         // Apply experts
         var y = experts(x, inds)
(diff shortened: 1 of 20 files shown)


davidkoski (approved):
Nice idea, thank you!

PR #325, opened 2026-05-31
Preserve aspect ratio in Gemma4 image preprocessing

Penrose0v0:
## Summary

This updates Gemma4 image preprocessing to choose an aspect-ratio-preserving target size instead of resizing every image to the fixed square `config.fixedSize`.

The target size is aligned to `patchSize * poolingKernelSize` and kept within the existing pre-pooling patch budget derived from `imageSeqLength`, so the model still receives the same fixed number of vision soft tokens.

This better matches the Gemma4 / mlx-vlm preprocessing behavior and avoids distorting non-square screenshots, dashboards, and tall images.

## Testing

- Added unit tests for `gemma4VisionAspectPreservingTarget`
  - verifies target sizes align to the pooling stride
  - verifies patch count stays within budget
  - verifies non-square images are not forced to square
  - verifies degenerate extents fall back

### DIFF
--- a/Libraries/MLXVLM/Models/Gemma4.swift
+++ b/Libraries/MLXVLM/Models/Gemma4.swift
@@ -9,6 +9,7 @@ import MLXNN
 private enum Gemma4Error: LocalizedError {
     case imageTokenCountMismatch(expectedVisionTokens: Int, actualPromptTokens: Int)
     case multimodalTokenCountMismatch(kind: String, featureTokens: Int, promptTokens: Int)
+    case missingImageFrames
 
     var errorDescription: String? {
         switch self {
@@ -18,6 +19,9 @@ private enum Gemma4Error: LocalizedError {
         case .multimodalTokenCountMismatch(let kind, let featureTokens, let promptTokens):
             return
                 "Gemma4 \(kind) token count mismatch: encoder produced \(featureTokens) soft tokens, but the prompt contains \(promptTokens) \(kind) tokens."
+        case .missingImageFrames:
+            return
+                "Gemma4 received image pixels without per-image frames; the vision path needs each image's (height, width) to reconstruct the flattened pixels."
         }
     }
 }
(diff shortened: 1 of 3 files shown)


Libraries/MLXVLM/Models/Gemma4.swift
```diff
@@ -1836,11 +1885,17 @@ public struct Gemma4Processor: UserInputProcessor {
...
         MLXArray, THW
     ) {
-        var userProcessing = processing ?? UserInput.Processing()
-        let targetSize = config.fixedSize
-        userProcessing.resize = targetSize
+        let targetSize =
+            images.first.map {
+                gemma4VisionAspectPreservingTarget(
+                    width: Double($0.extent.width), height: Double($0.extent.height),
+                    imageSeqLength: config.imageSeqLength, fallback: config.fixedSize)
```
davidkoski, line 2705: This looks good, but there is one potential flaw:  if you have multiple input images with different aspect ratios this will resize them independently and will later fail in concatenating them (probably with a failure to broadcast is my guess).

A few things I could think of:

- multiple images use the old path
- it finds a fit-in size and resizes everything to fit in that with padding (not sure how well this works)
- cropping to fit the aspect ratio

I am not sure how often this case will come up.  Portrait vs landscape.  Screenshots.

Penrose0v0: Thanks for pointing this out! You're right that the previous version could produce different `H x W` tensors for multiple images and then try to concatenate them directly.

I've updated the code so each image is preprocessed independently, flattened, and paired with its `THW` metadata. The model side then reconstructs each image from the flattened buffer using those frame sizes before running the vision tower. This avoids the shape mismatch while preserving the aspect-ratio resize path for both single-image and multi-image inputs.

I also reproduced the same shape issue in my own downstream test project, and this update fixes it there.

davidkoski:
Thank you for the work on this but let's go with #405 -- this uses the same path with the Gemma4Unified and the python reference and removes the padding item from #290.

PR #327, opened 2026-06-03
Add Gemma 4 12B unified

adrgrondin:
## Proposed changes

Add Gemma 4 12B unified (https://huggingface.co/google/gemma-4-12B-it) ported from https://github.com/Blaizzy/mlx-vlm/commit/608ce45561a95996f7d5f33d0a6a051b61b23b03.

I kept the audio parts since the model is "unified" but did not test audio, I don't think it's also supported by mlx-swift-lm.

## Checklist

### DIFF
--- a/Libraries/MLXLLM/LLMModelFactory.swift
+++ b/Libraries/MLXLLM/LLMModelFactory.swift
@@ -32,6 +32,7 @@ public enum LLMTypeRegistry {
         "gemma3_text": create(Gemma3TextConfiguration.self, Gemma3TextModel.init),
         "gemma3n": create(Gemma3nTextConfiguration.self, Gemma3nTextModel.init),
         "gemma4": create(Gemma4Configuration.self, Gemma4Model.init),
+        "gemma4_unified": create(Gemma4Configuration.self, Gemma4Model.init),
         "gemma4_text": create(Gemma4TextConfiguration.self, Gemma4TextModel.init),
         "qwen2": create(Qwen2Configuration.self, Qwen2Model.init),
         "qwen3": create(Qwen3Configuration.self, Qwen3Model.init),
--- a/Libraries/MLXVLM/Models/Gemma4.swift
+++ b/Libraries/MLXVLM/Models/Gemma4.swift
@@ -8,12 +8,16 @@ import MLXNN
 
 private enum Gemma4Error: LocalizedError {
     case imageTokenCountMismatch(expectedVisionTokens: Int, actualPromptTokens: Int)
+    case multimodalTokenCountMismatch(kind: String, featureTokens: Int, promptTokens: Int)
 
     var errorDescription: String? {
         switch self {
         case .imageTokenCountMismatch(let expectedVisionTokens, let actualPromptTokens):
             return
                 "Gemma4 image token count mismatch: vision encoder produced \(expectedVisionTokens) soft tokens, but the prompt contains \(actualPromptTokens) image tokens."
+        case .multimodalTokenCountMismatch(let kind, let featureTokens, let promptTokens):
+            return
+                "Gemma4 \(kind) token count mismatch: encoder produced \(featureTokens) soft tokens, but the prompt contains \(promptTokens) \(kind) tokens."
         }
     }
 }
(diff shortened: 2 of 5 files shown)


aleroot:
Thanks for this!
Audio is coming https://github.com/ml-explore/mlx-swift-lm/pull/298

gregorym:
Can we merge this before #298 ?

notatestuser:
Nice work on the text path. One bug in the image branch: `Gemma4UnifiedProcessorConfiguration.fixedSize` takes the `if let size` branch because every Gemma 4 `processor_config.json` ships `size: {224, 224}`. With `model_patch_size: 48` that's `(224/48)*48 = 192` → a 4×4 = **16 soft-token** grid, so the model can't actually read the image and just makes things up (no error — the 16 placeholders still match).

The fix is to resize per-image into the `max_soft_tokens` budget (aspect-preserving, like HF's `get_aspect_ratio_preserving_size`) instead of using `size`. Locally that takes a 2174×1356 image from 16 → 273 tokens and it reads the on-screen numbers verbatim instead of fabricating them. Happy to send a PR.

notatestuser:
Opened a fix against the PR branch: adrgrondin/mlx-swift-lm#2

joshuaejackson:
Very excited for this!

I've been running this port (M3 Pro, 18 GB · `mlx-community/gemma-4-12B-it-4bit`) — the branch's changes brought into my app's vendored copy of the library, plus notatestuser's essential resize fix from adrgrondin#2, and one key modification: chunked prefill on the text-only path.
 
With that setup, the text path works well even on my meager 18gb of memory. Coherent single and multi-turn output and exact recall on a 2,513-token needle prompt.
 
But the prefill modification is important: `prepare` inherits the unchunked pattern from #336, which will really hurt at 12B. Important note: image prompts can't be chunked - they must stay single-shot (the vision mask needs the whole sequence - chunking could split an image block). 

With the #297/#337 shape applied to text-only prompts, my GPU peak stayed ~1.2 GB over weights and is nearly flat from 4k to 16k tokens. Happy to PR that commit (with chunk-size-invariance tests) against this branch.

adrgrondin:
> Very excited for this!
> 
> I've been running this port (M3 Pro, 18 GB · `mlx-community/gemma-4-12B-it-4bit`) — the branch's changes brought into my app's vendored copy of the library, plus notatestuser's essential resize fix from [adrgrondin#2](https://github.com/adrgrondin/mlx-swift-lm/pull/2), and one key modification: chunked prefill on the text-only path.
> 
> With that setup, the text path works well even on my meager 18gb of memory. Coherent single and multi-turn output and exact recall on a 2,513-token needle prompt.
> 
> But the prefill modification is important: `prepare` inherits the unchunked pattern from #336, which will really hurt at 12B. Important note: image prompts can't be chunked - they must stay single-shot (the vision mask needs the whole sequence - chunking could split an image block).
> 
> With the #297/#337 shape applied to text-only prompts, my GPU peak stayed ~1.2 GB over weights and is nearly flat from 4k to 16k tokens. Happy to PR that commit (with chunk-size-invariance tests) against this branch.

@joshuaejackson did an update to address this!

Libraries/MLXVLM/Models/Gemma4.swift
```diff
@@ -1825,60 +2029,24 @@ public final class Gemma4: Module, VLMModel, KVCacheDimensionProvider {
...
-                    nil,
-                    cache: convertedCache,
-                    inputsEmbeds: allEmbeds[0..., range, 0...],
-                    perLayerInputs: allPerLayerInputs.map { $0[0..., range, 0..., 0...] }
-                )
-                asyncEval(cache)
-                processed += chunkLength
-            }
-            // Single sync after the loop to flush any remaining async work.
-            eval(cache)
```
davidkoski, line 1852: Is this deletion intended?

Libraries/MLXLLM/LLMModelFactory.swift
```diff
@@ -32,6 +32,7 @@ public enum LLMTypeRegistry {
         "gemma3_text": create(Gemma3TextConfiguration.self, Gemma3TextModel.init),
         "gemma3n": create(Gemma3nTextConfiguration.self, Gemma3nTextModel.init),
         "gemma4": create(Gemma4Configuration.self, Gemma4Model.init),
+        "gemma4_unified": create(Gemma4Configuration.self, Gemma4Model.init),
```
davidkoski, line 35: This works ok?  The LLM side doesn't need e.g. `gemma4PrepareTextOnly()`?

Libraries/MLXVLM/Models/Gemma4.swift
```diff
@@ -1052,21 +1223,52 @@ final class Gemma4TextBackbone: Module {
...
+                if let tokenTypeIds {
+                    (tokenTypeIds .== 3).asType(.int32).sum().item(Int.self) > 0
+                } else {
+                    false
+                }
+            let useBidirectionalVision =
+                config.useBidirectionalAttention == "vision"
+                && tokenTypeIds != nil
+                && !hasAudioTokens
+                && h0.dim(1) > 1
```
davidkoski, line 1253: mlx-vlm has a `has_visual_tokens` check:

```python
        has_visual_tokens = (
            mm_token_type_ids is not None
            and int(mx.sum((mm_token_type_ids == 1) | (mm_token_type_ids == 2)).item())
            > 0
        )
```

I suspect that should be included here.

Libraries/MLXVLM/Models/Gemma4.swift
adrien-lms: Yes. This is after @joshuaejackson comments.

Text-only prompts: chunked prefill
Image/multimodal prompts: single-shot prefill

adrgrondin: Yes. This is after @joshuaejackson  comments.

Text-only prompts: chunked prefill
Image/multimodal prompts: single-shot prefill

Libraries/MLXLLM/LLMModelFactory.swift
adrgrondin: Yes. There's also not a lot of LM only Gemma 12B, only found this one: mlx-community/gemma-3-12b-it-qat-abliterated-lm-4bit

Libraries/MLXVLM/Models/Gemma4.swift
adrgrondin: Included in https://github.com/ml-explore/mlx-swift-lm/commit/eea29086d1bad825f01440087cdd12d8964b5728

davidkoski (approved):
This looks good to me, thank you!

PR #329, opened 2026-06-05
Add variance-normalized KV cache

aleroot:
## Proposed changes

This PR adds an opt-in, KVarN-inspired variance-normalized KV for reducing KV-cache memory pressure during long-context generation.

This introduces `kvQuantizationStrategy: .varianceNormalized`, which keeps completed KV tiles compressed and uses cache-native attention over quantized rotated tiles instead of repeatedly materializing full K/V tensors.

The cache applies:

- Hadamard rotation before tile compression
- log-domain dual-axis variance normalization with clipped scale updates
- best-imbalance scale tracking
- asymmetric K/V quantization, e.g. 4-bit K with 2-bit V
- compact completed-tile storage with fused quantization and variance scales
- cache-native attention over completed quantized tiles

## Benefits

- Reduces KV cache memory pressure for long-context generation.
- Avoids materializing completed compressed tiles on the decode hot path.

## Checklist

## Notes

The current implementation is slower than the default cache, so the intended use case is memory-constrained long-context decoding, not latency-sensitive generation.

Reference: https://arxiv.org/abs/2606.03458

### DIFF
--- a/Libraries/MLXLMCommon/VarianceNormalizedKVCache.swift
+++ b/Libraries/MLXLMCommon/VarianceNormalizedKVCache.swift
@@ -1,0 +1,826 @@
+// Copyright © 2024 Apple Inc.
+
+import Foundation
+import MLX
+import MLXNN
+
+private struct VarianceNormalizedKVTile {
+    var keyWeight: MLXArray
+    var keyScales: MLXArray
+    var keyBiases: MLXArray
+    var keyColumnScales: MLXArray
+
+    var valueWeight: MLXArray
+    var valueScales: MLXArray
+    var valueBiases: MLXArray
+    var valueColumnScales: MLXArray
+}
+
+private let varianceNormalizationEpsilon: Float = 1e-8
+private let varianceNormalizationMinimumStandardDeviation: Float = 1e-3
+private let varianceNormalizationMaximumStandardDeviation: Float = 1e3
+private let varianceNormalizationMinimumLogScale: Float = -0.3
+private let varianceNormalizationMaximumLogScale: Float = 10
+
+private func isPowerOfTwo(_ value: Int) -> Bool {
+    value > 0 && (value & (value - 1)) == 0
+}
+
+private func varianceNormalizedValueGroupSize(valueHeadDim: Int) -> Int? {
+    resolvedKVQuantizationGroupSize(
+        requested: min(128, max(32, valueHeadDim)),
+        keyHeadDim: valueHeadDim,
+        valueHeadDim: valueHeadDim
+    )
+}
+
+func supportsVarianceNormalizedKVCache(
+    keyHeadDim: Int,
+    valueHeadDim: Int,
+    tileSize: Int
+) -> Bool {
+    [32, 64, 128].contains(tileSize)
+        && keyHeadDim > 0
+        && valueHeadDim > 0
+        && isPowerOfTwo(keyHeadDim)
+        && isPowerOfTwo(valueHeadDim)
+        && varianceNormalizedValueGroupSize(valueHeadDim: valueHeadDim) != nil
+}
+
+extension KVCacheSimple {
+    /// Convert to a variance-normalized tile cache.
+    public func toVarianceNormalized(
+        tileSize: Int = 128,
+        keyBits: Int = 4,
+        valueBits: Int = 2,
+        sinkhornIterations: Int = 4
+    ) -> VarianceNormalizedKVCache {
+        let cache = VarianceNormalizedKVCache(
+            tileSize: tileSize,
+            keyBits: keyBits,
+            valueBits: valueBits,
+            sinkhornIterations: sinkhornIterations)
+
+        if let keys = self.keys, let values = self.values {
+            let currentKeys = keys[.ellipsis, ..<offset, 0...]
+            let currentValues = values[.ellipsis, ..<offset, 0...]
+            _ = cache.update(keys: currentKeys, values: currentValues)
+        } else {
+            cache.offset = self.offset
+        }
+
+        return cache
+    }
+}
+
+/// Variance-normalized KV cache following KVarN's core tile-compression steps.
+///
+/// This is intentionally implemented as a regular ``KVCache`` so model attention code can
+/// use it without changes. When called through
+/// ``attentionWithCacheUpdate(queries:keys:values:cache:scale:mask:)``, completed tiles stay
+/// quantized on the hot path: attention scores and value products are computed with `quantizedMM`
+/// in the rotated domain, and only the final output is inverse-rotated. The tile normalization uses
+/// clipped log-domain dual-axis scale balancing with best-imbalance tracking, matching the central
+/// KVarN VarN algorithm more closely while remaining an MLX Swift implementation. The RTN affine
+/// scale/bias are fused into the matching variance-normalization axis so each completed tile stores
+/// compact K/V records: packed weights, fused affine scale, fused affine bias, and the remaining
+/// variance scale.
+public class VarianceNormalizedKVCache: BaseKVCache, KVCacheAttentionProtocol,
+    CustomDebugStringConvertible
+{
+    static let compactTileStateCount = 8
+    static let legacyTileStateCount = 10
+    static let tailStateCount = 2
+
+    private var tiles: [VarianceNormalizedKVTile] = []
+    private var tailKeys: MLXArray?
+    private var tailValues: MLXArray?
+    private var restoredTileCount: Int?
+    private var restoredTailLength: Int?
+
+    public let tileSize: Int
+    public let keyBits: Int
+    public let valueBits: Int
+    public let sinkhornIterations: Int
+
+    public init(
+        tileSize: Int = 128,
+        keyBits: Int = 4,
+        valueBits: Int = 2,
+        sinkhornIterations: Int = 4
+    ) {
+        precondition(
+            (try? VarianceNormalizedKVCacheConfiguration(
+                keyBits: keyBits,
+                valueBits: valueBits,
+                tileSize: tileSize,
+                sinkhornIterations: sinkhornIterations)) != nil,
+            "VarianceNormalizedKVCache requires supported bit widths and tile size, plus a positive iteration count"
+        )
+        self.tileSize = tileSize
+        self.keyBits = keyBits
+        self.valueBits = valueBits
+        self.sinkhornIterations = sinkhornIterations
+        super.init()
+    }
+
+    public override func innerState() -> [MLXArray] {
+        state
+    }
+
+    private func rotate(_ x: MLXArray) -> MLXArray {
+        let dimensions = x.dim(-1)
+        guard isPowerOfTwo(dimensions) else { return x }
+
+        let originalShape = x.shape
+        var rotated = x.asType(.float32)
+        var stride = 1
+
+        while stride < dimensions {
+            let prefixShape = Array(rotated.shape.dropLast())
+            let paired = rotated.reshaped(prefixShape + [dimensions / (2 * stride), 2, stride])
+            let halves = split(paired, parts: 2, axis: -2)
+            let left = halves[0].squeezed(axis: -2)
+            let right = halves[1].squeezed(axis: -2)
+            rotated = concatenated([left + right, left - right], axis: -1).reshaped(originalShape)
+            stride *= 2
+        }
+
+        return (rotated / sqrt(Float(dimensions))).asType(x.dtype)
+    }
+
+    private func inverseRotate(_ x: MLXArray) -> MLXArray {
+        // Sylvester Hadamard is symmetric and orthonormal, so the inverse is itself.
+        rotate(x)
+    }
+
+    private func balanced(
+        original: MLXArray,
+        columnScales: MLXArray,
+        rowScales: MLXArray
+    ) -> MLXArray {
+        original / columnScales / rowScales
+    }
+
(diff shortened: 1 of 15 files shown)


Libraries/MLXLMCommon/VarianceNormalizedKVCache.swift
```diff
@@ -0,0 +1,826 @@
...
+            .transposed(0, 1, 3, 2)
+
+        let valueHeadDim = tile.valueColumnScales.dim(-1)
+        let valueGroupSize = validatedValueGroupSize(valueHeadDim)
+        let scaledValues = dequantized(
+            tile.valueWeight, scales: tile.valueScales, biases: tile.valueBiases,
+            groupSize: valueGroupSize, bits: valueBits, mode: .affine)
+        let rotatedValues = scaledValues * tile.valueColumnScales
+
+        return (rotatedKeys.asType(.float16), rotatedValues.asType(.float16))
```
davidkoski: Is there a case where converting to float16 would produce `Inf`, which leads to `NaN`?  As far as I can tell the original dtype could be float32 or bfloat16 and both of these can easily exceed the range here.

Can it use the original dtype?  does `supportsVarianceNormalizedKVCache` need to guard against this?

Libraries/MLXLMCommon/VarianceNormalizedKVCache.swift
```diff
@@ -0,0 +1,826 @@
...
+        }
+
+        offset += keys.dim(2)
+        absorbTail()
+    }
+
+    public override func update(keys: MLXArray, values: MLXArray) -> (MLXArray, MLXArray) {
+        append(keys: keys, values: values)
+
+        guard let state = materializedState() else {
```
davidkoski, line 895: Is there a case where we want to allow callers to use this?  If this is accidentally called does it signal a bug?  

This looks like an expensive call: it dequantizes and combines all the tiles.

Libraries/MLXLMCommon/VarianceNormalizedKVCache.swift
```diff
@@ -0,0 +1,826 @@
...
+                bits: valueBits,
+                mode: .affine
+            ) * tile.valueColumnScales
+        }
+    }
+
+    private func stackedTileScores(
+        rotatedQueries: MLXArray,
+        scale: Float
+    ) -> MLXArray {
```
davidkoski, line 656: The code here looks like it combines all the tiles to do a big quantizedMM.  This is called per token from attention, so the cost here is O(n^2) where n is the number of tokens.

The caller looks like this:

```swift
            if tiles.count > 1 {
                [stackedTileScores(rotatedQueries: rotatedQueries, scale: scale)]
            } else {
                tiles.map { tile in
                    quantizedTileScores(rotatedQueries: rotatedQueries, tile: tile, scale: scale)
                }
            }
```

This suggests that the calling code could handle an array of tiles - otherwise why put the scores in an array?  Both paths end up producing an array of size 1, so I wonder if that is an oversight?

I am not sure if this can be done any more efficiently -- potentially knowing something about the mask would help, but we don't want an eval to peak inside.  Having the calling code consume tiles would eliminate some copying, but I am not sure if that helps either.

Anyway, just poking around to see if we really need this work on every token.

aleroot: Fixed: completed tiles now use geometrically grown packed storage, eliminating per-token stacking while preserving batched quantizedMM calls.

Libraries/MLXLMCommon/VarianceNormalizedKVCache.swift
aleroot: Callers do need this compatibility path because several specialized attention implementations invoke KVCache.update directly. It is intentionally expensive as the standard attention path dispatches to updateAndAttend and keeps completed tiles quantised.

aleroot:
Today I had the opportunity to work on this and I have added a new optimisation commit: "bounded attention slab scheduling",  which addresses a performance cliff found while benchmarking the variance-normalized KV cache across different context lengths.

Previously, cache compression and attention slab formation both used 32-tile batches. Until the cache accumulated all 32 tiles, attention could require up to 31 separate GPU dispatches. This produced a noticeable slowdown immediately before the 4,096-token boundary.

The implementation now separates these concerns:

- Compression still processes up to 32 tiles together, preserving its batching efficiency.
- Attention slabs are formed every 4 tiles and compacted hierarchically.
- The quantization algorithm, cached representation, and attention calculation are unchanged.

This is therefore a scheduling optimization rather than a numerical or format change.

### Decode benchmark

Apple M2, Q8/KV4, head dimension 128, K4/V2, 128-token tiles. Times are median one-token decode latency(lower is better).

| Context | Previous partitions | New partitions | Previous | New | Improvement |
|---:|---:|---:|---:|---:|---:|
| 512 | 4 | 1 | 1.926 ms | 1.198 ms | 37.8% |
| 1,024 | 8 | 2 | 2.967 ms | 1.512 ms | 49.0% |
| 2,048 | 16 | 4 | 5.114 ms | 2.211 ms | 56.8% |
| 3,968 | 31 | 10 | 8.971 ms | 4.052 ms | 54.8% |
| 4,096 | 1 | 1 | 1.364 ms | 1.290 ms | 5.4% |
| 8,192 | 2 | 2 | 2.124 ms | 1.991 ms | 6.3% |
| 32,768 | 1 | 1 | 2.089 ms | 2.017 ms | 3.4% |

The important result is that the worst pre-boundary case becomes approximately **2.2× faster**, while the already-efficient 4K, 8K, and 32K boundaries do not regress.

### Real-model A/B benchmark

Model: `mlx-community/Qwen3-4B-4bit`  
Workload: 3,968-token natural-language prompt followed by 64 greedy decode tokens.

| VarN K4/V2 implementation | Prefill | Decode | Peak memory delta |
|---|---:|---:|---:|
| Previous 32-tile base slab | 108.19 tok/s | 10.96 tok/s | 1,347.19 MiB |
| New 4-tile base slab | 148.41 tok/s | 12.95 tok/s | 1,347.19 MiB |
| Improvement | **+37.2%** | **+18.2%** | Unchanged |

The generated-token sequence had **100% agreement** with the FP16 KV-cache reference for all 64 decode tokens in this test.

For context, the complete backend comparison on the same workload was:

| KV-cache backend | Prefill | Decode | Token agreement vs FP16 |
|---|---:|---:|---:|
| FP16 | 183.22 tok/s | 20.74 tok/s | Reference |
| Affine 4-bit | 171.35 tok/s | 18.93 tok/s | 100% |
| TurboQuant K4/V2 | 174.53 tok/s | 17.96 tok/s | 43.75% |
| VarN K4/V2 | 149.76 tok/s | 12.90 tok/s | 100% |

These results do not claim that VarN is faster than every existing backend. They show that this change removes a substantial internal dispatch bottleneck without changing its output quality, cache format, or peak working memory.

aleroot:
### Memory consumption

Two different memory measurements should be distinguished:

1. **Persistent KV-cache storage**: memory retained for cached keys and values.
2. **Peak process memory**: includes model weights, MLX allocator reservations, temporary normalization buffers, attention workspaces, and the KV cache.

In the isolated 32K, one-layer storage test:

| Representation | KV storage |
|---|---:|
| Raw FP16 keys and values | 64.00 MiB |
| VarN K4/V2 compact storage | 13.50 MiB |
| Reduction | **78.9% / 4.74× smaller** |

VarN therefore delivers the expected reduction in persistent KV-cache storage. This becomes increasingly relevant for long contexts and models with many global-attention layers.

However, the full Qwen3-4B benchmark showed a different result for peak process memory:

| KV-cache backend | Peak memory delta | Difference vs FP16 |
|---|---:|---:|
| FP16 | 1,067.30 MiB | Reference |
| Affine 4-bit | 1,067.30 MiB | No measurable change |
| TurboQuant K4/V2 | 1,067.30 MiB | No measurable change |
| VarN K4/V2 | 1,347.19 MiB | **+279.89 MiB / +26.2%** |

This means the compact VarN representation is substantially smaller once stored, but the current implementation uses enough temporary workspace during normalization and compression to produce a higher process high-water mark on this workload.

The slab-scheduling optimization in this update improves prefill and decode performance but does **not** reduce that peak workspace: both the previous and new versions measured approximately 1,347 MiB.

Consequently, the memory claim for this PR should be precise:

- VarN substantially reduces persistent KV-cache storage.
- It does not yet reduce end-to-end peak process memory on Qwen3-4B.
- Reducing temporary compression and materialization buffers remains a separate optimization opportunity.

This does not block merging the scheduling improvement because it introduces no memory regression relative to the previous VarN implementation. It does, however, mean VarN should remain opt-in until its transient workspace behavior has been improved and evaluated at longer full-model context lengths. 

> This implementation still has clear opportunities for further optimization, particularly around temporary workspace and peak memory usage. I believe it is a good foundation to merge in its current opt-in form and optimize incrementally in follow-up PRs.

aleroot:
Following the earlier review, I investigated the temporary-memory peak rather than treating compact KV storage as equivalent to low process memory. The peak came from variance-normalizing all eight KV heads in one lazy FP32 graph. The cache now normalizes at most four KV heads at a time while retaining the existing 32-tile compression batching.

This does not change:

- The VarN algorithm or Sinkhorn iteration count.
- Quantized values or cache format.
- Attention behavior.
- Prompt-cache compatibility.
- The path used by models with four or fewer KV heads.

The compact results from each head batch are concatenated into the same final cache representation.

### Real-model test

Hardware: Apple M2  
Model: `mlx-community/Qwen3-4B-4bit`  
Prompt: 3,968 natural-language tokens  
Generation: 64 greedy tokens

| Configuration | Prefill | Decode | Peak memory delta | Agreement vs FP16 |
|---|---:|---:|---:|---:|
| FP16 KV | 146.38 tok/s | 18.22 tok/s | 1,067.30 MiB | Reference |
| VarN K4/V2 | 121.86 tok/s | 10.87 tok/s | **1,067.30 MiB** | **100%** |

Before bounded-head normalization, the same VarN configuration reached a peak delta of 1,347.19 MiB.

The change therefore removes:

- **279.89 MiB** of peak process memory.
- The previous **26.2% peak overhead relative to FP16**.
- **20.8% of VarN’s previous measured peak**.

The objective here is memory-constrained long-context operation rather than outperforming FP16 latency. VarN remains slower, but its measured process peak is no longer higher than FP16 for this workload.

### Direct 32K workspace A/B

A Qwen3-4B-like synthetic configuration was used:

- 32 query heads.
- 8 KV heads.
- Head dimension 128.
- 32,768 cached tokens.
- K4/V2 compression.

| Measurement | All 8 heads together | 4-head batches | Change |
|---|---:|---:|---:|
| Peak normalization workspace | 946.7 MiB | **558.9 MiB** | **−41.0%** |
| 32K ingest | 1.286 s | 1.321 s | approximately +2.7% |
| Raw FP16 KV storage | 128.0 MiB | 128.0 MiB | — |
| Compact VarN storage | 27.0 MiB | 27.0 MiB | **78.9% smaller than raw** |

The previous unbounded implementation fails the 768 MiB workspace regression guard at this shape. The bounded implementation passes with substantial margin.

### Correctness evidence

The head batching is valid because variance normalization operates independently for each KV head. This is now enforced by a test comparing:

- One eight-head cache using bounded batches.
- Eight independently compressed one-head caches.

The reconstructed keys and values match with relative RMSE below `1e-6`.

Additional quality results:

- Qwen3-4B greedy agreement with FP16: **64/64 tokens**.
- Deterministic model test agreement: **100/100 tokens**.
- FP16 perplexity: `5.768397`.
- VarN K4/V2 perplexity: `5.744203`.
- Long-decode, GQA, masked-attention, trimming, serialization, dtype and prompt-cache tests pass.

### Performance work included in the latest commits

The preceding slab-scheduling change also removed the dispatch cliff immediately before 4,096 tokens:

| Context | Previous decode | Current decode | Improvement |
|---:|---:|---:|---:|
| 512 | 1.926 ms | 1.198 ms | 37.8% |
| 1,024 | 2.967 ms | 1.512 ms | 49.0% |
| 2,048 | 5.114 ms | 2.211 ms | 56.8% |
| 3,968 | 8.971 ms | 4.052 ms | 54.8% |
| 4,096 | 1.364 ms | 1.290 ms | 5.4% |
| 8,192 | 2.124 ms | 1.991 ms | 6.3% |
| 32,768 | 2.089 ms | 2.017 ms | 3.4% |

### Memory-footprint summary

The memory results measure two different things:

| Measurement | FP16 / previous VarN | Current VarN K4/V2 | Result |
|---|---:|---:|---:|
| Qwen3-4B peak process delta at 3,968 tokens | FP16: 1,067.30 MiB | 1,067.30 MiB | No measurable peak overhead vs FP16 |
| VarN peak process delta before head batching | 1,347.19 MiB | 1,067.30 MiB | −279.89 MiB / −20.8% |
| 32K normalization workspace | 946.7 MiB | 558.9 MiB | −41.0% |
| 32K retained KV storage, keys and values | FP16: 128.0 MiB | 27.0 MiB | −78.9% / 4.74× smaller |

The full-model peak includes model weights, activations, allocator reservations, attention workspaces, and temporary compression buffers. The retained-storage measurement isolates the memory occupied by the KV representation itself. _The important result is that the current implementation preserves VarN’s approximately 4.74× reduction in retained KV storage while eliminating the previously measured full-process peak overhead for this Qwen3-4B workload._


> With these changes, I think I  will start to really use VarN for my **local coding-agent**. 
@davidkoski I think now it is ready for another review. Thanks.

davidkoski (approved):
The change looks good to me.  Thank you for all the followup here!

PR #331, opened 2026-06-06
Align Falcon H1 with upstream MLX inference

aleroot:
## Proposed changes

Update Falcon H1/H1R inference behavior to match current mlx-lm more closely.

This fixes tied embedding output projection and scaling, routes attention through the common cache-aware causal path, advances Mamba cache metadata during generation, reports per-layer KV head counts correctly, and chunks SSM prefill to reduce long-prompt memory and latency.

Also mirror upstream ArraysCache length handling for hybrid and batched paths: preserve left padding and lengths through filter, extend, copy, and serialization; forward length preparation through CacheList; and use per-row lengths in Falcon H1 convolution plus chunked SSM state updates.

## Checklist

### DIFF
--- a/Libraries/MLXLLM/Models/SSM.swift
+++ b/Libraries/MLXLLM/Models/SSM.swift
@@ -14,7 +14,7 @@ import MLXNN
 public func computeDt(_ dt: MLXArray, _ dtBias: MLXArray, _ timeStepLimit: (Float, Float))
     -> MLXArray
 {
-    let dt = softplus(dt + dtBias)
+    let dt = softplus(dt.asType(.float32) + dtBias)
     return MLX.clip(dt, min: timeStepLimit.0, max: timeStepLimit.1)
 }
 
@@ -50,7 +50,7 @@ private func makeSSMKernel() -> MLXFast.MLXFastKernel? {
                 auto idx = d_idx * Ds + s_idx;
                 auto dB_by_x = x_ * dt_ * static_cast<float>(B_[s_idx]);
                 auto state = dA * i_state[idx] + dB_by_x;
-                o_state[idx] = static_cast<T>(state);
+                o_state[idx] = static_cast<U>(state);
                 acc += state * C_[s_idx];
             }
             acc = simd_sum(acc);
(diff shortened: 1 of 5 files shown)


Libraries/MLXLLM/Models/SSM.swift
```diff
@@ -150,47 +152,101 @@ public func ssmAttn(
...
+            dtA: dtA[0..., i ..< end],
+            B: B[0..., i ..< end],
+            C: C[0..., i ..< end],
+            state: state,
+            mask: maskChunk,
+            lengths: lengths
+        )
+        ys.append(y)
+        state = nextState
+        if let currentLengths = lengths {
```
davidkoski, line 242: This looks like prefill code -- I wonder if asyncEval inside the loop would be a good idea?

e.g. https://github.com/ml-explore/mlx-swift-lm/pull/337/changes#r3392157172

Libraries/MLXLLM/Models/SSM.swift
```diff
@@ -150,47 +152,101 @@ public func ssmAttn(
...
+            decay = MLX.takeAlong(decay, pos, axis: 2)
+        } else {
+            decay = decay[0..., 0..., (-1)..., 0...]
+        }
+        decay = decay.transposed(0, 3, 1, 2)
+        B = MLX.repeated(B, count: h / g, axis: 1).swappedAxes(2, 3)
+        var dtxdecay = dtx * decay
+        dtxdecay = dtxdecay.swappedAxes(1, 2).swappedAxes(2, 3)
+
+        var nextState = dtxdecay.matmul(B)
```
davidkoski, line 204: Not sure if this is a problem or not -- this ends up promoting nextState to `float32` (via `computeDt` -> `dt` -> `dtx`).  This is an output of the function and I think it propagates into various places.

aleroot: That's right: the chunked path now preserves the recurrent state dtype when returning `nextState`, so fp32 intermediate math no longer leaks into the cache state. I also added a regression test covering bf16 chunked SSM state dtype preservation.

Libraries/MLXLLM/Models/SSM.swift
aleroot: I added `asyncEval(y, nextState)` inside the chunk loop and a single `eval(y, state)` after the loop, matching the LLMModel/Gemma4 pattern.

davidkoski (approved):
Looks good, thank you!

PR #332, opened 2026-06-06
Fix LFM2-MoE sigmoid routing and expert-bias selection

ykhrustalev:
## Proposed changes

Fix LFM2-MoE routing to match the sigmoid-gated behavior used by LFM2.5 checkpoints.

The current Swift path applies softmax to router logits and folds `expert_bias` into the combination weights. LFM2-MoE uses per-expert sigmoid scores: `expert_bias` should affect top-k selection only, while the returned weights stay gathered from the unbiased sigmoid scores, optionally normalized by `norm_topk_prob`, then scaled by `routed_scaling_factor`.

This mirrors the same routing fix in ml-explore/mlx-lm#1354.

This also adds focused routing regression tests for the sigmoid gate, expert-bias selection-only behavior, and top-k normalization.

The dense LFM2-MoE MLP path now compiles the SwiGLU activation product, matching the existing compiled SwitchGLU path used by the MoE experts.

## Checklist

### DIFF
--- a/Libraries/MLXLLM/Models/LFM2MoE.swift
+++ b/Libraries/MLXLLM/Models/LFM2MoE.swift
@@ -12,6 +12,12 @@ import MLX
 import MLXLMCommon
 import MLXNN
 
+private let compiledLFM2MoESiluProduct: @Sendable (MLXArray, MLXArray) -> MLXArray = compile(
+    shapeless: true
+) { gate, up in
+    MLXNN.silu(gate) * up
+}
+
 public struct LFM2MoEConfiguration: Codable, Sendable {
     public let modelType: String
     public let vocabularySize: Int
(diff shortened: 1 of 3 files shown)


ykhrustalev:
cc @nastya236

Libraries/MLXLLM/Models/LFM2MoE.swift
```diff
@@ -12,6 +12,12 @@ import MLX
 import MLXLMCommon
 import MLXNN
 
+private let compiledLFM2MoESiluProduct: @Sendable (MLXArray, MLXArray) -> MLXArray = compile(
```
davidkoski: Let's use this:

```swift
private let compiledSiluProduct: @Sendable (MLXArray, MLXArray) -> MLXArray = ...
```

from SwichLayers.swift -- identical function.  Promote that to public (like the `weightedExpertSum` next to it).

Other than this small item, looks good!

davidkoski (approved):
Thank you!

PR #334, opened 2026-06-09
Add MLXFoundationModels: an MLX-backed FoundationModels LanguageModel

ctymoszek:
MLXLanguageModel conforms to FoundationModels.LanguageModel, so locally-run MLX models can be used through the FoundationModels framework. It supports chat, tool calling, and guided generation (JSON-schema-constrained sampling via a vendored copy of xgrammar).

The adapter is gated behind two default-on SwiftPM traits (FoundationModelsIntegration, GuidedGenerationSupport) and is @available(iOS 27.0, macOS 27.0, visionOS 27.0). Platform floors are unchanged from upstream (.macOS(.v14), .iOS(.v17), .tvOS(.v17), .visionOS(.v1)), so existing consumers and anyone who disables the traits are unaffected.

To learn more about the `LanguageModel` protocol this adapter conforms to, see *[Bring an LLM provider to the Foundation Models framework](https://www.youtube.com/watch?v=u06ZVpSl0J4)* from WWDC26.

## Checklist


IntegrationTesting/IntegrationTestingTests/MLXFoundationModelsIntegration/TextGeneration/ApplyChatTemplateProbeTests.swift
davidkoski, line 1: It would be great to group these into one or more directories for organization purposes.

Libraries/MLXFoundationModels/GuidedGeneration/GuidedGenerationError.swift
davidkoski, line 1: Not really for this file, but anchoring it here in the GuidedGeneration directory.  I think it would be nice if we had GuidedGeneration and FoundationModels as two separate libraries.  FM could depend on GuidedGeneration.

I think most of the GuidedGeneration code doesn't need to be in MLXLMCommon -- e.g. the Embedders do not need it.  That isn't to say it _can't_ go there, but it may make more sense in GuidedGeneration if that is its own library.

A split like this would allow people to opt-in to GuidedGeneration without using any of the FM code.  For example on Linux the FM code won't be available at all but the GuidedGeneration piece would be interesting.

Libraries/MLXFoundationModels/GuidedGeneration/GuidedGenerationLoop.swift
```diff
@@ -0,0 +1,540 @@
...
+                    if let wsBias = whitespaceBias, whitespaceTracker.isActive {
+                        activeBias = activeBias.map { $0 + wsBias } ?? wsBias
+                    }
+                }
+                let token: UInt32 = mask.mask.withUnsafeBufferPointer { buffer in
+                    let ptr: UnsafePointer<UInt32>? =
+                        mask.needsApply
+                        ? UnsafeRawPointer(buffer.baseAddress!).assumingMemoryBound(to: UInt32.self)
+                        : nil
+                    return applyMaskAndSample(
```
davidkoski: This reads from `logits` (via `item()`) before the `asyncEvals` below are executed.  This may cause a stall where we wait for the current graph to finish before scheduling any new work.  Consider the normal eval loop:

```swift
        // save current value -- this will be returned
        let previousY = y

        // compute the next state and async eval the next token
        let token = step(previous: previousY)
        y = .init(tokens: token)
        asyncEval(token)

        tokenCount += 1

        return previousY.tokens.item(Int.self)
```

`tokens` is still an MLXArray (lazy) at the point where `asyncEval` is called.

Libraries/MLXFoundationModels/GuidedGeneration/GuidedGenerationLoop.swift
```diff
@@ -0,0 +1,540 @@
...
+                    // Kick off GPU computation asynchronously
+                    asyncEval(logits)
+
+                    // Overlap: compute next mask on CPU while GPU runs
+                    mask = try constraint.computeMask()
+
+                    // Wait for GPU to finish (may already be done)
+                    eval(logits)
+                }
+            }
```
davidkoski: The normal eval loop has this:

```swift
        // Apply dynamic cache quantization after each step
        maybeQuantizeKVCache(
            cache: &cache,
            kvBits: kvBits,
            kvGroupSize: kvGroupSize,
            quantizedKVStart: quantizedKVStart
        )
```

Do we need it here inside the generate loop?  Not every model uses it, but I think it is important for memory use in models that do.

Libraries/MLXFoundationModels/GuidedGeneration/GuidedGenerationLoop.swift
```diff
@@ -0,0 +1,540 @@
...
+            vocabSize: Int,
+            closingBias: MLXArray? = nil
+        ) -> UInt32 {
+            // Extract last-position logits: [batch, seq, vocab] -> [vocab]
+            var logits = rawLogits[0..., -1, 0...]
+
+            if let maskPtr = sampleMask {
+                let logitDim = logits.shape[logits.ndim - 1]
+                let maskArray = bitmaskToMLXArray(
+                    maskPtr, maskBitCount: vocabSize, totalCount: logitDim)
```
davidkoski: I wonder if this can be cached in the eval loop and passed in?

Libraries/MLXFoundationModels/ModelCustomizer.swift
```diff
@@ -0,0 +1,55 @@
+// Copyright © 2025 Apple Inc.
+
+#if FoundationModelsIntegration
+    #if canImport(FoundationModels, _version: 2)
+
+        import Foundation
```
davidkoski: This indentation isn't my favorite.  What do you think about doing this in `.swift-format`?

```diff
diff --git a/.swift-format b/.swift-format
index 8892e9f..b9a726a 100644
--- a/.swift-format
+++ b/.swift-format
@@ -4,4 +4,5 @@
         "spaces": 4
     },
     "spacesAroundRangeFormationOperators": true,
+    "indentConditionalCompilationBlocks": false,
 }
```

Libraries/MLXFoundationModels/MLXLanguageModel.swift
```diff
@@ -0,0 +1,1847 @@
...
+                                        tokenizer: context.tokenizer
+                                    )
+                                    let completionReserve = Swift.max(
+                                        structuralReserve * 3, maxTokens / 4)
+                                    let hardReserve = structuralReserve * 8
+
+                                    let (whitespaceBias, whitespaceTokenIDs) =
+                                        WhitespaceTokenBias.compute(
+                                            tokenizer: context.tokenizer
+                                        )
```
davidkoski: This is computed per response -- should it be held in the `ModelCache`?

I think WhitespaceTokenBias and also friends.

Libraries/MLXFoundationModels/MLXLanguageModel.swift
```diff
@@ -0,0 +1,1847 @@
...
+            case structuralTag
+        }
+
+        // MARK: - Model Cache Actor
+
+        /// Thread-safe model cache using Swift actor isolation.
+        /// Prevents race conditions when multiple concurrent requests try to load the model.
+        /// Supports caching multiple models by their identifiers.
+        private actor ModelCache {
+            private var containers: [String: ModelContainer] = [:]
```
davidkoski: A few of these can grow without bound, e.g. in a long running model server.  `evictAll()` / `evictAllModels()` can remove them but that removes everything.  I wonder if this needs finer grained eviction support?

Libraries/MLXFoundationModels/MLXLanguageModel.swift
```diff
@@ -0,0 +1,1847 @@
...
+                /// `warmUp()` pre-created the tokenizer.
+                static func hasCachedXGTokenizer(modelID: String) async -> Bool {
+                    await cache.hasCachedXGTokenizer(modelID: modelID)
+                }
+            #endif
+
+            /// Evicts all cached models, tokenizers, and constraint templates.
+            /// Frees GPU memory held by model weights. Subsequent requests will
+            /// reload models from disk cache.
+            static func evictAllModels() async {
```
davidkoski: Should this be public?

I guess similar question for each of these static functions -- should they be public?

Libraries/MLXLMCommon/GuidedGeneration/CompositeLogitProcessor.swift
```diff
@@ -0,0 +1,38 @@
...
+import MLX
+
+/// Chains multiple `LogitProcessor` instances, applying them in order.
+///
+/// Grammar processors should come first (hard constraints that mask invalid tokens),
+/// followed by soft preference processors (repetition penalty, temperature scaling).
+///
+/// Thread safety: marked `@unchecked Sendable` because all access is serialized
+/// through `ModelContainer.perform`.
+public struct CompositeLogitProcessor: LogitProcessor, @unchecked Sendable {
```
davidkoski: Should this be `@unchecked Sendable`?  `LogitProcessor` is not `Sendable` and the processors can be stateful -- I don't think this is correct.

Libraries/MLXFoundationModels/ModelCustomizer.swift
ctymoszek: I'm in favor, that would look nicer - but maybe that change should be separate from this PR since it affects one other file (`WiredMemoryPolicies.swift`)

aleroot:
From a Swift package design perspective, this turns a focused Swift ML library into the owner of a substantial native dependency: a large C++ source snapshot, third-party headers, a custom C shim, unsafe compiler flags, symbol-isolation rules, update tooling, and new public API surface. That is not just implementation detail. It changes the package’s maintenance contract for every downstream user.

`mlx-swift-lm` is used as a library by many apps and packages, so the default dependency graph matters. Consumers should not inherit a vendored C++ grammar engine, its compile-time cost, warning policy, ABI/linking risks, and security/update burden just because they depend on MLX language model utilities or a FoundationModels adapter.

If guided generation needs xgrammar, I think it should be isolated behind a separate package/product with a clear ownership and release policy, or consumed as an external dependency. That would let users opt into the native grammar engine deliberately, while keeping the core package’s build surface predictable and Swift-native.

I would strongly prefer this PR be split so the FoundationModels adapter can be reviewed independently from the much larger decision of whether this project should own and ship a vendored C++ dependency...

As designed, this couples an optional adapter feature to a large native dependency decision that should be made explicitly and independently.

davidkoski:
> From a Swift package design perspective, this turns a focused Swift ML library into the owner of a substantial native dependency: a large C++ source snapshot, third-party headers, a custom C shim, unsafe compiler flags, symbol-isolation rules, update tooling, and new public API surface. That is not just implementation detail. It changes the package’s maintenance contract for every downstream user.

There is some missing back story -- this is the implementation from this:

- https://developer.apple.com/videos/play/wwdc2026/339/

"WWDC26 Bring an LLM provider to the Foundation Models framework"

The goal is that the code will go into new libraries that you can opt-in to.  If you don't use it then it shouldn't build.  It does have the xgrammar dependency (which will be pulled) but also not built if you don't use it.

We are experimenting with Traits as well -- they can be used control opt-in / out as well, but I think just picking which libraries link should control what builds.

I appreciate your feedback and agree that a separate repository would make this simpler, but we have bureaucratic constraints and cannot easily add new repositories.

aleroot:
> > From a Swift package design perspective, this turns a focused Swift ML library into the owner of a substantial native dependency: a large C++ source snapshot, third-party headers, a custom C shim, unsafe compiler flags, symbol-isolation rules, update tooling, and new public API surface. That is not just implementation detail. It changes the package’s maintenance contract for every downstream user.
> 
> There is some missing back story -- this is the implementation from this:
> 
> * https://developer.apple.com/videos/play/wwdc2026/339/
> 
> "WWDC26 Bring an LLM provider to the Foundation Models framework"
> 
> The goal is that the code will go into new libraries that you can opt-in to. If you don't use it then it shouldn't build. It does have the xgrammar dependency (which will be pulled) but also not built if you don't use it.
> 
> We are experimenting with Traits as well -- they can be used control opt-in / out as well, but I think just picking which libraries link should control what builds.
> 
> I appreciate your feedback and agree that a separate repository would make this simpler, but we have bureaucratic constraints and cannot easily add new repositories.

Ah ok, sorry. This is coming from Apple itself 😀. I didn't know 😀

rudrankriyam:
Xcode 27 exposes image input as a standard Foundation Models capability through .vision and Transcript.Segment.attachment. The current TranscriptConverter skips non-text segments, although MLX already supports UserInput.Image and VLMs such as Qwen3-VL.

Is vision support intentionally outside this PR’s scope? If so, would you prefer a focused follow-up PR that maps Foundation Models image attachments into MLX image inputs and adds .vision coverage?

ctymoszek:
> Xcode 27 exposes image input as a standard Foundation Models capability through .vision and Transcript.Segment.attachment. The current TranscriptConverter skips non-text segments, although MLX already supports UserInput.Image and VLMs such as Qwen3-VL.
> 
> Is vision support intentionally outside this PR’s scope? If so, would you prefer a focused follow-up PR that maps Foundation Models image attachments into MLX image inputs and adds .vision coverage?

Not at all outside the scope - this is in flight (PR currently open on my fork, to be merged into this branch shortly).

Libraries/MLXFoundationModels/GuidedGeneration/GuidedGenerationError.swift
ctymoszek: This is complete - thanks @thechriswebb !

Libraries/MLXLMCommon/GuidedGeneration/CompositeLogitProcessor.swift
ctymoszek: You're right - fixed.

IntegrationTesting/IntegrationTestingTests/MLXFoundationModelsIntegration/TextGeneration/ApplyChatTemplateProbeTests.swift
ctymoszek: Sure, they're organized now.

Libraries/MLXFoundationModels/MLXLanguageModel.swift
ctymoszek: Yes, that's reasonable.

Here's a little inventory after I checked the nearby static funcs:
- eviction-related: yes, should be public
- loadContainer: Sure, I suppose people may want to re-use an existing cache and there's no way to accomplish this currently. There is a wrinkle though: It has this `suppressDownloadingState` param. This is used by the internal `warmUp()` but I'm not sure we want to expose that publicly. Perhaps a public overload is a good compromise?
- makeContainerLoader/makeXGTokenizer/makeConstraint: I don't think so
- isDownloadingInCache/lastLoadErrorInCache: This functionality is accessed using the `availability` APIs so no need to make these public

Libraries/MLXFoundationModels/MLXLanguageModel.swift
ctymoszek: Yes, that's a good idea. I created 
`public func evict()` so that a given instance can be evicted.
I also changed `evictAllModels()` to `evictAll()` to better match `evict()`.

Something to consider in the future: maybe a `static evict(modelID:)` in case someone wants to do more sophisticated cache management.

Libraries/MLXFoundationModels/ModelCustomizer.swift
davidkoski: OK, if you want, make a small PR and we can get that in on main, then rebase here.

Libraries/MLXFoundationModels/MLXLanguageModel.swift
davidkoski: I wonder if model caching is something that MLXLMCommon should provide?  Then we can inject the cache?  We don't need to solve that here, but something to think about.

Package.swift
```diff
@@ -28,13 +28,35 @@ let package = Package(
...
             name: "IntegrationTestHelpers",
             targets: ["IntegrationTestHelpers"]),
     ],
+    traits: [
+        // Gates the MLXLanguageModel adapter for Apple's FoundationModels
+        // framework. Default-on. Disabling the trait compiles MLXFoundationModels
+        // to an effectively empty library (only MLXDownloadProgress survives):
+        // the entire `MLXLanguageModel` / `MLXLanguageModel.Executor` surface
+        // requires FoundationModels types that are not available on platforms
+        // older than iOS/macOS/visionOS 27.0. Consumers targeting older floors
```
davidkoski: Is "older floors" a typo?

Package.swift
```diff
@@ -146,6 +168,128 @@ let package = Package(
...
+                // Rename the vendored C++ namespaces at compile time so this
+                // target's symbols cannot collide with another xgrammar in the
+                // same binary (e.g. CoreAI's prebuilt copy). Token-level
+                // substitution: it rewrites bare `xgrammar` / `picojson`
+                // identifiers (namespace decls and `::` uses) but not header
+                // names, string literals, `XGRAMMAR_*` macros, or `xg_*` tokens.
+                .define("xgrammar", to: "mlx_xgrammar"),
+                .define("picojson", to: "mlx_picojson"),
+                // xgrammar throws -- exceptions must stay enabled.
+                .unsafeFlags(["-std=c++17", "-fexceptions"]),
```
davidkoski: There is standard package flag that controls which C++ language standard is used (see mlx-swift):

```
    cxxLanguageStandard: .gnucxx20
```

For exceptions, what if we don't enable this (`-fexceptions`)?  AFAIK mlx (the C++ core) also uses exceptions and compiles fine without this.

Libraries/MLXFoundationModels/MLXLanguageModel.swift
ctymoszek: Sure, how about we cache a bias struct like this:
```
struct TokenizerBias {
      let closing: MLXArray
      let whitespace: MLXArray
      let whitespaceTokenIDs: Set<Int>
  }
```

aleroot:
Just a comment not really related to the changes of this PR but in relation, should a new stable release be cut before merging this ? It is awhile there is no release, maybe before merging this is a good time?

davidkoski:
> Just a comment not really related to the changes of this PR but in relation, should a new stable release be cut before merging this ? It is awhile there is no release, maybe before merging this is a good time?

Yes, that was the plan.  This should land with no impact on existing libraries, but you never know.

There are still a lot of open PRs, but I think this backlog will take a while -- more updates keep coming in :-)  Maybe monday.

Libraries/MLXFoundationModels/MLXLanguageModel.swift
ctymoszek: Heads up: loadContainer is now an instance method following [this PR](https://github.com/ctymoszek/mlx-swift-lm/pull/8/)

Libraries/MLXFoundationModels/GuidedGeneration/GuidedGenerationLoop.swift
ctymoszek: This won't quite work for guided gen... Chris' caching work already improved the performance of the eval loop, but this sequencing problem is harder to solve.

davidkoski: It isn't a correctness issue, so IMHO get it in working and then think about how to solve it in a followup (if possible).

Libraries/MLXFoundationModels/GuidedGeneration/GuidedGenerationLoop.swift
ctymoszek: This was added - [PR here](https://github.com/ctymoszek/mlx-swift-lm/pull/7).

Libraries/MLXFoundationModels/GuidedGeneration/GuidedGenerationLoop.swift
ctymoszek: Not exactly cached - for guided generation, that won't work, but it's been relocated so this now happens during the forward pass ([PR here](https://github.com/ctymoszek/mlx-swift-lm/pull/7))

Package.swift
ctymoszek: Not a typo, just a silly jargon-y way to say "minimum version". Reworded.

Package.swift
ctymoszek: Addressed in [this PR](https://github.com/ctymoszek/mlx-swift-lm/pull/6)

davidkoski (approved):
Changes look good, thank you!

PR #335, opened 2026-06-09
Use MaterializedArray for Sendable conformance

davidkoski:
## Proposed changes

- adopt changes from https://github.com/ml-explore/mlx-swift/pull/418
- we don't need private box types -- the technique becomes general
- it also opens up some potential for synchronous evaluation
- `ModelContext` is now Sendable and `ModelContainer` is deprecated
- the model in `ModelContext` is now immutable -- suitable for inference, but not training/fine tuning
- `TrainableLanguageModel` and `TrainableModelContext` provide mutable models
- but require separate API (`loadTrainable`) to load

## Breaking API Changes

- `BaseLanguageModel` is not a `Module` any more -- not all `LanguageModels` are mutable
- see `TrainableLanguageModel` if mutation is needed (past load)
- `UserInput.Image`/`Audio`, `MLXArray` -> `MaterializedArray`
- `ModelContainer` is deprecated (`ModelContext` is now Sendable) and contains a `TrainableModelContext` rather than a `ModelContext` (now immutable)
    - only code that references the types directly will need to change

See https://github.com/ml-explore/mlx-swift-examples/pull/488 for adoption of the new API.

## Checklist

### DIFF
--- a/IntegrationTesting/IntegrationTestingTests/MTPRung4TokenParityTests.swift
+++ b/IntegrationTesting/IntegrationTestingTests/MTPRung4TokenParityTests.swift
@@ -36,13 +36,6 @@ private func mtpFixturesDirOrSkip(name: String) async -> URL? {
 // expensive (~30–60s); the `@Suite(.serialized)` wrapper below ensures
 // the three Rung 4 case methods run sequentially, and this module-level
 // cache reuses one (drafter, target) pair across all three.
-//
-// `Gemma4AssistantDraftModel` is deliberately non-Sendable (see the
-// design note at Gemma4Assistant.swift:153–155 — Embedding is a
-// reference type and cross-domain access must go through
-// `MTPDrafterContainer.perform`). `nonisolated(unsafe)` is appropriate
-// here because the suite is serialized — only one test runs at a time,
-// and the cache becomes read-only after first population.
 
 private struct Rung4BoundDrafter {
     let drafter: Gemma4AssistantDraftModel
--- a/Libraries/BenchmarkHelpers/BenchmarkHelpers.swift
+++ b/Libraries/BenchmarkHelpers/BenchmarkHelpers.swift
@@ -413,7 +413,7 @@ public func benchmarkEmbeddingLoading(
         id: "mlx-community/Qwen3-Embedding-0.6B-8bit"),
     runs: Int = BenchmarkDefaults.loadingRuns
 ) async throws -> BenchmarkStats {
-    _ = try await EmbedderModelFactory.shared.loadContainer(
+    _ = try await EmbedderModelFactory.shared.load(
         from: downloader, using: tokenizerLoader, configuration: configuration
     ) { _ in }
     Memory.clearCache()
@@ -421,7 +421,7 @@ public func benchmarkEmbeddingLoading(
     var times: [Double] = []
     for i in 1 ... runs {
         let start = CFAbsoluteTimeGetCurrent()
-        _ = try await EmbedderModelFactory.shared.loadContainer(
+        _ = try await EmbedderModelFactory.shared.load(
             from: downloader, using: tokenizerLoader, configuration: configuration
         ) { _ in }
         let elapsed = (CFAbsoluteTimeGetCurrent() - start) * 1000
@@ -542,7 +542,7 @@ public func benchmarkLLMGeneration(
     runs: Int = BenchmarkDefaults.loadingRuns
 ) async throws -> LLMGenerationStats {
     let configuration = MLXLMCommon.ModelConfiguration(id: modelId)
-    let container = try await LLMModelFactory.shared.loadContainer(
+    let context = try await LLMModelFactory.shared.load(
         from: downloader, using: tokenizerLoader, configuration: configuration
     ) { _ in }
 
@@ -558,9 +558,9 @@ public func benchmarkLLMGeneration(
     // Warm-up: prime caches and weights. UserInput is `sending`, so build a
     // fresh instance for each call.
     do {
-        let warmInput = try await container.prepare(input: UserInput(prompt: benchmarkPrompt))
-        let warmStream = try await container.generate(
-            input: warmInput, parameters: parameters)
+        let warmInput = try await context.processor.prepare(
+            input: UserInput(prompt: benchmarkPrompt))
+        let warmStream = try generate(input: warmInput, parameters: parameters, context: context)
         for await _ in warmStream {}
     }
 
@@ -570,8 +570,8 @@ public func benchmarkLLMGeneration(
     var generationTokenCount = 0
 
     for i in 1 ... runs {
-        let input = try await container.prepare(input: UserInput(prompt: benchmarkPrompt))
-        let stream = try await container.generate(input: input, parameters: parameters)
+        let input = try await context.processor.prepare(input: UserInput(prompt: benchmarkPrompt))
+        let stream = try generate(input: input, parameters: parameters, context: context)
         var info: GenerateCompletionInfo?
         for await item in stream {
             if case .info(let i) = item { info = i }
--- a/Libraries/MLXEmbedders/EmbedderModelContainer.swift
+++ b/Libraries/MLXEmbedders/EmbedderModelContainer.swift
@@ -31,10 +31,10 @@ import MLXLMCommon
 ///
 /// ## Implementation Note
 ///
-/// Previously the `ModelContainer` held the `EmbedderModelContext` in a `SerialAccessContainer` -- an internal type
-/// that provided a lock-like exclusive access for ``perform(_:)-((ModelContext)->R)`` and ``update(_:)``.
+/// Previously the `EmbedderModelContainer` held the `EmbedderModelContext` in a `SerialAccessContainer` -- an internal type
+/// that provided a lock-like exclusive access for ``perform(_:)-((EmbedderModelContext)->R)`` and ``update(_:)``.
 /// The `EmbedderModelContext` was not `Sendable` and this provided the `@unchecked Sendable` protection needed.
-/// In practice, some code would use ``perform(_:)-((ModelContext)->R)`` to
+/// In practice, some code would use ``perform(_:)-((EmbedderModelContext)->R)`` to
 /// _borrow_ the model.  This wouldn't have been safe if
 /// another thread was mutating through the reference, of course.
 ///
@@ -146,5 +146,3 @@ public final class EmbedderModelContainer: @unchecked (Sendable) {
         }
     }
 }
-
-public typealias EmbedderModelContainerConstraint = EmbedderModelContainer
--- a/Libraries/MLXEmbedders/EmbedderModelContainer.swift
+++ b/Libraries/MLXEmbedders/EmbedderModelContainer.swift
@@ -5,6 +5,8 @@ import MLXLMCommon
 
 /// Container for embedder models that guarantees single threaded access.
 ///
+/// * Important: `EmbedderModelContext` is now `Sendable` and can be used directly.
+///
 /// Wrap models used by e.g. the UI in a ModelContainer. Callers can access
 /// the model and/or tokenizer (any values from the ``EmbedderModelContext``):
 ///
@@ -26,53 +28,84 @@ import MLXLMCommon
 ///     return result.map { $0.asArray(Float.self) }
 /// }
 /// ```
-public final class EmbedderModelContainer: Sendable {
-    private let context: SerialAccessContainer<EmbedderModelContext>
+///
+/// ## Implementation Note
+///
+/// Previously the `ModelContainer` held the `EmbedderModelContext` in a `SerialAccessContainer` -- an internal type
+/// that provided a lock-like exclusive access for ``perform(_:)-((ModelContext)->R)`` and ``update(_:)``.
+/// The `EmbedderModelContext` was not `Sendable` and this provided the `@unchecked Sendable` protection needed.
+/// In practice, some code would use ``perform(_:)-((ModelContext)->R)`` to
+/// _borrow_ the model.  This wouldn't have been safe if
+/// another thread was mutating through the reference, of course.
+///
+/// The new code uses an `NSLock` to guard access to the `EmbedderModelContext`.  The context is now `Sendable`
+/// and the model itself is immutable.  This now allows concurrent _use_ of the context -- all reads of the struct
+/// itself are done under the lock.  Callers to ``update(_:)`` can modify the context (to a lesser extent than before)
+/// and this is done with the lock held.
+///
+/// Ideally, all use cases will move to use `EmbedderModelContext` directly, but in the meantime be aware of this
+/// change in implementation.
+@available(*, deprecated, message: "use EmbedderModelContext instead")
+public final class EmbedderModelContainer: @unchecked (Sendable) {
+    private var _context: EmbedderModelContext
+    private let lock = NSLock()
+    private var context: EmbedderModelContext {
+        get {
+            lock.lock()
+            defer { lock.unlock() }
+            return _context
+        }
+        set {
+            lock.lock()
+            defer { lock.unlock() }
+            _context = newValue
+        }
+    }
 
     public var configuration: ModelConfiguration {
-        get async {
-            await context.read { $0.configuration }
-        }
+        context.configuration
     }
 
     public var tokenizer: Tokenizer {
-        get async {
-            await context.read { $0.tokenizer }
-        }
+        context.tokenizer
     }
 
     public var poolingStrategy: Pooling.Strategy {
-        get async {
-            await context.read { $0.pooling.strategy }
-        }
+        context.pooling.strategy
     }
 
     public init(context: consuming EmbedderModelContext) {
-        self.context = .init(context)
+        self._context = context
     }
 
     /// Perform an action on the ``EmbedderModelContext``.
     /// Callers _must_ eval any `MLXArray` before returning as `MLXArray` is not `Sendable`.
     ///
-    /// - Note: The closure receives `EmbedderModelContext` which is not `Sendable`. This is intentional -
-    ///   the closure runs within the actor's isolation, ensuring thread-safe access to the model.
     /// - Note: The `sending` keyword indicates the return value is transferred (not shared) across
     ///   isolation boundaries, allowing non-Sendable types to be safely returned.
     public func perform<R: Sendable>(
         _ action: @Sendable (EmbedderModelContext) async throws -> sending R
     ) async rethrows -> sending R {
-        try await context.read {
-            try await action($0)
-        }
+        try await action(context)
+    }
+
+    /// Perform an action on the ``EmbedderModelContext``.
+    ///
+    /// This is the synchronous form of ``perform(_:)`` and has
+    /// fewer restrictions.
+    public func perform<R>(
+        _ action: @Sendable (EmbedderModelContext) throws -> R
+    ) rethrows -> R {
+        try action(context)
     }
 
     @available(*, deprecated, message: "use perform(_: (EmbedderModelContext) -> R) instead")
     public func perform<R: Sendable>(
         _ action: @Sendable (EmbeddingModel, Tokenizer, Pooling) async throws -> sending R
     ) async rethrows -> sending R {
-        try await context.read {
-            try await action($0.model, $0.tokenizer, $0.pooling)
-        }
+        try await action(
+            context.model, context.tokenizer, context.pooling
+        )
     }
 
     /// Perform an action on the ``EmbedderModelContext`` with additional (non `Sendable`) context values.
@@ -82,17 +115,18 @@ public final class EmbedderModelContainer: Sendable {
         nonSendable values: consuming V,
         _ action: @Sendable (EmbedderModelContext, V) async throws -> R
     ) async rethrows -> sending R {
-        let values = SendableBox(values)
-        return try await context.read {
-            try await action($0, values.consume())
-        }
+        try await action(context, values)
     }
 
     /// Update the owned `EmbedderModelContext`.
     /// - Parameter action: update action
+    @available(
+        *, deprecated,
+        message: "mutate EmbedderModelContext before passing to EmbedderModelContainer"
+    )
     public func update(_ action: @Sendable (inout EmbedderModelContext) -> Void) async {
-        await context.update {
-            action(&$0)
+        lock.withLock {
+            action(&_context)
         }
     }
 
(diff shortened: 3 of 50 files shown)


Libraries/MLXEmbedders/EmbedderModelContainer.swift
```diff
@@ -27,52 +29,52 @@ import MLXLMCommon
 /// }
 /// ```
 public final class EmbedderModelContainer: Sendable {
-    private let context: SerialAccessContainer<EmbedderModelContext>
+    private let context: EmbedderModelContext
```
davidkoski: We don't need this as the Context is now Sendable

Libraries/MLXEmbedders/EmbedderModelContainer.swift
```diff
@@ -27,52 +29,52 @@ import MLXLMCommon
...
 
     public var configuration: ModelConfiguration {
-        get async {
-            await context.read { $0.configuration }
-        }
+        context.configuration
```
davidkoski, line 66: These become synchronous accessors.

Libraries/MLXEmbedders/EmbedderModelContainer.swift
```diff
@@ -27,52 +29,52 @@ import MLXLMCommon
...
+        try await action(context)
+    }
+
+    /// Perform an action on the ``EmbedderModelContext``.
+    ///
+    /// This is the synchronous form of ``perform(_:)`` and has
+    /// fewer restrictions.
+    public func perform<R>(
+        _ action: @Sendable (EmbedderModelContext) throws -> R
+    ) rethrows -> R {
```
davidkoski, line 98: And we can have a synchronous `perform()`

Libraries/MLXEmbedders/EmbedderModelContainer.swift
```diff
@@ -27,52 +29,52 @@ import MLXLMCommon
 /// }
 /// ```
 public final class EmbedderModelContainer: Sendable {
```
davidkoski: Is this type still needed?  Maybe.  `EmbedderModelContext` is a struct, so this gives reference semantics -- users share the instance.  Potentially `EmbedderModelContext` could become a class and we remove this?  It would need to be immutable to do so.  Keeping it for now.

Libraries/MLXEmbedders/EmbeddingModel.swift
```diff
@@ -48,3 +48,31 @@ extension EmbeddingModel {
             attentionMask: attentionMask)
     }
 }
+
+extension MaterializedModule: EmbeddingModel, BaseLanguageModel where LayerType: EmbeddingModel {
```
davidkoski: This is how we make `MaterializedModule` usable as a `EmbeddingModel`

Libraries/MLXEmbedders/ModelFactory.swift
```diff
@@ -97,18 +97,18 @@ public class EmbedderRegistry: AbstractModelRegistry, @unchecked Sendable {
 ///
 /// This is created using a ``EmbedderModelFactory`` and often used
 /// inside a ``EmbedderModelContainer``.
-public struct EmbedderModelContext {
+public struct EmbedderModelContext: Sendable {
     public var configuration: ModelConfiguration
-    public var model: any EmbeddingModel
+    public var model: any EmbeddingModel & Sendable
     public var tokenizer: any Tokenizer
     public let pooling: Pooling
```
davidkoski, line 132: Potentially we want this to be `var`

Libraries/MLXEmbedders/Pooling.swift
```diff
@@ -63,7 +63,7 @@ func loadPooling(modelDirectory: URL, model: EmbeddingModel) -> Pooling {
 ///
 /// `Pooling` takes the sequence of hidden states from a transformer model and collapses them
 /// into a single vector using strategies like mean, max, or token selection.
-open class Pooling: Module {
+public struct Pooling: Sendable {
```
davidkoski, line 83: I don't see why this should be `Module`.  It isn't attached to the model itself (so e.g. `update()` semantics are not needed).  I changed it like this and everything built cleanly.

Package.swift
```diff
@@ -36,7 +36,7 @@ let package = Package(
             targets: ["IntegrationTestHelpers"]),
     ],
     dependencies: [
-        .package(url: "https://github.com/ml-explore/mlx-swift", .upToNextMinor(from: "0.31.4")),
+        .package(url: "https://github.com/ml-explore/mlx-swift", branch: "materialized-array"),
```
davidkoski: Pick up that branch for now.

Libraries/MLXEmbedders/EmbedderModelContainer.swift
```diff
@@ -27,52 +29,52 @@ import MLXLMCommon
...
-    ///   the closure runs within the actor's isolation, ensuring thread-safe access to the model.
     /// - Note: The `sending` keyword indicates the return value is transferred (not shared) across
     ///   isolation boundaries, allowing non-Sendable types to be safely returned.
     public func perform<R: Sendable>(
         _ action: @Sendable (EmbedderModelContext) async throws -> sending R
     ) async rethrows -> sending R {
-        try await context.read {
-            try await action($0)
-        }
+        try await action(context)
```
aleroot, line 89: This seems to drop the serialization guarantee previously provided. The old `SerialAccessContainer.read` held an async mutex for the full duration of the closure, including suspension points, so concurrent `perform` calls could not overlap on the same model/tokenizer/pooling context.

With this direct call, two tasks can now enter `perform` concurrently and use the same underlying context values at the same time, while the type-level documentation still says the container “guarantees single threaded access.” Even if `MaterializedModule` makes the wrapped model sendable, that does not by itself preserve the previous exclusive-access contract for all work reachable through the context.

This may not be a real issue, except a documentation kind of inconsistency

Libraries/MLXEmbedders/ModelFactory.swift
```diff
@@ -97,18 +97,18 @@ public class EmbedderRegistry: AbstractModelRegistry, @unchecked Sendable {
...
 
     public init(
-        configuration: ModelConfiguration, model: any EmbeddingModel,
+        configuration: ModelConfiguration, model: some EmbeddingModel,
         tokenizer: any Tokenizer, pooling: Pooling
     ) {
         self.configuration = configuration
-        self.model = model
+        self.model = MaterializedModule(model)
```
aleroot, line 139: `context.model` is still exposed as `any EmbeddingModel & Sendable`; `EmbeddingModel` inherits `BaseLanguageModel`, which inherits the `Module` mutation APIs. This now stores a `MaterializedModule`, whose `update`/`apply`/`train`/`freeze` overrides trap with `fatalError`. Existing callers mutating the model through `perform { try $0.model.update(...) }` can still compile through the existential API but now crash at runtime.

Libraries/MLXEmbedders/EmbedderModelContainer.swift
davidkoski: I think that is a good point -- it won't matter as much for the Embedders because they don't have state, but for example the KVCache _is_ state.  It has to have exclusive access or be some kind of copy-in/copy-out setup.

For example in `ChatSession` it had to play some games with private classes and knowledge of what was safe:

```swift
    private func streamMap<R: Sendable>() {
                try await cache.update { cache in

                    let model = await model.perform { context in
                        SendableBox(context.model)
                    }.consume()
```

This would get exclusive access to the KVCache and then "borrow" the model/weights -- it treated them as Sendable, but they were not represented that way in the type system.  This had to be internal implementation because it required care to make sure it was thread safe.

My hope is that providing Sendable pieces would allow anyone to write this code and do it safely.

But your point about what is serial access vs not is important.

Libraries/MLXEmbedders/ModelFactory.swift
davidkoski: Yes, I would prefer it to be typed as `MaterializedModule` and then these functions are unavailable.  That would require either the `ModelContext` to be generic OR `MaterializedModule` to not be generic (probably subclasses).

I need to play around with that and see if one works better than the other.  Just getting this to compile took some effort -- perhaps it can be done better.  This change has to be done in the context of integrating it!

The fact that the Materialized variants violate Liskov substitution isn't great, but I think the benefits of documenting and obtaining Sendable in a way that is interoperable is probably worth it, but I would appreciate feedback from people who use it!

davidkoski: I should note that `MaterializedModule` has these methods marked as `unavailable` so the compile time warning would be available if it were typed like that.

davidkoski: A non-generic `MaterializedModule` won't work -- it has to be `final` for the `Sendable` to work on it.  I can't make it a `struct` because you can't subclass a struct.  

I will look at making `ModelContext` generic, but that will change things all over the place.

We may have to live with some limitations on `MaterializedModule`, but perhaps if it is encapsulated inside things like `ModelContext` it won't be an issue.

IntegrationTesting/IntegrationTestingTests/MTPRung4TokenParityTests.swift
```diff
@@ -34,13 +34,6 @@ private func mtpFixturesDirOrSkip(name: String) async -> URL? {
 // expensive (~30–60s); the `@Suite(.serialized)` wrapper below ensures
 // the three Rung 4 case methods run sequentially, and this module-level
 // cache reuses one (drafter, target) pair across all three.
-//
-// `Gemma4AssistantDraftModel` is deliberately non-Sendable (see the
-// design note at Gemma4Assistant.swift:153–155 — Embedding is a
-// reference type and cross-domain access must go through
-// `MTPDrafterContainer.perform`). `nonisolated(unsafe)` is appropriate
-// here because the suite is serialized — only one test runs at a time,
-// and the cache becomes read-only after first population.
```
davidkoski, line 45: This comment is obsolete, though the cache here looks like it probably needs at least a lock -- it isn't thread safe.

Libraries/BenchmarkHelpers/BenchmarkHelpers.swift
```diff
@@ -558,9 +558,9 @@ public func benchmarkLLMGeneration(
     // Warm-up: prime caches and weights. UserInput is `sending`, so build a
     // fresh instance for each call.
     do {
-        let warmInput = try await container.prepare(input: UserInput(prompt: benchmarkPrompt))
-        let warmStream = try await container.generate(
-            input: warmInput, parameters: parameters)
+        let warmInput = try await context.processor.prepare(
+            input: UserInput(prompt: benchmarkPrompt))
+        let warmStream = try generate(input: warmInput, parameters: parameters, context: context)
```
davidkoski, line 563: Use ModelContext instead of ModelContainer.  The latter still works, but is deprecated.

Libraries/MLXEmbedders/EmbedderModelContainer.swift
```diff
@@ -26,53 +28,67 @@ import MLXLMCommon
 ///     return result.map { $0.asArray(Float.self) }
 /// }
 /// ```
-public final class EmbedderModelContainer: Sendable {
-    private let context: SerialAccessContainer<EmbedderModelContext>
+@available(*, deprecated, message: "use EmbedderModelContext instead")
```
davidkoski, line 48: We can start removing this.  Everything should still work but it is no longer needed for Sendable conformance.

Libraries/MLXEmbedders/EmbedderModelContainer.swift
```diff
@@ -26,53 +28,67 @@ import MLXLMCommon
...
+public final class EmbedderModelContainer: @unchecked (Sendable) {
+    private var _context: EmbedderModelContext
+    private let lock = NSLock()
+    private var context: EmbedderModelContext {
```
davidkoski, line 52: Because the API allows mutation (of a value type) we do it under lock so the mutation itself is thread safe.

Libraries/MLXLMCommon/ChatSession.swift
```diff
@@ -40,8 +40,8 @@ import MLX
 public struct SpeculativeDecodingConfig: Sendable {
```
davidkoski, line 43: I will file an issue for this, but I think this needs to be a reference type -- as it is written, if you have multiple ChatSessions with deferred loads on the draft model, each of them will load it into memory.  It works but is probably not the intent (since the whole idea around it is being sensitive to memory pressure).

Libraries/MLXLMCommon/ChatSession.swift
```diff
@@ -407,9 +412,9 @@ public final class ChatSession {
     public func respond(
         to prompt: String,
         role: Chat.Message.Role = .user,
-        images: consuming [UserInput.Image],
-        videos: consuming [UserInput.Video],
-        audios: consuming [UserInput.Audio]
+        images: [UserInput.Image],
+        videos: [UserInput.Video],
+        audios: [UserInput.Audio]
```
davidkoski, line 424: These types are all Sendable now so this doesn't have to be consuming.

Libraries/MLXLMCommon/ChatSession.swift
```diff
@@ -565,51 +570,30 @@ public final class ChatSession {
...
-                    // should not be used in calling code.
-                    //
-                    // The benefit is that callers can be running multiple
-                    // ChatSessions in parallel, as long as the instances
-                    // are distinct.  In particular the KVCache cannot
-                    // be shared and that is the lock that is held here.
-
-                    let model = await model.perform { context in
-                        SendableBox(context.model)
-                    }.consume()
```
davidkoski, line 618: This fancy dance is no longer needed since the model itself is now Sendable.  I think this is one of the biggest gains -- you no longer need tricky code and long explanations about why you think it is safe.

Libraries/MLXLMCommon/Evaluate.swift
```diff
@@ -1190,21 +1190,7 @@ public func generate(
...
-    let input = LMInput(tokens: tokens)
-    let configuration = ModelConfiguration(id: "stand-in", extraEOSTokens: extraEOSTokens ?? [])
-    let context = ModelContext(
-        configuration: configuration, model: model, processor: StandInUserInputProcessor(),
-        tokenizer: tokenizer)
-
-    return generate(
-        input: input, context: context, iterator: iterator,
-        didGenerate: didGenerate)
+    fatalError("not implemented")
```
davidkoski, line 1276: This has been deprecated for a while.  I think it _could_ be ported, but ModelContext would have to accept (I think) MaterializedModule also.  Easier to mark as unavailable, but if it becomes an issue then we look at bringing it back.

Libraries/MLXLMCommon/LanguageModel.swift
```diff
@@ -29,6 +33,14 @@ extension BaseLanguageModel {
     }
 }
 
+extension BaseLanguageModel where Self: Module {
+    public var parameterNBytes: Int {
+        parameters().reduce(0) { $0 + $1.nbytes }
+    }
+}
+
+public typealias TrainableBaseLanguageModel = BaseLanguageModel & Module
```
davidkoski, line 53: BaseLanguageModel (LanguageModel) is no longer a Module.  That looks fine except in the LoRA case, where it is a breaking change.  We want this because we want to wrap the trainable (Module) based models with MaterializedModule to make them Sendable (for inference cases).

Libraries/MLXLMCommon/UserInput.swift
```diff
@@ -75,10 +75,10 @@ public struct UserInput {
     }
 
     /// Representation of an image resource.
-    public enum Image {
+    public enum Image: Sendable {
         case ciImage(CIImage)
         case url(URL)
-        case array(MLXArray)
+        case array(MaterializedArray)
```
davidkoski, line 113: Potential source break

Libraries/MLXEmbedders/ModelFactory.swift
davidkoski: OK, MaterializedModule is no longer a Module and I think that removes a lot of API surface that would e confusing.

It does mean if you want to mutate the model after load you have to indicate that when you load it (`loadTrainable` vs `load`).  There is a `TrainableModelContext` and a `ModelContext` to keep these separate.

Everything should work as it did with the exception of LoRA style fine tuning -- the fact that the model is immutable on load breaks that path.  I will update the PR description to reflect the breaks.

davidkoski:
@ronaldmannak @aleroot @manojmahapatra FYI -- I think this is ready to go, along with the equivalent https://github.com/ml-explore/mlx-swift/pull/418

There are a couple of breaking changes -- we can't have both mutable and Sendable (which we did before -- it was fine unless you used it!)

Libraries/MLXHuggingFace/Macros.swift
```diff
@@ -135,6 +137,24 @@ public macro huggingFaceLoadModel(
...
+/// import MLXHuggingFace
+/// import HuggingFace
+/// import Tokenizers
+///
+/// let modelContext = try await huggingFaceLoadTrainabledModel(
+///     configuration: modelConfiguration
+/// ) { progres in ... }
+/// ```
+@freestanding(expression)
+public macro huggingFaceLoadTrainabledModel(
```
manojmahapatra: Question: `huggingFaceLoadTrainabledModel` looks like this is a typo?

davidkoski: Doh!

Libraries/MLXHuggingFace/Macros.swift
```diff
@@ -135,6 +137,24 @@ public macro huggingFaceLoadModel(
 ) -> ModelContext =
     #externalMacro(module: "MLXHuggingFaceMacros", type: "LoadContextMacro")
 
+/// Load a `TrainableModelContext` using default hub client and tokenizer loader with progress.
```
manojmahapatra, line 140: Question: Do we also need to add a no progressHandler overload of `huggingFaceLoadTrainableModel`?

davidkoski: For consistency we probably should, but let me see if I can make it work with a default on the progress handler (or maybe nullable).  The macro side actually supplies a default if not present.

manojmahapatra: > The macro side actually supplies a default if not present.

yea, I see that.

manojmahapatra:
I think overall the changes look good to me. 👍

davidkoski:
> @ronaldmannak @aleroot @manojmahapatra FYI -- I think this is ready to go, along with the equivalent [ml-explore/mlx-swift#418](https://github.com/ml-explore/mlx-swift/pull/418)
> 
> There are a couple of breaking changes -- we can't have both mutable and Sendable (which we did before -- it was fine unless you used it!)

I was thinking about the breaking change for `ModelContainer`.  I think I might be able to hold a mutable model (behind a lock) and callers can `update()` it.  Then when done and you try to run inference it can switch it to a `MaterializedModule`.  It can't switch back (you could reload it) but this would work for the examples -- they train or augment, then do inference.

I will give that a try in the morning.  This API break is the only thing that gives me pause.  I think the rest of it cleans up a lot of sensibility issues nicely.

manojmahapatra:
> I was thinking about the breaking change for ModelContainer. I think I might be able to hold a mutable model (behind a lock) and callers can update() it. Then when done and you try to run inference it can switch it to a MaterializedModule. It can't switch back (you could reload it) but this would work for the examples -- they train or augment, then do inference.

That sounds reasonable to me. one question - would the state machine be explicit here? In short, the container stays mutable until the first inference call, then transitions one way into a materialized state afterward?

aleroot (approved):

davidkoski:
> That sounds reasonable to me. one question - would the state machine be explicit here? In short, the container stays mutable until the first inference call, then transitions one way into a materialized state afterward?

Yes, exactly.

Though looking at mlx-swift-examples, I am not sure it will work.  It does the `LoRAContainer` augmentation in `perform()`.  This "works" because it doesn't mutate the `ModelContext`, it mutates the `Module` itself.

In fact there are no calls to `update()` at all because it mutate the reference type.

Maybe better just to take the API hit and fix this correctly.  😭

Libraries/MLXLMCommon/Load.swift
```diff
@@ -12,7 +12,7 @@ import MLXNN
 /// applies optional quantization, and
 /// updates the model with the weights.
 public func loadWeights(
-    modelDirectory: URL, model: BaseLanguageModel,
```
manojmahapatra, line 43: Question: I pulled the PR locally and it fails to build because `ModelConversion.convert(...)` still takes `model` as `BaseLanguageModel`, while the implementation uses `Module` APIs like `parameters()`, `leafModules()`, and `update(...)`. Should this be changed to `TrainableBaseLanguageModel` / `BaseLanguageModel & Module`?

Libraries/MLXLMCommon/ModelContainer.swift
```diff
@@ -6,6 +6,8 @@ import MLXNN
 
 /// Container for models that guarantees single threaded access.
```
manojmahapatra, line 7: Suggestion: The “guarantees single threaded access” wording feels misleading now. `perform` only reads `_context` under the lock; the closure itself runs after that access is complete, so multiple `perform` calls can execute concurrently. Should we update this doc comment to describe `ModelContainer` as a deprecated lightweight wrapper around `ModelContext` instead?

Libraries/MLXLMCommon/Load.swift
davidkoski: Yes, that sounds right -- it it curious that it was building for me.  I will take a look today.  Thanks!

Libraries/MLXLMCommon/ModelContainer.swift
davidkoski: Yes, good idea.  I can document how it used to work vs how it works now.  In practice things like ChatSession actually used it this way (the current way) but had to do so very very carefully.

davidkoski:
- [x] investigate ModelConversion.convert
- [x] update docs on ModelContainer

Libraries/MLXLMCommon/Load.swift
davidkoski: Rebased on main & this issue is fixed.

Libraries/MLXLMCommon/ModelContainer.swift
davidkoski: Updated the documentation on `ModelContainer`.

davidkoski:
OK, I thought about `ModelContainer` and came up with a new approach that is _mostly_ not source breaking.  Details are in the class docs for it.

`llm-tool` from mlx-swift-examples requires two changes:

```diff
--- a/Tools/llm-tool/LLMTool.swift
+++ b/Tools/llm-tool/LLMTool.swift
@@ -196,7 +196,7 @@ struct GenerateArguments: ParsableArguments, Sendable {
     }
 
     func prepare(
-        _ context: inout ModelContext
+        _ context: inout TrainableModelContext
     ) {
         if let extraEosToken {
             context.configuration.extraEOSTokens.insert(extraEosToken)
@@ -204,7 +204,7 @@ struct GenerateArguments: ParsableArguments, Sendable {
     }
 
     func generate(
-        input: LMInput, context: ModelContext
+        input: LMInput, context: ModelContextProviding
     ) async throws -> (GenerateCompletionInfo, String) {
         var output = ""
         for await item in try MLXLMCommon.generate(
```

The LoRATrainingExample (app) requires none.

Issue #336, opened 2026-06-10
[BUG] MLXVLM models ignore windowSize — prompt prefill is unbounded, memory scales with prompt length


joshuaejackson:
`GenerateParameters.prefillStepSize` flows through `TokenIterator` into `LanguageModel.prepare(_:cache:windowSize:)`. `LLMModel.prepare` honors it (chunks the prompt with `eval(cache)` between chunks), but 13 of 14 MLXVLM model implementations ignore it and run the entire prompt in a single forward pass — two (`Qwen35`, `Qwen3VL`) declare the parameter as `windowSize _: Int?`. Only `Idefics3` chunks (fixed in #297; `SmolVLM2` inherits it via typealias).
 
Affected: FastVLM, Gemma3, Gemma4, GlmOcr, LFM2VL, Mistral3, Paligemma, Pixtral, Qwen25VL, Qwen2VL, Qwen35 (+ Qwen35MoE), Qwen3VL. The porting template in `Libraries/MLXVLM/README.md` also shows an unchunked `prepare`, so new ports inherit the pattern.
 
Single-pass prefill allocates transient buffers proportional to prompt length (~1.7 MB/token measured on Gemma4 E4B 4-bit). Likely the cause of #209.

**To Reproduce:**

```swift
// Any MLXVLM model; measured with mlx-community/gemma-4-e4b-it-4bit on an M3 Pro / 18 GB.
let params = GenerateParameters(prefillStepSize: 512, temperature: 0)

// generate() with a text-only prompt of ~8k tokens, then:
print(MLX.Memory.peakMemory)  // ~17.7 GB — prefillStepSize had no effect

// With #297-style chunking in Gemma4.prepare: ~5.1 GB for the same prompt.
```

**Expected behavior:**
Prefill memory bounded by `prefillStepSize`, as in `LLMModel.prepare` and Idefics3.

**Desktop:**
 - OS Version: macOS 26.1
 - Device: M3 Pro MacBook Pro, 18 GB
 - Version: main (also 3.31.3)

**Additional context:**
Measured on Gemma4 E4B 4-bit (5 iterations/point, temp 0, medians):
 
| prompt | peak MLX active, single-pass | chunked (512) |
|---|---|---|
| 4k tokens | 11.2 GB | 5.0 GB |
| 8k tokens | 17.7 GB | 5.1 GB |
| 16k tokens | crashes (≈33 GB projected) | 5.2 GB |
 
[...]

PR #337, opened 2026-06-10
fix Gemma4 prepare() to honor windowSize -- prompt prefill is never chunked

joshuaejackson:
## Proposed changes
Addresses the Gemma4 instance of #336 (MLXVLM models ignore `windowSize` — same fix shape as #297).

`Gemma4.prepare(_:cache:windowSize:)` accepts `windowSize` but never uses it: the whole prompt runs through the language model in one forward pass, so transient prefill memory scales with prompt length (~1.7 MB/token measured on E4B 4-bit — a 4k-token prompt peaks at 11.2 GB MLX active against 4.3 GiB of weights; 16k crashes an 18 GB machine).

This applies the same chunking #297 added for Idefics3, to both Gemma4 branches: the text path chunks raw tokens; the image path chunks the merged embeddings together with the aligned per-layer-input slices. The KV cache is evaluated between chunks; the final position returns the first-token logits.

Gemma4-specific notes: KV-shared layers read the source layer's cache-augmented keys/values, so per-chunk forwards still see full history; per-layer inputs are per-token; RoPE offsets and masks derive from `cache.offset`. 

## Verification (Gemma4 E4B 4-bit, M3 Pro 18 GB)

- **Memory** (5 iters/point, temp 0, medians): peak MLX active 11.2→5.0 GB at 4k, 17.7→5.1 GB at 8k; 16k was un-runnable before (~33 GB projected) and peaks at 5.2 GB after. TTFT improved 25.6→17.9 s (4k) and 70.5→47.1 s (8k).
- **Output equality:** greedy outputs byte-identical before/after at 34/492/573/733-token prompts (spanning the 512 chunk boundary). At 4k/8k, long free-form greedy generations are deterministic per build and can diverge at near-tie tokens (floating-point evaluation order — same class of variation as changing `prefillStepSize` on `LLMModel`); needle-recall scoring was 5/5 in both configurations at all lengths.

## Checklist

### DIFF
--- a/Libraries/MLXVLM/Models/Gemma4.swift
+++ b/Libraries/MLXVLM/Models/Gemma4.swift
@@ -1728,9 +1728,10 @@ public final class Gemma4: Module, VLMModel, KVCacheDimensionProvider {
             let (allEmbeds, allPerLayerInputs) = try getInputEmbeddings(
                 inputIds: input.text.tokens, pixelValues: imagePixels)
             // Prefill the merged image+text embeddings (and the aligned
-            // per-layer inputs) in windowSize-sized chunks, evaluating the
-            // KV cache between chunks; the final position yields the
-            // first-token logits. Matches LLMModel.prepare and #297.
+            // per-layer inputs) in windowSize-sized chunks; the final
+            // position yields the first-token logits. Matches
+            // LLMModel.prepare and #297. asyncEval lets the CPU build
+            // chunk N+1's graph while the GPU evaluates chunk N.
             let totalPositions = allEmbeds.dim(1)
             var processed = 0
             while totalPositions - processed > 1 {
@@ -1742,9 +1743,11 @@ public final class Gemma4: Module, VLMModel, KVCacheDimensionProvider {
                     inputsEmbeds: allEmbeds[0..., range, 0...],
                     perLayerInputs: allPerLayerInputs.map { $0[0..., range, 0..., 0...] }
                 )
-                eval(cache)
+                asyncEval(cache)
                 processed += chunkLength
             }
+            // Single sync after the loop to flush any remaining async work.
+            eval(cache)
             let result = languageModel(
                 nil,
                 cache: convertedCache,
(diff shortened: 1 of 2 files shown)


Libraries/MLXVLM/Models/Gemma4.swift
```diff
@@ -1723,18 +1723,54 @@ public final class Gemma4: Module, VLMModel, KVCacheDimensionProvider {
...
+                let range = processed ..< (processed + chunkLength)
+                _ = languageModel(
+                    nil,
+                    cache: convertedCache,
+                    inputsEmbeds: allEmbeds[0..., range, 0...],
+                    perLayerInputs: allPerLayerInputs.map { $0[0..., range, 0..., 0...] }
+                )
+                eval(cache)
+                processed += chunkLength
+            }
```
davidkoski, line 1748: Since we are here touching it (and I suspect this pattern will be copied), let's use the asyncEval + eval pattern from LLMModel:

```swift
        // Prepare the prompt in chunks if larger than the prefill size.
        // asyncEval lets the CPU build chunk N+1's graph while the GPU evaluates
        // chunk N.
        var state: LMOutput.State?
        while y.tokens.size > prefillStepSize {
            let input = y[.newAxis, ..<prefillStepSize]
            let output = self(input, cache: cache.isEmpty ? nil : cache, state: state)
            state = output.state
            asyncEval(cache)
            y = y[prefillStepSize...]
        }

        // Single sync after the loop to flush any remaining async work.
        eval(cache)
```

asyncEval inside the loop and eval after

joshuaejackson: Thanks!! Done. Switched to asyncEval in the loop with a single eval(cache) after, in both branches. (Gemma4 doesn't carry decoder state, so I skipped the state threading from your snippet.)

I also added a test file — it builds a tiny random-weight Gemma4 and checks that chunked and single-pass prefill give the same answer. If the pattern gets copied, the test can get copied.

davidkoski (approved):

davidkoski:
Thank you for the contribution!

PR #340, opened 2026-06-11
Raise swift-syntax floor to 602.0.0 for prebuilt artifact resolution

roryford:
Narrowed from the original proposal. This PR is now only the swift-syntax floor bump, flagged as useful
  
Why: swift.org publishes signed prebuilt swift-syntax artifacts only for >= 602 tags on current toolchains. A 600.x/601.x resolution silently falls back to compiling swift-syntax from source. Measured on Swift 6.3.2 / Xcode 26.5: forcing source compile adds ~46 Swift-compile tasks / ~29s of compile CPU per cold build (SwiftCompile 24.1s → 53.2s) that the prebuilt skips entirely.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -37,7 +37,10 @@ let package = Package(
     ],
     dependencies: [
         .package(url: "https://github.com/ml-explore/mlx-swift", .upToNextMinor(from: "0.31.4")),
-        .package(url: "https://github.com/swiftlang/swift-syntax.git", "600.0.0" ..< "604.0.0"),
+        // 602.0.0 floor: swift.org publishes signed prebuilt swift-syntax artifacts only for
+        // >= 602 tags on current toolchains; a 600.x/601.x resolution falls back to the full
+        // source compile of swift-syntax.
+        .package(url: "https://github.com/swiftlang/swift-syntax.git", "602.0.0" ..< "604.0.0"),
     ],
     targets: [
         .target(


Package.swift
```diff
@@ -37,7 +40,12 @@ let package = Package(
     ],
     dependencies: [
         .package(url: "https://github.com/ml-explore/mlx-swift", .upToNextMinor(from: "0.31.4")),
-        .package(url: "https://github.com/swiftlang/swift-syntax.git", "600.0.0" ..< "604.0.0"),
+        // 602.0.0 floor: swift.org publishes signed prebuilt swift-syntax artifacts only for
+        // >= 602 tags on current toolchains; a 600.x/601.x resolution falls back to the full
+        // source compile of swift-syntax.
+        .package(url: "https://github.com/swiftlang/swift-syntax.git", "602.0.0" ..< "604.0.0"),
```
davidkoski, line 43: This looks useful!

Package.swift
```diff
@@ -37,7 +40,12 @@ let package = Package(
...
+        .package(url: "https://github.com/huggingface/swift-transformers", from: "1.3.0"),
+        .package(url: "https://github.com/huggingface/swift-huggingface", from: "0.9.0"),
```
davidkoski: We can't do this -- see #118 .  The package resolution can't be turned off and name collisions are not allowed.  There is some aliasing capability, but it was unable to solve it.  

This is how we ended up with the macros.  Or you can use the external (to mlx-explore) implementations.

davidkoski (approved):
Thank you!

PR #326, opened 2026-06-03
fix speculative decode tests

davidkoski:
## Proposed changes

- fix #315
- run models in float16 (more typical) and avoid the float32/tf32 mismatch

## Checklist

Put an `x` in the boxes that apply.

- [x] I have read the [CONTRIBUTING](https://github.com/ml-explore/mlx/blob/main/CONTRIBUTING.md) document
- [x] I have run `pre-commit run --all-files` to format my code / installed pre-commit prior to committing changes
- [x] I have added tests that prove my fix is effective or that my feature works
- [ ] I have updated the necessary documentation (if needed)

PR #344, opened 2026-06-12
fix MLXVLM prepare(): honor windowSize for chunked prefill (FastVLM / Gemma3 / Qwen2VL / LFM2VL / Pixtral / Mistral3)

### LINKED ISSUE #336: [BUG] MLXVLM models ignore windowSize — prompt prefill is unbounded, memory scales with prompt length

john-rocky:
Fixes #336.

Idefics3 was fixed in #297, SmolVLM2 inherits via typealias, Gemma4 was fixed in #337. This PR extends the same chunked-prefill pattern to the 6 remaining models that still run the full prompt in a single forward pass.

## Problem

`GenerateParameters.prefillStepSize` flows through `TokenIterator` into `VLMModel.prepare(_:cache:windowSize:)`, but 6 of the MLXVLM implementations ignore `windowSize` and run the entire merged embedding in one shot. Single-pass prefill allocates transient buffers proportional to prompt length.

Measured by the issue reporter on Gemma4 E4B 4-bit (M3 Pro 18 GB, 8k-token prompt):

| prefill | peak MLX active |
|---------|----------------|
| single-pass (pre-fix) | ~17.7 GB |
| chunked, windowSize=512 (post-#337) | ~5.1 GB |

## Fix

All 6 models follow the Gemma4 / #337 loop structure:

```swift
let totalPositions = allEmbeds.dim(1)
var processed = 0
while totalPositions - processed > 1 {
    let chunkLength = min(prefillStepSize, totalPositions - processed - 1)
    let range = processed ..< (processed + chunkLength)
    _ = languageModel(nil, cache: cache, inputEmbedding: allEmbeds[0..., range, 0...])
    asyncEval(cache)
    processed += chunkLength
}
eval(cache)
let result = languageModel(nil, cache: cache, inputEmbedding: allEmbeds[0..., processed..., 0...])
```

`asyncEval(cache)` lets the CPU build chunk N+1's graph while the GPU evaluates chunk N. A single `eval(cache)` sync before the last position keeps the semantics identical to the single-pass path.

Models fixed and their call pattern:

| Model | call pattern |
|-------|-------------|
| FastVLM | `languageModel(nil, cache:, inputEmbedding:)` |
| Gemma3 | embed path + text-only token-chunk path |
| Qwen2VL | `languageModel(nil, cache:, inputEmbedding:)` |
| LFM2VL | `languageModel(nil, cache:, inputsEmbeds:)` |
| Pixtral | `languageModel(inputIds, cache:, inputsEmbeds:)` — slices inputIds + embeds |
| Mistral3 | same as Pixtral |

README porting template updated to show chunked `prepare()`.

## Models deferred (different architectures — separate follow-up)

- **Qwen25VL / GlmOcr**: MROPE `positionIds` tensor `[3, batch, seq_len]` requires an additional slice axis; fixing without regressions needs more care.
- **Qwen35 / Qwen35MoE / Qwen3VL**: `languageModel` receives raw `pixelValues` / `deepstackEmbeds` directly rather than a pre-merged embedding; the chunking hook is inside the backbone, not in `prepare()`.
- **Paligemma**: uses a full 4D bidirectional attention mask `[batch, heads, seq_len, seq_len]` that would need to be sliced in both query and key dimensions.

## Tests

```
xcodebuild test -scheme mlx-swift-lm-Package -destination 'platform=macOS'
```

137/137 tests pass (including SpeculativeDecodingTests and Gemma4ChunkedPrefillTests added in #337).

### DIFF
--- a/Libraries/MLXVLM/Models/FastVLM.swift
+++ b/Libraries/MLXVLM/Models/FastVLM.swift
@@ -1135,7 +1135,18 @@ public class FastVLM: Module, VLMModel, KVCacheDimensionProvider {
             pixelValues: input.image?.pixels,
             mask: input.text.mask
         )
-        let result = languageModel(nil, cache: cache, inputEmbedding: embeddings)
+        let prefillStepSize = windowSize ?? 512
+        let totalPositions = embeddings.dim(1)
+        var processed = 0
+        while totalPositions - processed > 1 {
+            let chunkLength = min(prefillStepSize, totalPositions - processed - 1)
+            let range = processed ..< (processed + chunkLength)
+            _ = languageModel(nil, cache: cache, inputEmbedding: embeddings[0..., range, 0...])
+            asyncEval(cache)
+            processed += chunkLength
+        }
+        eval(cache)
+        let result = languageModel(nil, cache: cache, inputEmbedding: embeddings[0..., processed..., 0...])
         return .logits(result)
     }
 
--- a/Libraries/MLXVLM/Models/FastVLM.swift
+++ b/Libraries/MLXVLM/Models/FastVLM.swift
@@ -1146,7 +1146,8 @@ public class FastVLM: Module, VLMModel, KVCacheDimensionProvider {
             processed += chunkLength
         }
         eval(cache)
-        let result = languageModel(nil, cache: cache, inputEmbedding: embeddings[0..., processed..., 0...])
+        let result = languageModel(
+            nil, cache: cache, inputEmbedding: embeddings[0..., processed..., 0...])
         return .logits(result)
     }
 
--- a/Libraries/MLXVLM/Models/Gemma3.swift
+++ b/Libraries/MLXVLM/Models/Gemma3.swift
@@ -985,11 +985,25 @@ public class Gemma3: Module, VLMModel, KVCacheDimensionProvider {
     public func prepare(_ input: LMInput, cache: [any KVCache], windowSize: Int?) throws
         -> PrepareResult
     {
+        let prefillStepSize = windowSize ?? 512
+        let convertedCache = cache.compactMap { $0 as KVCache }
+
         guard let imagePixels = input.image?.pixels else {
-            // Text-only input
-            let convertedCache = cache.compactMap { $0 as KVCache }
+            var tokens = input.text.tokens
+            if tokens.ndim == 1 { tokens = tokens.expandedDimensions(axis: 0) }
+            let totalPositions = tokens.dim(1)
+            var processed = 0
+            while totalPositions - processed > 1 {
+                let chunkLength = min(prefillStepSize, totalPositions - processed - 1)
+                _ = languageModel(
+                    tokens[0..., processed ..< (processed + chunkLength)],
+                    cache: convertedCache, inputEmbedding: nil, mask: nil)
+                asyncEval(cache)
+                processed += chunkLength
+            }
+            eval(cache)
             let result = languageModel(
-                input.text.tokens, cache: convertedCache, inputEmbedding: nil, mask: nil)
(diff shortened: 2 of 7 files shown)


davidkoski:
Rebase to fix merge conflict

davidkoski (approved):
Change looks good, thank you!

davidkoski:
That CI failure looks like the one @aleroot fixed in #323 which is similar to what I fixed in #326 (though 326 really only applies on M5 and the CI runners are M4).

davidkoski:
I ran that test in a loop and I can reproduce the failure.  Perhaps something related to random state?  I think the fix to serialize it maybe made it happen less -- not sure.  The concurrent execution should have been ok, but it might have made it happen faster (more random samples?)

I will try to investigate this.  I don't think it is related to this PR, I think it just got unlucky and hit it.

davidkoski:
The CI failure happens on main, just not reliably.  Filed #353 for further investigation.  I think we can take this now and deal with that as an independent issue.

PR #345, opened 2026-06-13
Qwen2-VL: implement M-RoPE in the language model (mirror of #239)

mnmly:
PR #239 fixed exactly this for `Qwen2.5-VL` but left `Qwen2-VL` untouched. This is pretty much a straight port of that fixes into `Qwen2VL.swift`; the LM spatial-position scheme is identical between the two models. 

I've added explicit inline-comments referring which parts of Qwen2.5-VL M-RoPE adaptations are ported here, but I will clean it all once this PR is reviewed and ready for merge.


davidkoski (approved):
Change looks good.  I am not sure the verbose comments are needed post-PR, but we can clean them out if needed later.  Thank you!

PR #348, opened 2026-06-13
Add TranslateGemma support (Gemma 3 translation template)

beshkenadze:
## What

Adds first-class support for **TranslateGemma** — Google's translation-tuned fine-tunes of Gemma 3 (4B/12B/27B). They load through the existing `gemma3` text path; this PR adds the translation chat-template handling and the registry presets.

## How (revised after review)

The first cut hung the translation template on `Gemma3TextModel`, which applied it to every gemma3 text checkpoint. It is now opt-in per model configuration, as suggested in review:

- `ModelConfiguration.messageGenerator: (any MessageGenerator)?` overrides the generator the model would otherwise supply; `nil` keeps the model's own default, so nothing changes for existing configurations.
- Honored in `_load()` of **both** `LLMModelFactory` and `VLMModelFactory`. VLM processors bake message generation into their own `prepare(input:)`, so there the override is applied by wrapping the model's processor in `MessageGeneratorUserInputProcessor` rather than by touching every processor.
- The property travels through `ResolvedModelConfiguration`, which is what the factories actually read.
- `Gemma3TextModel` no longer overrides `messageGenerator(tokenizer:)` — plain Gemma 3 is untouched.
- `Gemma3MessageGenerator` → `TranslateGemma3MessageGenerator`, set explicitly on the six `translategemma_*` presets.

TranslateGemma's chat template requires each user turn tagged with `source_lang_code` / `target_lang_code` (ISO 639-1), read from `UserInput.additionalContext`:

```swift
UserInput(
    chat: [.user("Hello, how are you?")],
    additionalContext: ["source_lang_code": "en", "target_lang_code": "fr"])
```

`ChatSession` takes the same `additionalContext` and forwards it into the `UserInput` it builds, so the translation path is reachable from `ChatSession` as well. Without the codes the generator is byte-identical to `DefaultMessageGenerator`.

Two notes for the reviewer:

- `ModelConfiguration.==` becomes hand-written, because the existential is not `Equatable`. All conformers are stateless, so concrete-type identity is the meaningful comparison; a comment above `==` flags that new stored properties must be added by hand.
- `MessageGeneratorUserInputProcessor` is currently `public` in `MLXLMCommon`, but its only in-repo consumer is `VLMModelFactory`. Happy to narrow it to `package` if you'd rather not commit to it as public API.

## Changes

- `Libraries/MLXLMCommon/ModelConfiguration.swift`: `messageGenerator` property, defaulted parameters on both initializers, hand-written `==`.
- `Libraries/MLXLMCommon/Downloader.swift`: `messageGenerator` on `ResolvedModelConfiguration`.
- `Libraries/MLXLMCommon/UserInput.swift`: `MessageGeneratorUserInputProcessor`.
- `Libraries/MLXLLM/LLMModelFactory.swift`: honor the override in `_load()`; `translategemma_{4b,12b,27b}_it_{4bit,8bit}` presets with the generator and `extraEOSTokens: ["<end_of_turn>"]`.
- `Libraries/MLXVLM/VLMModelFactory.swift`: honor the override in `_load()`.
- `Libraries/MLXLLM/Models/Gemma3Text.swift`: drop the model-level override; rename the generator.
- `Tests/MLXLMTests/TranslateGemmaTests.swift`: tests below.

## Validation

`swift build` clean; `swift test --filter MLXLMTests` → 405 XCTest cases and 581 swift-testing tests, 0 failures. `TranslateGemmaTests` is 10 tests:

- config decode from nested `text_config`, `sanitize` (`language_model.` prefix + tied `lm_head`), preset ids;
- generator behavior with and without language codes;
- `translategemma_*` presets carry `TranslateGemma3MessageGenerator`, `gemma3_*` presets carry none;
- `Gemma3TextModel` still returns `DefaultMessageGenerator` — the regression guard for the first review comment, verified red/green by temporarily restoring the override;
- the override survives `resolved(modelDirectory:tokenizerDirectory:)`;
- `MessageGeneratorUserInputProcessor` hands the wrapped processor the generated messages.

Also verified end-to-end on `mlx-community/translategemma-4b-it-4bit`: greedy `en→fr` matches a full-precision `transformers` reference ("Hello, how are you?" → "Bonjour, comment allez-vous ?").

## Benchmark

FLORES+ `devtest` quick set (64 sentences per direction), greedy, `mlx-community/translategemma-4b-it-4bit` on an M1 Max, using the translation template this PR adds. 16 directions across 9 European languages plus Russian: chrF++ 45.4–63.9, BLEU 17.2–38.1, 16.2–23.0 tok/s, 0 errors. Per-direction table with links to each scored run is in [this comment](https://github.com/ml-explore/mlx-swift-lm/pull/348#issuecomment-5278819375).

> These runs apply the translation template (the feature). omni-bench's registered FLORES+ Task is raw-prompt only (hosts may not add a chat template), so the runs are tagged with a template-specific implementation id and are not comparable to raw-prompt FLORES+ numbers.


Libraries/MLXLLM/Models/Gemma3Text.swift
```diff
@@ -459,10 +459,70 @@ public class Gemma3TextModel: Module, LLMModel {
         return .tokens(y)
     }
 
+    public func messageGenerator(tokenizer: Tokenizer) -> MessageGenerator {
+        Gemma3MessageGenerator()
```
davidkoski, line 463: This will override the message generator for _all_ gemma3 (text) models, not just the translate gemma models.

Libraries/MLXLLM/Models/Gemma3Text.swift
```diff
@@ -459,10 +459,70 @@ public class Gemma3TextModel: Module, LLMModel {
...
     /// Prefill chunk size when the caller sets none, tuned for this path on Apple Silicon.
     private static let defaultPrefillChunkSize = 128
 }
 
+/// Message generator for the Gemma 3 text path.
+///
+/// TranslateGemma's chat template needs each user turn tagged with `source_lang_code` /
+/// `target_lang_code` (ISO 639-1); supply them via `UserInput.additionalContext`. Without
+/// those codes this behaves exactly like `DefaultMessageGenerator`.
+public struct Gemma3MessageGenerator: MessageGenerator {
```
davidkoski, line 475: Perhaps this should have a more specific name -- I don't think we want this for all gemma3 models as it is too restrictive.  Perhaps `TranslateGemma3MessageGenerator`?

Libraries/MLXLLM/LLMModelFactory.swift
```diff
@@ -212,6 +212,42 @@ public class LLMRegistry: AbstractModelRegistry, @unchecked Sendable {
         extraEOSTokens: ["<turn|>"]
     )
 
+    static public let translategemma_4b_it_4bit = ModelConfiguration(
+        id: "mlx-community/translategemma-4b-it-4bit",
+        defaultPrompt: "Hello, how are you?",
+        extraEOSTokens: ["<end_of_turn>"]
+    )
```
davidkoski, line 219: Given my previous comments, I think we need a way for particular models to override the MessageGenerator -- not use the default for the model.  What if we add a `messageGenerator: MessageGenerator?` property to `ModelConfiguration` and then honor that in the `_load()` methods (`LLMModelFactory` and `VLMModelFactory`)

davidkoski (changes requested):
Looks good but see my comments -- I think this will apply too widely as-is.

abra-code:
Additional info from integrator perspective in case it helps the design question here.

I run TranslateGemma in [mlx-agent](https://github.com/abra-code/mlx-agent) which uses mlx-swift-lm today without TranslateGemma-specific code by passing the structured content part through as data: messages carrying `{"type": "text", "source_lang_code": ..., "target_lang_code": ..., "text": ...}`, handed to `tokenizer.applyChatTemplate(messages:)`.

That path can't start from `ChatSession`, since `Chat.Message.content` is a `String` which, as I read it, is waht `Gemma3MessageGenerator` change is proposing.  
This is a design choice: a per-model, renamed generator, which would keep it off every other Gemma 3, or letting `Chat.Message` carry structured content parts, which would make a TranslateGemma-specific generator unnecessary since the template would know the format.

Separately, on `extraEOSTokens`: the presets are fixing a real bug becuase `mlx-community/translategemma-27b-it-4bit` comes with `generation_config.json` with no `eos_token_id` at all, so nothing stops on `<end_of_turn>`.

My implementation in mlx-agent, if it's useful as a reference:
https://github.com/abra-code/mlx-agent/commit/35bc0da694773fdcb44fe2bf360325e06057d601

mlx-agent tool is employed in just released [Interpreter.app](https://github.com/abra-code/InterpreterApp/releases), a specialized macOS translator app built for TranslateGemma and other models. It chunks the document on paragraph (or sentence) boundaries, translates chunk by chunk, and stitches the result back.

beshkenadze:
## TranslateGemma multilingual benchmark (FLORES+ quick)

Model: `mlx-community/translategemma-4b-it-4bit` (4bit, artifact `sha256:113acb0c2999…`)
Backend: `mlx_swift_lm` @ `upstream-translategemma` · adapter `MLXLMGenerator`
Hardware: Apple M1 Max, 32 GB · macOS 26.2.0
Profile: `text_generation.batch_single.v1` · delivery `batch`, concurrency 1, warmup 0
Decoding: temperature 0, top_p 1, seed 0, max_output_tokens 128
Scope: 16 directions × 64 samples, 0 errors · environment label `pr-348-multilingual-20260812`

| Direction | chrF++ | BLEU | tok/s | p50 s | p99 s | peak RSS GB | Result |
|---|---:|---:|---:|---:|---:|---:|---|
| de-en | 61.49 | 34.65 | 16.8 | 1.57 | 2.88 | 2.58 | [view](https://bench.bshk.app/result/sha256%3Aa5c571b4390789febac141d651384de36322e7f9cc4720caf84c90942b2432b9) |
| en-de | 57.99 | 30.44 | 17.6 | 2.00 | 4.04 | 1.64 | [view](https://bench.bshk.app/result/sha256%3A05bac09d7d304a9ca5d902c883964325118cc4a6424a609acf589c992569ae14) |
| en-es | 53.24 | 27.34 | 21.6 | 1.51 | 2.96 | 2.61 | [view](https://bench.bshk.app/result/sha256%3Aecca639dc75142b338bc7ecc270ebeb19fdd5d82932e4a3b79f3d7a5cbb4c382) |
| en-fr | 62.16 | 36.01 | 21.4 | 1.67 | 3.32 | 2.61 | [view](https://bench.bshk.app/result/sha256%3A43245f617f32b795e56a070a558357aa4a790997e3532690f5b844044e9e1ba3) |
| en-it | 53.91 | 26.61 | 20.8 | 1.65 | 4.51 | 2.61 | [view](https://bench.bshk.app/result/sha256%3A56739c9b8280ffd75f07e6720279b4cc1d94b1a895a87a4b9c11633e719f9746) |
| en-nl | 51.54 | 21.74 | 18.0 | 2.02 | 4.47 | 2.61 | [view](https://bench.bshk.app/result/sha256%3A0081b6b72bca9973dae9fac7ff8432911d296b3ec2cf3734156013a5091b3df6) |
| en-pl | 45.43 | 17.21 | 23.0 | 1.70 | 3.15 | 2.61 | [view](https://bench.bshk.app/result/sha256%3A124ebc494fdea622c10e35b010097448964a2bed409be02593698c0f6fa57fec) |
| en-pt | 61.69 | 37.69 | 20.1 | 1.69 | 3.05 | 2.61 | [view](https://bench.bshk.app/result/sha256%3Aae7f76b993d80fd782fe93b5a71998b56551b9feadda59488d90fdf917285eb8) |
| en-ru | 50.00 | 22.91 | 19.4 | 1.81 | 3.70 | 2.61 | [view](https://bench.bshk.app/result/sha256%3Aed33140d514579ba8396d1d357a7e945743b97ad30a265cdf0b977405cba7cc0) |
| es-en | 54.33 | 23.72 | 19.9 | 1.41 | 2.17 | 2.61 | [view](https://bench.bshk.app/result/sha256%3A4501763f5720331346f387838e8126adeafd1092d1c60211aa3f27f1c9f8a5a1) |
| fr-en | 61.72 | 34.50 | 16.5 | 1.59 | 3.99 | 2.48 | [view](https://bench.bshk.app/result/sha256%3A5cda70b292184e699161b035450aea0574f94f37c6323ec42999d68c54165d57) |
| it-en | 56.73 | 26.96 | 16.2 | 1.66 | 3.88 | 2.61 | [view](https://bench.bshk.app/result/sha256%3Af46598f1543d182c2cfbdb7bc1ce7d9abe423bb5737e749caa3cc1371cc23984) |
| nl-en | 56.26 | 28.35 | 16.4 | 1.58 | 3.40 | 2.51 | [view](https://bench.bshk.app/result/sha256%3A3baf0afa45d3e4287283ebcff3aefb4aaec612264a973642c9682996c53c4c4f) |
| pl-en | 52.39 | 22.96 | 19.1 | 1.46 | 2.21 | 2.61 | [view](https://bench.bshk.app/result/sha256%3Aaab21e6eb0248d6c9a680963dd6c5b1f03e02203091bd360417c2e8515e6d91f) |
| pt-en | 63.87 | 38.13 | 17.4 | 1.53 | 2.68 | 2.61 | [view](https://bench.bshk.app/result/sha256%3A264421d9e73a6816580e350f212e74be577b051b3985811eac1e404c851106bb) |
| ru-en | 56.05 | 28.62 | 18.8 | 1.46 | 2.47 | 0.92 | [view](https://bench.bshk.app/result/sha256%3A9e7b8498f31f48d6f8fbd0524f2218acbdc0acd8e80ddc7597bc785ea6cef476) |

Ranges: chrF++ 45.43–63.87 · BLEU 17.21–38.13 · throughput 16.2–23.0 tok/s. Each row is one Task `textgen.mt.flores_plus.<direction>.quick.v1`; every Result is scored and schema-validated by omni-bench.

Libraries/MLXLLM/Models/Gemma3Text.swift
beshkenadze: Removed — `Gemma3TextModel` no longer overrides `messageGenerator(tokenizer:)`, so plain Gemma 3 keeps `DefaultMessageGenerator`. The translation template is now opt-in per configuration (see the reply on the `LLMModelFactory` comment). `testGemma3TextModelKeepsDefaultMessageGenerator` locks this in; I verified it red/green by temporarily restoring the override.

Libraries/MLXLLM/Models/Gemma3Text.swift
beshkenadze: Renamed to `TranslateGemma3MessageGenerator`, and the doc comment now says it is for TranslateGemma rather than "the Gemma 3 text path". Only the six `translategemma_*` presets reference it.

davidkoski (approved):
Changes look good to me, thank you for the followup!

PR #349, opened 2026-06-13
Expose ModelTypeRegistry.contains(_:) for pre-load support checks

jfranknichols:
`ModelTypeRegistry` can instantiate a model type via `createModel`, but there is
no way to ask whether a creator is registered for a given `model_type` without
attempting that throwing, allocating call. The `creators` dictionary is private,
so the check has to live on the type.

This adds a cheap `contains(_:)` so a caller can, for example, decide before a
multi-GB download whether a Hub repo's `model_type` is runnable by the factory.

### DIFF
--- a/Libraries/MLXLMCommon/Registries/ModelTypeRegistry.swift
+++ b/Libraries/MLXLMCommon/Registries/ModelTypeRegistry.swift
@@ -31,4 +31,12 @@ public actor ModelTypeRegistry<T> {
         return try creator(configuration)
     }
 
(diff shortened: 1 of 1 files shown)


davidkoski (approved):
Looks good, thank you!

davidkoski:
Just waiting for CI

jfranknichols:
Thanks David, 

I appreciate the quick turn around!

Frank

davidkoski:
If they were all this straightforward it would help a lot :-)

PR #350, opened 2026-06-13
Gemma4Text: load conversions that omit redundant KV-shared-layer K/V projections

jfranknichols:
## Problem

Gemma 4's last `num_kv_shared_layers` layers reuse an earlier layer's keys and
values — they have no `k_proj`/`v_proj`/`k_norm` of their own. `Gemma4Attention`
built those modules unconditionally for every layer, so loading a conversion
that *omits* those redundant tensors fails with:

```
keyNotFound(path: ["...", "layers", "24", "self_attn", "k_norm", "weight"], ...)
```

Stock mlx-community conversions ship the redundant weights (loaded but never used
on shared layers), so this only surfaces with space-optimized conversions that
drop them — e.g. `mlx-optiq` mixed-precision quants.

## Fix

- Make `kProj`/`kNorm` optional and skip building K/V/k_norm for KV-shared
  layers. The backbone already feeds those layers an earlier layer's KV, so the
  per-layer recompute path is unreachable for them (guarded with a clear error).
- Drop the redundant `k_proj`/`v_proj`/`k_norm` tensors for shared layers in
  `sanitize`, so stock conversions that still ship them load past
  `verify(.noUnusedKeys)`.

Net effect: conversions that omit the weights now load, and standard conversions
are unchanged in behavior (those shared-layer projections were dead weight)
while using slightly less memory.

## Verification

Loaded both through `LLMModelFactory`'s `Gemma4Model`:
- `mlx-community/gemma-4-e4b-it-OptiQ-4bit` (omits the weights) — now loads.
- `mlx-community/gemma-4-e4b-it-8bit` (ships them) — still loads.

## Note

`MLXVLM/Models/Gemma4.swift` has the same KV-shared gap. It already carries a
`kvSharedOnly` constructor flag for this, but the VLM backbone doesn't wire it
on, so it hits the same `keyNotFound`. This PR fixes the `MLXLLM` text path; the
VLM could get the matching treatment (wire `kvSharedOnly` from the backbone plus
the same sanitize drop) as a follow-up.

### DIFF
--- a/Libraries/MLXLLM/Models/Gemma4Text.swift
+++ b/Libraries/MLXLLM/Models/Gemma4Text.swift
@@ -204,14 +204,15 @@ private class Gemma4Attention: Module {
     let nKvHeads: Int
     let useKeqV: Bool
     let scale: Float
+    let isKVSharedLayer: Bool
 
     @ModuleInfo(key: "q_proj") var qProj: Linear
-    @ModuleInfo(key: "k_proj") var kProj: Linear
+    @ModuleInfo(key: "k_proj") var kProj: Linear?
     @ModuleInfo(key: "v_proj") var vProj: Linear?
     @ModuleInfo(key: "o_proj") var oProj: Linear
 
     @ModuleInfo(key: "q_norm") var qNorm: RMSNorm
-    @ModuleInfo(key: "k_norm") var kNorm: RMSNorm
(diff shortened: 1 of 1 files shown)


davidkoski:
I think this is a dup of #342 (already merged).  Please reopen (or @ me if you can't) if you disagree.  Thank you!

PR #351, opened 2026-06-15
Add Qwen3.5 MTP speculative decoding

aleroot:
## Proposed changes

Adds Qwen3.5/Qwen3.6 MTP speculative decoding support for text and VLM models.

## Checklist


Libraries/MLXLMCommon/MTPSpeculativeTokenIterator.swift
```diff
@@ -332,21 +335,35 @@ public struct MTPSpeculativeTokenIterator: TokenIteratorProtocol {
...
+            // Hybrid models such as Qwen3.5/Qwen3.6 carry Mamba-style caches
+            // that cannot be rewound after a rejected speculative tail. Verify
+            // on a copied cache, then commit only the real prefix into the
+            // canonical cache: bonus token plus the accepted draft tokens. The
+            // finalToken is the next token to feed, so it is intentionally not
+            // replayed here.
+            var commitState = mainState ?? state
+            commitState[mtpEmitFlagKey] = true
+            let committedTokens = verifyTokens[0 ..< (accepted + 1)]
+            let committed = mainModel(
```
davidkoski: Is there an optimized path if we are accepting all the tokens?  Does that happen in real life?

Libraries/MLXLMCommon/QwenMTP.swift
```diff
@@ -0,0 +1,109 @@
...
+    forward: (
+        _ inputsEmbeds: MLXArray, _ hiddenStates: MLXArray, _ cache: KVCache?,
+        _ stepIndex: Int, _ positionOffset: Int
+    ) -> MLXArray
+) -> MLXArray {
+    precondition(blockSize >= 2, "blockSize must be >= 2")
+
+    var tok = lastToken.ndim == 1 ? lastToken.reshaped([lastToken.dim(0), 1]) : lastToken
+    var hidden = lastHidden
+    let cache = newCache()
```
davidkoski: from mlx-vm:

```python
class Qwen3_5MTPDraftModel(nn.Module):
...
    def accept_verified_tokens(
        self,
        verify_hidden: mx.array,
        draft_tokens: mx.array,
        accepted: int,
        new_tokens: List[int],
        sampler,
        token_dtype: mx.Dtype = mx.int32,
        greedy: bool = False,
    ) -> None:
        keep_appended = min(int(accepted), self._round_appended)
        trim = self._round_appended - keep_appended
        if trim > 0:
            for cache in self._cache:
                cache.trim(trim)
```

Keeping the cache as state in the model won't work for mlx-swift-lm (and I wonder about that in mlx-vlm for server cases, but the mlx-vlm MTP models are generally stateful).

`gemm4_assistant.py` notes that it has no cache of its own, and that matches the swift code.

Anyway, I wonder what impact creating the cache has here.  Does this mean the draft model misses out on all the context?  Do we need an `LMOuput.State` here for the cache?

Libraries/MLXLMCommon/MTPSpeculativeTokenIterator.swift
aleroot: Yes, good point. Fixed.

aleroot: This can happen in real use, especially with greedy/low-temperature sampling, small block sizes, repetitive/structured continuations, or a strong MTP drafter.

Libraries/MLXLMCommon/QwenMTP.swift
aleroot: Good catch. Creating a fresh Qwen MTP cache per draft round would lose the predictor’s own accepted-prefix context. I changed this so Qwen MTP uses iterator-owned per-stream drafter state.

Libraries/MLXLMCommon/MTPSpeculativeTokenIterator.swift
```diff
@@ -393,21 +431,55 @@ public struct MTPSpeculativeTokenIterator: TokenIteratorProtocol {
...
+                // the rejected count, in lockstep. The verify pass's state
+                // emission spans the full verify chunk — stale tail rows must
+                // not survive into the next round's draftBlock.
+                let trimmed = mainCacheStorage.trim(rejected)
+                trimSharedKVState(&mainState, numTokens: trimmed)
+            }
+        } else {
+            if rejected == 0 {
+                if !nativeHybridRewind {
+                    mainCache = verifyCache
```
davidkoski: I think this requires `KVCacheStorage` to be used in order for the caller to observe the changes in the cache, however that is still `package` scoped and `ChatSession` doesn't use this iterator.

aleroot: You are completely right, I am still working on this set of changes, check back tomorrow and it should be in a better shape. I should have probably put it on the draft mode

aleroot: This should be solved now, thanks.

davidkoski (approved):
Changes look good, thank you!

PR #352, opened 2026-06-15
Base Gemma diffusion implementation

aleroot:
## Proposed changes

Basic Gemma diffusion model implementation.

## Checklist

### DIFF
--- a/Libraries/MLXLMCommon/BlockDiffusionTokenIterator.swift
+++ b/Libraries/MLXLMCommon/BlockDiffusionTokenIterator.swift
@@ -244,7 +244,7 @@ public struct BlockDiffusionTokenIterator: TokenIteratorProtocol {
         var draftCanvas = currentCanvas
         argmaxCanvasHistory = nil
 
-        for curStep in stride(from: maxDenoisingSteps, through: 1, by: -1) {
+        denoisingLoop: for curStep in stride(from: maxDenoisingSteps, through: 1, by: -1) {
             let rawLogits =
                 if prefersLogitsSelfConditioning {
                     model.diffusionLogits(
@@ -293,7 +293,7 @@ public struct BlockDiffusionTokenIterator: TokenIteratorProtocol {
 
                 if draftRevealMask.all().item(Bool.self) {
                     argmaxCanvas = draftCanvas
-                    break
+                    break denoisingLoop
                 }
             }
 
--- a/Libraries/MLXLMCommon/BlockDiffusionTokenIterator.swift
+++ b/Libraries/MLXLMCommon/BlockDiffusionTokenIterator.swift
@@ -1,0 +1,384 @@
+// Copyright © 2026 Apple Inc.
+
+import Foundation
+import MLX
+import MLXNN
+
+/// Generator for block-diffusion language models.
+///
+/// The model first encodes the prompt into a KV cache, then repeatedly denoises
+/// a full canvas. The iterator exposes the finalized canvas tokens one by one
+/// so existing detokenization, EOS handling, tool parsing, and timing code can
+/// remain shared with autoregressive iterators.
+public struct BlockDiffusionTokenIterator: TokenIteratorProtocol {
+    let model: any BlockDiffusionLanguageModel
+    let cacheStorage: KVCacheStorage
+    var cache: [KVCache] {
+        get { cacheStorage.cache }
+        set { cacheStorage.replace(with: newValue) }
+    }
+    var kvCachePlan: KVCachePlan { cacheStorage.plan }
+    let prefillStepSize: Int?
+    let minCanvasLength: Int
+    let maxCanvasLength: Int
+    let maxDenoisingSteps: Int
+    let entropyBound: Float
+    let temperatureMin: Float
+    let temperatureMax: Float
+    package let denoiserTemperature: Float
+    let diffusionSampler: BlockDiffusionParameters.Sampler
+    let prefersLogitsSelfConditioning: Bool
+    let selfConditioningWeight: MLXArray?
+    let stabilityThreshold: Int
+    let confidenceThreshold: Float
+
+    public var tokenCount = 0
+    public let maxTokens: Int?
+    public var promptPrefillTime: TimeInterval = 0.0
+
+    private var pendingTokens = [Int]()
+    private var pendingIndex = 0
+    private var randomState: MLXRandom.RandomState
+    private var committedPendingIndex = 0
+    private var argmaxCanvasHistory: [MLXArray]?
+
+    public init(
+        input: LMInput,
+        model: any BlockDiffusionLanguageModel,
+        cache: [KVCache]? = nil,
+        parameters: GenerateParameters
+    ) throws {
+        let plan = try parameters.kvCachePlan()
+        try self.init(
+            input: input, model: model,
+            cacheStorage: KVCacheStorage(
+                cache ?? (try model.newCache(parameters: parameters)), plan: plan),
+            parameters: parameters)
+    }
+
+    package init(
+        input: LMInput,
+        model: any BlockDiffusionLanguageModel,
+        cacheStorage: KVCacheStorage,
+        parameters: GenerateParameters
+    ) throws {
+        let promptTokenCount = try Self.validatePrompt(input.text, model: model)
+        let kvCachePlan = cacheStorage.plan
+        let cacheStorage = try kvCachePlan.validated(cacheStorage)
+
+        self.model = model
+        self.cacheStorage = cacheStorage
+        self.prefillStepSize = parameters.prefill.stepSize
+        self.maxTokens = parameters.maxTokens ?? model.diffusionDefaultMaxTokens
+        let resolvedCanvasBounds: (minimum: Int, maximum: Int) =
+            switch parameters.diffusion.canvas {
+            case .full:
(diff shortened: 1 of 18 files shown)


Libraries/MLXLMCommon/Evaluate.swift
```diff
@@ -707,6 +767,314 @@ public struct TokenIterator: TokenIteratorProtocol {
     }
 }
 
+/// Generator for block-diffusion language models.
+///
+/// The model first encodes the prompt into a KV cache, then repeatedly denoises
+/// a full canvas. The iterator exposes the finalized canvas tokens one by one
+/// so existing detokenization, EOS handling, tool parsing, and timing code can
+/// remain shared with autoregressive iterators.
+public struct BlockDiffusionTokenIterator: TokenIteratorProtocol {
```
davidkoski: This looks great, but i wonder if this should be in its own file?  Evaluate.swift is getting unwieldy.

Libraries/MLXLMCommon/Evaluate.swift
```diff
@@ -778,6 +1146,15 @@ public struct SpeculativeTokenIterator: TokenIteratorProtocol {
...
         numDraftTokens: Int
     ) throws {
+        guard !(mainModel is any BlockDiffusionLanguageModel) else {
+            throw GenerateError.unsupportedSpeculativeDecoding(
+                String(describing: type(of: mainModel)))
+        }
+        guard !(draftModel is any BlockDiffusionLanguageModel) else {
+            throw GenerateError.unsupportedSpeculativeDecoding(
+                String(describing: type(of: draftModel)))
+        }
```
davidkoski: I think this is fine for now but we might want some kind of capability property down the road.  maybe long these lines:

https://developer.apple.com/documentation/foundationmodels/languagemodelcapabilities/capability/vision

then we can test directly vs using `is` / `as`

Libraries/MLXLMCommon/KVCache.swift
```diff
@@ -669,7 +669,7 @@ public class RotatingKVCache: BaseKVCache, CustomDebugStringConvertible {
                     values[.ellipsis, ..<offset, 0...],
                 ]
             } else {
-                return [keys, values]
+                return [temporalOrder(keys), temporalOrder(values)]
```
davidkoski: I think this is desired for this diffusion model, but what about other users of RotatingKVCache?  this will be a semantic change for them.  Also this interacts with serialization.

Maybe it is fine -- this is used in one path for updates, but i think it needs closer examination.

Libraries/MLXLMCommon/Evaluate.swift
```diff
@@ -707,6 +767,314 @@ public struct TokenIterator: TokenIteratorProtocol {
...
+    private mutating func processLogits(_ logits: MLXArray, curStep: Int) -> MLXArray {
+        let temperature = MLXArray(temperature(curStep: curStep))
+        guard processor != nil else {
+            return logits / temperature
+        }
+
+        let length = logits.dim(1)
+        let processed = (0 ..< length).map { index in
+            var tokenLogits = logits[0..., index, 0...]
+            tokenLogits = processor?.process(logits: tokenLogits) ?? tokenLogits
```
davidkoski: It looks like the mlx-vlm side doesn't even support logit processors:

```python
                schedule_temperature = _diffusion_linear_temperature(
                    cur_step,
                    max_denoising_steps,
                    temperature_config,
                )
                if schedule_temperature is not None:
                    processed_logits = processed_logits / schedule_temperature

                argmax_canvas = mx.argmax(processed_logits, axis=-1).astype(
                    input_ids.dtype
                )
                if cur_step == 1 and not diffusion_show_unmasking:
                    break

                denoiser_canvas = (
                    argmax_canvas
                    if temperature <= 0
                    else _diffusion_sample_canvas(
                        processed_logits,
                        input_ids.dtype,
                        temperature,
                    )
                )
```

I am not sure this is safe to do -- the logit processors can be stateful (e.g. repetition penalties) and applying them to the entire canvas would (I think) leak state into the processor even if we don't end up using the token.

Check my math on that, but if it is correct do we:

- it is fine
- need a notion of a stateless logit processor that could be applied?
- be able to back out logits that do not make it?
- apply that processor after we accept the logits?

I am not sure I fully wrap my head around the diffusion path so I might be off base here.

aleroot: This is an important point. For the moment I will remove it, but in the future there should probably be an explicit new API:

- StatelessLogitProcessor: safe for pure masks/biases that do not depend on generated history.
- DiffusionLogitProcessor: receives canvas position, reveal mask, accepted mask/history, denoising step, etc.

Apply penalties only after token acceptance: safe for state, but too late to affect selection, so mostly useless for logits processing.

Libraries/MLXLMCommon/KVCache.swift
aleroot: Good point. I changed this so `RotatingKVCache.state` keeps the existing ring-buffer storage order.

Libraries/MLXLMCommon/BlockDiffusionTokenIterator.swift
```diff
@@ -0,0 +1,384 @@
...
+                currentCanvas = MLX.where(
+                    draftRevealMask | acceptedMask,
+                    acceptedCanvas,
+                    makeInitialCanvas(length: length))
+                draftRevealMask = draftRevealMask | acceptedMask
+                draftCanvas = acceptedCanvas
+
+                if draftRevealMask.all().item(Bool.self) {
+                    argmaxCanvas = draftCanvas
+                    break
```
davidkoski: I wonder if this break is meant to break out of the loop?  I think this is the matching python code:

```python
                if diffusion_sampler == "confidence-threshold" and bool(
                    mx.all(draft_reveal_mask).item()
                ):
                    accepted_canvas = draft_canvas
                    break
```

I think since this swift code is in a switch it only breaks the switch itself, not an exit from the loop.

Libraries/MLXVLM/Models/DiffusionGemma.swift
```diff
@@ -0,0 +1,701 @@
...
+        DiffusionGemmaConfiguration(
+            textConfig: textConfig,
+            generationConfig: generationConfig,
+            canvasLength: canvasLength,
+            tieWordEmbeddings: tieWordEmbeddings,
+            modelType: modelType)
+    }
+}
+
+public final class DiffusionGemma: Module, VLMModel, BlockDiffusionLanguageModel,
```
davidkoski, line 88: Does this support tools?  It doesn't define a `toolCallFormat`

Libraries/MLXLMCommon/Models/DiffusionGemma.swift
```diff
@@ -0,0 +1,1167 @@
...
+        selfConditioningEmbeddings: MLXArray? = nil
+    ) -> MLXArray {
+        var h = embeddings(canvasTokens)
+
+        let signal: MLXArray
+        if let selfConditioningEmbeddings {
+            signal = selfConditioningEmbeddings.asType(h.dtype)
+        } else if let selfConditioningLogits {
+            let probabilities = softmax(
+                selfConditioningLogits.asType(h.dtype), axis: -1, precise: true)
```
davidkoski: Spotted by claude in comparison with mlx-vlm.  The corresponding python code is:

```python
    def _embed_canvas(
        self,
        canvas_ids,
        self_conditioning_logits=None,
        self_conditioning_embeddings=None,
    ):
        inputs_embeds = self.embed_tokens(canvas_ids) * self.embed_scale
..
        else:
            probs = mx.softmax(self_conditioning_logits, axis=-1, precise=True)
```

notice the lack of type conversion.  The is another python function that does match this:

```python
    def diffusion_self_conditioning(
        self,
        processed_logits: mx.array,
        embedding_weight: Optional[mx.array],
    ) -> mx.array:
        if self.prefers_logits_self_conditioning:
            return processed_logits
        # Match the HF generation path: self-conditioning logits are stored in
        # the embedding dtype before the next decoder softmax.
        probs = mx.softmax(
            processed_logits.astype(embedding_weight.dtype),
            axis=-1,
            precise=True,
        )
```

but I think that matches `selfConditioningEmbeddings()` in this file.

Libraries/MLXLMCommon/Models/DiffusionGemma.swift
```diff
@@ -0,0 +1,1167 @@
...
+    }
+
+    let state =
+        if let rotatingCache = cache as? RotatingKVCache {
+            rotatingCache.temporalState
+        } else {
+            cache.state
+        }
+
+    guard state.count == 2 else {
```
davidkoski: Can this model be configured with a QuantizedKVCache?  That can return state with more than 2 elements.

The state is really meant for doing `eval`, I wonder if we need API that exposes this properly?  The ordering of the values in state is not specified.

Libraries/MLXLMCommon/Evaluate.swift
```diff
@@ -1057,6 +1138,15 @@ public struct SpeculativeTokenIterator: TokenIteratorProtocol {
         numDraftTokens: Int,
         components: GenerationComponents = .init()
     ) throws {
+        guard !mainModel.capabilities.contains(.blockDiffusion) else {
```
davidkoski, line 1141: I think MTPSpeculativeTokenIterator may need the same checks.

PR #358, opened 2026-06-19
Fix KV cache quantization: cache updates lost due to value-type propagation

somegeekintn:
**Background**

`maybeQuantizeKVCache` takes `cache: inout [KVCache]` and replaces array elements with new `QuantizedKVCache` instances when the quantization threshold is crossed. Because Swift arrays are value types, this replacement was only visible inside `TokenIterator`'s local copy — the caller's array continued holding the original `KVCacheSimple` references. The quantized caches were deallocated at the end of generation, silently discarding all context generated after the quantization threshold. See [issue #312](https://github.com/ml-explore/mlx-swift-lm/issues/312) for additional details and reproduction steps.

**What this PR does**

Introduces `KVCacheContainer`, a reference-type (`class`) wrapper that owns the `[KVCache]` array. `TokenIterator` now holds a `KVCacheContainer` rather than a bare array, so element replacements during quantization are visible to any holder of the same container. `KVCacheContainer.maybeQuantize` also uses the new `Bool` return value from `maybeQuantizeKVCache` to set a `canQuantize` flag to `false` after the first successful quantization pass, eliminating the redundant per-step scan of all cache layers that occurred in the original code.

The `QuantizableKVCache` protocol is introduced to formalize which cache types support conversion to quantized form. `KVCacheSimple` conforms to it. `RotatingKVCache` is not yet supported (matching the Python implementation) with a TODO stub in place. Several quantization helpers (`asQuantizableWithGroupSize`, `canQuantize(at:groupSize:)`, `isValidQuantizedGroupSize`, `resolvedKVQuantizationGroupSize`) are moved from a private free function into a `KVCache` extension, making them accessible for reuse and testing. The dead-code `!(firstQuantizable is QuantizedKVCache)` check noted in the original issue is removed.

`maybeQuantizeKVCache` gains `@discardableResult` and now returns `Bool` to indicate whether quantization occurred.

`TokenIterator`'s multiple `init` paths are consolidated to funnel through a single canonical initializer, reducing duplication. The previously-fixed inits also pick up optional `processor` and `sampler` override parameters for consistency. `tokenCount` is tightened to `private(set) public`.

**Out of scope**

This PR does not address the `ChatSession` side of cache ownership, `SpeculativeTokenIterator`, or `RotatingKVCache` quantization support. Those are candidates for follow-up work.

**Unrelated changes**

Two minor compiler warning fixes are included: removal of an unused `hasExplicitCache` variable in `Gemma4.swift`, and a `var` → `let` correction for `freqsT` in `Qwen25VL.swift`.

_PR description crafted with assist from Claude since I'm a bad explainer_

## Checklist

### DIFF
--- a/Libraries/MLXLMCommon/Evaluate.swift
+++ b/Libraries/MLXLMCommon/Evaluate.swift
@@ -533,11 +533,12 @@ public struct TokenIterator: TokenIteratorProtocol {
     var state: LMOutput.State?
 
     var y: LMInput.Text
-    var cache: [KVCache]
+    var cacheContainer: KVCacheContainer
+    var cache: [KVCache] { cacheContainer.cache }
     var processor: LogitProcessor?
     let sampler: LogitSampler
 
-    public var tokenCount = 0
+    private(set) public var tokenCount = 0
     public let maxTokens: Int?
 
     // Cache quantization parameters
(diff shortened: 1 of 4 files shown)


fdagostino:
We validated this approach downstream (cherry-picked onto our fork after hitting #312 in production — a multi-turn voice app with `kvBits` set, where the session went amnesic about its own replies): `KVCacheContainer` fixes the direct `TokenIterator` path as advertised.

One gap worth flagging for the next step: **`ChatSession` is still affected**. It stores its own `[KVCache]` array and passes it to `TokenIterator`, so the element replacements the iterator makes through its container (KVCacheSimple → QuantizedKVCache) never reach the session's storage — every token generated after the quantization swap is lost, and the next turn continues from a stale prefix. Downstream we fixed it by having the session create the `KVCacheContainer`, pass it via the container-based init, and adopt `container.cache` back into its stored state after each generation pass — plus a regression test that fails without the sync-back, reproducing the #312 symptom exactly (the session ends up holding the stale `KVCacheSimple` while the iterator's `QuantizedKVCache` is dropped).

Happy to submit that as a follow-up PR once this one merges (one step at a time, as discussed). The `SpeculativeTokenIterator`/MTP paths hold plain `[KVCache]` arrays too and would lose replacements the same way — probably further steps down the same road.

davidkoski:
Sorry for the delay getting to this -- a lot of PRs coming in!  I like the shape of this but I think I am going to take #453 over this.  It has the same approach but also covers the ChatSession side as well as dealing with a number of other KVCache issues.

Thank you for the contribution!

PR #360, opened 2026-06-22
Support correlated tool call transcripts

aleroot:
## Proposed changes

This PR adds the missing transcript representation and default message-generation support for correlated tool calls. Model-specific generators can build on this where needed.

## Checklist

### DIFF
--- a/Libraries/MLXLMCommon/Tool/Parsers/JSONToolCallParser.swift
+++ b/Libraries/MLXLMCommon/Tool/Parsers/JSONToolCallParser.swift
@@ -31,8 +31,7 @@ public struct JSONToolCallParser: ToolCallParser, Sendable {
 
         guard
             let data = jsonStr.data(using: .utf8),
-            let normalizedData = normalizedToolCallData(from: data),
-            let function = try? JSONDecoder().decode(ToolCall.Function.self, from: normalizedData)
+            let toolCall = parseToolCall(from: data)
         else { return nil }
 
         // If tool schemas are provided, only accept calls to declared tools.
@@ -40,7 +39,7 @@ public struct JSONToolCallParser: ToolCallParser, Sendable {
             var isDeclaredTool = false
             for tool in tools {
                 let functionSpec = tool["function"] as? [String: any Sendable]
-                if functionSpec?["name"] as? String == function.name {
+                if functionSpec?["name"] as? String == toolCall.function.name {
                     isDeclaredTool = true
                     break
                 }
@@ -51,13 +50,19 @@ public struct JSONToolCallParser: ToolCallParser, Sendable {
             }
         }
 
-        return ToolCall(function: function)
+        return toolCall
     }
 
-    private func normalizedToolCallData(from data: Data) -> Data? {
+    private func parseToolCall(from data: Data) -> ToolCall? {
         guard var jsonObject = try? JSONSerialization.jsonObject(with: data) as? [String: Any]
         else {
             return nil
+        }
+
+        var id = jsonObject["id"] as? String
+        if let functionObject = jsonObject["function"] as? [String: Any] {
+            id = id ?? functionObject["id"] as? String
+            jsonObject = functionObject
         }
 
         if let stringifiedArguments = jsonObject["arguments"] as? String {
@@ -69,6 +74,11 @@ public struct JSONToolCallParser: ToolCallParser, Sendable {
             jsonObject["arguments"] = argumentsObject
         }
 
-        return try? JSONSerialization.data(withJSONObject: jsonObject)
+        guard
+            let normalizedData = try? JSONSerialization.data(withJSONObject: jsonObject),
+            let function = try? JSONDecoder().decode(ToolCall.Function.self, from: normalizedData)
+        else { return nil }
+
+        return ToolCall(function: function, id: id)
     }
 }
--- a/Libraries/MLXLMCommon/Tool/ToolCall.swift
+++ b/Libraries/MLXLMCommon/Tool/ToolCall.swift
@@ -20,13 +20,21 @@ public struct ToolCall: Hashable, Codable, Sendable {
             self.name = name
             self.arguments = arguments.mapValues { JSONValue.from($0) }
         }
+
+        var argumentsObject: [String: any Sendable] {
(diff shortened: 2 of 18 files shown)


ronaldmannak:
As mentioned in #246:

Thanks for the follow-up @davidkoski and @aleroot.

I think the best way forward is to close this PR and use #360 as the foundation, since it seems to model and enforce message correctness better.

That said, I think #360 still needs one follow-up improvement: **missing tool-call IDs should be normalized/generated before the calls are exposed to the caller**.

Suggested behavior:

```swift
if parsedToolCall.id == nil {
    parsedToolCall.id = generateToolCallID()
}
```

The generated ID should then stay attached to the assistant ToolCall, and callers should use that same ID when appending the corresponding tool-result message.

It may also be worth adding tests for:
- preserving provider/model-supplied tool-call IDs;
- generating IDs for parsed tool calls that do not include one;
- ensuring generated IDs are distinct for multiple tool calls in one assistant turn;
- serializing tool-result messages with the matching tool_call_id;
- preventing or detecting duplicate IDs in a single assistant turn;
- preventing orphan tool-result messages that reference no prior assistant tool call.

aleroot:
@ronaldmannak Thanks for the feedback, I have updated.

Tests/MLXLMTests/ToolCallIdTests.swift
```diff
@@ -0,0 +1,173 @@
...
+        #expect(calls.count == 2)
+
+        let weather = calls[0]
+        #expect(weather["id"] as? String == "call_123")
+        #expect(weather["type"] as? String == "function")
+
+        let weatherFunction = try #require(weather["function"] as? [String: any Sendable])
+        #expect(weatherFunction["name"] as? String == "get_weather")
+
+        #expect(weatherFunction["arguments"] as? String == #"{"location":"Paris"}"#)
```
davidkoski: Should this be structured?  The chat template may be expecting a json structure.  From mlx-lm server.py:

```python
        if tool_calls := message.get("tool_calls"):
            for tool_call in tool_calls:
                if func := tool_call.get("function"):
                    if args := func.get("arguments"):
                        func["arguments"] = json.loads(args)
```

It converts the string to structured.  At least some chat templates expect that, e.g. https://huggingface.co/mlx-community/Meta-Llama-3.1-8B-Instruct-bf16/blob/main/tokenizer_config.json

```
            {%- for arg_name, arg_val in tool_call.arguments | items %}
                {{- arg_name + '=\"' + arg_val + '\"' }}
                {%- if not loop.last %}
                    {{- \", \" }}
                {%- endif %}
                {%- endfor %}
            {{- \")\" }}
```

Is there an equivalent here?  Should there be?

Libraries/MLXLMCommon/Tool/ToolCallProcessor.swift
```diff
@@ -437,6 +438,36 @@ public class ToolCallProcessor {
...
+
+    private func normalizedToolCall(_ call: ToolCall) -> ToolCall {
+        if let id = call.id, !id.isEmpty, emittedToolCallIDs.insert(id).inserted {
+            return call
+        }
+
+        return ToolCall(function: call.function, id: generateToolCallID())
+    }
+
+    private func generateToolCallID() -> String {
```
davidkoski, line 461: I wonder if this needs to be configurable, e.g. via an enum or the ModelConfiguration level?  Per gemini, Mistral models have restrictions on this and won't handle this format:

> Format: Tool call IDs must match the regex `^[a-zA-Z0-9]{9}$`. They must consist only of lowercase letters, uppercase letters, and digits.
> Length: IDs must be exactly 9 characters long.
> Forbidden Characters: No underscores ( _ ), hyphens ( - ), or prefixes (like call_ ) are allowed in the ID string.

Libraries/MLXLMCommon/Tool/ToolCall.swift
```diff
@@ -20,13 +20,32 @@ public struct ToolCall: Hashable, Codable, Sendable {
...
+            encoder.outputFormatting = [.sortedKeys]
+
+            do {
+                let data = try encoder.encode(arguments)
+                guard let string = String(data: data, encoding: .utf8) else {
+                    preconditionFailure("Tool call arguments must encode as UTF-8 JSON.")
+                }
+                return string
+            } catch {
+                preconditionFailure("Tool call arguments must be JSON encodable: \(error)")
```
davidkoski: Should this throw?  Would this trigger if the model produced arguments that didn't encode properly -- the should probably be a runtime error rather than a programmer error.

Maybe the same for the UTF-8, though I would expect _all_ JSON to be UTF-8, so maybe not.

Libraries/MLXLMCommon/Tool/Parsers/JSONToolCallParser.swift
```diff
@@ -51,15 +50,20 @@ public struct JSONToolCallParser: ToolCallParser, Sendable {
...
-    private func normalizedToolCallData(from data: Data) -> Data? {
+    private func parseToolCall(from data: Data) -> ToolCall? {
         guard var jsonObject = try? JSONSerialization.jsonObject(with: data) as? [String: Any]
         else {
             return nil
         }
 
+        let id = jsonObject["id"] as? String
+        if let functionObject = jsonObject["function"] as? [String: Any] {
+            jsonObject = functionObject
```
davidkoski, line 65: In this nested call structure, would the id be inside the `function: { id: ... }`?

Libraries/MLXLMCommon/Tool/ToolCallProcessor.swift
aleroot: I had totally missed this. Addressed by deriving generated IDs from ToolCallFormat. The default formats keep OpenAI-style call_ IDs, while .mistral now generates exactly 9 alphanumeric characters. Thanks

Libraries/MLXLMCommon/Tool/Parsers/JSONToolCallParser.swift
aleroot: Good catch. The parser now preserves a nested function.id when the top-level id is absent.

davidkoski (approved):

PR #363, opened 2026-06-22
Gemma4: drop gemma4_unified vision_embedder weights in sanitize

neuromechanist:
The `gemma4_unified` (12B) family is multimodal: its checkpoints carry an encoder-free vision module under the `vision_embedder` prefix, alongside the already-skipped `embed_vision` / `embed_audio`. `gemma4_unified` is registered to the text `Gemma4Model` in `LLMModelFactory`, but `Gemma4Model.sanitize` does not drop `vision_embedder`, so loading a multimodal checkpoint (e.g. `mlx-community/gemma-4-12B-it-4bit`) through the text path fails with:

```
Unhandled keys ["vision_embedder"] in Gemma4Model
```

This adds the one missing prefix to the existing vision/audio skip list. Verified by loading + greedy-generating `mlx-community/gemma-4-12B-it-4bit` and matching mlx-vlm's text path token-for-token.

### DIFF
--- a/Libraries/MLXLLM/Models/Gemma4.swift
+++ b/Libraries/MLXLLM/Models/Gemma4.swift
@@ -68,10 +68,13 @@ public class Gemma4Model: Module, LLMModel, KVCacheDimensionProvider {
             let startsWithModel = k.hasPrefix("model.")
             k = k.replacingOccurrences(of: "model.", with: "", options: .anchored)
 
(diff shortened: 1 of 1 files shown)


davidkoski (approved):
Same fix as #366 (which got merged first) but this adds a nice comment.  Thanks!

PR #364, opened 2026-06-22
MLXLLM Gemma4Text: add MoE block (router + experts) for the text path

neuromechanist:
The MLXLLM `Gemma4Text` text path can load dense and (since #342) KV-shared Gemma4 checkpoints, but not **MoE** ones (26B-A4B): loading fails with `Unhandled keys ["experts", "router", "post_feedforward_layernorm_1", "post_feedforward_layernorm_2", "pre_feedforward_layernorm_2"]`. The complete MoE implementation already exists in `MLXVLM/Models/Gemma4.swift` (#180, #228); this brings it to the LLM text path.

Adds: `Gemma4TextRouter`, `Gemma4TextExperts` (SwitchGLU), the `enable_moe_block` config + dense/sparse decoder branch, and the fused-expert weight remap in `sanitize` (`experts.gate_up_proj`/`down_proj` -> `switch_glu.*`). Pairs with the KV-sharing already merged in #342.

Verified by loading + generating `mlx-community/gemma-4-26b-a4b-it-4bit` through the LLM text path (correct output at native context). Ported from ml-explore's own MLXVLM implementation.

### DIFF
--- a/Libraries/MLXLLM/Models/Gemma4Text.swift
+++ b/Libraries/MLXLLM/Models/Gemma4Text.swift
@@ -65,6 +65,11 @@ public struct Gemma4TextConfiguration: Codable, Sendable {
     var attentionKeqV: Bool = false
     var finalLogitSoftcapping: Float = 30.0
     var useDoubleWideMlp: Bool = true
+    // MoE block (E-series: enable_moe_block=true)
+    var enableMoEBlock: Bool = false
+    var numExperts: Int?
+    var topKExperts: Int?
+    var moeIntermediateSize: Int?
     var layerTypes: [String] = []
     var tieWordEmbeddings: Bool = true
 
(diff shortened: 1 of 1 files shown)


Libraries/MLXLLM/Models/Gemma4Text.swift
```diff
@@ -740,7 +881,37 @@ public class Gemma4TextModel: Module, LLMModel, KVCacheDimensionProvider {
             {
                 continue
             }
-            sanitized[k] = v
+
+            // Remap .experts.down_proj -> .experts.switch_glu.down_proj.weight
```
aleroot, line 885: The original sanitized[k] = value pass-through runs before the remap checks. For MoE keys, the original key (e.g. .experts.down_proj) is inserted into the dictionary alongside the remapped key (.experts.switch_glu.down_proj.weight). The continue only prevents the second pass-through it doesn't remove the first assignment. MoE keys should be remapped instead of passed through, not in addition.

davidkoski (approved):
Change looks good, thank you!

PR #365, opened 2026-06-23
Lfm2.5 bidirectional embedders

ronaldmannak:
## Proposed changes

Adds support for LiquidAI's **LFM2.5 bidirectional encoders** to `MLXEmbedders`:

- **LFM2.5-Embedding-350M** — CLS-pooled 1024-d dense vector (cosine similarity).
- **LFM2.5-ColBERT-350M** — per-token `Dense(1024→128)` projection (MaxSim late interaction).

A single `LFM2BidirectionalModel: EmbeddingModel` branches on the config `mlx` head. Both models share `model_type: "lfm2"` and are registered in `EmbedderTypeRegistry` plus six `EmbedderRegistry` configurations (`mlx-community/LFM2.5-…-{bf16,4bit,8bit}`).

### Implementation notes
- The encoder differs from the causal generative `LFM2` in `MLXLLM`: **non-causal** GQA attention (additive pad-only mask, no causal mask), a **centered** depthwise short-conv, and a pooling/projection head (no LM head). The short-conv runs **unmasked** over the full sequence — the checkpoints were trained that way on the eager/SDPA path, and zeroing padded / ColBERT query-expansion positions would shift per-token embeddings and hurt MaxSim; padding is handled only by the attention key-padding mask. Kept self-contained in `MLXEmbedders` per the module's convention (Qwen3/Gemma3 are reimplemented there rather than imported from `MLXLLM`).
- The ColBERT head returns **raw per-token vectors**; pad-dropping, query-expansion, and the punctuation skiplist live in the retrieval layer (as in PyLate), not the model.
- `sanitize` applies only the depthwise-conv transpose guard — deliberately **not** a blanket `model.` prefix (that would corrupt the bare `dense.weight`).

### Files
- `Libraries/MLXEmbedders/Models/LFM2Bidirectional.swift` (new) — config, bidirectional backbone, model, `sanitize`.
- `Libraries/MLXEmbedders/ModelFactory.swift` — `"lfm2"` type creator + six model configurations.
- `Tests/MLXLMTests/LFM2BidirectionalTests.swift` (new) —  config-decode + forward-shape tests.

---

### DIFF
--- a/Libraries/MLXEmbedders/ModelFactory.swift
+++ b/Libraries/MLXEmbedders/ModelFactory.swift
@@ -78,22 +78,22 @@ public class EmbedderRegistry: AbstractModelRegistry, @unchecked Sendable {
 
     /// LFM2.5 Embedding 350M — CLS-pooled 1024-d dense vectors (cosine).
     public static let lfm2_embedding_350m = ModelConfiguration(
-        id: "ronaldmannak/LFM2.5-Embedding-350M-bf16")
+        id: "mlx-community/LFM2.5-Embedding-350M-bf16")
     /// LFM2.5 Embedding 350M — 4-bit quantized.
     public static let lfm2_embedding_350m_4bit = ModelConfiguration(
-        id: "ronaldmannak/LFM2.5-Embedding-350M-4bit")
+        id: "mlx-community/LFM2.5-Embedding-350M-4bit")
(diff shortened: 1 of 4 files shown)


davidkoski (approved):
This looks great, thank you!

PR #367, opened 2026-06-24
Avoid replaying static chat prompt state

aleroot:
## Proposed changes

Avoid replaying static prompt material in long-lived `ChatSession` instances.

`ChatSession` now renders session instructions and prompt tool schemas only when the KV cache is first built, or after `clear()`. Later turns append only incremental messages/tool results, while parser tool schemas remain available for tool-call parsing.

#### Why

For chatbots and agents, static instructions are already represented in the KV cache after the first turn. Re-rendering them on every response wastes context and prefill work. For agents, repeated tool schemas can be especially expensive.

The existing `tools:` API remains source-compatible. A grouped `ToolConfiguration` supports advanced prompt/parser schema separation without expanding every initializer.

#### Notes

`clear()` resets the session cache, so the next response rebuilds the static prefix. Apps that change instructions mid-session should call `clear()` or create a new session.

## Checklist

### DIFF
--- a/Libraries/MLXHuggingFaceMacros/HuggingFaceIntegrationMacros.swift
+++ b/Libraries/MLXHuggingFaceMacros/HuggingFaceIntegrationMacros.swift
@@ -113,9 +113,25 @@ public struct TokenizerAdaptorMacro: ExpressionMacro {
                         tools: [[String: any Sendable]]?,
                         additionalContext: [String: any Sendable]?
                     ) throws -> [Int] {
+                        try applyChatTemplate(
+                            messages: messages,
+                            tools: tools,
+                            additionalContext: additionalContext,
+                            addGenerationPrompt: true)
+                    }
+
+                    func applyChatTemplate(
+                        messages: [[String: any Sendable]],
+                        tools: [[String: any Sendable]]?,
+                        additionalContext: [String: any Sendable]?,
+                        addGenerationPrompt: Bool
+                    ) throws -> [Int] {
                         do {
                             return try upstream.applyChatTemplate(
-                                messages: messages, tools: tools, additionalContext: additionalContext)
+                                messages: messages,
+                                addGenerationPrompt: addGenerationPrompt,
+                                tools: tools,
+                                additionalContext: additionalContext)
                         } catch Tokenizers.TokenizerError.missingChatTemplate {
                             throw MLXLMCommon.TokenizerError.missingChatTemplate
                         }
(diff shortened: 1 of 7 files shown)


davidkoski:
Is this the correct behavior?  I agree it is represented in the KVCache, but consider this from mlx-lm:

```python
            if args.system_prompt is not None:
                messages.append({"role": "system", "content": args.system_prompt})
            messages.append({"role": "user", "content": query})
            prompt = tokenizer.apply_chat_template(
                messages,
                add_generation_prompt=True,
            )
            last_response = None
            for response in stream_generate(...
```

That is inside the loop and it always injects the system prompt -- my understanding was this is what was expected.  Gemini was a little wishy washy on answering :-)

aleroot:
> Is this the correct behavior? I agree it is represented in the KVCache, but consider this from mlx-lm:
> 
> ```python
>             if args.system_prompt is not None:
>                 messages.append({"role": "system", "content": args.system_prompt})
>             messages.append({"role": "user", "content": query})
>             prompt = tokenizer.apply_chat_template(
>                 messages,
>                 add_generation_prompt=True,
>             )
>             last_response = None
>             for response in stream_generate(...
> ```
> 
> That is inside the loop and it always injects the system prompt -- my understanding was this is what was expected. Gemini was a little wishy washy on answering :-)

 Good catch. The mlx_lm.chat loop does indeed add the system prompt every turn. Because it reuses the same prompt cache and stream_generate performs no prefix comparison, those tokens are physically appended again; the cache does not deduplicate them.
After checking the broader implementation, I do not think that CLI behavior is a sufficient generic contract. mlx_lm.server instead renders the complete conversation, finds the exact cached token prefix, and sends only the remaining suffix.
My goal to avoid duplicate prefill is still fine, but the current “omit instructions/tools after the first turn” implementation is too broad. Real templates can emit BOS or a default system message for an isolated user message, and swift-transformers can switch from tool_use to the default template when tools becomes nil.
I’m going to revise this around exact full-prompt token-prefix reuse rather than treating static material as a one-time rendering operation. The system/tools should remain part of every canonical logical render; only their already-cached tokens should be skipped during prefill. I will update this later today ideally.

aleroot:
@davidkoski  my initial approach here was a bit optimistic 🙂, in fact simplistic.

I have now reworked the solution around exact token-prefix reuse rather than omitting static prompt after the first turn. The session still renders the canonical full prompt with instructions and prompt-side tool schemas, then verifies that the static-only render is an exact token prefix of that full prompt. On the first turn we prefill the full prompt and remember that reusable prefix; on later turns, if the prompt configuration is unchanged and the KV cache is safe to reuse, we pass only the suffix tokens that are not already represented in the cache.

So the behavior should now match the right semantic model: instructions/tools remain part of every canonical prompt render, but already-cached static prefix tokens are not replayed through the model. This should improve performances when using the library from a coding agent.

When eventually https://github.com/ml-explore/mlx-swift-lm/pull/360 will be merged, this can be improved further.

aleroot:
I am still working on some issues, I will update this PR further and add some more comments.

PR #368, opened 2026-06-24
Falcon-H1R: correctness, cache, and preset fixes

aleroot:
## Proposed changes

- Correct Falcon-H1R attention key multiplier handling.
- Correct Q/K/V sanitizer scaling for attention_in_multiplier.
- Preserve full logits for direct model calls, while generation uses num_logits_to_keep.
- Propagate Falcon mamba_chunk_size into SSM update.
- Make Falcon hybrid cache respect attention-only KV limits and KV quantization.
- Add recursive cache quantization for CacheList without touching Mamba state.
- Add conversion metadata so converted Falcon checkpoints are not re-sanitized.
- Add tiiuae/Falcon-H1R-7B to LLMRegistry in the same style as other models.

## Checklist


davidkoski (approved):
Changes look good, thank you!

PR #369, opened 2026-06-24
Fix Python finfo.min port in quantized attention masking

ronaldmannak:
## Proposed changes

Fixes a likely mistranslation from Python `mlx-lm` in `quantizedScaledDotProductAttention`.

The Python reference masks boolean/causal logits with `mx.finfo(scores.dtype).min`, a large negative finite value that suppresses masked positions before softmax. The Swift port used `Float.leastNormalMagnitude`, which is instead a tiny positive value, so masked positions could remain competitive or dominate when valid scores are negative.

This updates the Swift quantized attention mask fill value to match the intended Python semantics and adds regression coverage for masked quantized attention.

## Checklist


ronaldmannak:
That turned out to be a little bit more complex than I thought, but should be good to go now. Also in this PR: a related CQA fix in the same KVCache method

ronaldmannak:
One open question: I added extensions to MLXArray and DType to this repo. You could argue those belong in MLX-Swift. Happy to open a separate PR on MLX-Swift for that

davidkoski:
> One open question: I added extensions to MLXArray and DType to this repo. You could argue those belong in MLX-Swift. Happy to open a separate PR on MLX-Swift for that

Yeah, that would probably be the best approach.  There is a very _thin_ finfo on there already

ronaldmannak:
@davidkoski Done: https://github.com/ml-explore/mlx-swift/pull/429

davidkoski (approved):
Changes look good, I think this just needs a rebase for CI.  Thank you!

ronaldmannak:
@davidkoski done!

PR #370, opened 2026-06-24
Reuse cached chat transcript prefixes

aleroot:
## Proposed changes

Changes ChatSession to reuse the KV cache for chat prefixes automatically.
Instead of repeatedly prefilling prompt-rendered chat-prefix tokens on later turns, the session keeps an exact token ledger for the prompt and generated assistant output. On the next turn it re-renders the full transcript, checks that the cached tokens are an exact prefix and that KV cache offsets match, then prefills only the new suffix. If anything does not line up, it falls back to full prefill or throws when the prompt configuration changed.
Reuse is intentionally disabled for unsafe paths like logit processors, speculative decoding, bounded/non-reusable caches, and multimodal or masked inputs.

This should improve performance on coding agents using the library.

## Notes

**This is the first step to make a Codex-like app based on the library.** At the moment prefix reuse is enabled just for `KVCacheSimple` with no logit processors, no speculative decoding, and no dynamic KV quantization. The safest subset. Each opt-out is a known performance floor with a clear unlock path:

- **`RotatingKVCache`** — The transcript prefix eventually gets overwritten once `offset > maxSize`. A future two-tier cache (static prefix + rotating suffix) would enable reuse for the dominant production config (anyone setting `maxKVSize`).

- **Logit processors** — `PenaltyProcessor`'s `TokenRing` state is stale when skipping the prefix. On a prefix-reused turn, calling `processor.prompt()` with the prefix tokens before generation would sync the rings. Small mechanical change once plumbing is ready.

- **Speculative decoding** — `trimPromptCache` after rejected draft tokens can desync KV offsets from the ledger. The speculative iterator already tracks accepted/rejected per round — wiring `GenerationTrace.committedTokens` to the net effect would unlock this.

The `supportsStaticPrefixReuse` / `allowsPrefixReuse` gates are single-point switches designed for independent enablement.

#### Important 

 Once [#360](https://github.com/ml-explore/mlx-swift-lm/pull/360) lands, `PromptLedger.Message` should carry the structured tool metadata, not just `role` and `content`, so cached replay preserves correlated tool calls exactly.

## Checklist

### DIFF
--- a/Libraries/MLXLMCommon/ChatSession.swift
+++ b/Libraries/MLXLMCommon/ChatSession.swift
@@ -688,6 +688,12 @@ public final class ChatSession {
             return [.system(instructions)] + turnMessages
         }
 
+        static func containsMedia(_ messages: [Chat.Message]) -> Bool {
+            messages.contains {
+                !$0.images.isEmpty || !$0.videos.isEmpty || !$0.audios.isEmpty
+            }
+        }
+
         static func suffixInput(fullInput: LMInput, dropping prefixTokens: [Int]) -> LMInput? {
             guard fullInput.text.tokens.ndim == 1, fullInput.text.mask == nil else {
                 return nil
@@ -728,6 +734,9 @@ public final class ChatSession {
                 })
             else {
                 return .unavailable
+            }
+            guard !containsMedia(turnMessages) else {
+                return .invalid
             }
 
             let fullUserInput = UserInput(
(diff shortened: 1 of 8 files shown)


aleroot:
I have reworked completely from my previous https://github.com/ml-explore/mlx-swift-lm/pull/367 since that was too simplistic

aleroot:
Closing this in favor of #472.
After comparing the implementations, #472 provides the same central user benefit proposed here: the session retains the transcript, renders the complete conversation for template correctness, verifies the cached token prefix, and normally sends only the uncached suffix through model prefill.
This also benefits callers using the existing respond(to: String, role:) and streaming APIs because those overloads use the same internal continuation path as the structured-message APIs.
#472 is based on the newer implementation and additionally handles structured tool metadata, cache mismatch rebuilding/common-prefix trimming, speculative decoding alignment, media fallbacks, carried model state, and raw-cache compatibility. Therefore, merging both PRs would duplicate and conflict in the same cache-management code.

PR #371, opened 2026-06-25
Validate RoPE model configurations

RNT56:
## Summary
- validate RoPE scaling configuration before model construction
- apply the validation through the LLM and VLM model factory paths that consume RoPE config
- add regression coverage for unsupported RoPE types and malformed multimodal RoPE sections

## Validation
- swift test --filter RoPEConfigurationValidationTests
- swift build --target MLXVLM
- pre-commit run --all-files
- git diff --check


RNT56:
This should be ready for review now.

I kept this PR focused on RoPE config validation. I also added tests for valid current configs, so this should catch bad configs without blocking the supported ones.

Local validation:

- `swift test --filter RoPEConfigurationValidationTests`
- `swift build --target MLXLMCommon`
- `git diff --check`

davidkoski (approved):
Looks good, thank you!

PR #372, opened 2026-06-25
Add runtime stop string handling

RNT56:
## Summary
- decode stop string metadata from generation_config.json and model registry entries
- propagate stop strings through resolved model configuration and generation handlers
- stop decoded text generation when a stop string spans token chunks without emitting the stop sequence

## Scope
This PR is intentionally limited to runtime stop-string handling. It does not include the broader DFlash, MTP, or TurboQuant stack from the staging branch this was split from.

## Validation
- swift build --target MLXLMCommon
- swift build --target MLXLLM
- swift build --target MLXVLM
- swift test --filter StopStringTests
- pre-commit run --all-files
- git diff --check

### DIFF
--- a/Libraries/MLXLLM/LLMModelFactory.swift
+++ b/Libraries/MLXLLM/LLMModelFactory.swift
@@ -127,13 +127,15 @@ public class LLMRegistry: AbstractModelRegistry, @unchecked Sendable {
     static public let phi3_5_4bit = ModelConfiguration(
         id: "mlx-community/Phi-3.5-mini-instruct-4bit",
         defaultPrompt: "What is the gravity on Mars and the moon?",
-        extraEOSTokens: ["<|end|>"]
+        extraEOSTokens: ["<|end|>"],
+        stopStrings: ["<|end|>"]
     )
 
     static public let phi3_5MoE = ModelConfiguration(
         id: "mlx-community/Phi-3.5-MoE-instruct-4bit",
         defaultPrompt: "What is the gravity on Mars and the moon?",
-        extraEOSTokens: ["<|end|>"]
+        extraEOSTokens: ["<|end|>"],
+        stopStrings: ["<|end|>"]
     )
 
     static public let gemma2bQuantized = ModelConfiguration(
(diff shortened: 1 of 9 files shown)


RNT56:
This should be ready for review now.

I kept this PR limited to runtime stop-string handling. The added tests cover the config path and a case where the stop string is split across decoded chunks.

Local validation:

- `swift test --filter StopStringTests`
- `swift build --target MLXLMCommon`
- `git diff --check`

davidkoski:
What is the difference between EOS and stop strings?  Is it just token vs potentially sequence of tokens?

RNT56:
Yes, that is basically the difference.

EOS is token based: generation stops when the model emits an EOS token id.

Stop strings are text based: after decoding, we look for configured strings. They can be one token, multiple tokens, or split across decoded chunks, so this path also has to hold back a small suffix to avoid returning part of a stop string.

So this PR keeps EOS unchanged and adds the text-level stop-string layer on top.

Libraries/MLXLLM/LLMModelFactory.swift
```diff
@@ -127,13 +127,15 @@ public class LLMRegistry: AbstractModelRegistry, @unchecked Sendable {
     static public let phi3_5_4bit = ModelConfiguration(
         id: "mlx-community/Phi-3.5-mini-instruct-4bit",
         defaultPrompt: "What is the gravity on Mars and the moon?",
-        extraEOSTokens: ["<|end|>"]
+        extraEOSTokens: ["<|end|>"],
+        stopStrings: ["<|end|>"]
```
davidkoski: Should stopStrings be nullable and nil by default -- just use the `extraEOSTokens` if not specified?  It looks like all the entries do just that.

Libraries/MLXVLM/VLMModelFactory.swift
```diff
@@ -196,43 +204,50 @@ public class VLMRegistry: AbstractModelRegistry, @unchecked Sendable {
...
+        extraEOSTokens: ["<end_of_turn>"],
+        stopStrings: ["<end_of_turn>"]
     )
 
     static public let gemma4_E2B_it_4bit = ModelConfiguration(
         id: "mlx-community/gemma-4-e2b-it-4bit",
         defaultPrompt: "Describe the image in English",
-        extraEOSTokens: ["<end_of_turn>"]
+        extraEOSTokens: ["<end_of_turn>"],
+        stopStrings: ["<end_of_turn>"]
```
davidkoski: Not a bug in this PR, but noticed here:  the gemma4 end token is `<turn|>` (see the LLMModelFactory).

RNT56:
Thanks, I´ve updated this.

Changes made so far:

- `stopStrings` is now optional on `ModelConfiguration`.
- If it is not set, runtime text stops fall back to `extraEOSTokens`.
- If it is set explicitly, that value is used instead, including an explicit empty set.
- Removed the duplicated registry `stopStrings` entries that only repeated `extraEOSTokens`.

I also fixed the Gemma4 VLM registry defaults you noticed. The Gemma4 VLM entries now use `<turn|>`, matching the LLM registry and the model metadata. I hope it fits.

Libraries/MLXVLM/VLMModelFactory.swift
RNT56: Should I open a standalone Gemma4 token PR or is this fine in #372 ?

davidkoski: Cleanest to make a new PR, but I think fine to fix here.

RNT56:
I split this out as per your comment @davidkoski 

I removed the Gemma4 VLM token change from this PR, so this is back to just the stop-string handling.

The Gemma4 VLM token fix is now in #373.

davidkoski (approved):
This looks good to me, thank you!

PR #373, opened 2026-06-26
Fix Gemma4 VLM end token defaults

RNT56:
This fixes the Gemma4 VLM registry defaults to use `<turn|>`.

The LLM Gemma4 entries already use this token. This makes the VLM entries match that.

Tests:
- `swift test --filter VLMRegistryTests`
- `swift build --target MLXVLM`
- `git diff --check`


davidkoski (approved):
Looks good, just waiting on CI.  Thank you!

PR #375, opened 2026-06-28
Add reranker API support

aleroot:
## Proposed changes

- Adds shared reranker request/result types and public rerank APIs.
- Supports encoder sequence-classification rerankers and Jina listwise reranking.
- Handles token limits, marker validation, score selection, and reranker registry entries.
- Adds focused reranker tests covering scoring, truncation, model factories, and error 

## Checklist

### DIFF
--- a/Libraries/MLXEmbedders/Models/Bert.swift
+++ b/Libraries/MLXEmbedders/Models/Bert.swift
@@ -18,6 +18,8 @@ private class BertEmbedding: Module {
 
     /// The number of distinct token types the model can distinguish.
     let typeVocabularySize: Int
+    let padTokenID: Int
+    let usesPaddingAwarePositionIDs: Bool
 
     /// The embedding lookup table for the vocabulary.
     @ModuleInfo(key: "word_embeddings") var wordEmbeddings: Embedding
@@ -35,6 +37,8 @@ private class BertEmbedding: Module {
     /// - Parameter config: The `BertConfiguration` containing dimensions, vocabulary sizes, and epsilon values.
     init(_ config: BertConfiguration) {
         typeVocabularySize = config.typeVocabularySize
+        padTokenID = config.padTokenID
+        usesPaddingAwarePositionIDs = config.usesPaddingAwarePositionIDs
 
         // Initialize word embeddings based on vocab size and embedding dimension
         _wordEmbeddings.wrappedValue = Embedding(
@@ -69,13 +73,11 @@ private class BertEmbedding: Module {
         positionIds: MLXArray? = nil,
         tokenTypeIds: MLXArray? = nil
     ) -> MLXArray {
-        // Generate position IDs [0, 1, ..., N] and broadcast to the input shape if not provided
-        let posIds = positionIds ?? broadcast(MLXArray.arange(inputIds.dim(1)), to: inputIds.shape)
+        let posIds = positionIds ?? defaultPositionIDs(for: inputIds)
 
         // Combine word and position embeddings
         var words = wordEmbeddings(inputIds) + positionEmbeddings(posIds)
 
-        // Add token type (segment) embeddings if both IDs and the layer exist
         if let tokenTypeIds, let tokenTypeEmbeddings {
             words += tokenTypeEmbeddings(tokenTypeIds)
         }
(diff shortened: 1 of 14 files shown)


aleroot:
Validated just now the numerical tests for BGE, Qwen3, and Jina against python reference implementation, it looks like we are good. 

@davidkoski Whenever you got some time I would appreciate your validation or insights on the API shape, next week I would like to start the implementation in my app Lampo.

davidkoski:
I will take a look -- I need to understand rerankers in a bit more detail.

Libraries/MLXLMCommon/Reranker.swift
```diff
@@ -0,0 +1,1332 @@
...
+            documents: documents,
+            instruction: nil,
+            options: .init())
+    }
+
+    /// Score documents without changing their original order.
+    public func scores(
+        query: String,
+        documents: [String],
+        instruction: String?
```
davidkoski: since this isn't a protocol requirement, should this fold into the function above and simply be `instruction: String? = nil`?

Libraries/MLXLMCommon/Reranker.swift
```diff
@@ -0,0 +1,1332 @@
...
+            documents: documents,
+            instruction: instruction,
+            options: .init())
+    }
+
+    /// Score documents without changing their original order using custom execution options.
+    public func scores(
+        query: String,
+        documents: [String],
+        options: RerankExecutionOptions
```
davidkoski: similarly, should this fold into the above as `options: RerankExecutionOptions = .init()`?

Libraries/MLXLMCommon/Reranker.swift
```diff
@@ -0,0 +1,1332 @@
...
+    /// Identifier of the loaded model.
+    var modelID: String { get }
+
+    /// Semantic meaning and numeric range of returned scores.
+    var scoreKind: RerankScoreKind { get }
+
+    /// Score documents while preserving their original order.
+    func scores(
+        query: String,
+        documents: [String],
```
davidkoski, line 233: Is there a case where you would want to provide documents as cached embedding space?

Perhaps not, it looks like they are all considered together in the input: 

```swift
        let prompt = renderPrompt(query: sanitizedQuery, documents: sanitizedDocuments)
        var tokenIds = tokenizer.encode(text: prompt, addSpecialTokens: false)
```

producing:

```swift
package struct RerankerInput: Sendable {
    /// Token IDs passed to the model.
    public var tokenIds: [Int]
```

Libraries/MLXLMCommon/Reranker.swift
davidkoski: I am worried about proliferation of methods, e.g. ChatSession and generate.

Libraries/MLXEmbedders/Models/Bert.swift
```diff
@@ -69,20 +73,45 @@ private class BertEmbedding: Module {
...
 
         // Combine word and position embeddings
         var words = wordEmbeddings(inputIds) + positionEmbeddings(posIds)
 
-        // Add token type (segment) embeddings if both IDs and the layer exist
-        if let tokenTypeIds, let tokenTypeEmbeddings {
-            words += tokenTypeEmbeddings(tokenTypeIds)
+        if let tokenTypeEmbeddings {
+            let typeIDs = bertTokenTypeIDs(inputIDs: inputIds, tokenTypeIDs: tokenTypeIds)
+            words += tokenTypeEmbeddings(typeIDs)
```
davidkoski: This looks like a behavior change if `tokenTypeIds` is omitted -- now it will run zeros through the embedding layer and add it in.  I am not sure what the effect would be, but it seems like it would produce different output than before.

Is this desired?  It affects all bert models (e.g. non-reranker models).

davidkoski (changes requested):
This API looks clean and makes sense to me.

Take a look at the question on Bert and if folding some of the methods down using default arguments makes sense.

Otherwise this looks good to me.  Thanks!

Libraries/MLXLMCommon/Reranker.swift
aleroot: Agree. I consolidated the convenience overloads into a single forwarding method with instruction: String? = nil.

Libraries/MLXLMCommon/Reranker.swift
aleroot: The same method now uses options: RerankExecutionOptions = .init(), removing the overload proliferation.

Libraries/MLXLMCommon/Reranker.swift
aleroot: Yes, that's correct.

Libraries/MLXEmbedders/Models/Bert.swift
aleroot: Good catch. I restored the existing behavior: when tokenTypeIds is omitted, token-type embeddings are not added.

aleroot:
> This API looks clean and makes sense to me.
> 
> Take a look at the question on Bert and if folding some of the methods down using default arguments makes sense.
> 
> Otherwise this looks good to me. Thanks!

Folded the convenience overloads into default arguments, and restored BERT so omitted  tokenTypeIds  does not add token-type embeddings. Ready for another look.
Thanks

davidkoski (approved):
Slipped through the cracks :-)

Looks good to me!

PR #376, opened 2026-06-29
Decode PoolingConfiguration with optional pooling_mode flags

### LINKED ISSUE #26: PoolingConfiguration fails to decode when pooling_mode_lasttoken is missing
Decoding [PoolingConfiguration](https://github.com/ml-explore/mlx-swift-examples/blame/main/Libraries/Embedders/Pooling.swift#L8) fails for models like [sentence-transformers/all-MiniLM-L6-v2](https://huggingface.co/sentence-transformers/all-MiniLM-L6-v2) because their [1_Pooling/config.json](https://huggingface.co/sentence-transformers/all-MiniLM-L6-v2/blob/main/1_Pooling/config.json) does not include the `pooling_mode_lasttoken` field.

The struct currently requires this key as a non-optional Bool.

**Impact**: As a result, embeddings are generated without any pooling, differing subtly from those produced by other frameworks.

**Workaround**: As far as I can tell, there’s also no way to explicitly configure a pooling strategy as a workaround.

GoodOlClint:
## Proposed changes

Fixes #26. Some sentence-transformers pooling configs omit `pooling_mode_*` keys — e.g. all-MiniLM-L6-v2 has no `pooling_mode_lasttoken`. Today the synthesized `Codable` decoder requires all four flags, so `PoolingConfiguration` decoding throws and `loadPooling` silently falls back to the model's built-in strategy instead of using the config file.

This adds a custom `init(from:)` that decodes each `pooling_mode_*` flag with `decodeIfPresent(...) ?? false` — an absent key means that mode is off — while keeping `word_embedding_dimension` required. It mirrors the optional-decode pattern already used by `Qwen3Configuration` / `Gemma3Configuration`.

Re: the open question on the default — `false` matches the sentence-transformers convention that an absent `pooling_mode_*` key means that mode is disabled.

Behavior note: a config missing *all four* flags now resolves to `.first` (previously it threw and fell back to the model strategy); configs that set at least one flag (the common case, incl. all-MiniLM) are unaffected beyond no longer throwing.

Test: `EmbeddingPoolingTests` decodes an all-MiniLM-L6-v2-shaped config (no `pooling_mode_lasttoken`) and asserts it resolves to `.mean` pooling. Verified via `xcodebuild test ... -only-testing:MLXLMTests/EmbeddingPoolingTests` → **TEST SUCCEEDED**.

## Checklist

### DIFF
--- a/Libraries/MLXEmbedders/Pooling.swift
+++ b/Libraries/MLXEmbedders/Pooling.swift
@@ -33,6 +33,23 @@ public struct PoolingConfiguration: Codable {
         case poolingModeMaxTokens = "pooling_mode_max_tokens"
         case poolingModeLastToken = "pooling_mode_lasttoken"
     }
+
+    public init(from decoder: any Decoder) throws {
+        let container = try decoder.container(keyedBy: CodingKeys.self)
+        // `word_embedding_dimension` is always present in the sentence-transformers spec.
+        dimension = try container.decode(Int.self, forKey: .dimension)
+        // The `pooling_mode_*` flags are optional: some sentence-transformers configs (e.g.
+        // all-MiniLM-L6-v2) omit keys such as `pooling_mode_lasttoken`. An absent key means
+        // that mode is off, so default each missing flag to `false`.
+        poolingModeClsToken =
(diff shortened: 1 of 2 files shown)


davidkoski (approved):
Change looks good, thank you.

PR #377, opened 2026-06-29
Add optional seed to GenerateParameters for reproducible sampling

GoodOlClint:
## Proposed changes

Adds an optional `seed: UInt64?` to `GenerateParameters`, threaded into `TopPSampler` and `CategoricalSampler`, so that a given `(seed, prompt, parameters)` deterministically reproduces the same sampled tokens. This brings reproducible sampling to mlx-swift-lm, matching the seed support already in Python `mlx-lm`.

Fully backward-compatible: `seed` defaults to `nil`, which preserves the existing entropy-seeded `MLXRandom.RandomState()` behavior. The seed is inert at `temperature == 0` (the argmax path has no RNG).

Tests (`SampleTests`): the same seed reproduces a 24-draw sequence; distinct seeds diverge; `GenerateParameters` threads the seed through to the constructed sampler; and a `nil` seed still produces valid in-range draws. Verified via `xcodebuild test ... -only-testing:MLXLMTests/SampleTests` → 24 tests, 0 failures, **TEST SUCCEEDED**.

## Checklist

### DIFF
--- a/Libraries/MLXLMCommon/Evaluate.swift
+++ b/Libraries/MLXLMCommon/Evaluate.swift
@@ -89,6 +89,13 @@ public struct GenerateParameters: Sendable {
     /// Min-p sampling threshold relative to the highest probability token (0 disables)
     public var minP: Float
 
+    /// Optional seed for reproducible sampling. When set, the sampler's RNG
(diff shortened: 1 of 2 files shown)


davidkoski (approved):
Change looks good, thank you!

PR #379, opened 2026-06-29
Add DeepSeek-V2 model

GoodOlClint:
## Proposed changes

Adds the **DeepSeek-V2** architecture (`model_type: "deepseek_v2"`) to MLXLLM — a port of mlx-lm's [`deepseek_v2.py`](https://github.com/ml-explore/mlx-lm/blob/main/mlx_lm/models/deepseek_v2.py). Covers DeepSeek-V2 / V2-Lite, which had no Swift implementation.

It reuses the in-tree **DeepSeek-V3** MLA machinery (`q_a`/`q_b` + `kv_a_proj_with_mqa`/`kv_b` low-rank compression, qk nope/rope split, the shared `initializeRope` YaRN path) and `SwitchGLU` MoE with shared experts.

It differs from V3 only in the router (`DeepseekV2MoEGate`):
- plain **softmax** scoring (no sigmoid + `e_score_correction_bias`),
- optional `group_limited_greedy` group masking by each group's **max** score (vs V3's top-2-per-group sum),
- an unconditional `routed_scaling_factor`.

Dense layers below `first_k_dense_replace` use the standard MLP. `sanitize()` stacks per-expert gate/up/down projections (`weight`/`scales`/`biases`) into `switch_mlp`.

Tests (`DeepseekV2Tests`, no model download): forward-pass logit shape; a **deterministic group-limited-greedy routing** test (synthetic gate weights → the winning group confines top-k); and `sanitize()` expert-stacking. Verified via `xcodebuild test ... -only-testing:MLXLMTests/DeepseekV2Tests` → 3 tests, 0 failures, **TEST SUCCEEDED**.

> Note: validation is structural/unit-level (shape + routing + weight-remap); end-to-end numerical parity against a downloaded DeepSeek-V2 checkpoint has not been run.

## Checklist

### DIFF
--- a/Libraries/MLXLLM/Models/DeepseekV2.swift
+++ b/Libraries/MLXLLM/Models/DeepseekV2.swift
@@ -17,7 +17,9 @@ public struct DeepseekV2Configuration: Codable, Sendable {
     var nRoutedExperts: Int?
     var routedScalingFactor: Float = 1.0
     var kvLoraRank: Int = 512
-    var qLoraRank: Int = 1536
+    // Optional: DeepSeek-V2-Lite sets `q_lora_rank: null` (direct q_proj, no
+    // low-rank q compression); full DeepSeek-V2 sets it to 1536. nil ⇒ q_proj.
+    var qLoraRank: Int?
     var qkRopeHeadDim: Int = 64
     var vHeadDim: Int = 128
     var qkNopeHeadDim: Int = 128
@@ -77,7 +79,7 @@ public struct DeepseekV2Configuration: Codable, Sendable {
         self.routedScalingFactor =
             try c.decodeIfPresent(Float.self, forKey: .routedScalingFactor) ?? 1.0
         self.kvLoraRank = try c.decodeIfPresent(Int.self, forKey: .kvLoraRank) ?? 512
-        self.qLoraRank = try c.decodeIfPresent(Int.self, forKey: .qLoraRank) ?? 1536
+        self.qLoraRank = try c.decodeIfPresent(Int.self, forKey: .qLoraRank)
         self.qkRopeHeadDim = try c.decodeIfPresent(Int.self, forKey: .qkRopeHeadDim) ?? 64
         self.vHeadDim = try c.decodeIfPresent(Int.self, forKey: .vHeadDim) ?? 128
         self.qkNopeHeadDim = try c.decodeIfPresent(Int.self, forKey: .qkNopeHeadDim) ?? 128
@@ -121,7 +123,7 @@ class DeepseekV2Attention: Module {
     init(config: DeepseekV2Configuration) {
         self.config = config
         self.numHeads = config.numAttentionHeads
-        self.qLoraRank = config.qLoraRank == 0 ? nil : config.qLoraRank
+        self.qLoraRank = (config.qLoraRank ?? 0) > 0 ? config.qLoraRank : nil
         self.qkRopeHeadDim = config.qkRopeHeadDim
         self.kvLoraRank = config.kvLoraRank
         self.vHeadDim = config.vHeadDim
--- a/Libraries/MLXLLM/Models/DeepseekV2.swift
+++ b/Libraries/MLXLLM/Models/DeepseekV2.swift
@@ -23,7 +23,7 @@ public struct DeepseekV2Configuration: Codable, Sendable {
     var qkRopeHeadDim: Int = 64
     var vHeadDim: Int = 128
     var qkNopeHeadDim: Int = 128
-    var topkMethod: String = "gready"
+    var topkMethod: String = "greedy"
     var nGroup: Int?
     var topkGroup: Int?
     var numExpertsPerTok: Int?
@@ -83,7 +83,7 @@ public struct DeepseekV2Configuration: Codable, Sendable {
         self.qkRopeHeadDim = try c.decodeIfPresent(Int.self, forKey: .qkRopeHeadDim) ?? 64
         self.vHeadDim = try c.decodeIfPresent(Int.self, forKey: .vHeadDim) ?? 128
         self.qkNopeHeadDim = try c.decodeIfPresent(Int.self, forKey: .qkNopeHeadDim) ?? 128
-        self.topkMethod = try c.decodeIfPresent(String.self, forKey: .topkMethod) ?? "gready"
+        self.topkMethod = try c.decodeIfPresent(String.self, forKey: .topkMethod) ?? "greedy"
         self.nGroup = try c.decodeIfPresent(Int.self, forKey: .nGroup)
         self.topkGroup = try c.decodeIfPresent(Int.self, forKey: .topkGroup)
         self.numExpertsPerTok = try c.decodeIfPresent(Int.self, forKey: .numExpertsPerTok)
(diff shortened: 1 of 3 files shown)


GoodOlClint:
**End-to-end validation (real weights) — caught and fixed a load bug.**

Loading `mlx-community/DeepSeek-V2-Lite-Chat-4bit-mlx` initially failed: V2-Lite sets `q_lora_rank: null` (direct `q_proj`, no low-rank q compression), but the config defaulted a null/absent value to `1536`, taking the `q_a`/`q_b` LoRA path and mismatching the checkpoint's `q_proj`. Fixed by making `qLoraRank` optional (nil ⇒ `q_proj`); added a regression test.

After the fix, forward-pass parity on a fixed token sequence `[1..8]` vs the Python `mlx_lm` reference:

| | argmax | top5(last) |
|---|---|---|
| Python `mlx_lm` | `[185,1,4,5,6,826,4698,575]` | `[575,2,185,4,746]` |
| this PR (Swift) | `[185,1,4,5,6,826,4698,575]` | `[575,2,185,4,746]` |

**Byte-exact match** (MLA + YaRN + group-routed MoE + quantized loading).

Libraries/MLXLLM/Models/DeepseekV2.swift
```diff
@@ -0,0 +1,413 @@
...
+        }
+
+        let queries = concatenated([qNope, qPe], axis: -1)
+
+        let output = attentionWithCacheUpdate(
+            queries: queries, keys: keys, values: values,
+            cache: cache, scale: scale, mask: mask
+        )
+        .transposed(0, 2, 1, 3)
+        .reshaped(B, L, -1)
```
davidkoski: It looks like this updates the KVCache twice per attention call -- probably not desired.

DeepseekV3 has the same issue.

Libraries/MLXLLM/Models/DeepseekV2.swift
```diff
@@ -0,0 +1,413 @@
...
+    var nRoutedExperts: Int?
+    var routedScalingFactor: Float = 1.0
+    var kvLoraRank: Int = 512
+    // Optional: DeepSeek-V2-Lite sets `q_lora_rank: null` (direct q_proj, no
+    // low-rank q compression); full DeepSeek-V2 sets it to 1536. nil ⇒ q_proj.
+    var qLoraRank: Int?
+    var qkRopeHeadDim: Int = 64
+    var vHeadDim: Int = 128
+    var qkNopeHeadDim: Int = 128
+    var topkMethod: String = "gready"
```
davidkoski: typo: greedy

Libraries/MLXLLM/Models/DeepseekV2.swift
```diff
@@ -0,0 +1,413 @@
...
+        self.nSharedExperts = try c.decodeIfPresent(Int.self, forKey: .nSharedExperts)
+        self.nRoutedExperts = try c.decodeIfPresent(Int.self, forKey: .nRoutedExperts)
+        self.routedScalingFactor =
+            try c.decodeIfPresent(Float.self, forKey: .routedScalingFactor) ?? 1.0
+        self.kvLoraRank = try c.decodeIfPresent(Int.self, forKey: .kvLoraRank) ?? 512
+        self.qLoraRank = try c.decodeIfPresent(Int.self, forKey: .qLoraRank)
+        self.qkRopeHeadDim = try c.decodeIfPresent(Int.self, forKey: .qkRopeHeadDim) ?? 64
+        self.vHeadDim = try c.decodeIfPresent(Int.self, forKey: .vHeadDim) ?? 128
+        self.qkNopeHeadDim = try c.decodeIfPresent(Int.self, forKey: .qkNopeHeadDim) ?? 128
+        self.topkMethod = try c.decodeIfPresent(String.self, forKey: .topkMethod) ?? "gready"
```
davidkoski: same typo

davidkoski (changes requested):
Looks good, but take a look at the question on the double KVCache update.  DeepseekV3 would need the same fix, assuming this is an issue.

Libraries/MLXLLM/Models/DeepseekV2.swift
davidkoski: The equivalent python code:

```python
        if cache is not None:
            kv_latent, k_pe = cache.update_and_fetch(kv_latent, k_pe)

        pe_scores = (q_pe * self.scale) @ k_pe.swapaxes(-1, -2)
        if mask is not None:
            pe_scores = mx.where(
                mask,
                pe_scores,
                mx.array(mx.finfo(pe_scores.dtype).min, pe_scores.dtype),
            )

        if L == 1:
            q_nope = self.embed_q(q_nope)
            k = v = kv_latent
        else:
            k = self.embed_q(kv_latent, transpose=False)
            v = self.unembed_out(kv_latent)

        output = scaled_dot_product_attention(
            q_nope, k, v, cache=cache, scale=self.scale, mask=pe_scores
        )
        if L == 1:
            output = self.unembed_out(output)
```

It doesn't have the attentionWithCacheUpdate equivalent (which handles quantization on top of normal updates).

GoodOlClint: Confirmed and fixed in 8e11ca5 — thanks, this was a real bug.

`attentionWithCacheUpdate` performs the update itself, so the explicit `cache.update(...)` beforehand meant the returned history got appended a second time. Removed the manual update and let the helper own the single one, matching Llama/Qwen3/Cohere and your Python excerpt. That also routes the quantized-cache path properly, which the manual update bypassed.

Pinned with a regression test: a 5-token forward pass must leave every layer's cache at offset 5. Before the fix it reported **10** on both layers, so the test fails without the change.

Worth flagging why my earlier end-to-end validation didn't catch this: it ran `model(inputs, cache: nil)`, so neither update fired. The parity numbers were real but only covered prefill without a cache.

On DeepseekV3 — you're right, it has the identical pattern (`DeepseekV3.swift:207` + `:215`). Since that's existing code unrelated to this PR, I'll send it as a separate PR rather than widen this one. Happy to fold it in here instead if you'd prefer.

Libraries/MLXLLM/Models/DeepseekV2.swift
GoodOlClint: Fixed in 8e11ca5. (Carried over from the Python port, which has the same misspelling — it's cosmetic here since the only comparison is against `group_limited_greedy` and real configs set `topk_method` explicitly.)

Libraries/MLXLLM/Models/DeepseekV2.swift
GoodOlClint: Fixed in 8e11ca5. (Carried over from the Python port, which has the same misspelling — it's cosmetic here since the only comparison is against `group_limited_greedy` and real configs set `topk_method` explicitly.)

Libraries/MLXLLM/Models/DeepseekV2.swift
GoodOlClint: DeepseekV3 follow-up is up as #457.

Worth flagging what turned up there: V3 also had `kvHeads` declared as `[Int] = []` and never assigned, so `newCache` returned an empty array and the layer loop's `cache?[i]` trapped with "Index out of range". V3 couldn't generate at all — which is presumably why the double-update never surfaced there. #457 fixes both and adds the first `DeepseekV3Tests`.

davidkoski (approved):
New model looks good, thank you!

davidkoski (approved):
New

PR #381, opened 2026-06-29
Fix silent state truncation for non-multiple-of-32 GDN key dim

GoodOlClint:
## Proposed changes

The fused gated-delta kernel distributes the key head dim `Dk` over exactly 32 simd lanes (`n_per_t = Dk / 32`, integer-truncating). When `Dk` is **not** a multiple of 32 the kernel covers only dims `0..<(32 * n_per_t)` and silently drops the trailing `Dk % 32` state dimensions, producing wrong output with no error.

Stock Qwen3.5 uses `Dk = 192` (a multiple of 32) so this never fires there, but the dimension is config-driven, so a model with a non-multiple-of-32 key dim would hit it.

This guards `gatedDeltaUpdate` so any non-multiple-of-32 `Dk` routes to the ops fallback (`gatedDeltaOps`), which handles an arbitrary key dimension correctly (slower, but never truncating).

## Test

Adds `testGatedDeltaNonMultipleOf32KeyDimUsesAllDims`: builds inputs with `Dk = 48`, perturbs **only** the trailing `Dk % 32` dims (32..<48) of q/k, and asserts the output moves. The truncating kernel ignores those dims (max abs diff `0.0`); the ops fallback uses them. Verified the test fails without the guard and passes with it.

## Checklist

### DIFF
--- a/Libraries/MLXLMCommon/GatedDelta.swift
+++ b/Libraries/MLXLMCommon/GatedDelta.swift
@@ -298,7 +298,14 @@ public func gatedDeltaUpdate(
         state = state.asType(.float32)
     }
 
-    if GatedDeltaKernelManager.shared.kernel != nil {
+    // The fused kernel distributes Dk over exactly 32 simd lanes
+    // (`n_per_t = Dk / 32`, integer-truncating). A key head dim that is not a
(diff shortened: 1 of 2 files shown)


davidkoski (approved):
Change looks good, thank you!

Issue #282, opened 2026-05-12
Gemma 4 model family: three independent loader gaps blocking Apple Silicon adoption


swinxx:
# Gemma 4 model family: three independent loader gaps blocking Apple Silicon adoption

Filing a bundled report on three separate `mlx-swift-lm` gaps that all block Gemma 4 family adoption on a downstream Swift app. The three gaps are independent loader/runtime defects in the same stack layer — listing them together so they can be triaged as a group, since fixing one likely lays an architectural foundation for the others.

## Environment

- `mlx-swift-lm` `3.31.3` (revision `1c05248bb0899e2a7a4962b84d319cf12f4e12aa`)
- macOS 26.0+, Apple Silicon (M-series)
- Swift 6, App-target `SWIFT_STRICT_CONCURRENCY=complete`
- Loading via `LLMModelFactory.shared.loadContainer(from:using:configuration:)` with a `HubApi`-backed `Downloader` + a `Tokenizers.AutoTokenizer` bridge to `MLXLMCommon.Tokenizer`.

---

## Gap 1 — `model_type: gemma4_assistant` not registered (MTP drafter for speculative decoding)

**Symptom:**

```
LLMModelFactory.shared.loadContainer(
    from: downloader,
    using: tokenizerLoader,
    configuration: ModelConfiguration(id: "mlx-community/gemma-4-E4B-it-assistant-bf16")
)
// throws: MLXLLMError.unsupportedModelType("gemma4_assistant")
```

**Repro:** Self-contained Swift Testing probe (~130 LoC) available in our spike phase artifact, runs in ~100 s including a one-time ~182 MB download. Reproduces deterministically against `3.31.3`.

**Root cause:** `gemma4_assistant` is the architecture key Google released alongside Gemma 4 (May 2026) for the MTP (multi-token-prediction) drafter family that powers speculative decoding for E2B/E4B/26B/31B. The `mlx-swift-lm` LLM factory does not register a loader for this key. A code scan of `Libraries/MLXVLM/` and `Libraries/MLXLLM/` confirms only `gemma4` (in `MLXVLM/VLMModelFactory.swift:92`) is registered.

**Why it's not an alias of `gemma4`:** The drafter is a structurally different model — see `mlx-community/gemma-4-E4B-it-assistant-bf16/config.json`:

- `architectures: ["Gemma4AssistantForCausalLM"]`
- `model_type: "gemma4_assistant"`
- `backbone_hidden_size: 2560`
[...]

PR #384, opened 2026-06-30
Fix MLXVLM Gemma4 loader to honor num_kv_shared_layers (E-series targets)

GoodOlClint:
## Proposed changes

Fixes loading of the Gemma 4 **E-series** targets (E2B/E4B), which currently fail
with `keyNotFound … layers.{15,24}.self_attn.k_proj.weight`. This is independent
of MTP — plain loading (`speculative=false`) fails identically.

**Root cause.** The E-series checkpoints use `num_kv_shared_layers > 0`: the last
N decoder layers reuse an earlier layer's K/V and ship no `k_proj`/`v_proj`. They
carry `vision_config`, so they load through `VLMModelFactory` →
`MLXVLM/Models/Gemma4.swift`. That backbone builds every decoder layer with the
default `kvSharedOnly: false`, and `Gemma4TextAttention.init` gated K/V projection
construction on `kvSharedOnly` alone — even though it already computes
`isKVSharedLayer`. So the shared-tail layers declared a `k_proj` the checkpoint
doesn't have, and `loadWeights` threw. The text-only backbone
(`MLXLLM/Models/Gemma4Text.swift`) already gates on the shared-layer predicate;
this was drift between the two copies.

**Fix.** Gate K/V projection construction on `!kvSharedOnly && !isKVSharedLayer`,
matching the text backbone. The runtime already passes `sharedKV` to these layers
and `sanitize` already drops redundant K/V weights for PTQ checkpoints, so both
QAT (omit) and PTQ (ship+drop) layouts load. Dense / 26B-A4B MoE
(`num_kv_shared_layers = 0`) are unaffected.

Relates to #282 (Gemma 4 family loader gaps). Complements the E-series MTP
centroid-embedder PR — that fixes the drafter; this fixes the target it pairs with.

## Checklist

- [x] I have added tests that prove my fix is effective
  - `testGemma4TextSharedLayersDeclareNoKProj` (unit, no checkpoint): a 4-layer
    model with 2 KV-shared layers declares `k_proj` only on the KV-owning layers.
  - `Gemma4EseriesLoadIntegrationTests` (cache-gated): the real E2B/E4B 4bit
    targets load through `VLMModelFactory` and generate. Verified locally — both
    load and produce output (the path that threw pre-fix).
- [ ] Documentation updates — n/a

### DIFF
--- a/IntegrationTesting/IntegrationTestingTests/Gemma4EseriesLoadIntegrationTests.swift
+++ b/IntegrationTesting/IntegrationTestingTests/Gemma4EseriesLoadIntegrationTests.swift
@@ -1,0 +1,64 @@
+// Copyright © 2026 Apple Inc.
+
+import Foundation
+import HuggingFace
+import IntegrationTestHelpers
+import MLXHuggingFace
+import MLXLMCommon
+import MLXVLM
+import Testing
+import Tokenizers
+
+// MARK: - E-series (KV-sharing) target load regression
+//
+// The Gemma 4 "effective" targets (E2B/E4B) use num_kv_shared_layers > 0: the
+// last N decoder layers reuse an earlier layer's K/V and ship no k_proj/v_proj.
+// They carry vision_config so they load through VLMModelFactory. Before the
+// loader fix, the VLM Gemma4 backbone declared k_proj on the shared tail and
+// loadWeights failed `keyNotFound layers.{15,24}.self_attn.k_proj.weight`
+// (independent of MTP — plain load). This exercises the real cached checkpoint
+// end to end: load + a short greedy generate (which also drives the runtime
+// shared-KV forward on those layers).
+//
+// Cache-gated: each test enables only when its checkpoint is already in the HF
+// cache, so it runs locally where the weights exist and disables (does not fail)
+// in CI without them — no env var, no multi-GB auto-download.
+
+private let e4bTarget = "mlx-community/gemma-4-e4b-it-4bit"
+private let e2bTarget = "mlx-community/gemma-4-e2b-it-4bit"
+
+@Suite(.serialized)
+struct Gemma4EseriesLoadIntegrationTests {
+
+    @Test(.enabled(if: hfSnapshotDir(modelId: e4bTarget) != nil))
+    func testE4BTargetLoadsAndGenerates() async throws {
+        try await loadAndGenerate(modelId: e4bTarget)
+    }
+
+    @Test(.enabled(if: hfSnapshotDir(modelId: e2bTarget) != nil))
+    func testE2BTargetLoadsAndGenerates() async throws {
+        try await loadAndGenerate(modelId: e2bTarget)
+    }
+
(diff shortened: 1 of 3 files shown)


mandrael:
Independently hit this exact bug and can confirm the fix.

**Repro:** Loading `mlx-community/gemma-4-E4B-it-qat-4bit` through `VLMModelFactory` fails with `UpdateError.keyNotFound` at `language_model.model.layers.24.self_attn.k_proj.weight`. E4B has `num_hidden_layers=42` / `num_kv_shared_layers=18`, so layers 24–41 are KV-shared and the checkpoint ships no k_proj/v_proj/k_norm for them, yet `Gemma4TextAttention.init` builds those modules for every layer. The PTQ `...-4bit` build fails the same way one step earlier, at `layers.24.self_attn.k_norm.weight`: `sanitize()` (merged #330) correctly drops the redundant tensors, which leaves the wrongly-declared modules unfilled. So #330 alone doesn't cover this — the module construction still needs the gate this PR adds.

**Confirmation:** With this PR applied, `gemma-4-E4B-it-qat-4bit` loads and generates coherent output through `VLMModelFactory`. No regression for non-E-series configs: `isKVSharedLayer` is `false` whenever `num_kv_shared_layers == 0`, so those models build every layer exactly as before.

As a bonus this also unblocks the E-series `gemma4_assistant` MTP drafter end-to-end (paired with #383): with both applied, the E4B target + assistant runs speculative decoding losslessly (byte-identical greedy output) at ~2× tokens/s on an M-series Mac.

Thanks for the fix!

davidkoski (dismissed):
Changes look good but it is failing swift-format -- please rerun that.  Thank you!

GoodOlClint:
Done — rerun and pushed (785f963).

The failing file was actually an unrelated one: `Tests/MLXLMTests/MTPSpeculativeTokenIteratorTests.swift`, not the KV-sharing change. Root cause is a swift-format version skew: the CI `lint` job builds swift-format from `releases/latest` (currently **603.0.0**), whose `OrderedImports` rule groups attributed imports (`@_spi`/`@testable`) into a trailing block, whereas the repo's `pre-commit` hook is `language: system` and picks up whatever's on the contributor's PATH (603 vs my local 602.x disagree on exactly this file). So it had drifted on `main` and `pre-commit run --all` (whole-repo) tripped it on this PR. Reformatted to 603's ordering here.

Heads-up for the maintainers: because CI tracks `releases/latest` while `pre-commit` is unpinned, every swift-format release that changes a rule can red-flag otherwise-untouched PRs until someone reformats. Pinning the CI `swift-format` tag to match a pinned pre-commit `rev` would stop the recurring churn — happy to send that as a separate PR if useful.

davidkoski:
Ugh, trying to get CI to rerun on this.

GoodOlClint:
Claude thinks it sees the issue with the CI, PR incoming.
#404 should hopefully unblock CI, looks like it's been failing across the board for a week.

davidkoski (approved):
Changes look good, thank you!

PR #385, opened 2026-06-30
Set indentConditionalCompilationBlocks to false in .swift-format

ctymoszek:
Aligns the body of #if/#else/#endif blocks with the directive instead of indenting it one level deeper.

## Checklist


davidkoski (approved):
Change looks good, waiting on CI.  Thank you!

PR #386, opened 2026-07-02
ci: pin swift-format version so lint doesn't drift across PRs

GoodOlClint:
## What

- Pin the CI `lint` job's swift-format to a fixed tag (`603.0.0`) instead of resolving `releases/latest` at run time.
- Reformat the one file that the 603.0.0 rules changed (`Tests/MLXLMTests/MTPSpeculativeTokenIteratorTests.swift` — attributed-import grouping).

## Why

The `lint` job builds swift-format from `releases/latest`:

```sh
LATEST_TAG=$(curl -s https://api.github.com/repos/swiftlang/swift-format/releases/latest | ...)
```

So CI silently adopts each new swift-format release. When **603.0.0** shipped, its `OrderedImports` rule began grouping attributed imports (`@_spi` / `@testable`) into a trailing block. That reformatted `MTPSpeculativeTokenIteratorTests.swift` — a file no in-flight PR had touched — and because the step runs `pre-commit run --all` over the **whole repo**, every open PR's lint went red at once with a diff unrelated to its own change (observed on #383 and #384).

Pinning makes lint **reproducible**: the rules only change when someone deliberately bumps the tag (and reformats in the same PR), not whenever upstream cuts a release. `603.0.0` is exactly what `latest` resolves to today, so this is a no-op to the tool version — it just freezes it.

## Note

The `pre-commit` hook is `language: system`, so contributors' local swift-format still isn't locked to CI's version. Matching the pinned tag locally is the reliable way to reproduce lint; if you'd prefer, a follow-up could switch the hook to a pinned swift-format rev so local and CI are identical.

### DIFF
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -30,12 +30,12 @@ jobs:
         id: swift-format
         shell: sh
         run: |
-          cd /tmp
-          LATEST_TAG=$(curl -s https://api.github.com/repos/swiftlang/swift-format/releases/latest | \
-            grep '"tag_name":' | \
-            sed -E 's/.*"([^"]+)".*/\1/')
-          echo "swift-format $LATEST_TAG"
-          echo "SWIFT_FORMAT_VERSION=$LATEST_TAG" >> $GITHUB_OUTPUT
+          # Pinned instead of releases/latest: a new swift-format release can
+          # change formatting rules and reformat files no PR touched, turning
+          # the whole-repo `pre-commit run --all` red on every open PR at once.
+          # Bump this deliberately (with the matching reformat) rather than
+          # tracking latest. 603.0.0 is what `latest` resolved to at pin time.
+          echo "SWIFT_FORMAT_VERSION=603.0.0" >> $GITHUB_OUTPUT
 
       - name: Cache swift-format build
         uses: actions/cache@v4


davidkoski:
I think this is a good idea, but I wonder if we need to do it a different way -- we don't want to dictate what version of swift-format people have installed (e.g. if they have different projects that need different versions).  Maybe we need a more complete configuration file.  `swift-format dump-configuration` can give the full config.  When a new version appears we just need to figure out what the new config option is and add it to match what it used to do.

Thinking about it a different way, I think maybe we need both.  We pin the version AND we specify a full config.  If the version updates and picks up new options we can update the config.  Then people can use whatever version they like as long as it is at or newer than the pin.

aleroot:
> we don't want to dictate what version of swift-format people have installed

Absolutely, agree 💯

davidkoski:
Rebase on main to pick up CI fix.

davidkoski (approved):
Thank you!

davidkoski:
I was playing around with 602.0.0 vs 603.0.0.  There are only two configuration settings that are new:

```
  "indentBlankLines" : false,
  "orderedImports" : {
    "includeConditionalImports" : true
  },
```

Neither of these affect this, so this version of swift-format gives us a hard break that we can only solve via pinning the version.  We can't use newer versions and control them with config.

So is sticking with an old version the right thing?  No easy answer.  Filed:

https://github.com/swiftlang/swift-format/issues/1242

PR #389, opened 2026-07-03
Fix cancellation submission race

### LINKED ISSUE #382: [BUG] Cancelling generation can still submit one more GPU evaluation (iOS/iPadOS crash)
## TL;DR

On iOS/iPadOS, cancelling MLX generation as the app backgrounds can submit one more GPU evaluation *after* the cancel; the system rejects background GPU work and MLX turns that into a process-killing crash. Cause: `generateLoopTask` calls `iterator.next()` — which submits the next token's GPU work via `asyncEval(...)` — *before* checking `Task.isCancelled`. Fix: check cancellation first (`while !Task.isCancelled { guard let token = iterator.next() else { break } }`). Present on the latest `3.31.4` / `main`; fixed and tested on a branch off `3.31.4`.

**Discovered** by backgrounding an iPad/iPhone app mid-generation: the app deactivates, we cancel the generation task, and the one already-primed evaluation lands in the background → `kIOGPUCommandBufferCallbackErrorBackgroundExecutionNotPermitted` → `SIGABRT`.

## Status

- **Present on the latest release `3.31.4` and current `main`**, and the whole 2.x line. Every async generation path is affected — they all route through `generateLoopTask` in `Libraries/MLXLMCommon/Evaluate.swift`.
- **Fixed and tested** on a branch off `3.31.4`: the cancellation-ordering change below, with 4 settlement tests passing under Xcode/Metal.

## The problem

iOS/iPadOS only allow GPU submission while the app is foregrounded. If a generation is cancelled as the app backgrounds, one *extra* GPU evaluation can still be submitted after the cancel — it lands in the background, and MLX's C++ Metal completion handler raises an uncaught `std::runtime_error` that terminates the process:

```text
[METAL] Command buffer execution failed: Insufficient Permission
(to submit GPU work from background)
(00000006:kIOGPUCommandBufferCallbackErrorBackgroundExecutionNotPermitted)
```

## The cause

`generateLoopTask` checks `Task.isCancelled` too late. On `3.31.4` / `main` the loop is:

```swift
tokenLoop: while let token = iterator.next() {          // next() runs first
    if Task.isCancelled { stopReason = .cancelled; break }   // checked only after
    // … process token (enum-returning handler) …
}
```

`TokenIterator.next()` is not a plain iterator — it calls `asyncEval(...)`, submitting the next token's GPU work as a side effect. So `next()` primes one more evaluation *before* the loop sees `Task.isCancelled`. That extra submission is what reaches the background and crashes the process. (The older 2.x `for token in iterator` shape has the identical ordering.)

## The fix

Check cancellation *before* asking for the next token:

```swift
tokenLoop: while !Task.isCancelled {
    guard let token = iterator.next() else { break }
    // … process token …
}
```

The post-loop block already assigns `stopReason = .cancelled` on a cancelled exit, and the existing `Stream().synchronize()` still settles anything already in flight (including the one evaluation primed in `TokenIterator.init`).

**Not a verbatim cross-line patch.** The 2.x fix targeted a `for token in iterator` loop; `3.x` uses a labeled `tokenLoop:` with an enum-returning handler. Same idea, different surrounding code — it must be re-applied to each line's loop shape, not cherry-picked blindly.

**Honest scope.** This does not guarantee *zero* post-cancel submissions — a cancel can still land in the small window between the `while` check and `next()` returning — but it removes the *guaranteed* extra submission. Contract: once the task returned by `generateTask(...)` finishes, no evaluation from that generation remains in flight.

## Environment

- Latest affected: **mlx-swift-lm `3.31.4`** (and current `main`); first seen on `2.31.3`.
- iPadOS 26.5, iPad (M-series); Xcode 26.5, Swift 6 (strict concurrency: complete).
- Model: `mlx-community/Qwen3-4B-Instruct-2507-4bit`. Lifecycle: active → inactive → background.

## Verification

4 settlement tests pass under Xcode/Metal on the `3.31.4` fix branch: settle-after-cancel (immediate cancel and stream-break cancel), `.cancelled` reported as the stop reason, and `.length` (not `.cancelled`) at maxTokens. Note `swift test` cannot run these — it does not stage the Metal `metallib`, so MLX aborts at GPU init; `xcodebuild` stages it and the tests run.

SpiraMira:
## Proposed changes

Fixes #382 

## Checklist

### DIFF
--- a/Libraries/MLXLMCommon/Evaluate.swift
+++ b/Libraries/MLXLMCommon/Evaluate.swift
@@ -1861,12 +1861,15 @@ private func generateLoopTask<Handler: TokenLoopHandler>(
                 tokenizer: tokenizer
             )
 
-            tokenLoop: while let token = iterator.next() {
-                // Check for cancellation on every loop iteration.
-                if Task.isCancelled {
-                    stopReason = .cancelled
-                    break
-                }
+            // Check cancellation BEFORE iterator.next(): next() calls asyncEval() to
+            // pipeline the next GPU evaluation, so checking after it (the previous
+            // `while let token = iterator.next()` form) allowed one extra asyncEval to be
+            // submitted post-cancellation — which faults if the app has backgrounded
+            // (kIOGPUCommandBufferCallbackErrorBackgroundExecutionNotPermitted). The
+            // post-loop block below assigns `.cancelled`; Stream().synchronize() still
+            // settles any in-flight evaluation at the end of the task body.
+            tokenLoop: while !Task.isCancelled {
+                guard let token = iterator.next() else { break }
 
                 if promptTime == 0 {
                     let now = Date.timeIntervalSinceReferenceDate
--- a/Tests/MLXLMTests/CancellationTests.swift
+++ b/Tests/MLXLMTests/CancellationTests.swift
@@ -1,0 +1,134 @@
+// Copyright © 2026 Apple Inc.
+
+import MLX
+import MLXLLM
+import MLXLMCommon
+import MLXNN
+import XCTest
+
+/// Tests for the cancellation-before-next() ordering fix in generateLoopTask.
+///
+/// These tests are in a separate commit so they can be pruned before upstream PR
+/// submission if the project prefers to keep tests closer to a model integration suite.
+final class CancellationTests: XCTestCase {
+
+    private func makeTinyModel() -> LlamaModel {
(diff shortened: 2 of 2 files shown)


davidkoski:
Rebase on current main.

davidkoski (approved):
Change looks good, thank you!

PR #392, opened 2026-07-04
Gemma 4: native audio input (port the gemma4_audio USM-Conformer encoder)

fdagostino:
# Gemma 4: native audio (+ video) input for the `Gemma4` VLM

**Draft — builds on #390 (`fix/gemma4-vlm-kv-shared-construct`), which the E4B checkpoint needs to load.**

Wires **native audio** (and video) into the `Gemma4` VLM class (`model_type: "gemma4"`,
e.g. `mlx-community/gemma-4-e4b-it-4bit`), previously text + vision only. The audio
encoder (`audio_tower`, a USM-Conformer) is ported from `Blaizzy/mlx-vlm`
`models/gemma4/audio.py` and **cross-checked against three independent implementations**
that all agree: mlx-vlm (Python), `VincentGourbin/gemma-4-swift-mlx` (Swift), and the
**merged** llama.cpp PR #21421 (`tools/mtmd/models/gemma4a.cpp`, C++).

## Changes

**`Libraries/MLXVLM/Models/Gemma4Audio.swift`** (new) — the `audio_tower`:
SSCP subsample projection (2× `Conv2d` 3×3 stride-2 → flatten(F·C) → `Linear`), macaron
`ConformerFeedForward` (0.5 residual), chunked-local `AudioAttention` (Transformer-XL
relative shift, base-2 `qScale`/`kScale`, `per_dim_scale` softplus, tanh softcap 50,
float32 math), depthwise-causal `LightConv1d`, ×12 `ConformerBlock`, `output_proj`.

**`Libraries/MLXVLM/Models/Gemma4AudioFeatures.swift`** (new) — `Gemma4AudioFeatureExtractor`:
faithful MLX port of HF `Gemma4AudioFeatureExtractor` (raw 16 kHz waveform → log-mel
`[1,T,128]` + validity mask). Periodic Hann, HTK mel (128 bins, 0–8 kHz), `rfft` (n=512),
magnitude spectrum, `log(mel + 1e-3)`, semicausal pad. `rfft` is re-exported by `MLX`
(no new package dependency).

**`Libraries/MLXVLM/Models/Gemma4.swift`:**
- `Gemma4Configuration` decodes `audio_config`, `video_token_id`, `boa`/`eoa`,
  `audio_soft_tokens_per_image`, `audio_ms_per_token`.
- `Gemma4` builds `audio_tower`/`embed_audio` when `audio_config` is present; runs the
[...]

PR #395, opened 2026-07-05
Add regression coverage for mixed-precision (per-layer) quantized checkpoint loading

fdagostino:
## What

Test-only PR: adds `MixedPrecisionQuantLoadTests`, end-to-end regression coverage for loading checkpoints that mix quantization precisions via per-module overrides in `config.json` — the layout every mlx-community **Gemma 4 QAT** upload uses (`gemma-4-12B-it-qat-4bit`, `gemma-4-E4B-it-qat-4bit`: 4-bit attention/embeddings, 8-bit `mlp.{gate,up,down}_proj`, declared as 144/126 per-module entries interleaved with the global quantization dict).

## Why

Reports such as ollama/ollama#16740 attribute empty/garbage generation from these checkpoints to loaders reading the 8-bit MLP payload as 4-bit. While investigating exactly that symptom on device, I verified that **mlx-swift-lm handles this correctly today**: `QuantizationContainer` decodes the interleaved dict, `PerLayerQuantization` resolves each module path, and `loadWeights` applies the per-module bits (the same semantics as Python mlx-lm's `class_predicate`). Verified on macOS against the real `mlx-community/gemma-4-12B-it-qat-4bit`: after load the module tree has exactly 144 × `QuantizedLinear(bits: 8)` for the MLP projections, 4-bit everywhere else, and greedy generation is correct.

However, nothing in `Tests/` exercises this path end to end (only the config-decode unit tests in `BaseConfigurationTests`), so the behavior the whole Gemma 4 QAT family depends on was unpinned. This PR pins it.

## The tests

Built on a tiny text-only `gemma4_unified` model whose own initial weights are quantized into a synthetic on-disk checkpoint with the QAT layout:

1. **Per-module quantization overrides are applied when loading** — decodes a QAT-style `quantization` dict (global 4-bit + per-module 8-bit MLP overrides), runs `loadWeights`, and asserts each MLP projection comes back as `QuantizedLinear(bits: 8, groupSize: 64)`, everything else quantizable at 4-bit, with an 8-bit dequantization round-trip check.
2. **Ignoring the per-module overrides fails loudly, not silently** — loading the same checkpoint with only the global 4-bit settings must throw (packed-shape mismatch) rather than silently mis-decode the 8-bit payload into garbage logits — the failure mode described in the ollama issue must never regress into silence.

Validated to bite: replacing the per-layer lookup in `loadWeights` with the global default makes test 1 fail with `mismatchedSize` on `layers.0.mlp.gate_proj` (`[128, 8]` vs `[128, 16]`).

## Related

Follows the Gemma 4 series #390 (KV-shared construction), #391 (video), #392 (audio encoder).

### DIFF
--- a/Tests/MLXLMTests/MixedPrecisionQuantLoadTests.swift
+++ b/Tests/MLXLMTests/MixedPrecisionQuantLoadTests.swift
@@ -1,0 +1,206 @@
+// Copyright © 2026 Apple Inc.
+
+import Foundation
+import MLX
+import MLXLMCommon
+import MLXNN
+import Testing
+
+@testable import MLXVLM
+
+/// End-to-end regression coverage for mixed-precision quantized checkpoints.
+///
+/// The mlx-community Gemma 4 QAT uploads (e.g. `gemma-4-12B-it-qat-4bit`,
+/// `gemma-4-E4B-it-qat-4bit`) quantize attention and embeddings at 4-bit but
+/// every `mlp.{gate,up,down}_proj` at 8-bit, declared in `config.json` as a
+/// global `quantization` dict interleaved with per-module overrides:
+///
+/// ```json
+/// "quantization": {
+///     "group_size": 64, "bits": 4, "mode": "affine",
+///     "language_model.model.layers.0.mlp.gate_proj": {"group_size": 64, "bits": 8},
+///     ...
+/// }
+/// ```
+///
+/// `loadWeights` must quantize each module with its per-module bits (matching
+/// Python mlx-lm's `class_predicate`), and a load that ignores the overrides
+/// must fail loudly on the packed-shape mismatch rather than silently
+/// mis-decoding 8-bit payloads as 4-bit (see ollama/ollama#16740 for the
+/// confusion these checkpoints caused across loaders).
+struct MixedPrecisionQuantLoadTests {
+
+    private static let groupSize = 64
+
+    private func tinyUnifiedConfig() throws -> Gemma4UnifiedConfiguration {
+        // Text-only Gemma4 Unified sized so every quantizable input dimension
+        // is divisible by the quantization group size.
+        try JSONDecoder.json5().decode(
+            Gemma4UnifiedConfiguration.self,
+            from: Data(
+                """
+                {
+                  "model_type": "gemma4_unified",
+                  "vocab_size": 64,
+                  "image_token_id": 31,
+                  "audio_token_id": 30,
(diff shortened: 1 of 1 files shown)


Tests/MLXLMTests/MixedPrecisionQuantLoadTests.swift
```diff
@@ -0,0 +1,222 @@
...
+            let module = modules[path] as? QuantizedLinear
+        {
+            let loaded = dequantized(
+                module.weight, scales: module.scales, biases: module.biases,
+                groupSize: Self.groupSize, bits: 8)
+            let expected = dequantized(
+                arrays["\(path).weight"]!,
+                scales: arrays["\(path).scales"]!,
+                biases: arrays["\(path).biases"],
+                groupSize: Self.groupSize, bits: 8)
```
davidkoski: I think `module.weight` and `arrays["\(path).weight"]` are the same value -- `makeCheckpoint()` creates the arrays, `writeCheckpoint` writes them to files, and `loadWeights` loads them back in.  The values that are being dequantized are the same, assuming I am reading that right.

The title of the test:

> Per-module quantization overrides are applied when loading

suggests that the desire is to verify this code:

```swift
    // quantize if needed
    if quantization != nil || perLayerQuantization != nil {
        quantize(model: model) { path, module in
            if weights["\(path).scales"] != nil {
                if let perLayerQuantization {
                    return perLayerQuantization.quantization(layer: path)?.asTuple
                } else {
                    return quantization?.asTuple
                }
            } else {
                return nil
            }
        }
    }
```

from `loadWeights`.  I wonder if the loops above exercise that sufficiently?

davidkoski (changes requested):
Change looks good, but see my question.  It might be better off without that part, but see what you think.

fdagostino:
Good catch — you are right, that assertion was vacuous. `module.weight` and `arrays["<path>.weight"]` are the same tensor round-tripped through `writeCheckpoint`/`loadWeights`, so all it verified was safetensors save/load fidelity, not the per-module override path.

Dropped it. The override behavior is already pinned by the loops above (`bits == 8` on the mlp projections, `bits == 4` everywhere else), and those do bite: replacing the per-layer lookup in `loadWeights` with the global default fails the first test with `mismatchedSize` on `layers.0.mlp.gate_proj` ([128, 8] vs [128, 16]). Also removed the matching claim from the commit message.

davidkoski (approved):
Changes look good, thank you!

PR #398, opened 2026-07-05
Qwen3VL: default per-image resolution to a 1,280 vision-token budget

### LINKED ISSUE #396: Qwen3.5/Qwen3-VL: uncapped per-image resolution lets the ViT allocate tens-of-GB attention buffers
## Problem

The qwen3_5 vision tower runs **global O(patches²) attention** — one SDPA over `P = total patches` with a `[1, P, P]` mask and no window structure. Nothing in the processor caps the per-image patch count for model configs whose pixel budgets never clamp: several published checkpoints (e.g. Qwen3.6-27B variants) ship a `size.longest_edge` around 16 MP, so a single full-resolution image produces enough patches that the `[16 heads, P, P]` bf16 score matrix alone is ~29 GB — on top of an ~18 GB quantized model.

Measured on an M-series Mac with a 27B Qwen3.6 checkpoint: one full-screen screenshot in the prompt → **47 GB peak memory and 121 s prefill**. With a per-image vision-token cap: **~21 GB peak**, prefill time proportionally down. (The language-model side is already chunked; the blowup is entirely the ViT.)

Multi-image prompts make it worse: patches are concatenated and attended jointly (block-diagonal mask, still one `[16, ΣP, ΣP]` allocation).

## Proposal

Default each image (and each video frame) to a **vision-token budget** — `min(config.maxPixels, N * factor²)` with N ≈ 2,560, mirroring the Qwen2.5-VL clamp — and honor an explicit `processing.maxPixels` override so callers who want full resolution can still opt in. Only what the tower sees is downscaled; stored image bytes are untouched.

Whether N=2,560 is the right *default* is negotiable — the important part is that a budget exists and is configurable. Happy to adjust the shape of the knob to whatever fits the library's conventions.

PR with an implementation incoming.

spokvulcan:
Fixes #396.

The qwen3_5 ViT runs global O(patches²) attention with no window structure, so an uncapped full-resolution image (several published configs ship a ~16 MP longest_edge that never clamps) can allocate a tens-of-GB attention score matrix. Measured on a 27B Qwen3.6 checkpoint: 47 GB peak / 121 s prefill for one full-screen screenshot; ~21 GB with a 2,560-token cap, lower still at the current default.

- Default each image — and each video frame — to a 1,280 vision-token budget: `min(config.maxPixels, 1280 * factor²)`, mirroring the Qwen2.5-VL clamp.
- Honor an explicit `processing.maxPixels` override, so full resolution stays available as an opt-in.
- Only what the tower sees is downscaled; stored image bytes are untouched.
- The budget multiply is guarded with `multipliedReportingOverflow` so a pathological patch/merge size degrades to the config ceiling instead of trapping, and the image and video preprocess paths clamp through the same `config.maxPixels` / `config.minPixels` accessors so the two clamp sites cannot drift.


spokvulcan:
Lint fixed.

Libraries/MLXVLM/Models/Qwen3VL.swift
```diff
@@ -33,6 +33,19 @@ public struct Qwen3VLProcessor: UserInputProcessor {
             .normalized(mean: config.imageMeanTuple, std: config.imageStdTuple)
     }
 
+    /// Pixels for the default 2,560 vision-token budget (`2560 * factor²`),
+    /// clamped to `ceiling`. Overflow-safe: a pathological config (an absurd
+    /// `patch_size`/`merge_size`) yields `ceiling` instead of trapping on the
+    /// unchecked multiply, so bad model metadata degrades to the config ceiling
+    /// rather than crashing preprocess (tesseract ADR-0014).
```
davidkoski: Question on the 2560 value -- is this too large?  It looks like the mlx-vlm side uses 1280:

processing_qwen3_vl.py

```python
def _smart_resize_image(
...
    max_pixels: int = 14 * 14 * 4 * 1280,
```

as does the 25VL in swift:

Qwen25VL.swift

```swift
        // Qwen2.5-VL ships a very large max_pixels; default to the budget the
        // model card recommends (1280 * 28 * 28), override-able via `processing`.
        let maxPixels = processing?.maxPixels ?? min(config.size.maxPixels, 1280 * factor * factor)
```

The code looks good but I am curious about the number.

spokvulcan: You're right, and it turns out 1280 is even a touch generous. The mlx-vlm constant (`14 * 14 * 4 * 1280`) carries patch-14 geometry from the qwen2 lineage, while qwen3_vl is patch 16, so on this model it works out to roughly 1.0 MP, about a 980-token budget. Qwen25VL.swift is the same pixel count (`1280 * 28²`). `1280 * factor²` here lands at 1.31 MP, slightly above both precedents, with the same expression as the 25VL clamp.

Before changing it I ran the halved budget against real workloads in our app. Compared with 2560 the extractions come out the same. The one caveat I found is on a 5 MP Hacker News screenshot, where two usernames in the small gray byline text came back off by one glyph (bilsbie read as bilskie, SirOibaf as SirOlbaf). Titles, domains, points, timestamps and comment counts all survived intact. That seems like the right trade for a default, since `processing.maxPixels` stays available when someone needs full fidelity on large dense images.

Default is now 1280 (637e9f2), title and description updated.

<details>
<summary>Evidence: 5 MP dense-text screenshot read at the 1280 default</summary>

Source image:



Model transcription:




</details>

<details>
<summary>Evidence: 1.7 MP marketing page read at the 1280 default</summary>




</details>

davidkoski (approved):
Looks good, thank you!

NivDvir:
1,280 keeps showing up across these defaults for a reason: it's the resolution range the model was tuned in. The cap isn't really a compute limit — it's a distribution limit. Go far past it and the extra pixels aren't detail the model can use; position statistics drift out of the range it knows, and box geometry goes first.

I measured this on Qwen2.5-VL-7B-4bit over ScreenSpot-Pro screenshots. Large targets (>5% of the frame, n=36) peak exactly at 1,280 — 0.643 IoU, down to 0.526 by 16,384. More budget makes the common case worse, not better.

Your off-by-one glyphs sit on the other curve. Small targets (0.1–1%) keep gaining up to ~4,096 (0.111 → 0.179), and tiny text is still climbing at 16,384. So the default lands right, and ~4,096 is a sensible override for dense screenshots.



Full size-binned table + method: https://dev.to/nivdvir/maxpixels-is-a-token-budget-in-disguise-and-the-right-cap-depends-on-the-size-of-what-youre-l0c

davidkoski:
@NivDvir great info thank you!

PR #399, opened 2026-07-05
Qwen3.5/3.6: windowed prefill + state-threaded warm continuation (fixes multi-turn M-RoPE drift)

spokvulcan:
### Proposed changes

Three related changes, reshaped by the review (thanks @davidkoski, both
questions turned out to be the right ones):

1. **Windowed prefill for Qwen3.5/3.6** (the item deferred from #344).
   `prepare` drives the language model in chunks of `windowSize ?? 512`,
   same default and shape as Gemma3 and `LLMModel`. This bounds the
   full-attention scratch to `[heads, window, L]` instead of
   `[heads, L, L]`. The vision tower runs once; chunks slice the merged
   embeddings. Batched inputs keep the single-shot path.

2. **`prepare` now takes `state: LMOutput.State?`**, and the
   `WindowedVisionContinuation` protocol from earlier revisions is
   removed. `prepareContinuation` is private to `Qwen35` now.

   This fixes a live bug: `ChatSession` reuses the KV cache across turns,
   but the M-RoPE state died with each turn's `TokenIterator`, so Qwen3.5
   recomputed positions from zero on every warm continuation. On a tiny
   random-weight model the warm turn-2 logits diverge from a cold full
   prefill by 0.43 max-abs, against a 8.3e-07 decode-path noise floor.

   `TokenIterator` takes an optional seed state and exposes the
   post-prefill state, `ChatSession` carries it across turns and
   tool-call restarts, and every other model ignores the new parameter
   (one line each). Speculative/MTP iterators are untouched. Qwen2.5-VL
   and Qwen3-VL have the same cross-turn drop, filed as #420.

   Known limitation, out of scope here: `savePromptCache` doesn't persist
   `LMOutput.State`, so a disk-restored image prefix still needs the
   caller to hold the state. Happy to follow up on the file format if
   there's interest.

3. **Processor config: `size.{longest_edge, shortest_edge}`** decode
   support, unchanged from earlier revisions (newer Qwen3-VL configs only
   ship the new-style keys).

### Tests

`Qwen35ContinuationTests`: tiny random-weight model, in-process, no
downloads (pattern per `NemotronHTests`):

- warm text continuation matches full prefill (the drift regression, red
  before the fix)
- warm image continuation matches full prefill (the rope delta carry)
- three-turn resume: text, then an image arriving mid-continuation, then
  text again
- windowed matches single-shot, for text and for an image straddling a
  window boundary

Plus `Qwen3VLProcessorConfigTests` for (3). Full MLXLMTests suite green
locally.

### Checklist

- [x] I have updated the documentation accordingly (API doc comments + MLXVLM README example)

### DIFF
--- a/Libraries/MLXLMCommon/LanguageModel.swift
+++ b/Libraries/MLXLMCommon/LanguageModel.swift
@@ -218,7 +218,7 @@ public struct LMOutput {
     }
 }
 
-/// The result of the call to ``LanguageModel/prepare(_:cache:windowSize:)``
+/// The result of the call to ``LanguageModel/prepare(_:cache:state:windowSize:)``
 public enum PrepareResult {
     /// tokens to process by the ``TokenIterator``
     case tokens(LMInput.Text)
@@ -227,48 +227,31 @@ public enum PrepareResult {
     case logits(LMOutput)
 }
 
-/// Opt-in capability for a vision language model that can continue an already
-/// warmed KV cache *through* a new image, without recomputing the cached
-/// prefix and without a single-shot `[heads, L, L]` full-attention scratch.
-///
-/// `prepareContinuation` runs the vision tower and the image→token merge
-/// once, then drives the language-model forward in windows of `windowSize`,
-/// bounding the scratch to `[heads, windowSize, L]`. A cold cache is the
-/// degenerate case (anchor 0), so a conforming model's `prepare` can delegate
-/// its `windowSize` chunking here. Only families that have wired the windowed
-/// image path conform; callers feature-detect via
-/// `as? WindowedVisionContinuation`.
-public protocol WindowedVisionContinuation {
-    /// Continue `cache` (warmed to its current offset `P`) through `input`'s
-    /// image-bearing remainder, in windows of `windowSize`.
-    ///
-    /// - Parameter state: carries the Position Anchor's rope delta for the
-    ///   images cached before `P` (absent / zero for an image-free prefix). The
-    ///   new image's M-RoPE positions are computed from that anchor instead of
-    ///   resetting to zero.
-    /// - Returns: `.logits` whose `state` carries the rope delta the post-image
-    ///   text tail resumes with, for a caller threading state end-to-end.
-    func prepareContinuation(
-        _ input: LMInput, cache: [KVCache], state: LMOutput.State?, windowSize: Int
-    ) throws -> PrepareResult
-}
-
 /// Interface for all Language Models (e.g. LLM, VLM).
 ///
 /// The language model is typically called by the ``TokenIterator`` and it:
 ///
 /// - consumes the ``LMInput``
-/// - calls ``prepare(_:cache:windowSize:)`` to initialize the KVCache and consume the prompt
+/// - calls ``prepare(_:cache:state:windowSize:)`` to initialize the KVCache and consume the prompt
 /// - calls ``callAsFunction(_:cache:state:)-9kuvf`` for each token, producing an ``LMOutput``
 /// - the ``TokenIterator`` accumulates this information into a ``GenerateResult``
 public protocol LanguageModel: BaseLanguageModel {
 
     /// Prepare the cache state and consume the ``LMInput``.
     ///
+    /// `state` is the ``LMOutput/state`` a caller carried over from earlier
+    /// evaluation against the same `cache` — present when `cache` is already
+    /// warm (a multi-turn chat, a tool-call restart, a restored prompt
+    /// cache). Models that keep per-call positional state (e.g. the M-RoPE
+    /// `ropeDeltas` of the Qwen VLM families) use it to anchor the new
+    /// tokens' positions at the cache offset; models without such state can
+    /// ignore it. In the typical cold call it is `nil`.
+    ///
     /// This can return:
     /// - ``PrepareResult/tokens(_:)`` if the caller should evaluate the (remaining) tokens normally
     /// - ``PrepareResult/logits(_:)`` to produce the next token from the prompt
-    func prepare(_ input: LMInput, cache: [KVCache], windowSize: Int?) throws -> PrepareResult
+    func prepare(_ input: LMInput, cache: [KVCache], state: LMOutput.State?, windowSize: Int?)
+        throws -> PrepareResult
 
     /// Primary entry point to produce a step (single token) from the model
     func callAsFunction(_ input: LMInput.Text, cache: [KVCache]?, state: LMOutput.State?)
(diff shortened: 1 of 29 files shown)


Libraries/MLXVLM/Models/Qwen35.swift
```diff
@@ -1043,6 +1071,113 @@ public class Qwen35: Module, VLMModel {
...
+            eval(typedCache)
+        }
+
+        // Seed the post-image text tail's anchor. The vendor's flat-continuation
+        // branch positions tail token j at `tailCacheOffset + ropeDeltas + j`;
+        // after this remainder `tailCacheOffset = P + remainderLength`, so the
+        // delta the tail needs is the offset-frame `getRopeIndex` delta minus
+        // `P` (which `getRopeIndex` implicitly counted into `remainderLength`).
+        var resumeState = LMOutput.State()
+        resumeState[ropeDeltasKey] = ropeDeltas - MLXArray(Int32(cacheOffset))
```
davidkoski, line 1187: I think this needs a change in Evaluate.swift:

```swift
    mutating func prepare(input: LMInput, windowSize: Int? = nil) throws {
...
        switch try model.prepare(input, cache: cache, windowSize: windowSize) {
...
        case .logits(let result):
            y = .init(tokens: convertToToken(logits: result.logits))
            asyncEval(y.tokens)
```

This doesn't consume the `state` and probably needs to.  It looks like state should be writable there.

Libraries/MLXLMCommon/LanguageModel.swift
```diff
@@ -227,6 +227,32 @@ public enum PrepareResult {
...
+/// prefix and without a single-shot `[heads, L, L]` full-attention scratch.
+///
+/// `prepareContinuation` runs the vision tower and the image→token merge
+/// once, then drives the language-model forward in windows of `windowSize`,
+/// bounding the scratch to `[heads, windowSize, L]`. A cold cache is the
+/// degenerate case (anchor 0), so a conforming model's `prepare` can delegate
+/// its `windowSize` chunking here. Only families that have wired the windowed
+/// image path conform; callers feature-detect via
+/// `as? WindowedVisionContinuation`.
+public protocol WindowedVisionContinuation {
```
davidkoski: It looks like this is just called through from the normal prepare path -- do we need to protocol?

Libraries/MLXLMCommon/LanguageModel.swift
```diff
@@ -227,6 +227,32 @@ public enum PrepareResult {
...
+    /// image-bearing remainder, in windows of `windowSize`.
+    ///
+    /// - Parameter state: carries the Position Anchor's rope delta for the
+    ///   images cached before `P` (absent / zero for an image-free prefix). The
+    ///   new image's M-RoPE positions are computed from that anchor instead of
+    ///   resetting to zero.
+    /// - Returns: `.logits` whose `state` carries the rope delta the post-image
+    ///   text tail resumes with, for a caller threading state end-to-end.
+    func prepareContinuation(
+        _ input: LMInput, cache: [KVCache], state: LMOutput.State?, windowSize: Int
```
davidkoski: The current prepare path doesn't take `state` -- should it?  I am unsure as the prepare is typically used to run the prompt through but I guess is also used to restore session state.  In a typical call there _shouldn't_ be previous state.  Can there be for this case?

davidkoski (changes requested):
Looks good but see question about discarding state during prepare

Libraries/MLXVLM/Models/Qwen35.swift
davidkoski: See #419 -- exactly this fix!

NivDvir:
Since #419 landed, the `resumeState` rope-delta carry on the `.logits` path threads through cleanly — the remaining soft spot looks like test coverage for the new windowed cold-prefill branch: it produces first-token logits from the chunked forward rather than the single-shot path, and a 1-position M-RoPE drift there wouldn't move loss, but at temp 0 it can silently flip a dense grounding prompt's bbox JSON into an empty code fence — a case short-prompt tests miss.

Might be worth asserting the windowed and single-shot paths produce identical first-token logits on a long dense-image prompt; happy to run that comparison on a Qwen2.5/3-VL grounding set if useful.

Libraries/MLXVLM/Models/Qwen35.swift
spokvulcan: Merged main to pick up #419, so the prefill state now flows through end to end. The new `Qwen35ContinuationTests` lock the behavior in. Thanks for connecting the dots to that fix.

Libraries/MLXLMCommon/LanguageModel.swift
spokvulcan: You were right, the protocol is gone. `prepareContinuation` is now a private detail of `Qwen35`, and the capability rides the standard entry point instead: `prepare` now takes a `state:` parameter. Details on the sibling thread, which asked exactly that question.

Libraries/MLXLMCommon/LanguageModel.swift
spokvulcan: It turned out to be a live bug, not a hypothetical. `ChatSession` keeps the KV cache warm across turns, but each turn builds a fresh `TokenIterator`, so the M-RoPE state dies with the previous turn. On the next turn the model sees a warm cache with no rope deltas and recomputes positions from zero.

Measured on a tiny random-weight Qwen35 (now in `Qwen35ContinuationTests`): a warm turn-2 prefill diverges from a cold full prefill by 0.43 max-abs in next-token logits. The decode path, which does carry state, sits at 8.3e-07 on the same comparison. That gap is the position restart, not numerics.

So `prepare` now takes `state:`, the same `LMOutput.State` that `callAsFunction` already takes. Other models just ignore the parameter for now (one line each), `TokenIterator` accepts a seed state and exposes the post-prefill state, and `ChatSession` carries it across turns. Passing nil behaves exactly as before.

Qwen2.5-VL and Qwen3-VL drop state across turns the same way. Filed #420 for those so this PR stays reviewable.

spokvulcan:
Both halves of your comment are now covered in CI on a tiny random-weight model (`Qwen35ContinuationTests`, no downloads): windowed vs single-shot logits for plain text and for an image straddling a window boundary, plus warm-continuation vs full-prefill checks for the state carry your #419 completed. A grounding run on real weights would still be very welcome, especially for the models #420 covers, which this PR intentionally doesn't touch.

davidkoski:
Lint is failing due to conflicts on the swift-format version.  (see the top level README for details)

Just scanning through this now.  I can make the swift-format change in a few minutes.

davidkoski:
> Known limitation, out of scope here: savePromptCache doesn't persist LMOutput.State, so a disk-restored image prefix still needs the caller to hold the state. Happy to follow up on the file format if there's interest.

An interesting point -- probably worth an issue at least.

davidkoski (approved):
Changes look good, just waiting on CI.

PR #401, opened 2026-07-05
Add custom LogitProcessor injection to GenerateParameters / ChatSession

xocialize:
## Motivation

`TokenIterator` already exposes a direct `init(input:model:cache:processor:sampler:prefillStepSize:maxTokens:)` for callers who want to supply their own ``LogitProcessor``. But `GenerateParameters.processor()` only ever produces the built-in penalty processors, and `ChatSession` (like the other high-level paths) drives generation exclusively through the parameters-based initializers — so a custom processor (e.g. grammar-constrained / structured decoding) cannot ride `ChatSession` at all.

In practice this forces downstream packages to bypass `ChatSession` for structured-output requests and re-implement the prompt-formatting + iterator plumbing just to inject one processor. We carry exactly that bypass in two production packages and would like to retire it.

## API

```swift
var parameters = GenerateParameters(temperature: 0)
parameters.customLogitProcessorFactory = { GrammarProcessor(grammar: grammar) }

// rides ChatSession (and every other parameters-driven path) unchanged
let session = ChatSession(model, generateParameters: parameters)
```

- `GenerateParameters.customLogitProcessorFactory: (@Sendable () -> any LogitProcessor)?` — new stored property (plus a defaulted trailing `init` parameter).
- `GenerateParameters.processor()` composes the factory's product **after** the built-in `PenaltyProcessor`.
- `ChainedLogitProcessor` — small public processor that applies multiple `LogitProcessor`s in order; used for the composition and available for users combining several custom processors.

## Design notes

- **Factory, not a stored processor.** `LogitProcessor` is stateful (`didSample`), so each generation needs a fresh instance — important for multi-turn `ChatSession` where the same `GenerateParameters` drives many generations. A `@Sendable` closure also keeps `GenerateParameters: Sendable` intact, which a stored `any LogitProcessor` would not.
- **No plumbing changes needed.** `TokenIterator`, `SpeculativeTokenIterator`, and `MTPSpeculativeTokenIterator` all build their processor via `parameters.processor()`, so `ChatSession`, `generate(...)`, and the speculative paths pick this up for free.
- **Custom-after-penalty ordering** matches Python mlx-lm, where user `logits_processors` append after the ones from `make_logits_processors`.
- Purely additive — behavior is unchanged when the factory is `nil` (`processor()` returns exactly what it did before).

## Testing

New tests, all passing via `xcodebuild test -scheme mlx-swift-lm-Package`:

- `SampleTests.testGenerateParametersCustomLogitProcessorFactory` — nil/custom-only/chained composition of `processor()`.
- `SampleTests.testGenerateParametersAppliesCustomProcessorAfterPenalties` — numeric check that penalties apply before the custom processor.
- `SampleTests.testChainedLogitProcessorForwardsPromptAndDidSample` — `prompt`/`didSample` forwarding through the chain.
- `ChatSessionTests.testChatSessionUsesCustomLogitProcessor` — end-to-end: a masking processor on a tiny randomly-initialized Gemma3 (existing test pattern, no downloads) forces every generated token.


Libraries/MLXLMCommon/Evaluate.swift
```diff
@@ -132,7 +152,8 @@ public struct GenerateParameters: Sendable {
         frequencyPenalty: Float? = nil,
         frequencyContextSize: Int = 20,
         prefillStepSize: Int = 512,
-        seed: UInt64? = nil
+        seed: UInt64? = nil,
+        customLogitProcessorFactory: (@Sendable () -> any LogitProcessor)? = nil
```
aleroot: This constructor got telescopic, but more than that I think that placing a behavioral factory closure inside a pure data configuration struct introduces a leak. GenerateParameters is a value type meant strictly for static generation metrics (temperature, tokens, etc.). A LogitProcessor controls execution behavior. Mixing the two is not a great idea...
GenerateParameters should remain strictly focused on configuration parameters. A logitProcessorProvider should be introduced in the ChatSession, I believe...

aleroot: Even better would be having something like this in the futue: 

```
public struct GenerationComponents: Sendable {
    public var logitProcessorProvider: (@Sendable () -> any LogitProcessor)?
    public var stoppingCriteriaProvider: (@Sendable () -> any StoppingCriteria)?
    public var onTokenSampled: (@Sendable (Int) -> Void)?
    }
```

Then the session: 

```
public class ChatSession {
    private let model: any LanguageModel
    private let parameters: GenerateParameters
    private let components: GenerationComponents
public init(
        _ model: any LanguageModel,
        parameters: GenerateParameters = .init(),
        components: GenerationComponents = .init()
    ) {
        self.model = model
        self.parameters = parameters
        self.components = components
    }
    }
```

xocialize: Good point, and I agree `GenerationComponents` is the cleaner long-term shape — behavioral concerns (logit processor, stopping criteria, per-token callback) do belong together and separate from static config.

The reason it currently lives on `GenerateParameters` is that `processor()` is the single composition point every generation path builds from — `TokenIterator`, `SpeculativeTokenIterator`, `MTPSpeculativeTokenIterator`, and therefore `generate(...)`, `ChatSession`, and the speculative paths all pick it up with zero plumbing. That's specifically the bypass we're trying to retire downstream: a custom (e.g. grammar-constrained) processor can currently only be injected via the direct `TokenIterator.init(...processor:...)`, so structured-output callers have to re-implement the prompt-formatting + iterator wiring instead of riding `ChatSession`.

A `GenerationComponents` on `ChatSession` is nicer conceptually, but as sketched it only reaches `ChatSession` — `generate(...)` and the speculative iterators still wouldn't see the processor, since they compose off `parameters.processor()`. To keep parity we'd need to thread `components` through those iterator inits too, which is a larger, more invasive change.

(One small note on the "leak" framing: the factory is `@Sendable () -> any LogitProcessor`, so it keeps `GenerateParameters: Sendable` intact — the concern is separation of concerns rather than a Sendability break.)

I'm happy to go the `GenerationComponents` route if that's the preferred direction — but since the additive version here is fully behavior-preserving (`processor()` is unchanged when the factory is `nil`) and unblocks the downstream bypass today, could we keep this PR minimal and track `GenerationComponents` (with the iterator-path threading) as a follow-up refactor? Glad to take whichever scope the maintainers prefer.

aleroot: `parameters.processor()` is currently the common composition point for the parameter-driven generation paths, and I understand why adding the factory there gives broad coverage with a small diff. My concern is that this makes an implementation convenience part of the long-term API contract.

This is ultimately a call for @davidkoski  to make but I think the right fix is to introduce the separate concept now, even if the first version only carries a logit processor factory:

```swift
public struct GenerationComponents: Sendable {
    public var logitProcessorFactory: (@Sendable () -> any LogitProcessor)?

    public init(
        logitProcessorFactory: (@Sendable () -> any LogitProcessor)? = nil
    ) {
        self.logitProcessorFactory = logitProcessorFactory
    }
}
```

Then thread that through the generation entry points that need parity:

```swift
TokenIterator(..., parameters: parameters, components: components)
SpeculativeTokenIterator(..., parameters: parameters, components: components)
MTPSpeculativeTokenIterator(..., parameters: parameters, components: components)
ChatSession(..., generateParameters: parameters, generationComponents: components)
```

The composition can still happen in one place, just not by storing behavioral state inside `GenerateParameters`:

```swift
let processor = compose(
    parameters.processor(),
    components.logitProcessorFactory?()
)
```

I know this is more work than the current patch, but I think it is the better direction for future maintainability and API clarity. If we add the factory to GenerateParameters now, we are accepting that shape as public API; moving it later would become a breaking change.

Sorry to be an intruder into your pull request, but this looks like an API-impacting change, and GenerateParameters has already grown quite a bit.
Refactoring now would be something easier than later, once you change the public API then it becomes a breaking change ...

xocialize: That's a fair point and it changes my mind — threading `components` alongside `parameters` into the iterator/session entry points gives the same broad coverage without storing behavioral state in `GenerateParameters`, so my "single composition point" concern is fully addressed by your sketch. And the breaking-change argument is the decisive one: once the factory ships on `GenerateParameters` it's public API, and moving it later is a source break, whereas the split is essentially free right now. No intrusion at all — this is exactly the kind of API-shape call worth making before release.

So I'm on board with `GenerationComponents`. Before I rework the PR I'd just like a steer from @davidkoski on the shape, since it touches public API on several entry points:

- naming (`GenerationComponents`, `components:` / `generationComponents:` labels)
- whether the free `generate(...)` functions should also gain a `components:` parameter, or whether it's scoped to the iterators + `ChatSession`
- default-arg placement on the existing entry points

Happy to do the full threading (`TokenIterator` / `SpeculativeTokenIterator` / `MTPSpeculativeTokenIterator` / `ChatSession`, with `compose(parameters.processor(), components.logitProcessorFactory?())` as the one composition point) once the preferred shape is confirmed — I'd rather build it once to spec than guess. @davidkoski, any preference on the above?

davidkoski: Just catching up with this.  My take is that `GenerateParameters` _is_ the factory for `LogitProcessor` and `LogitSampler`:

```swift
public struct GenerateParameters: Sendable {

    public func sampler() -> LogitSampler {
    public func processor() -> LogitProcessor? {
```

It is a convenient way to get pre-configured samplers and processors.  If you wanted _custom_ sampler/processor you could use this init on `TokenIterator`:

```swift
public struct TokenIterator: TokenIteratorProtocol {

    public init(
        input: LMInput, model: any LanguageModel, cache: [KVCache]? = nil,
        processor: LogitProcessor?, sampler: LogitSampler, prefillStepSize: Int = 512,
        maxTokens: Int? = nil
    ) throws {
```

So I see:

- GenerateParameters is not extensible
- a custom TokenIterator is really only usable from the _lowest_ level API (direct calls to `generate()`)

My inclination would be to _replace_ `GenerateParameters` with `GenerationComponents` -- make `GenerationComponents` be the protocol and have `GenerateParameters` implement it.  I don't think we need both.

That would mean `GenerationComponents` either has properties that describe `maxTokens` etc. or has a function/property that produces a struct that contains it. **OR** we need something on the token generation (GenerationComponents) and a second type for the token _iteration_ (maxTokens, etc.).

That said, this may make the API on ChatSession and others a bit dirty.  It could either be like this:

```swift
    public init(
        _ model: ModelContainer,
        instructions: String? = nil,
        speculativeDecoding: SpeculativeDecodingConfig? = nil,
        generateParameters: GenerationComponents = ...,
```

where `generateParameters` matches with `GenerationComponents` or we would have yet another init.

This might be the point where we add `ChatParameters` and attach various properties to it.  Then we could stop with the proliferation of `init()` on `ChatSession`.

Perhaps the next thing to think of is whether `ChatSession`'s properties need to be `var`.  If they were all `let` it could potentially be `Sendable` or a `struct`.

The last point is perhaps out of scope for this discussion but the second to last might not be.

Thoughts on this?

I guess one other thing to think about:  how does the `TokenLoopHandler` fit into this?  Do they need to be related?  Should the Handler be (perhaps optionally) produced by GenerationComponents?  Conceptually they seem to be related as they both control how the tokens are generated.

xocialize: Thanks @davidkoski — this all makes sense, and I agree with the framing: `GenerateParameters` is the *factory* for pre-configured processor/sampler, and the only real gap is reach — a custom processor can only be injected through the low-level `TokenIterator.init`, so it can't ride `ChatSession` (or the speculative paths). Making `GenerationComponents` the first-class home for behavioral concerns is the right move, and I think `TokenLoopHandler` belongs there too — it's "how tokens are generated," same as a custom processor/sampler, so having `GenerationComponents` optionally vend it fits cleanly.

On the one fork I'd offer an opinion: I'd lean **compose rather than make `GenerateParameters` a protocol**. Turning a concrete, widely-used value type into a protocol is a source break for anyone constructing or copying it, pushes callers toward `any GenerationComponents` existentials on a hot path, and gives up the value semantics that make `GenerateParameters` easy to pass around and mutate. Concretely:

- `GenerateParameters` stays the value type for *generation config* (temperature, penalties, seed) — it already knows how to vend the built-in processor/sampler.
- `GenerationComponents` is a small `Sendable` struct for *behavior*: `logitProcessorFactory`, `samplerFactory`, `stoppingCriteria`, and (optionally) a `TokenLoopHandler` factory. Factories, not instances, so each generation gets a fresh stateful object and the struct stays `Sendable`.
- The composition point stays single: `compose(parameters.processor(), components.logitProcessorFactory?())`, threaded alongside `parameters` into `TokenIterator` / the speculative iterators / `ChatSession`.

That's non-breaking and keeps "both," but as *composition* rather than a type hierarchy — which I think reads cleaner than a protocol for value types. You're right that `maxTokens`/`prefillStepSize` are really *iteration* knobs rather than *generation*; I'd keep them where they are for now and treat a generation-vs-iteration split (and `ChatParameters` to tame the `ChatSession` init proliferation, and the `let`/`Sendable`-`ChatSession` idea) as follow-ups so this PR can land minimal and additive.

If you'd rather go the protocol-replacement route despite the above, I'm happy to build that instead — I just didn't want to guess. We opened this, so we're glad to do the work either way: tell me which shape you want (compose vs protocol-replace, and whether `TokenLoopHandler` should be in scope for v1) and I'll rework the PR to match.

davidkoski: It may be fine if maxTokens and friends are in the same place with the rest -- they do seem separate, but if we scope this to "token generation" then they would clearly fit in (I think).

So are you thinking that `GenerateParameters` could produce a `GenerationComponents`?

I agree that source breaking wouldn't be a good path.  GenerateParameters should not become a protocol, but it might _implement_ the new protocol.  I don't think existentials are too much a problem here: dynamic method dispatch to produce these values is insignificant compared to the metal kernel dispatch to actually produce a single token.  That said, simple structs as you said may be the way to go.

I think we need _something_ that can be passed down through `ChatSession` and `generate()` that can encapsulate all these pieces.  We need a reasonable default implementation for temperature (like GenerateParameters).  This could either produce the value or be the value.

I think folding TokenLoopHandler in would be nice but it would make it generic as it has an associated type.  That might complicate things and make it necessary to keep them separate.  Unless they need to be coupled to actually work.

I am not tied to any specific implementation -- protocols were what came to mind, but they are not the only way.  I agree we can do the ChatSession work as a followup.

I think my ideas are still a bit hand-wavy.  I can imagine the general shape but I think it needs to be tried to see how it all fits.  I think we need some code to look at to decide.  If you want to try your approach, go for it!  If I get a chance I will try playing around a bit with this too and can post notes or a draft PR so we can see the diffs.

xocialize: Sounds good — I'll take a first cut so we have a concrete diff to react to rather than staying hand-wavy. Plan for v1, matching what we landed on:

- Keep `GenerateParameters` concrete/non-breaking (it stays the default and still vends `.processor()`/`.sampler()`).
- Add a small `GenerationComponents: Sendable` struct of behavioral factories — `logitProcessorFactory` to start, with room for `samplerFactory`/`stoppingCriteria` — defaulting to empty so behavior is identical when unused.
- Thread an optional `components:` through `TokenIterator` / the two speculative iterators / `ChatSession`, with a single composition point `compose(parameters.processor(), components.logitProcessorFactory?())`.
- Leave `TokenLoopHandler` out of v1 (the associated type makes it generic, as you noted) and treat the `ChatSession` init cleanup / `ChatParameters` as the follow-up.

I'll push a commit to the PR branch so you can see how it actually fits. Please still poke at it / post notes or a draft if you get the chance — happy to converge on whichever diff reads best.

aleroot: Have you completed, the changes ? 
I would need to add the thinking budget, for example: 

```
public struct ThinkingBudgetConfiguration: Sendable, Equatable {
    public let maximumTokenCount: Int
    public let closingPrefix: String

    public init(
        maximumTokenCount: Int,
        closingPrefix: String = "\n"
    ) throws
}
```

and then plug that in like : 

```
var components = GenerationComponents()

components = try components.applyingThinkingBudget(
    ThinkingBudgetConfiguration(maximumTokenCount: 512),
    reasoning: reasoningConfig,
    tokenizer: tokenizer
)

let session = ChatSession(
    model,
    generateParameters: parameters,
    components: components
)
```
If you can complete  and merge this pull request then I will open mine. Thanks. 
@davidkoski what do you think of the Thinking budget ?

xocialize:
@aleroot Yes — the rework is complete and pushed (2026-07-14). The branch is up to date with `main` and mergeable; there's nothing outstanding on my side.

What's on it now, per the direction you and @davidkoski landed on:

- `GenerationComponents: Sendable` — behavioral hooks kept out of `GenerateParameters`, which is back to being a pure value type (its `processor()` is penalty-only again, as before this PR).
- A single composition point, `components.logitProcessor(parameters:)`, composing penalty-then-custom.
- `components:` threaded (additive + defaulted, so non-breaking) through `TokenIterator`, `SpeculativeTokenIterator`, `MTPSpeculativeTokenIterator`, all six `ChatSession` inits, and the non-deprecated `generate(...)` free functions.
- Tests covering the ChatSession path end-to-end, per-generation factory freshness (stateful processors must not leak across turns), and an empty-components regression guard.

It's waiting on two maintainer-side things, not on code: @davidkoski's call on the broader shape (he raised possibly folding this into a protocol and/or adding `ChatParameters`, and mentioned he might prototype it), and CI approval — the run has been sitting at `action_required` since the push, since fork PRs need a maintainer to start the workflow.

On the thinking budget: that's a good fit, and your snippet already reads against the current API exactly as-is (`GenerationComponents()` → `ChatSession(model, generateParameters:, components:)`). Worth noting v1 deliberately ships only `logitProcessorFactory` — but I think a thinking budget lands cleanly on that alone: forcing `closingPrefix` once the thinking-token budget is spent is logit processing, and the per-generation factory semantics give you a fresh counter each turn, which is exactly what you'd want for a budget in a multi-turn session. `applyingThinkingBudget` returning a modified copy also fits the value-type design nicely.

If a budget ends up wanting a stopping-criteria or sampler hook too, the struct was built to grow — those were left out of v1 only to keep the first change minimal, and your use case is a reasonable argument for adding them.

aleroot:
@xocialize ok, when this gets merged, I will then prepare the work for the thinking budget. thanks .

Libraries/MLXLMCommon/GenerationComponents.swift
```diff
@@ -0,0 +1,71 @@
...
+/// Every field is optional and defaulted, so an empty `GenerationComponents()`
+/// reproduces the default behavior and can be threaded through any entry point
+/// without changing existing results.
+///
+/// ```swift
+/// var components = GenerationComponents()
+/// components.logitProcessorFactory = { GrammarProcessor(grammar: grammar) }
+/// let session = ChatSession(model, components: components)
+/// ```
+public struct GenerationComponents: Sendable {
```
davidkoski, line 26: This isn't really new to this PR, but https://github.com/ml-explore/mlx-swift-lm/issues/508 -- it makes it easy :-)

Libraries/MLXLMCommon/GenerationComponents.swift
```diff
@@ -0,0 +1,71 @@
...
+    /// otherwise structured decoding -- ride any parameters-driven API,
+    /// including ``ChatSession`` and the parameters-based ``TokenIterator``
+    /// initializers.
+    ///
+    /// This is a factory rather than a stored ``LogitProcessor`` because
+    /// processors are stateful (see ``LogitProcessor/didSample(token:)``): it is
+    /// invoked once per generation so each generation gets a fresh instance and
+    /// state cannot leak across generations. A `@Sendable` closure also keeps
+    /// `GenerationComponents` `Sendable`.
+    public var logitProcessorFactory: (@Sendable () -> any LogitProcessor)?
```
davidkoski, line 41: We may want the same for `LogitSampler` but now we have the mechanism to do it.

davidkoski (approved):
Looks good, thank you!

Libraries/MLXLMCommon/Evaluate.swift
davidkoski: ThinkingBudget sounds interesting -- so it is a LogitProcessor that constrains the tokens to finish thinking as it gets to the token count?

Does this need a specific `applyingThinkingBudget()`?  Ideally`GenerationComponents` should be able to compose these generically.  I guess we can answer these questions in a followup PR!

PR #402, opened 2026-07-06
Fix Gemma4 VLM load of QAT checkpoints: wire kvSharedOnly through the backbone

freeman412:
`Gemma4TextAttention` / `Gemma4TextDecoderLayer` accept `kvSharedOnly` to skip building `k_proj`/`v_proj`/`k_norm`/`v_norm` on KV-shared layers — #330's sanitize drop relies on those module slots not existing — but `Gemma4TextBackbone` constructs every layer with the default (`false`). Checkpoints that omit the redundant tensors (e.g. `mlx-community/gemma-4-e2b-it-4bit`) fail to load:

```
keyNotFound(path: ["language_model", "model", "layers", "15", "self_attn", "k_proj", "weight"], …)
```

This passes `kvSharedOnly` for layers `>= hiddenLayers - numKVSharedLayers`, the same boundary the attention/MLP/cache-map already use.

Verified with `gemma-4-e2b-it-4bit` (35 layers, 20 shared): loads, generates text, and describes images correctly via the VLM processor.

### DIFF
--- a/Libraries/MLXVLM/Models/Gemma4.swift
+++ b/Libraries/MLXVLM/Models/Gemma4.swift
@@ -1112,8 +1112,11 @@ final class Gemma4TextBackbone: Module {
 
         self._embedTokens.wrappedValue = Embedding(
             embeddingCount: config.vocabularySize, dimensions: config.hiddenSize)
+        let firstKVSharedLayer = config.hiddenLayers - config.numKVSharedLayers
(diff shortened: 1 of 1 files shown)


davidkoski (approved):
Changes look good but it fails on swift-format.  I think this may need a rebase, see #386

davidkoski:
Looks like a dup of #409 and #390.  I am leaning toward 390 as it also has a test.

davidkoski:
Going with #390

PR #403, opened 2026-07-06
Fix Qwen3.5 VLM garbage output: gate the RMSNorm weight shift in sanitize

GoodOlClint:
## Proposed changes

`MLXVLM/Models/Qwen35.sanitize` applies the RMSNorm `+1` weight offset **unconditionally**, whereas `MLXLLM/Models/Qwen35.sanitize` gates it behind `hasMTPWeights || hasUnsanitizedConv1d`.

On an already-converted MLX checkpoint (conv1d weights already sanitized to a trailing dim of 1, no MTP tensors), the norm weights have already been shifted at conversion time. The VLM path shifts them a **second** time, doubling every `input_layernorm` / `post_attention_layernorm` / `q_norm` / `k_norm` / `model.norm`. That decoheres the model and produces garbage tokens.

The same weights loaded through the `MLXLLM` `Qwen35` path (which gates the shift) decode correctly — which isolates the divergence to `sanitize`, not the model math or the shared gated-delta kernel. Because Qwen3.5 is natively multimodal, every Qwen3.5 checkpoint routes through `VLMModelFactory` → the affected path, so all Qwen3.5 generation (text and multimodal) is broken on pre-converted checkpoints.

### Fix

Mirror the `MLXLLM` gate in the VLM `sanitize`: compute `shouldShiftNormWeights = hasMTPWeights || hasUnsanitizedConv1d` from the incoming weights and apply the `+1` only when true. This makes the VLM path reproduce the `MLXLLM` numerics.

### Reproduction

- Load a pre-converted Qwen3.5 VLM checkpoint via `VLMModelFactory` and greedily decode a trivial prompt → garbage before, coherent after.
- The same weights with `vision_config` removed route to `LLMModelFactory` and were already coherent (the control).

### Tests

Adds two `Qwen35SanitizeTests` cases: a pre-converted checkpoint (conv1d already sanitized) must **not** shift the norm weights; a raw checkpoint (unsanitized conv1d) must. Verified end-to-end on the previously-failing checkpoint (garbage → coherent).

```
xcodebuild test -scheme mlx-swift-lm-Package -destination 'platform=macOS' \
  -only-testing:MLXLMTests/Qwen35SanitizeTests
```

## Checklist

- [ ] I have updated the necessary documentation (not needed)

### DIFF
--- a/Libraries/MLXVLM/Models/Qwen35.swift
+++ b/Libraries/MLXVLM/Models/Qwen35.swift
@@ -1071,6 +1071,19 @@ public class Qwen35: Module, VLMModel {
     }
 
     public func sanitize(weights: [String: MLXArray]) -> [String: MLXArray] {
+        // Whether the checkpoint stores RMSNorm weights in the raw (un-shifted)
+        // convention that needs the `+1` offset. Mirrors the MLXLLM Qwen35 gate
+        // so the VLM and text paths agree: a pre-converted MLX checkpoint
+        // (conv1d already sanitized, no MTP tensors) has already been shifted at
+        // conversion time and must NOT be shifted again — doing so double-shifts
+        // every layernorm and produces garbage tokens. Computed on the incoming
+        // weights, before the `mtp.` filter below.
+        let hasMTPWeights = weights.keys.contains { $0.contains("mtp.") }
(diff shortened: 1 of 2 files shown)


davidkoski (approved):
Changes look good, thank you!

PR #405, opened 2026-07-07
Gemma 4: aspect-preserving image resize with per-image dynamic token counts; fix multi-image requests

prbme:
## What

Ports the Python mlx-vlm `Gemma4ImageProcessor` resize behavior to the Swift Gemma 4 port, replacing the fixed-size square preprocessing, and makes multi-image requests work.

- **Aspect-preserving resize.** Each image is resized to the largest dimensions that keep the patch count within `max_soft_tokens × pooling_kernel_size²` while leaving both sides divisible by `pooling_kernel_size × patch_size`. The pooling kernel is then exact for every aspect ratio and the pooled grid covers the image fully. Degenerate extreme-aspect inputs (e.g. an 8000×20 panorama) clamp to one pooling cell on the short side, capped at the token budget on the long side.
- **Per-image dynamic soft-token counts.** Prompt placeholders expand per image to `numPatches / kernel²` soft tokens instead of a fixed `image_seq_length`. The processor decodes `max_soft_tokens` / `patch_size` / `pooling_kernel_size` from either a flat `preprocessor_config.json` or the `image_processor` object nested in `processor_config.json`.
- **Multi-image requests.** Images keep their own sizes through `LMInput.ProcessedImage`: smaller images are zero-padded onto the request's largest canvas, real sizes ride in `frames`, and the model slices each image back out and runs the vision tower on it alone, concatenating pooled tokens in placeholder order. Previously any request with more than one image threw `imageTokenCountMismatch` (`N × 280` placeholders vs `dim(1) == 280`); mixed portrait+landscape requests crashed on concat. Both work now.
- **Padded-path machinery removed.** With every patch real, the padded/masked path (`-1` positions, encoder attention mask, pooler valid-count trimming) is deleted; the encoder runs dense attention and the pooler pools the full grid. This also removes `gemma4VisionPoolingKernel` from #290 — the kernel no longer needs to be recovered from a padded length because the resize guarantees `numPatches = outputLength × kernel²` exactly. The #290 regression tests are replaced by `Gemma4ResizeTests`, which pin the resize invariants that make that bug class impossible (side alignment, patch budget, exact pooling coverage, aspect preservation, extreme-aspect clamps).
- The vestigial `size` config entry is now ignored, matching the Python reference (models ship 224×224 there, which was never used by the Python processor).

The `gemma4_unified` path (`Gemma4Unified`, its embedder and processor) is untouched.

## Why

The fixed-size square resize distorts every non-square image (a phone screenshot is squashed ~2:1) and diverges from the reference implementation in Python mlx-vlm / HF transformers, which resize aspect-preserving with dynamic token counts. And multi-image prompts — which the chat template and processor placeholder expansion already support — currently throw before generation starts.

## Verification

- `xcodebuild test -scheme mlx-swift-lm-Package` passes (the pre-existing `Gemma4ChunkedPrefillTests/chunkSizeInvariance` failure reproduces identically on current `main` without this change).
- Runtime-verified against `mlx-community/gemma-4-26b-a4b-it-4bit`: single portrait, single landscape, and portrait+landscape in one message all produce correct spatial color descriptions (half-and-half color test images; the model names both halves correctly, including the bottom half).
- Target-size math cross-checked against the Python `mlx-vlm` formula over 100k random sizes with zero mismatches.


Libraries/MLXVLM/Models/Gemma4.swift
```diff
@@ -1892,32 +1865,25 @@ private final class Gemma4VisionModel: Module {
...
+        for y in 0 ..< patchesH {
+            for x in 0 ..< patchesW {
+                values.append(Int32(x))
+                values.append(Int32(y))
             }
         }
-
-        let count = realCount + paddedCount
-        return (MLXArray(values, [batch, count, 2]), realCount)
+        let positions = MLXArray(values, [1, patchesH * patchesW, 2])
```
davidkoski: This could potentially be done with `meshGrid`.  Something like this:

```swift
let xs = MLXArray.arange(0, patchesW)
let ys = MLXArray.arange(0, patchesH)
let (gx, gy) = meshGrid([xs, ys])
let positions = stacked([gx.flattened(), gy.flattened], axis: 1).reshaped(1, -1, 2)
```

That all runs on the GPU and the larger array is never realized on the CPU

Libraries/MLXVLM/Models/Gemma4.swift
```diff
@@ -1792,53 +1778,45 @@ private final class Gemma4VisionPatchEmbedder: Module {
...
+        if numPatches <= length {
+            return hiddenStates * scale
+        }
+
+        // All batch rows share one position grid, so a single [l, L] weight
+        // matrix pools every row. The processor's resize keeps both sides
+        // divisible by kernel * patchSize, so the pooled grid covers the
+        // image exactly.
+        let positions = patchPositions[0]
+        let maxX = Int(positions[0..., 0].max().item(Int32.self)) + 1
```
davidkoski: The 0 column in positions is `0 ..< patchesW`, so `maxX = patchesW - 1 + 1 = patchesW`

I think this is only done once per image, but avoiding item/eval in inference is a key to good performance.

(I note this is the same code as before, just different indentation, but since we are in here anyway)

davidkoski (changes requested):
See what you think about the two items I suggested. This looks good!

davidkoski:
Also, please rebase -- the lint issue is #416 (and others).  You may not have been following that saga -- the short version is that it will pin the version of swift-format to 602.0.0, which is what we have all been using.  603 came out recently but we can't configure it to behave like 602, so waiting for 604.

Libraries/MLXVLM/Models/Gemma4.swift
prbme: Done in `8028f43` — used your snippet almost verbatim. Two small deltas: `meshGrid` returns `[MLXArray]` rather than a tuple, so it's `grids[0]`/`grids[1]`, and I passed `dtype: .int32` to `arange` to keep the positions integral. Left the default `indexing: .xy` (x varies fastest), which matches the row-major patch order the embedder and pooler expect — added a comment noting that, since correctness depends on it.

Libraries/MLXVLM/Models/Gemma4.swift
prbme: Done in `8028f43` — `maxX` is gone. The pooler now takes `patchesW: Int` from the caller and uses it directly. You're right on both counts: it was pre-existing and my diff only reindented it, and the arithmetic checks out — x runs `0 ..< patchesW`, so `max + 1 == patchesW` exactly. No `.item()` left in the vision path.

prbme:
Rebased onto `main` (`4ca25fd`) and pushed — `8028f43`. Both review comments addressed.

On the rebased branch: builds clean, `Gemma4ResizeTests` passes (5 tests, incl. 1×1 / 20×8000 / 4032×3024), and `swift-format lint --strict` is clean on the touched files.

On the lint saga — looks like #404 landed 603 conformance rather than pinning 602, so `main` pins `603.0.0` now. Lint's clean here at that version.

davidkoski (approved):
Change looks good, thank you!

PR #407, opened 2026-07-08
Fix Gemma4 VLM weight loading: skip K/V modules for KV-shared layers

kingster:
## Problem

Loading a Gemma 4 VLM checkpoint that uses cross-layer KV sharing fails at weight loading:

```
MLXNN.UpdateError.keyNotFound(path: ["language_model", "model", "layers", "15", "self_attn", "k_norm", "weight"],
  modules: ["Gemma4", "Gemma4TextLanguageModel", "Gemma4TextBackbone", "Gemma4TextDecoderLayer",
            "Gemma4TextAttention", "Gemma4RMSNormZeroShift"])
```

Reproduced with `mlx-community/gemma-4-e2b-it-4bit` (`num_hidden_layers = 35`, `num_kv_shared_layers = 20`) via `VLMModelFactory.shared.loadContainer(from:using:)`. Present on `main` and the latest tag (3.31.4).

## Root cause

KV-shared layers (`layerIdx >= hiddenLayers - numKVSharedLayers`) reuse an earlier layer's K/V and ship **no** `k_proj`/`v_proj`/`k_norm`/`v_norm` weights. But `Gemma4Attention.init` gates building those modules only on `kvSharedOnly` — the speculative-decoding drafter flag — so the main model builds them for KV-shared layers. `loadWeights` then throws `keyNotFound` for the first shared layer's `k_norm`.

This is internally inconsistent with the rest of the model, which already assumes KV-shared layers have no such modules:

- `Gemma4TextBackbone` feeds KV-shared layers `sharedKV`, so `k_proj`/`k_norm`/`v_proj`/`v_norm` are never used there.
- `sanitize(weights:)` drops any `self_attn.{k,v}_proj` / `k_norm` for KV-shared text layers, with the comment: *"they reuse an earlier layer's K/V and own no K projection or K norm, so the module tree has none."*

The construction was the one place that didn't match that assumption.

## Fix

Gate construction of the K/V projections and norms on `!isKVSharedLayer` as well:

```swift
if !kvSharedOnly && !isKVSharedLayer {
    // k_proj / v_proj / k_norm / v_norm
}
```

`isKVSharedLayer` is already computed in `init` just above. One-line change; the forward path and `sanitize` already handle these layers.

## Testing

With this change, `mlx-community/gemma-4-e2b-it-4bit` loads and generates (vision + text) on an M4 iPad. Non-KV-shared checkpoints are unaffected (`isKVSharedLayer` is `false` when `numKVSharedLayers == 0`).

### DIFF
--- a/Libraries/MLXVLM/Models/Gemma4.swift
+++ b/Libraries/MLXVLM/Models/Gemma4.swift
@@ -824,7 +824,11 @@ final class Gemma4TextAttention: Module {
         self.isKVSharedLayer = layerIdx >= firstKVSharedLayer && firstKVSharedLayer > 0
 
         self._qProj.wrappedValue = Linear(config.hiddenSize, numHeads * headDim, bias: false)
-        if !kvSharedOnly {
+        // KV-shared layers reuse an earlier layer's K/V and own no K/V projections or norms
+        // (see `sanitize`, which drops these, and `Gemma4TextBackbone` which feeds them
+        // `sharedKV`). Skip building them here too, otherwise weight loading throws
+        // `keyNotFound` for e.g. `language_model.model.layers.15.self_attn.k_norm.weight`.
+        if !kvSharedOnly && !isKVSharedLayer {
             self._kProj.wrappedValue = Linear(
                 config.hiddenSize, numKVHeads * headDim, bias: false)
             if !useKEqV {
--- a/Tests/MLXLMTests/Gemma4KVSharedLoadTests.swift
+++ b/Tests/MLXLMTests/Gemma4KVSharedLoadTests.swift
@@ -1,0 +1,129 @@
+// Copyright © 2026 Apple Inc.
+
+import Foundation
+import MLX
+import MLXLMCommon
+import MLXNN
+import Testing
+
+@testable import MLXVLM
+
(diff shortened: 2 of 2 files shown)


davidkoski:
Looks like a dup of #402 and #390.  I am leaning toward 390 as it also has a test, but I like the approach in this one.

kingster:
Thanks for flagging the dups (#402, #390).  

I've now ported the regression tests from #390 (credit to @fdagostino) so this PR carries them too. Verified locally via `xcodebuild test` on macOS:

- **without** the change: both tests fail — `sanitizedCheckpointLoadsWithoutKeyNotFound` reproduces `keyNotFound(… self_attn.k_norm.weight …)`;
- **with** the change: both pass.

The tests assert the module tree omits the K/V projections for KV-shared tail layers (while head layers keep them) and that a sanitized checkpoint loads through the strict `verify: [.all]` loader — so they pin the construction side to `sanitize`/forward regardless of which side implements the gate.

davidkoski:
After reading through these carefully I agree both (407, 390) fix the issue.  I am going with 390 because it matches the structure from the python code in mlx-vlm (which can be useful).

Thank you for this work!

PR #409, opened 2026-07-09
Fix tool round-trip: record assistant tool_calls message before tool results

jacobwillemsma:
## Proposed changes

When `ChatSession` dispatches tool calls, it appends only the role:`tool` result messages before restarting generation. The assistant message that *made* the calls is never recorded. Chat templates that render tool results by forward-scanning from an assistant message carrying `tool_calls` — Gemma 4's template among them — therefore drop the results from the rendered prompt entirely (standalone role:`tool` messages are skipped by such templates). The restarted generation answers as if the tool returned nothing.

Observed in the wild with `mlx-community/gemma-4-e2b-it-4bit` and gemma-4-31b conversions: with a `read_url` tool whose dispatch verifiably ran and returned page text, the model consistently replied "I need the actual content of the page…". With this change it answers from the tool result.

The fix records the assistant turn before appending results, matching the transcript shape already asserted by `ToolCallIdTests` ("Assistant tool calls and tool results form a correlated transcript"):

```swift
messages.append(.assistant("", toolCalls: pendingToolCalls))
```

Templates that render role:`tool` standalone are unaffected apart from an empty-content assistant turn carrying the calls, which mirrors what OpenAI-style transcripts contain anyway.

**Test:** `ChatSessionToolRoundTripTests` drives a real `ChatSession` round trip with a scripted tokenizer (first generation pass decodes a JSON tool call, later passes plain text) and a recording `MessageGenerator`, then asserts the restarted render pairs the tool result with an assistant `tool_calls` message. On unmodified `main` it fails with "tool result was rendered with no preceding assistant tool_calls message"; with the fix it passes.

Local test run: `MLXLMTests` passes 173/173 XCTest cases. `Gemma4ChunkedPrefillTests.chunkSizeInvariance` and `Gemma4UnifiedTests.textOnlyPrefillChunkSizeInvariant` (swift-testing) fail identically on unmodified `main` on this machine (M5 Pro, Xcode 26.6) — pre-existing and unrelated; possibly connected to what #404 addresses. Tests run via `xcodebuild … -skipPackagePluginValidation` per the same issue.

## Checklist

- [x] I have run `swift-format` on the changed files (`pre-commit`-equivalent, per CONTRIBUTING)
- [x] I have updated the necessary documentation (no documentation changes needed)

### DIFF
--- a/Libraries/MLXLMCommon/ChatSession.swift
+++ b/Libraries/MLXLMCommon/ChatSession.swift
@@ -788,6 +788,7 @@ public final class ChatSession {
                         if let toolDispatch, !pendingToolCalls.isEmpty,
                             !Task.isCancelled
                         {
+                            messages.append(.assistant("", toolCalls: pendingToolCalls))
                             for toolCall in pendingToolCalls {
                                 let toolResult = try await toolDispatch(toolCall)
                                 messages.append(.tool(toolResult, id: toolCall.id))
--- a/Tests/MLXLMTests/ChatSessionToolRoundTripTests.swift
+++ b/Tests/MLXLMTests/ChatSessionToolRoundTripTests.swift
@@ -1,0 +1,198 @@
+// Copyright © 2026 Apple Inc.
+
+import Foundation
+import MLX
+import MLXLLM
+import MLXNN
+import XCTest
+
+@testable import MLXLMCommon
+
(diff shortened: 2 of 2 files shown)


davidkoski (approved):
Makes sense, thank you for the change!

PR #411, opened 2026-07-09
Qwen3VL: apply the sRGB tone curve in the image preprocess path

### LINKED ISSUE #410: Qwen3VL image preprocessing skips the sRGB tone curve — model receives linear-light values, dark content becomes unreadable
### The bug

`Qwen3VLProcessor.preprocess` (image path) feeds the model **linear-light** pixel values instead of the gamma-encoded sRGB values the HF reference preprocessing uses. The practical effect: near-black content — e.g. dark-theme UI text — reaches the ViT with ~12× less contrast than the reference implementation and the model simply cannot see it, while bright content is barely affected (which is why this survives casual testing on bright images).

### Root cause

CoreImage filters operate in a linear-light working space, and `MediaProcessing.asMLXArray` renders with `colorSpace: nil`, i.e. raw working-space values. Every other Qwen image path compensates with the linear→sRGB tone-curve step before handing values to the model:

- `Qwen25VL` image path: applies `MediaProcessing.inSRGBToneCurveSpace`
- `Qwen2VL` image path: same
- `Qwen3VL` **video** path (`preprocess(image:resizedSize:)`): applies `.toSRGB()`
- `Qwen3VL` **image** path: **missing** — `Libraries/MLXVLM/Models/Qwen3VL.swift` resamples and normalizes without it

So the sRGB byte value 23 (a faint dark-theme headline) becomes linear 0.0078 instead of 23/255 = 0.0902, and its 6/255 background becomes 0.0018: after mean/std 0.5 normalization the signal between text and background collapses from 0.133 (reference) to 0.011.

### Measured

Replicating the exact pipeline (bicubic resample → mean/std normalize → `.RGBAf` render with nil colorSpace) on a real dark-theme screenshot whose second headline line sits at luminance 23/255 on a 6/255 background:

| pipeline | text-vs-background signal (normalized units) |
| --- | --- |
| current image path (no tone curve) | **0.011** |
| with `inSRGBToneCurveSpace` before resample | **0.126** |
| HF reference (byte values, mean/std 0.5) | 0.133 |

### Reproduction / evidence

Send a dark screenshot with faint secondary text to Qwen3-VL through MLXVLM and through any reference Qwen deployment; the reference reads the faint text, MLXVLM does not. Before/after from an app that vendors this library (same model, same prompt, same image — only the tone-curve step differs):

**Input** — note the faint second headline line "Your machine.":

Dark landing page with a faint second headline line

| Without tone curve | With tone curve |
| --- | --- |
| Model reads only 'Your intelligence.', misses the faint line | Model reads 'Your intelligence. Your machine.' and notes the second line is faded |

The model goes from omitting the faint line entirely to transcribing it and remarking that it is "faded/dimmed".

### Fix

One-line change (plus a regression test): apply `MediaProcessing.inSRGBToneCurveSpace` before the resample in the Qwen3VL image path, exactly as the sibling paths do. PR incoming.

spokvulcan:
Fixes #410.

The Qwen3VL **image** path hands the model CoreImage's linear-light working-space values instead of the gamma-encoded sRGB values the HF reference preprocesses: CoreImage filters run in linear light and `MediaProcessing.asMLXArray` renders with a nil colorSpace. The sibling `Qwen25VL`/`Qwen2VL` image paths and this file's own **video** path all apply the linear→sRGB tone-curve step; the Qwen3VL port dropped it in the image path.

Bright content survives the error; dark content does not — near-black text (luminance 23/255 on a 6/255 background) reaches the ViT with ~12× less contrast than reference and is unreadable. Measured on the incident screenshot (text-vs-background signal after mean/std-0.5 normalization): **0.011** without the step, **0.126** with it, HF reference **0.133**. Before/after screenshots in #410.

The change is the one-line `inSRGBToneCurveSpace` wrap at the resample site, plus a regression test (`Qwen3VLToneCurveTests`) that runs a synthetic faint dark-theme fixture through `preprocess` and pins the model-side extrema to the gamma-encoded reference values — the linear-space failure mode lands far outside the tolerances.

Note: textually adjacent to #398 (both touch the resample site); whichever lands second is a trivial rebase.

### DIFF
--- a/Libraries/MLXVLM/Models/Qwen3VL.swift
+++ b/Libraries/MLXVLM/Models/Qwen3VL.swift
@@ -52,7 +52,18 @@ public struct Qwen3VLProcessor: UserInputProcessor {
 
         let targetSize = CGSize(width: resizedWidth, height: resizedHeight)
 
-        let resampled = processed.map { MediaProcessing.resampleBicubic($0, to: targetSize) }
+        // The sRGB tone-curve step must precede resample/normalize: CoreImage
+        // filters run in a *linear-light* working space and `asMLXArray`
+        // renders with a nil colorSpace (raw working-space values), so without
+        // it the model receives linearized values instead of the gamma-encoded
+        // bytes the HF reference preprocesses — crushing dark-region contrast
+        // ~12× (near-black text becomes invisible to the model). The sibling
+        // Qwen25VL/Qwen2VL image paths and this file's own video path apply
+        // the same step.
+        let resampled = processed.map {
+            MediaProcessing.resampleBicubic(
+                MediaProcessing.inSRGBToneCurveSpace($0), to: targetSize)
+        }
 
         let normalized =
             resampled
--- a/Tests/MLXLMTests/Qwen3VLToneCurveTests.swift
+++ b/Tests/MLXLMTests/Qwen3VLToneCurveTests.swift
@@ -1,0 +1,79 @@
+// Copyright © 2026 Apple Inc.
+
+import CoreGraphics
+import CoreImage
+import MLX
+import MLXLMCommon
+import XCTest
+
+@testable import MLXVLM
+
+/// The Qwen3VL image path must hand the model gamma-encoded sRGB values —
+/// the byte values the HF reference preprocesses — not CoreImage's
+/// linear-light working-space values. Without the tone-curve step, near-black
+/// content (e.g. dark-theme text at luminance 23/255 on a 6/255 background)
+/// reaches the ViT with ~12x less contrast than the reference and becomes
+/// unreadable, while bright content is barely affected.
+final class Qwen3VLToneCurveTests: XCTestCase {
+
+    private func makeProcessor() throws -> Qwen3VLProcessor {
+        let json = """
+            {
+              "image_mean": [0.5, 0.5, 0.5],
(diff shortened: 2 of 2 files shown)


Libraries/MLXVLM/Models/Qwen3VL.swift
```diff
@@ -52,7 +52,18 @@ public struct Qwen3VLProcessor: UserInputProcessor {
...
+        // renders with a nil colorSpace (raw working-space values), so without
+        // it the model receives linearized values instead of the gamma-encoded
+        // bytes the HF reference preprocesses — crushing dark-region contrast
+        // ~12× (near-black text becomes invisible to the model). The sibling
+        // Qwen25VL/Qwen2VL image paths and this file's own video path apply
+        // the same step.
+        let resampled = processed.map {
+            MediaProcessing.resampleBicubic(
+                MediaProcessing.inSRGBToneCurveSpace($0), to: targetSize)
+        }
```
davidkoski, line 66: I wonder if other models have the same issue?  From a quick scan it looks like Idefics3, MistralV3, and Paligemma might.

Qwen25VL, Gemma3 and Gemma4 all do the resample in gamma space (incorrect, but probably matches the python code) so they won't have this issue.

davidkoski (approved):
Change looks good.  See comment on other models.  Thank you.

PR #412, opened 2026-07-09
Generation: honor suppress_tokens and mask multimodal placeholder tokens

fdagostino:
### Problem

Two related gaps in the sampling path:

1. **`suppress_tokens` from `generation_config.json` was silently ignored** — `GenerationConfigFile` never parsed the field, so checkpoints that declare tokens which must never be sampled got no masking at all.
2. **Multimodal placeholder tokens were sampled as text.** Nothing masks the placeholder IDs of multimodal models (image/audio/video soft tokens and their begin/end markers), so they occasionally leak verbatim into generated text. On `gemma4_unified` (e.g. `mlx-community/gemma-4-12B-it-4bit`) we saw `<audio|>` (258882) show up in text-only replies — nasty for downstream consumers (imagine a TTS reading it aloud). The same symptom was reported independently against the same checkpoint family in another runtime: jundot/omlx#1670. The shipped `generation_config.json` only suppresses 258883/258882 on some checkpoints, and even that was ignored (gap 1).

### Fix

Mirrors the `transformers` design (`SuppressTokensLogitsProcessor`, driven by `generation_config.suppress_tokens`):

- `GenerationConfigFile`: parse `suppress_tokens`.
- New `SuppressedTokensProviding` protocol: models advertise token IDs that must never be sampled; the LLM/VLM model factories merge `generation_config.json`'s `suppress_tokens` into that set after the model is created.
- New `SuppressTokensProcessor` (port of the transformers processor): masks those IDs to `-inf` before sampling, chained after the parameter-derived penalty processor in `TokenIterator`, `SpeculativeTokenIterator` and `MTPSpeculativeTokenIterator` via a small `ChainedLogitProcessor`.
- `Gemma4` / `Gemma4Unified`: seed the set from the config placeholder IDs (image/audio/video soft tokens plus boi/eoi/boa/eoa markers).

Only sampled tokens are affected; prompt tokens are untouched, so multimodal *inputs* keep working unchanged.

### Testing

- 12 new tests in `Tests/MLXLMTests/SuppressTokensTests.swift` (config parsing, processor masking/broadcasting, chaining, factory merge, Gemma 4 seeding).
- `xcodebuild build-for-testing -scheme mlx-swift-lm-Package -destination 'platform=macOS'` + `xcrun xctest …/MLXLMTests.xctest` (same as CI) pass on this branch.
- Smoke-tested with `mlx-community/gemma-4-12B-it-4bit`: the placeholder leakage disappears; generation output otherwise unchanged.

### Planned follow-up (separate PR)

App-driven per-generation suppression via a `suppressedTokens` field on `GenerateParameters`, merged into the same `makeLogitProcessor` union (and reaching the MTP draft sampler through the same path). This PR intentionally keeps checkpoint-level semantics only — `generation_config.json` and model-adopted sets — mirroring how `transformers` scopes `suppress_tokens`; the per-generation field is for app decisions (e.g. suppressing a model's thinking-channel delimiters for a voice reply while another session of the same loaded model leaves them free).

### DIFF
--- a/Libraries/MLXLMCommon/SuppressTokens.swift
+++ b/Libraries/MLXLMCommon/SuppressTokens.swift
@@ -31,8 +31,16 @@ public protocol SuppressedTokensProviding: AnyObject {
 /// which is driven by `suppress_tokens` in `generation_config.json`.
 public struct SuppressTokensProcessor: LogitProcessor {
 
+    /// Suppressed vocabulary indices, sorted ascending. IDs at or beyond the
+    /// logits vocabulary size are dropped at ``process(logits:)`` time, so a
+    /// configuration that inherits placeholder defaults larger than the
+    /// model's vocabulary (e.g. tiny test configs) degrades to a no-op
+    /// instead of indexing out of range.
+    private let sortedIds: [Int]
+
     /// Suppressed vocabulary indices, shaped `[1, N]` for broadcasting
-    /// against `[B, vocab]` logits.
+    /// against `[B, vocab]` logits. Fast path used when every ID fits the
+    /// logits vocabulary.
     private let broadcastIndices: MLXArray
 
     /// `-inf` fill values, shaped `[1, N]` to match `broadcastIndices`.
@@ -40,11 +48,13 @@ public struct SuppressTokensProcessor: LogitProcessor {
 
     /// Create a processor masking the given token IDs.
     ///
-    /// Returns `nil` when `tokenIds` is empty (nothing to suppress).
+    /// Negative IDs are ignored. Returns `nil` when no valid IDs remain
+    /// (nothing to suppress).
     public init?(tokenIds: Set<Int>) {
-        guard !tokenIds.isEmpty else { return nil }
-        let sorted = tokenIds.sorted().map { UInt32($0) }
-        self.broadcastIndices = MLXArray(sorted)[.newAxis, 0...]
+        let sorted = tokenIds.filter { $0 >= 0 }.sorted()
+        guard !sorted.isEmpty else { return nil }
+        self.sortedIds = sorted
+        self.broadcastIndices = MLXArray(sorted.map { UInt32($0) })[.newAxis, 0...]
         self.negInfValues =
             MLXArray([Float](repeating: -Float.infinity, count: sorted.count))[
                 .newAxis, 0...]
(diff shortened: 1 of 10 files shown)


Libraries/MLXLMCommon/SuppressTokens.swift
```diff
@@ -0,0 +1,120 @@
...
+/// Merge `suppress_tokens` from `generation_config.json` into the model's
+/// ``SuppressedTokensProviding/suppressedTokenIds``.
+///
+/// No-op when the configuration declares no suppressed tokens or the model
+/// does not conform to ``SuppressedTokensProviding``. Called by the model
+/// factories after the model is created.
+public func mergeGenerationConfigSuppressedTokens(
+    _ generationConfig: GenerationConfigFile?, into model: any LanguageModel
+) {
+    guard let suppressTokens = generationConfig?.suppressTokens?.values else { return }
```
aleroot, line 155: This makes generation_config.suppress_tokens a no-op for every model that does not adopt SuppressedTokensProviding; currently only the two Gemma 4 VLMs do...

Libraries/MLXLMCommon/SuppressTokens.swift
```diff
@@ -0,0 +1,120 @@
...
+
+    /// `-inf` fill values, shaped `[1, N]` to match `broadcastIndices`.
+    private let negInfValues: MLXArray
+
+    /// Create a processor masking the given token IDs.
+    ///
+    /// Returns `nil` when `tokenIds` is empty (nothing to suppress).
+    public init?(tokenIds: Set<Int>) {
+        guard !tokenIds.isEmpty else { return nil }
+        let sorted = tokenIds.sorted().map { UInt32($0) }
```
aleroot: These IDs are not validated against the logits vocabulary. The tiny Gemma4Unified test configuration has vocab_size: 32 but inherits boi/boa defaults in the 255k range, so a text-only TokenIterator will pass out-of-range indices to putAlong. Please filter to 0..<model.vocabularySize (and handle negative IDs) before building the suppressor, with a generation regression test.

aleroot:
If going ahead with this PR, MTP draft proposals need the same suppression policy to avoid unnecessary verifier rejections and reduced speculative-decoding acceptance.

fdagostino:
Thanks for the careful review — all three points were valid and are addressed in cb9b074: (1) `generation_config.json`'s `suppress_tokens` is now honored for every model — non-conforming models are tracked in a weak side table that the factories merge into and `makeLogitProcessor` unions with the protocol set; (2) IDs are validated at `process()` time against `logits.dim(-1)` (negatives dropped at init), with regression tests including your exact scenario (255k-range defaults on a tiny-vocab config); (3) `draftBlock` now receives the sampler wrapped in a `SuppressTokensSampler`, so MTP drafts can't propose tokens the verifier would reject (the classic `SpeculativeTokenIterator` already applied the processor to draft logits).

fdagostino:
The red `mac_build_and_test` was not a test failure: the run died at **Verify documentation**, before tests executed. The doc comments on `Gemma4.suppressedTokenIds` / `Gemma4Unified.suppressedTokenIds` referenced ``SuppressedTokensProviding`` as a DocC symbol link, but the protocol lives in MLXLMCommon and DocC can't resolve cross-module symbol links from MLXVLM's pass — with `--warnings-as-errors` that failed the job.

Pushed b8da8c5 switching the reference to plain code voice (matching how MLXVLM docs refer to `LanguageModel`). `verify-docs.sh` now passes locally across all targets; the CI run is awaiting workflow approval. No code changes — this remains ready for re-review.

PR #414, opened 2026-07-09
ChatSession: keep the cached transcript canonical on continuation turns

fdagostino:
### Problem

`ChatSession` keeps prior turns in the KV cache as raw tokens, and two behaviors interact so that **continuation turns can append to a malformed transcript**:

1. **A truncated generation leaves a dangling model turn in the cache.** A caller that stops consuming mid-stream (e.g. voice barge-in) — or a generation capped by `maxTokens` — leaves the cached turn cut off mid-sentence, with no turn terminator. (A normally completed turn is fine: the sampled stop token is fed through the model before the generate loop recognizes it, so it is already in the cache.)
2. **Templates that begin with `{{ bos_token }}` re-emit `<bos>` on every render** (Gemma's chat template does), so every continuation prompt injects a spurious `<bos>` mid-transcript.

Chat templates only render the messages they are given, so nothing ever repairs this: after an interruption the effective transcript the model sees is

```
…truncated repl<bos><|turn>user
next message…
```

instead of the canonical

```
…truncated repl<turn|>
<|turn>user
next message…
```

In practice the duplicate `<bos>` alone is often invisible on short text-only chats, but after an interruption the malformed transcript can push Gemma 4 into emitting its `<|channel>thought` channel or degenerate output (reproduced with `mlx-community/gemma-4-12B-it-4bit`). With the repair in place, post-interruption turns stay clean — validated on-device in a voice-assistant app with real barge-in.

### Fix

The split follows the layering @aleroot suggested in review: **the session determines the semantics, the model side owns the template mechanics.** Callers never see or set template syntax.

- New `ModelConfiguration.ContinuationPolicy` (`truncatedTurnClosure`, `stripsRepeatedBOS`) — model-side knowledge, declared alongside the template knowledge that already lives there (`extraEOSTokens`, `toolCallFormat`). The Gemma 3 / Gemma 4 registrations declare it (`"<end_of_turn>\n"` / `"<turn|>\n"`).
- `ChatSession` only determines: is this pass a continuation (live KV-cache offset, so the tool-dispatch restart is covered), and did the previous turn complete or truncate (tracked `GenerateStopReason`, pessimistically reset at pass start). It then applies the model's policy: the turn closure **only for truncated turns** (completed turns already carry their terminator in the cache), the BOS strip only when the first prompt token equals the tokenizer's BOS.
- **No policy ⇒ no repair**: models that don't declare a `ContinuationPolicy` get zero behavior change, which also resolves the earlier concern about the BOS-strip default.
- `Tokenizer.bosTokenId` convenience, mirroring the existing `eosTokenId`.

The token-level repair itself runs in one place (`repairContinuation`), after `processor.prepare`, handling both token-tensor shapes (rank-1 text / rank-2 VLM) and preserving a `nil` mask.

### Testing

- Deterministic unit tests for `repairContinuation` (rank-1/rank-2, BOS edge cases, closure prepending, mask handling) plus policy-plumbing tests (default configuration has no policy; Gemma registrations carry the right closures).
- `xcodebuild build-for-testing` + `MLXLMTests` via `xcrun xctest` (same as CI) pass on this branch.
- Deterministic A/B at temperature 0 on Gemma 4 12B: without repair, continuation turns show `<bos>` duplication and post-interruption thought-channel leakage; with repair, transcripts are canonical and the leakage disappears. Validated on-device (iPad, voice barge-in scenario).

### DIFF
--- a/Libraries/MLXLMCommon/ChatSession.swift
+++ b/Libraries/MLXLMCommon/ChatSession.swift
@@ -163,32 +163,6 @@ public final class ChatSession {
     /// Speculative decoding configuration, nil if disabled.
     public let speculativeDecoding: SpeculativeDecodingConfig?
 
-    /// Text that closes a *truncated* previous model turn when a new turn is
-    /// appended to a non-empty KV cache (e.g. `"<turn|>\n"` for Gemma 4,
-    /// `"<end_of_turn>\n"` for Gemma 3).
-    ///
-    /// The KV cache stores prior turns as raw tokens. A normally completed
-    /// turn ends with its stop token already in the cache (the iterator feeds
-    /// each sampled token through the model before the generate loop
-    /// recognizes it as a stop), but a generation that was cancelled
-    /// mid-stream — or capped by `maxTokens` — leaves the cached turn dangling
-    /// mid-sentence. Chat templates only render the messages they are given,
-    /// so nothing ever closes that dangling turn: the continuation is appended
-    /// right after the truncated reply's last word, leaving the transcript
-    /// malformed (`…repl<|turn>user…`).
-    ///
-    /// When set, these tokens are prepended to the continuation prompt *only
-    /// when the previous generation did not finish with a stop token*, so the
-    /// cached transcript stays canonical (`…repl<turn|>\n<|turn>user…`)
-    /// without duplicating the terminator of completed turns. Off (nil) by
-    /// default.
-    public var continuationTurnClosure: String?
-
-    /// Chat templates that begin with `{{ bos_token }}` re-emit `<bos>` on
-    /// every render; on continuation turns that would inject a spurious BOS
-    /// mid-transcript, so it is stripped. On by default.
-    public var stripsContinuationBOS: Bool = true
-
     /// Stop reason of the most recent generation pass, kept across turns so
     /// continuation repair can tell a completed turn (stop token already in
     /// the KV cache) from a truncated one (cancelled or length-capped).
@@ -655,8 +629,6 @@ extension ChatSession {
         // and are only being sent to the inner async
         let inputMessages = SendableBox<[Chat.Message]>(messages)
 
-        let continuationTurnClosure = self.continuationTurnClosure
-        let stripsContinuationBOS = self.stripsContinuationBOS
         let task = Task {
             [
                 model,
@@ -715,6 +687,11 @@ extension ChatSession {
                     // prepare the input
                     messages.append(contentsOf: inputMessages.consume())
 
+                    // The model side owns the template-specific mechanics of
+                    // continuation repair; the session only determines the
+                    // semantics (continuation? truncated or completed?).
+                    let continuationPolicy = modelConfiguration.continuationPolicy
+
                     // loop can restart on tool calls
                     restart: while !messages.isEmpty {
                         // A continuation appends to a non-empty KV cache.
@@ -735,13 +712,16 @@ extension ChatSession {
                             processing: processing,
                             tools: tools, additionalContext: additionalContext)
                         let preparedInput = try await processor.prepare(input: userInput)
-                        let input =
-                            isContinuation
-                            ? repairContinuation(
+                        let input: LMInput
+                        if isContinuation, let continuationPolicy {
+                            input = repairContinuation(
                                 preparedInput, tokenizer: tokenizer,
-                                closure: previousTurnCompleted ? nil : continuationTurnClosure,
-                                stripBOS: stripsContinuationBOS)
-                            : preparedInput
+                                closure: previousTurnCompleted
+                                    ? nil : continuationPolicy.truncatedTurnClosure,
+                                stripBOS: continuationPolicy.stripsRepeatedBOS)
+                        } else {
+                            input = preparedInput
+                        }
                         // Pessimistic until this pass reports completion: a
                         // generation that dies mid-stream leaves a truncated
                         // turn behind.
--- a/Libraries/MLXLMCommon/ChatSession.swift
+++ b/Libraries/MLXLMCommon/ChatSession.swift
@@ -163,25 +163,39 @@ public final class ChatSession {
     /// Speculative decoding configuration, nil if disabled.
     public let speculativeDecoding: SpeculativeDecodingConfig?
 
-    /// Text that closes the previous model turn when a new turn is appended to
-    /// a non-empty KV cache (e.g. `"<turn|>\n"` for Gemma 4, `"<end_of_turn>\n"`
-    /// for Gemma 3).
+    /// Text that closes a *truncated* previous model turn when a new turn is
+    /// appended to a non-empty KV cache (e.g. `"<turn|>\n"` for Gemma 4,
+    /// `"<end_of_turn>\n"` for Gemma 3).
     ///
-    /// The KV cache stores prior turns as raw tokens and always ends mid
-    /// model-turn: the stop token is *sampled* but never fed through the model,
-    /// and a cancelled generation stops anywhere. Chat templates only render
-    /// the messages they are given, so nothing ever closes that dangling turn —
-    /// every continuation turn is appended right after the previous reply's
-    /// last word, leaving the transcript malformed (`…reply<|turn>user…`).
-    /// When set, these tokens are prepended to each continuation prompt so the
-    /// cached transcript stays canonical (`…reply<turn|>\n<|turn>user…`), which
-    /// also heals turns truncated by cancellation. Off (nil) by default.
+    /// The KV cache stores prior turns as raw tokens. A normally completed
+    /// turn ends with its stop token already in the cache (the iterator feeds
+    /// each sampled token through the model before the generate loop
+    /// recognizes it as a stop), but a generation that was cancelled
+    /// mid-stream — or capped by `maxTokens` — leaves the cached turn dangling
+    /// mid-sentence. Chat templates only render the messages they are given,
+    /// so nothing ever closes that dangling turn: the continuation is appended
+    /// right after the truncated reply's last word, leaving the transcript
+    /// malformed (`…repl<|turn>user…`).
+    ///
+    /// When set, these tokens are prepended to the continuation prompt *only
+    /// when the previous generation did not finish with a stop token*, so the
+    /// cached transcript stays canonical (`…repl<turn|>\n<|turn>user…`)
+    /// without duplicating the terminator of completed turns. Off (nil) by
+    /// default.
     public var continuationTurnClosure: String?
 
     /// Chat templates that begin with `{{ bos_token }}` re-emit `<bos>` on
     /// every render; on continuation turns that would inject a spurious BOS
     /// mid-transcript, so it is stripped. On by default.
     public var stripsContinuationBOS: Bool = true
+
+    /// Stop reason of the most recent generation pass, kept across turns so
+    /// continuation repair can tell a completed turn (stop token already in
+    /// the KV cache) from a truncated one (cancelled or length-capped).
+    /// `nil` until a pass reports completion — and reset to `nil` at the
+    /// start of each pass, so a generation that dies without reporting is
+    /// treated as truncated.
+    private let lastStopReason = SerialAccessContainer<GenerateStopReason?>(nil)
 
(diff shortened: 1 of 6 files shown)


Libraries/MLXLMCommon/ChatSession.swift
```diff
@@ -637,7 +697,15 @@ public final class ChatSession {
                             chat: messages,
                             processing: processing,
                             tools: tools, additionalContext: additionalContext)
-                        let input = try await processor.prepare(input: userInput)
+                        let preparedInput = try await processor.prepare(input: userInput)
+                        let input =
+                            isContinuation
+                            ? repairContinuation(
+                                preparedInput, tokenizer: tokenizer,
+                                closure: continuationTurnClosure,
```
aleroot: Does this double-close normally completed turns? TokenIterator.next() calls step(previous: previousY) before yielding previousY, and step feeds that token into the model/KV cache. Therefore a generated EOS token is already cached before generateLoopTask recognizes it as a stop token. Prepending <end_of_turn> / <turn|> here would add a second terminator on every normal completed response...

Libraries/MLXLMCommon/ChatSession.swift
```diff
@@ -564,6 +584,42 @@ public final class ChatSession {
...
+/// cached transcript stays canonical: strips the duplicate `<bos>` that
+/// templates re-emit on every render, and prepends `continuationTurnClosure`
+/// to close the previous model turn (whose stop token was sampled but never
+/// fed — or which was truncated by a cancelled generation).
+private func repairContinuation(
+    _ input: LMInput, tokenizer: any Tokenizer,
+    closure: String?, stripBOS: Bool
+) -> LMInput {
+    var tokens = input.text.tokens
+    guard tokens.ndim == 2, tokens.dim(1) > 0 else { return input }
```
aleroot, line 594: This makes the repair a no-op for the default LLMUserInputProcessor, which returns LMInput(tokens: MLXArray(promptTokens)) with rank-1 tokens. Standard text LLM sessions therefore receive neither BOS stripping nor the closure.

Libraries/MLXLMCommon/ChatSession.swift
```diff
@@ -637,7 +697,15 @@ public final class ChatSession {
...
-                        let input = try await processor.prepare(input: userInput)
+                        let preparedInput = try await processor.prepare(input: userInput)
+                        let input =
+                            isContinuation
+                            ? repairContinuation(
+                                preparedInput, tokenizer: tokenizer,
+                                closure: continuationTurnClosure,
+                                stripBOS: stripsContinuationBOS)
+                            : preparedInput
+                        isContinuation = false
```
aleroot: This remains false across the toolDispatch continue restart path, even though the same KV cache is non-empty. As a result, the prompt containing the automatic tool result bypasses repairContinuation.

aleroot:
I was looking into this issue as well, as I have encountered it with Gemma 12B, but I do not think this implementation is ready to merge yet: it makes assumptions about cache boundaries and token tensor shape that do not hold across the existing processor/session paths, and it may double-append a turn terminator after normal EOS completion.

In addition I strongly suggest to add unit tests as this is a tricky are to play with...

fdagostino:
You were right on all three counts — thanks, this was exactly the review it needed. (1) Confirmed: the sampled stop token *is* fed through the model before the loop recognizes it, so a completed turn is already closed in the cache and the unconditional closure duplicated the terminator. In f81ce34 the session tracks the last pass's `stopReason` and the closure now applies only to truncated turns (cancelled / length-capped / unreported) — which was the actual barge-in bug. The PR description's rationale was wrong on this point; I've corrected it and the doc comments accordingly. (2) Fixed: `repairContinuation` normalizes rank-1 token tensors and restores rank on the way out (and no longer fabricates a mask when the input had none). (3) Fixed: the continuation check is derived from the live cache offset on each restart iteration, so the tool-dispatch path is repaired too. Added 8 deterministic unit tests for the repair. Happy to flip the `stripsContinuationBOS` default to `false` if strict compatibility is preferred.

Libraries/MLXLMCommon/ChatSession.swift
```diff
@@ -163,6 +163,40 @@ public final class ChatSession {
...
+    /// so nothing ever closes that dangling turn: the continuation is appended
+    /// right after the truncated reply's last word, leaving the transcript
+    /// malformed (`…repl<|turn>user…`).
+    ///
+    /// When set, these tokens are prepended to the continuation prompt *only
+    /// when the previous generation did not finish with a stop token*, so the
+    /// cached transcript stays canonical (`…repl<turn|>\n<|turn>user…`)
+    /// without duplicating the terminator of completed turns. Off (nil) by
+    /// default.
+    public var continuationTurnClosure: String?
```
aleroot: I think there is still an architectural issue with these additions to ChatSession.
continuationTurnClosure exposes model-specific serialization details to callers. Applications must identify the model and hard-code values such as "<turn|>\n" or "<end_of_turn>\n". This couples a generic session API to concrete chat-template protocols and requires downstream changes whenever a model or template variant uses different framing.
ChatSession should only determine that it is continuing after a truncated turn. It could pass that semantic context to UserInputProcessor, for example .continuation(previousTurn: .truncated), and let the processor/tokenizer integration handle the exact terminator tokens and BOS behavior. That layer already owns the conversion from chat messages to model-specific LMInput, including masks and multimodal alignment.
Could we replace continuationTurnClosure and stripsContinuationBOS with a continuation-aware processor API, or with a policy automatically provided by ModelContext? This would keep ChatSession model-agnostic and prevent callers from needing to understand each model’s internal chat-template syntax.

fdagostino:
You're right on all three counts — thank you, this was exactly the review it needed. Addressed in f81ce34, and the PR description is updated accordingly:

1. **Double terminator**: correct — the sampled stop token *is* fed through the model before the loop recognizes it, so a completed turn already carries its terminator in the cache (the original description had this backwards). The closure is now gated on the previous pass's `GenerateStopReason`: only turns that ended cancelled, length-capped, or unreported get closed. The reason is reset pessimistically at pass start, so a generation that dies mid-stream is treated as truncated.
2. **Rank-1 no-op**: fixed — `repairContinuation` normalizes rank-1 `[N]` to `[1, N]` and restores the rank on the way out, so default text-only sessions get the repair too. It also preserves a `nil` mask instead of fabricating one.
3. **toolDispatch bypass**: fixed — the continuation check is derived from the live KV cache offset on every pass of the restart loop instead of a one-shot flag, so the tool-result restart is repaired as well.

Also added 8 deterministic unit tests for `repairContinuation` (rank-1/rank-2 shapes, BOS edge cases, closure prepending, mask handling).

fdagostino:
Agreed — the knobs were the wrong layer, thanks for pushing on this. Reworked in 281a7b6: `continuationTurnClosure` and `stripsContinuationBOS` are gone from `ChatSession`.

The template-specific mechanics now live on the model side as `ModelConfiguration.ContinuationPolicy` (`truncatedTurnClosure`, `stripsRepeatedBOS`), next to the template knowledge that already lives there (`extraEOSTokens`, `toolCallFormat`); the Gemma 3 / Gemma 4 registrations declare it. `ChatSession` only determines the semantics — continuation or not (live KV-cache offset), previous turn truncated or completed (tracked stop reason) — and applies whatever the model declares. Callers never see template syntax, and **no policy ⇒ no repair**, which also settles the earlier BOS-strip default concern: the out-of-the-box behavior is unchanged for every model that doesn't declare a policy.

On the two shapes you suggested: I went with the ModelContext/configuration-provided policy rather than a continuation-aware `UserInputProcessor` API because the repair is token-level and identical across models once the two template facts are known — pushing it into the processors would replicate the same logic across every implementation (or grow the protocol for all of them). Happy to move it into the processor layer instead if you think that's where it belongs long-term (it would compose well with processors owning turn serialization more broadly, per the structured tool-turns discussion).

Libraries/MLXLMCommon/ChatSession.swift
```diff
@@ -752,25 +840,37 @@ public final class ChatSession {
...
+                            if consumerTerminated { continue }
                             // collect tool calls for dispatch; if no
                             // toolDispatch the caller handles them via
                             // the transform (streamDetails path)
                             if let toolCall = item.toolCall, toolDispatch != nil {
                                 pendingToolCalls.append(toolCall)
                             } else if let value = transform(item) {
                                 if case .terminated = continuation.yield(value) {
-                                    break
+                                    consumerTerminated = true
```
davidkoski, line 861: This is a bit subtle.  I had to read it a few times and have claude analyze it to probe it before I was fully understanding it :-)

In the case where the caller of ChatSession terminates the loop early, the yield of the token will indicate the outgoing stream is terminated.  We set `consumerTerminated` and go to the top of the loop.  Simultaneously the `continuation.onTermination` would fire and cancel this task.

One thing I didn't realize, and why removing the `break` is safe:  the `await item in genStream` will return `nil` if the calling task is cancelled.  The loop will either give us one more token (potentially the EOT) or it will break from the task cancel before we get it.  Either way, the loop breaking will cause the token generation loop to shut down.

But, I think this logic is a bit hard to understand -- I can barely keep all the 3 tasks in my head and the iterator state after staring at this code for a bit.  I also think that it may be racy:  the generation task _may_ have committed the EOT to the KVCache AND the task cancellation caused this to return nil so we missed the info.

I think.

So, what if the `genTask` wasn't `Task<Void, Never>`, but `Task<Info, Never>`.  We are already awaiting the task completion (because we have to guarantee the borrow of the KVCache is complete).  The task knows if it reached the end of turn or not and we can return it like that.

Anything dealing with the stream is racy.  We could 1) yield a token/info into the Evaluate -> ChatSession stream but not consume it because of the outer task cancel or 2) break before yielding in the Evaluate loop:

```swift
            tokenLoop: while let token = iterator.next() {
                // Check for cancellation on every loop iteration.
                if Task.isCancelled {
                    stopReason = .cancelled
                    break
                }
```

In either case the generate loop knows the last token it committed to the KVCache, but ChatSession does not — the stream may never deliver it. When that lost token was the EOT, ChatSession wrongly concludes the turn was truncated and prepends a redundant terminator.

In all cases the generate loop _does_ know what token got committed to the KVCache and the ChatSession _does_ always await the generate task completion.  That is a safe way to indicate to the caller what happened, even if we don't know what tokens were consumed by any of the upper layers.

The returned info could be the stopReason, but more importantly, I think it needs to indicate explicitly if the EOT was committed to the KVCache.  The computation of the stop reason with regard to the EOT is a little bit complicated (handlers, includeStopToken, etc.). 

That would let us go back to the more direct approach for this loop and I think resolve the race.  What do you think?

davidkoski (changes requested):
Needs merge conflicts resolved and see my comment on the loop in ChatSession.  Overall looking good.

fdagostino:
Closing this: #472 solved the underlying problem, and solved it better than this PR did.

When I opened this, `ChatSession` appended only the newly rendered messages to a warm cache, so a continuation re-emitted `<bos>` and could leave the previous model turn unterminated. This PR patched that after the fact, by splicing tokens — stripping the duplicate BOS and prepending a model-specific closure.

Since then, #472 made the session render the **full** transcript every turn and reconcile it against a token ledger through `PromptCacheReusePolicy`. That makes both defects structurally impossible rather than repaired:

- `ExtendCachedPrefixRule` requires `promptTokens.starts(with: cachedTokens)`, so `<bos>` sits inside the common prefix and is never fed a second time.
- A truncated turn's terminator arrives on its own inside the appended suffix, because the next full render contains it. A cancelled turn sets `shouldRecord == false`, invalidates the ledger, and falls through to `.rebuild`.

I verified this against current `main` by exercising the policy directly with the three scenarios this PR was built around — duplicate BOS, length-capped turn, cancelled turn. None of them reproduce.

That also settles both review points here, without the costs you each flagged:

- @aleroot — your objection that `continuationTurnClosure` leaked model-specific framing (`"<end_of_turn>\n"`) into a generic session API. The terminator now comes from the template itself, which is what you were asking for.
- @davidkoski — your concern about inferring the outcome from a stream that may never deliver the last token. The generation task now returns the tokens actually committed to the KV cache, so the session no longer has to guess.

Thanks both for the reviews, and sorry for leaving this sitting in a conflicted state.

PR #415, opened 2026-07-10
Enable MTP speculative decoding for Gemma 4 12B (gemma4_unified target + gemma4_unified_assistant drafter)

fdagostino:
### Problem

The 12B Gemma 4 (`model_type: gemma4_unified`, class `Gemma4Unified`) cannot use MTP speculative decoding, for three stacked reasons:

1. Its drafter (`google/gemma-4-12B-it-assistant` and the mlx-community conversions) ships `model_type: gemma4_unified_assistant`, which is not registered with `MTPDrafterTypeRegistry`.
2. `Gemma4Unified` does not override `callAsFunction(_:cache:state:)`, so the `mtpEmitFlagKey` opt-in is discarded by the protocol-extension default and the target never emits drafter state — the MTP iterator **silently falls back to single-token passthrough** (no error, just no speedup).
3. `Gemma4AssistantDraftModel.draftBlock` only accepted a `Gemma4` target and trapped on `Gemma4Unified`.

### Fix

- Register `gemma4_unified_assistant` with the same creator — the drafter implementation is shared; the unified variant differs only in its `text_config`, which `Gemma4TextConfiguration` already decodes.
- Add the MTP-aware state entry point to `Gemma4Unified`, mirroring `Gemma4`.
- Make `draftBlock` target-agnostic: instead of dispatching on concrete classes, the target is cast to a new internal `Gemma4BackboneProviding` protocol exposing the shared text backbone. `Gemma4` and `Gemma4Unified` adopt it in `Gemma4.swift`, where the conformances can reach the private `languageModel` without widening its visibility.
- For a non-Gemma-4 target, `draftBlock` keeps the existing trap semantics (`fatalError` with the offending type); happy to convert that to a thrown error if you'd prefer.

### Testing

- Config-decode pin mirroring the published 12B assistant config, and a synthetic end-to-end pin: `Gemma4Unified` emits state through the MTP entry point, and `draftBlock` accepts the unified target and returns `[1, blockSize-1]`.
- `xcodebuild build-for-testing` + `MLXLMTests` via `xcrun xctest` (same as CI): 184 tests pass on this branch.
- Validated on-device that the wiring works end-to-end: ~62% draft acceptance on predictable text with the 12B target + assistant drafter.

### DIFF
--- a/Libraries/MLXVLM/Gemma4AssistantRegistration.swift
+++ b/Libraries/MLXVLM/Gemma4AssistantRegistration.swift
@@ -25,13 +25,21 @@ import MLXLMCommon
 /// ```
 public enum Gemma4AssistantRegistration {
     public static func register() async {
-        await MTPDrafterTypeRegistry.shared.registerModelType(
-            "gemma4_assistant",
-            creator: { data in
-                let config = try JSONDecoder().decode(
-                    Gemma4AssistantConfiguration.self, from: data)
-                return Gemma4AssistantDraftModel(config)
-            }
-        )
+        // `gemma4_assistant` (E-series, 26B-A4B, 31B drafters) and
+        // `gemma4_unified_assistant` (the 12B drafter, whose target is the
+        // `gemma4_unified` VLM) share the drafter implementation: the unified
+        // variant differs only in its `text_config` (`gemma4_unified_text`,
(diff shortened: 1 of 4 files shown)


davidkoski (approved):
Looks good to me, thanks!

PR #418, opened 2026-07-12
Qwen3 embedder: honor attentionMask instead of silently ignoring it

spokvulcan:
## What

`Qwen3Model.callAsFunction` (MLXEmbedders) accepts an `attentionMask` parameter and silently discards it — the backbone always runs with a plain causal mask, so padding tokens participate as attention **keys** for any position that comes after them.

## Why it matters

Qwen3-Embedding is a last-token-pooling model, and its Hugging Face reference tokenizes with `padding_side='left'`. Under left padding, *every real token* attends to the pads that precede it, corrupting the hidden state at the exact position pooling reads. Callers who pass a mask today reasonably believe they are protected; the parameter is an API promise the implementation doesn't keep.

We found this in production: combined with maskless pooling on our side, short texts embedded to a shared content-free direction ("I love cats." vs "He is from Europe." at cosine 0.96), and semantic recall over a personal memory store returned confidently-ranked noise.

## Fix

When a caller provides a `[B, L]` mask, fold it into the causal mask — `causal ∧ key-not-padding`, broadcast to `[B, 1, L, L]` — via the existing `createCausalMask` helper. No mask provided → behavior unchanged.

Right-padded causal batches are unaffected by construction (pads sit after the real tokens, which causality already hides) — verified: our embedding regression suite produces identical vectors before and after this change for right-padded batches.

`swift build --target MLXEmbedders` passes.

### DIFF
--- a/Libraries/MLXEmbedders/Models/Qwen3.swift
+++ b/Libraries/MLXEmbedders/Models/Qwen3.swift
@@ -262,14 +262,27 @@ private class Qwen3ModelInner: Module {
     /// - Parameters:
     ///   - inputs: An `MLXArray` of token IDs with shape `[Batch, Length]`.
     ///   - cache: An optional array of `KVCache` objects, one for each layer.
+    ///   - attentionMask: Optional `[Batch, Length]` key-padding mask
+    ///     (nonzero = attend, zero = padding). Folded into the causal mask so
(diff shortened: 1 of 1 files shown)


davidkoski (approved):
Looks good to me.  Thank you.

PR #242, opened 2026-04-25
Fix Qwen2-VL chat template: emit vision tokens before text

NivDvir:
Splits one fix out of #222 per @vade's earlier suggestion.

Qwen2VLMessageGenerator emitted content as `[text, image]`, but HuggingFace's `apply_chat_template` for Qwen2.5-VL emits `<|vision_start|><|image_pad|><|vision_end|>{text}` — image content first. Swapping the order removes a deterministic +9 / +8 / +5 px bbox-coordinate offset on left/top/bottom edges vs the Python `mlx-vlm` 0.4.3 reference.

**Effect on three canonical test images** (Qwen2.5-VL-7B-4bit, temperature=0):

| Image | Before (text-first) | After (image-first) | Python reference |
|---|---|---|---|
| LeetCode A (3078×2114 → 1260×868) | Q [10,154,421,631] / E [421,154,881,631] | Q [1,146,421,628] / E [421,146,881,628] | Q [1,146,421,626] / E [421,146,881,626] |
| LeetCode B (3600×2338 → 1260×812) | (drift) | Q [24,161,352,556] / E [364,161,749,556] | identical (0 px) |
| Layout test (3600×2338 → 1260×812) | (drift) | Q [106,24,1091,536] / E [16,556,464,794] | identical (0 px) |

After the fix: ≤2 px on every edge of every panel, bit-exact (0 px) on 2 of 3 images.

Reproducer (clones upstream, applies a complete patch, builds with xcodebuild, runs a strict ≤2 px gate against saved Python references — aborts loudly on any edge >2 px on any image): https://github.com/NivDvir/screen-overlay-toolkit/tree/main/scripts/parity

Diff: 1 file (`Qwen2VL.swift`), 1 function (`Qwen2VLMessageGenerator.generate`), +9/-10.

PR #419, opened 2026-07-13
Fix prefill LMOutput.State being dropped on TokenIterator's .logits path

NivDvir:
## Proposed changes

Qwen2.5-VL seeds its MROPE `positionIds`/`ropeDeltas` into the per-call `LMOutput.State` during prefill (the #283 pattern), but `TokenIterator.prepare(...)`'s `.logits` branch never carries `result.state` into `self.state`, so the first decode step runs without them — positions drift by the image-token count and greedy decode emits an empty ` ``` ` fence instead of the JSON body.

Reproducible with `mlx-community/Qwen2.5-VL-7B-Instruct-4bit` on any dense full-page screenshot (e.g. a 1280×900 capture of a Wikipedia article) with a standard `bbox_2d` grounding prompt at temperature 0: `main` returns a lone fence; with the state carried, the same input grounds correctly.

The `.tokens` branch already gets the carry via `step(previous:)`, and the speculative iterator's `.logits` branch does `mainState = result.state`; this makes the standard path consistent with both. +2 lines, one function.

Follows the Qwen2.5-VL fixes in #238 / #239 / #242 / #243 — #239 moved the MROPE state onto per-call `LMOutput.State`; this restores its prefill→decode handoff on the `.logits` path.

## Checklist

- [ ] I have added tests that prove my fix is effective or that my feature works — no new test; the repro above exercises the path end-to-end
- [ ] I have updated the necessary documentation (if needed) — not needed

### DIFF
--- a/Libraries/MLXLMCommon/Evaluate.swift
+++ b/Libraries/MLXLMCommon/Evaluate.swift
@@ -649,6 +649,8 @@ public struct TokenIterator: TokenIteratorProtocol {
             asyncEval(y.tokens)
 
         case .logits(let result):
+            // carry the prefill state into decode, as step(previous:) does for later steps
+            self.state = result.state
             y = .init(tokens: convertToToken(logits: result.logits))
             asyncEval(y.tokens)
 


davidkoski (approved):
Change looks good.  will be useful for #399 as well.  Just waiting on CI

PR #422, opened 2026-07-14
Make DeepseekV3Model.init public

### LINKED ISSUE #417: Make `DeepseekV3Model.init(_:)` public so downstream modules can wrap it
`DeepseekV3Model` and everything needed to use it are already `public` — the class, `callAsFunction`, `sanitize`, `kvHeads`, `model`, and `loraLayers`. The one thing that isn't is the initializer, so a separate Swift module can't actually construct one.

In `Libraries/MLXLLM/Models/DeepseekV3.swift`:

```swift
public class DeepseekV3Model: Module, LLMModel, KVCacheDimensionProvider, LoRAModel {
    public var kvHeads: [Int] = []
    var args: DeepseekV3Configuration
    public var model: DeepseekV3ModelInner
    @ModuleInfo(key: "lm_head") var lmHead: Linear

    init(_ args: DeepseekV3Configuration) {   // internal — this is the blocker
        ...
    }
```

Why I'm asking: I keep a small set of model overlays on top of mlx-swift-lm for a macOS app, and I'd like to add Kimi K2.5 (`model_type: kimi_k25`). K2.5 is a thin vision shell over DeepSeek V3 — its text tower is exactly your `DeepseekV3Model`, nested under a `language_model` key, and the only real work is a `sanitize` that strips the vision weights and then defers to DeepSeek V3's own `sanitize`. So the natural implementation is a ~40-line wrapper that holds a `DeepseekV3Model`, forwards `callAsFunction` / `kvHeads` / `loraLayers`, and composes `sanitize`. All of that surface is already public — I just can't call the initializer from another module, and reimplementing the ~1000-line DeepSeek V3 core to get around one access modifier isn't appealing.

The minimal change is one line:

```diff
-    init(_ args: DeepseekV3Configuration) {
+    public init(_ args: DeepseekV3Configuration) {
```

For precedent, `GLM4MoEModel.init(_:)` in `GLM4MOE.swift` is already `public`, which is what lets me wrap that architecture the same way. If you'd also like `DeepseekV3ModelInner.init(config:)` made public for symmetry that's fine, but the wrapper only strictly needs the outer initializer. Glad to send the PR if you're open to it.

magicnight:
`DeepseekV3Model` and everything needed to use it are already public — the class, `callAsFunction`, `sanitize`, `kvHeads`, `model`, and `loraLayers`. The initializer is the one exception, so another module can't actually construct one.

This widens just the initializer to `public`. The signature parameter (`DeepseekV3Configuration`) is already public, so this is a pure access-modifier change with no effect on existing call sites.

`GLM4MoEModel.init(_:)` is already public, which is the precedent — it's what lets you wrap that architecture from another module. I'd like to do the same for DeepSeek V3: a thin wrapper (for a model whose text tower is exactly `DeepseekV3Model`) that holds one and forwards `callAsFunction` / `kvHeads` / `loraLayers` and composes `sanitize`, rather than reimplementing the ~1000-line core to get around one access modifier.

Resolves #417.

### DIFF
--- a/Libraries/MLXLLM/Models/DeepseekV3.swift
+++ b/Libraries/MLXLLM/Models/DeepseekV3.swift
@@ -421,7 +421,7 @@ public class DeepseekV3Model: Module, LLMModel, KVCacheDimensionProvider, LoRAMo
     public var model: DeepseekV3ModelInner
     @ModuleInfo(key: "lm_head") var lmHead: Linear
 
-    init(_ args: DeepseekV3Configuration) {
+    public init(_ args: DeepseekV3Configuration) {
         self.args = args
         self.model = DeepseekV3ModelInner(config: args)
         self._lmHead.wrappedValue = Linear(args.hiddenSize, args.vocabSize, bias: false)


davidkoski (approved):
Looks good, just waiting on CI.  Thank you!

Issue #425, opened 2026-07-15
Prompt-lookup (n-gram) speculative decoding: draft-model-free speedup for repetition-heavy generation


fdagostino:
## Motivation

Speculative decoding currently requires a second model (draft model, or an MTP drafter for Gemma 4). There is a well-known draft-model-free variant — **prompt-lookup decoding** (PLD, a.k.a. n-gram speculation): propose the tokens that followed the last occurrence of the trailing n-gram in the prompt+generated history, and let the target model verify them in one batched pass. Proposals are free; no extra weights, no extra memory.

It pays off whenever the output re-uses spans of the context verbatim — summarization, code editing, RAG-grounded answers, multi-turn chat with recurring phrasing — and costs one cheap single-token step per round when the lookup misses (no wasted batched verify).

Ecosystem signal:

- Python `mlx-lm` has an open request for exactly this (ml-explore/mlx-lm#851), with llama.cpp's n-gram speculation as the reference; nothing exists on the Swift side.
- The original reference implementation is apoorvumang/prompt-lookup-decoding (adopted by `transformers` as `prompt_lookup_num_tokens`).
- Community numbers are consistent: large wins on echo-heavy workloads (~2–3x reported by mlx-serve users), ~1.0x (no regression) on creative generation.

A note on correctness: because the PLD proposal is deterministic (there is no draft distribution), accepting a proposed token iff it equals the token *sampled from the target's own distribution* is unbiased at any temperature — the output distribution is exactly the target model's. PLD therefore doesn't need the residual-sampling machinery that model-based drafting requires for temperature > 0.

## Proposal

A `PromptLookupTokenIterator` mirroring `SpeculativeTokenIterator`'s verify/accept/rewind machinery, with the n-gram lookup as the draft source (longest trailing pattern first, most recent occurrence wins). Two hardenings worth building in from day one:

1. Check rewind headroom per round — trimmable now *and* the round cannot wrap a window-bounded cache mid-verify — and permanently fall back to single-token generation when the headroom is gone, instead of corrupting the cache (the failure mode described in #424).
2. On a lookup miss, take a plain single step rather than a wasted batched verify.

I have a working implementation with deterministic tests (exact token-stream parity with `TokenIterator` on repetitive text, with strictly fewer model forwards; parity preserved when a rotating cache wraps mid-generation) — PR incoming.

Issue #424, opened 2026-07-14
RotatingKVCache becomes untrimmable once the window wraps: speculative-decoding rollback silently no-ops, and prompt-cache prefix reuse is defeated


fdagostino:
## Summary

`RotatingKVCache.isTrimmable` is `offset < maxCacheSize` — once a sliding-window layer wraps, it reports untrimmable for the rest of the session. Two consequences flow from that, one a silent correctness bug, one a performance cliff:

1. **Speculative decoding silently corrupts the cache post-wrap.** `SpeculativeTokenIterator.speculateRound()` rewinds rejected draft tokens with `trimPromptCache(mainCache, numTokens: numDraft - accepted)` and discards the result (`Evaluate.swift`). `trimPromptCache` guards on `canTrimPromptCache` — `allSatisfy { $0.isTrimmable }` — so once **one** sliding layer wraps, the whole rollback returns 0 *silently*: the rejected tokens' K/V stay in every layer (including the still-trimmable global layers), and generation continues on a transcript containing tokens that were never emitted. The iterator's init-time trimmability check passes (pre-wrap), so nothing ever surfaces; the window wraps mid-generation. On Gemma-family models the sliding window is small (e.g. 512), so a single long reply is enough. `MTPSpeculativeTokenIterator` already handles this explicitly (it trims by the amount the cache actually reports trimmed); the classic iterator does not.

2. **Prefix reuse / history edits degrade to full re-prefill post-wrap.** Any feature that rewinds to a common prefix (edited history, prompt-cache reuse across requests) hits the same all-or-nothing guard: with one wrapped layer, nothing can be trimmed, so the only correct option is discarding the entire cache and re-prefilling from scratch. This is the Swift mirror of ml-explore/mlx-lm#980 (closed there after system-prompt checkpoint caching landed; hybrid/sliding-window models still report the underlying limitation).

A third, smaller footgun surfaced while reproducing: `RotatingKVCache.trim()` does **not** self-guard — called directly on a wrapped buffer it still decrements `offset`/`idx` and returns a nonzero count, corrupting the circular-buffer mapping. The `canTrimPromptCache` guard in callers is the only protection.

## Reproduction

Two deterministic tests against `main` (no model download needed):

```swift
@Test func rotatingCacheIsUntrimmableAfterWrap() {
    let rotating = RotatingKVCache(maxSize: 32)
    fill(rotating, tokens: 16)          // pre-wrap: trimming works
[...]

PR #426, opened 2026-07-15
Add prompt-lookup (n-gram) speculative decoding

fdagostino:
Implements the proposal in #425: **prompt-lookup decoding** (PLD / n-gram speculation) — draft-model-free speculative decoding, as requested for Python in ml-explore/mlx-lm#851 and implemented in llama.cpp and `transformers` (`prompt_lookup_num_tokens`).

### What

- `promptLookupDraft(history:maxNGramLength:minNGramLength:maxTokens:)`: pure n-gram lookup over the prompt+generated token history. Longest trailing pattern first, most recent occurrence wins, linear scan (an n-gram hash index à la llama.cpp is a possible follow-up if profiling warrants it).
- `PromptLookupTokenIterator`: mirrors `SpeculativeTokenIterator`'s verify/accept/rewind machinery with the lookup as draft source. Public init parameters: `numDraftTokens` (default 10), `maxNGramLength` (3), `minNGramLength` (1) — matching the reference implementation's defaults.

### Correctness at temperature > 0

The PLD proposal is deterministic — there is no draft distribution `q`. Accepting a proposed token iff it equals the token sampled from the target's own distribution keeps the output distribution exactly the target's at any temperature, so PLD needs none of the residual-sampling machinery that model-based drafting requires. Documented in the iterator's doc comment.

### Two hardenings over the existing speculative machinery

1. **Per-round rewind headroom check**: the round only runs if every cache layer is trimmable *and* — for window-bounded caches — has enough headroom that the verify pass cannot wrap the window mid-round. When the headroom is gone (e.g. a Gemma sliding-window layer wrapped), the iterator permanently falls back to single-token generation instead of corrupting the cache. This is the guard proposed as fix (a) in #424 — the silent-rollback failure mode reported there cannot occur here by construction.
2. **Lookup miss ⇒ plain single step**, not a wasted batched verify: PLD costs nothing on non-repetitive stretches.

### Testing

- 6 deterministic unit tests for the lookup (longest-pattern preference, most-recent occurrence, clamping, degenerate histories).
- Iterator contract tests against a deterministic transition model (same contract as the existing speculative-decoding test model, plus a forward-call counter):
  - **Exact token-stream parity** with `TokenIterator` on cyclic text — with strictly fewer model forwards and `acceptedDraftTokenCount > 0`.
  - **Parity preserved when a rotating cache wraps mid-generation** (the fallback path).
- `xcodebuild build-for-testing` + `MLXLMTests` via `xcrun xctest` (same as CI): 193 tests pass on this branch.

### Follow-ups (not in this PR, happy to do either)

- `ChatSession` integration (a config mirroring `SpeculativeDecodingConfig`).
- N-gram hash index for long contexts.

### DIFF
--- a/Libraries/MLXLMCommon/PromptLookupDecoding.swift
+++ b/Libraries/MLXLMCommon/PromptLookupDecoding.swift
@@ -1,0 +1,347 @@
+// Copyright © 2026 Apple Inc.
+
+import Foundation
+import MLX
+
+/// Find a draft continuation for the current context by n-gram lookup.
+///
+/// Prompt-lookup decoding (PLD) proposes draft tokens by matching the
+/// trailing n-gram of the generated-so-far token history against an earlier
+/// occurrence in that same history, and proposing the tokens that followed
+/// it. No draft model is involved; the proposal is free. It pays off
+/// whenever the output re-uses spans of the input verbatim — summarization,
+/// code editing, retrieval-grounded answers, chat with recurring phrasing —
+/// and costs one wasted verify slot per round otherwise.
+///
+/// Longer patterns are preferred (tried from `maxNGramLength` down to
+/// `minNGramLength`), and among equal-length matches the most recent
+/// occurrence wins.
+///
+/// Linear scan; O(history × maxNGramLength) per round. Fine for on-device
+/// context sizes — an n-gram hash index (as in llama.cpp) is a follow-up if
+/// profiling ever shows this to matter.
+///
+/// - Returns: up to `maxTokens` proposed continuation tokens; empty when no
+///   pattern of at least `minNGramLength` recurs.
+func promptLookupDraft(
+    history: [Int], maxNGramLength: Int, minNGramLength: Int, maxTokens: Int
+) -> [Int] {
+    guard maxTokens > 0, history.count > minNGramLength else { return [] }
+
+    let upperLength = min(maxNGramLength, history.count - 1)
+    guard upperLength >= minNGramLength else { return [] }
+
+    for length in stride(from: upperLength, through: max(minNGramLength, 1), by: -1) {
+        let pattern = history.suffix(length)
+        // Scan candidate start positions from most recent to oldest. The
+        // match may not be the trailing occurrence itself, hence the
+        // `history.count - length - 1` upper bound.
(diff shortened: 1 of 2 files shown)


Libraries/MLXLMCommon/PromptLookupDecoding.swift
```diff
@@ -0,0 +1,347 @@
...
+
+        switch try model.prepare(input, cache: cache, state: state, windowSize: windowSize) {
+        case .tokens(let tokens):
+            y = tokens
+        case .logits(let result):
+            var logits = result.logits[0..., -1, 0...]
+            logits = processor?.process(logits: logits) ?? logits
+            let token = sampler.sample(logits: logits)
+            processor?.didSample(token: token)
+            y = .init(tokens: token)
```
davidkoski, line 185: Does this need the:

```swift
        pendingTokens.append(tokenValue)
        history.append(tokenValue)
```

like `singleStep()`.

fdagostino: Yes — good catch, it did. Fixed in 7947445.

The `.tokens` case hands the prompt back for the iterator to forward, so nothing is generated yet. The `.logits` case is different: `prepare` already evaluated the final prompt position, so the token sampled from those logits **is** the first generated token. Assigning it to `y` alone carried it into the next forward pass but never emitted it — and never entered the n-gram corpus the lookup draws proposals from, so it was invisible to matching too.

The regression test uses a model whose `prepare` returns `.logits`. Without the fix it reproduces cleanly — prompt lookup starts one position ahead of plain generation:

```
PLD   → [4, 5, 6, 7, 0, 1, ...]
plain → [3, 4, 5, 6, 7, 0, ...]
```

This is the same prefill-bonus handling `singleStep()` and `MTPSpeculativeTokenIterator` already do.

I also rebased onto current `main` and adapted three signatures that moved since this branch was written (cfadbb3): `TokenIteratorProtocol.state` is now a public requirement, `newCache` throws, and `prepare` takes `PrefillParameters` instead of a window size. No behavior change — the test model ignores the prefill parameters, as the other iterator test models do.

Full suite green locally (618 XCTest + 749 swift-testing), docs verified.

PR #431, opened 2026-07-15
Track the current SDK's SamplingMode.Kind case names

CharlieTLe:
## Proposed changes

FoundationModels renamed GenerationOptions.SamplingMode.Kind's .top/ .nucleus cases to .randomTopK/.randomProbabilityThreshold, which broke compilation against the newer SDK. This file only compiles against a FoundationModels 2.x SDK in the first place, and the self-hosted CI runner is still on the macOS 26 SDK where this whole file compiles out via the canImport gate above it, so there's no CI-compatibility reason to support both names at once -- just track what the currently required SDK declares.

## Checklist


Libraries/MLXFoundationModels/MLXLanguageModel.swift
```diff
@@ -704,16 +704,21 @@ public struct MLXLanguageModel: FoundationModels.LanguageModel, Sendable {
         /// default" -- so an unrecognized case never traps and never reaches the
         /// resolver. All value policy lives in `resolveSamplingParameters`; this
         /// shim is a pure 1:1 case translation.
+        ///
+        /// `Kind`'s case names track whatever the currently-required
+        /// FoundationModels SDK declares (`Kind.top`/`.nucleus` are
+        /// `.randomTopK`/`.randomProbabilityThreshold` as of this SDK); update
+        /// these cases if a newer SDK renames them again.
```
davidkoski: This is ok for a PR comment but we don't really need it in the code -- presumably the code _would_ match the SDK :-)

ctymoszek:
Fix looks good to me, validated on recent build. Thank you!

davidkoski (approved):
Change looks good, thank you!

PR #434, opened 2026-07-16
Hoist tool-schema $defs to the tool-calling envelope root (fixes #432)

### LINKED ISSUE #432: [BUG] MLXFoundationModels: tool calling fails with constraintCompilationFailed for any Tool whose arguments contain a nested @Generable type — the tool-calling envelope leaves $refs dangling
**Describe the bug**

Registering any `Tool` whose `Arguments` contain a **nested `@Generable` type** makes the session fail at the first tool-calling turn with:

```
constraintCompilationFailed("... json_schema_converter.cc:957: Check failed: ...
Cannot find field $defs in {"oneOf": ...
```

Flat tools (only primitive fields) work, which is presumably why this hasn't surfaced — demo-sized tools don't produce `$defs`.

**Root cause** (from the error's own schema dump): `GenerationSchema` serializes the nested type as `$defs` + a **root-anchored** pointer (`"$ref": "#/$defs/Traveler"`). `SchemaConverter.toolCallingEnvelopeObject` then embeds the tool's schema as a *nested object* under `oneOf[i].properties.arguments` — so the `$defs` end up buried inside `arguments`, while the `$ref` still points at the (empty) document root of the new `{"oneOf": ...}` envelope. xgrammar resolves JSON Pointers from the document root → dangling ref → hard check failure. The error message conveniently prints the whole malformed envelope, where both the buried `$defs` and the root-anchored `$ref` are visible.

**To Reproduce**

Minimal repro — nothing custom, this is the documented way to define a tool:

```swift
struct BookTripTool: Tool {
    let name = "book_trip"
    let description = "Books a trip"

    @Generable
    struct Traveler {
        @Guide(description: "Full name") var name: String
        @Guide(description: "Age") var age: Int
    }

    @Generable
    struct Arguments {
        @Guide(description: "Origin city") var origin: String
        @Guide(description: "Destination city") var destination: String
        @Guide(description: "The traveler") var traveler: Traveler   // ← nested type = $defs/$ref
    }

    func call(arguments: Arguments) async throws -> String { "booked" }
}

let session = LanguageModelSession(model: mlxModel /* any MLXLanguageModel */,
                                   tools: [BookTripTool()])
_ = try await session.respond(to: "Book a trip from A to B for Ana, 34, using book_trip.")
// → constraintCompilationFailed
```

Model used: `mlx-community/Qwen3-8B-4bit` (text — this is not a VLM issue).

**Expected behavior**

The tool-calling turn compiles its grammar and the tool fires, exactly as it does for tools with flat (primitive-only) argument schemas.

 - OS Version: macOS 27.0 beta (26A5378n), Xcode 27 beta
 - Device: M4 Max MacBook/Studio-class, Apple silicon
 - Version: main @ f1573a93 (`MLXFoundationModels`)

**Additional context**

**Fix that works for us** — PR incoming: in `toolCallingEnvelopeObject`, hoist each embedded schema's `$defs` to the envelope root, namespacing keys per tool (`<tool>__<def>`) and rewriting that tool's `$ref`s accordingly. One composition point covers both the envelope and structural-tag encoders. Note for implementers: do the `$ref` string rewrite on the raw `JSONEncoder` output — a `JSONSerialization` re-encode escapes `/` as `\/` and the pointer prefix no longer matches (we hit that on the first attempt). With this patch, the repro above passes and we've run several hundred tool-calling generations through it.

Also affected: any runtime-built `DynamicGenerationSchema` tool parameters with named nested schemas (how we first hit it), and — per `FinalAnswerTool`'s design, which embeds the developer's response schema as the synthetic tool's arguments — any `respond(schema:)`/`respond(generating:)` call in a session **with tools** whose response schema has nested types.

This investigation was AI-assisted; every reproduction in this report was run and verified on hardware.

jyauxi:
## Proposed changes

Fixes #432.

`GenerationSchema` serializes named sub-schemas (nested `@Generable` types, named `DynamicGenerationSchema` nodes) as root-level `$defs` with root-anchored `"$ref": "#/$defs/..."` pointers. `SchemaConverter.toolCallingEnvelopeObject` embeds each tool's parameters schema as a nested object under `oneOf[i].properties.arguments`, which buries the tool's `$defs` inside `arguments` while the refs stay anchored to the (defs-less) envelope root. xgrammar resolves JSON Pointers from the document root, so any tool with a non-flat argument schema fails grammar compilation with `constraintCompilationFailed("... Cannot find field $defs in {\"oneOf\": ...")` at the first tool-calling turn. Flat, primitive-only tools don't emit `$defs`, which is why small demos pass.

This PR hoists each tool's `$defs` into the envelope root, namespaced per tool (`<tool>__<def>`, so same-named defs across tools can't collide), and rewrites that tool's `$ref`s to match. The change is in `toolCallingEnvelopeObject` — the single composition point shared by both `encodeToolCallingEnvelopeJSON` and the structural-tag `encodeToolCallingGrammar`, so both encoders are covered.

Implementation note: the ref rewrite runs on the raw `JSONEncoder` output, where the `#/$defs/` prefix appears literally. Rewriting a `JSONSerialization` re-serialization does not work — it escapes `/` as `\/`, the prefix never matches, and the refs silently survive unrewritten while the defs get hoisted under prefixed keys.

Tests added to `ToolCallingSchemaTests` (model-free, in the existing unit target):
- a nested-`@Generable` fixture's def is hoisted to the envelope root and removed from the embedded `arguments` schema;
- with two tools sharing a same-named nested type, every `$ref` in the envelope resolves within the document and the namespaced defs don't collide;
- the nested-defs envelope compiles end-to-end with xgrammar (`GrammarConstraint`) — this is the direct regression test: it fails on `main` with the "Cannot find field $defs" check failure;
- the structural-tag grammar embeds the hoisted `$defs` in both the wrapped and bare arms.

Beyond unit tests, this patch has run several hundred tool-calling generations in our downstream harness (macOS 27.0 beta 26A5378n, M4 Max) without a grammar-compilation failure.

Local verification note: this beta seed names `GenerationOptions.SamplingMode.Kind` cases `randomTopK`/`randomProbabilityThreshold`, where `main` matches `.top`/`.nucleus` from a newer seed — local builds needed that 2-line rename as a scratch commit, which is **not** part of this PR. CI on the newer seed should compile unmodified.

## Checklist

### DIFF
--- a/Libraries/MLXFoundationModels/GuidedGeneration/SchemaConverter.swift
+++ b/Libraries/MLXFoundationModels/GuidedGeneration/SchemaConverter.swift
@@ -201,15 +201,10 @@ enum SchemaConverter {
             // can embed it as a nested object in the envelope we assemble via
             // JSONSerialization.data(withJSONObject:). Cheap: schemas are small.
             let paramsData = try encoder.encode(tool.parameters)
-            // Rewrite refs on the raw JSONEncoder output, where the
-            // "#/$defs/" prefix appears literally. (A JSONSerialization
-            // re-serialization escapes "/" as "\/", so the prefix would not
-            // match and the refs would silently survive unrewritten. Cover
-            // the escaped form too, defensively.)
-            let rewritten = String(data: paramsData, encoding: .utf8)!
-                .replacingOccurrences(of: "#/$defs/", with: "#/$defs/\(tool.name)__")
-                .replacingOccurrences(of: "#\\/$defs\\/", with: "#\\/$defs\\/\(tool.name)__")
-            var paramsAny = try JSONSerialization.jsonObject(with: Data(rewritten.utf8))
+            var paramsAny = rewriteDefsRefs(
+                in: try JSONSerialization.jsonObject(with: paramsData),
+                toolName: tool.name
+            )
             if var paramsObj = paramsAny as? [String: Any] {
                 if let defs = paramsObj.removeValue(forKey: "$defs") as? [String: Any] {
                     for (key, value) in defs {
@@ -233,6 +228,35 @@ enum SchemaConverter {
             envelope["$defs"] = hoistedDefs
         }
         return envelope
+    }
+
+    /// Rewrites every `"$ref": "#/$defs/<name>"` in a parsed schema tree to
+    /// the per-tool namespaced key (`#/$defs/<tool>__<name>`).
+    ///
+    /// The rewrite is structure-aware: only the string value directly under a
+    /// `$ref` key is touched, so other strings that merely mention the
+    /// pointer text — a `description`, `const`, `enum` entry, `default`, or
+    /// `pattern` containing "#/$defs/" — survive verbatim. Runs before the
+    /// `$defs` are hoisted out, and recurses through the whole tree because
+    /// refs can appear anywhere, including inside other `$defs` bodies.
+    private static func rewriteDefsRefs(in value: Any, toolName: String) -> Any {
+        switch value {
+        case let object as [String: Any]:
+            var result: [String: Any] = [:]
+            result.reserveCapacity(object.count)
+            for (key, nested) in object {
+                if key == "$ref", let ref = nested as? String, ref.hasPrefix("#/$defs/") {
+                    result[key] = "#/$defs/\(toolName)__" + ref.dropFirst("#/$defs/".count)
+                } else {
+                    result[key] = rewriteDefsRefs(in: nested, toolName: toolName)
+                }
+            }
+            return result
+        case let array as [Any]:
+            return array.map { rewriteDefsRefs(in: $0, toolName: toolName) }
+        default:
+            return value
+        }
     }
 
     enum SchemaConversionError: Error {
--- a/Libraries/MLXFoundationModels/GuidedGeneration/SchemaConverter.swift
+++ b/Libraries/MLXFoundationModels/GuidedGeneration/SchemaConverter.swift
@@ -48,9 +48,16 @@ enum SchemaConverter {
     ///       }
     ///     },
     ///     ...
-    ///   ]
+    ///   ],
+    ///   "$defs": { "<tool name>__<def name>": ... }
     /// }
     /// ```
+    ///
+    /// If a tool's parameters schema carries `$defs` (named sub-schemas such
+    /// as nested `@Generable` types), they are hoisted to the envelope root
+    /// under per-tool namespaced keys, with that tool's `$ref`s rewritten to
+    /// match — JSON Pointers resolve from the document root, so defs left
+    /// nested inside `arguments` would leave every ref dangling.
     ///
     /// This is the *inner* schema -- it describes one tool call JSON object.
     /// For end-to-end grammar generation that also encodes the model's native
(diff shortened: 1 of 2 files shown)


davidkoski:
@thechriswebb FYI

thechriswebb:
Thanks for tackling the `$defs` hoisting. The per-tool namespacing is a nice touch, and the xgrammar dangling-ref failure is exactly the kind of thing that's easy to miss. One edge case we should address before this lands.

The root of it is that the `$ref` rewrite is a blind text find-and-replace over the whole serialized schema rather than something that operates on the parsed structure:

```
.replacingOccurrences(of: "#\\/$defs\\/", with: "#\\/$defs\\/\(tool.name)__")
```

Because it matches the `#/$defs/` substring wherever it shows up (not just in $ref values), any string that happens to contain that text gets the tool namespace spliced into it. A @Guide description is enough to trigger it, and the tool doesn't even need a nested @Generable, so there's no real `$defs` to rewrite in the first place:

```
@Generable
private struct PointerDocArgs {
    @Guide(description: "A JSON Pointer such as #/$defs/Foo to resolve.")
    var pointer: String
}

let tool = Transcript.ToolDefinition(
    name: "probe_tool",
    description: "Resolves JSON Pointers",
    parameters: PointerDocArgs.generationSchema
)
let json = try SchemaConverter.encodeToolCallingEnvelopeJSON(tools: [tool])
```

The description comes back out mangled:

```
"description": "A JSON Pointer such as #/$defs/probe_tool__Foo to resolve."
```

It's a very niche situation, but the same thing would hit any const, enum, default, or pattern string that mentions `#/$defs/`, so it felt worth addressing.

Since the code already parses the schema right after, one option is to parse first and then walk the parsed tree, rewriting only the string under a "`$ref`" key. Worth noting `$defs` live at the root but the refs to them can be nested anywhere (including inside other `$defs` bodies), so the walk needs to cover the whole schema and run before the defs are hoisted out.

Here's a test that pins it down if it's useful:

```
@Test
func nonRefStringMentioningDefsPointerIsNotRewritten() throws {
    guard #available(iOS 27.0, macOS 27.0, visionOS 27.0, *) else { return }
    let probe = Transcript.ToolDefinition(
        name: "probe_tool",
        description: "Resolves JSON Pointers",
        parameters: PointerDocArgs.generationSchema
    )
    let json = try SchemaConverter.encodeToolCallingEnvelopeJSON(tools: [probe])
    let strings = collectStringValues(in: try parseAsDictionary(json))

    #expect(strings.contains { $0.contains("#/$defs/") })          // the pointer text made it into the schema
    #expect(strings.allSatisfy { !$0.contains("probe_tool__") })   // the description survived untouched
}
```

Thanks again for the work here!

jyauxi:
Thanks for the careful review and the pinning test — you're right that the string-level rewrite could splice the namespace into innocent strings. Addressed in 72616a6: the rewrite now walks the parsed tree and touches only the string under a `$ref` key, running before the defs are hoisted so refs inside other `$defs` bodies are covered too (added a doubly-nested fixture for that case). Your test is included as-is and passes; full unit suite green locally (406 tests).

thechriswebb:
LGTM. Thanks, @jyauxi!

davidkoski (approved):
Thank you for the fix!

PR #435, opened 2026-07-16
Prepare tool-calling input through the model's processor (fixes #433)

### LINKED ISSUE #433: [BUG] MLXFoundationModels: fatal "SmallVector out of range" (mlx-c array.cpp:335) when .toolCalling is declared on a VLM-loaded model — with or without an image in the prompt
**Describe the bug**

Any session on a VLM-container model with `capabilities: [.toolCalling]` (or `[.vision, .toolCalling]`) and ≥1 registered tool dies with a **fatal, process-killing** C++ abort the moment the constrained tool-calling turn starts:

```
MLX/ErrorHandler.swift:345: Fatal error: SmallVector out of range.
  at .../mlx-c/mlx/c/array.cpp:335
```

100% reproducible, ~1s in. Not catchable (ErrorHandler `fatalError`s on a dispatch queue; the originating frames are lost across the C boundary).

**Isolation matrix** (same session code, one variable changed per row):

| Model | Capabilities | Image in prompt | Result |
|---|---|---|---|
| Qwen3-8B (LLM factory) | `[.toolCalling]` | no | ✅ tool fires |
| Qwen3-VL-4B (VLM factory) | `[.vision]` | yes | ✅ |
| Qwen3-VL-4B (VLM factory) | `[.vision, .guidedGeneration]` | yes | ✅ schema-valid output |
| Qwen3-VL-4B (VLM factory) | `[.vision, .toolCalling]` | yes | 💥 fatal |
| Qwen3-VL-4B (VLM factory) | `[.toolCalling]` | **no** | 💥 fatal (same abort) |

So the trigger is the **VLM container + the constrained tool-calling path**, independent of whether an image is actually present. Guided generation over the same VLM works, and the same tool works on a text model — which points at the tool-envelope grammar/mask path specifically when the model came through the VLM factory (vocab/logit dimensioning?), though we have not root-caused it (the ErrorHandler swallows the origin frames).

**To Reproduce**

```swift
struct ReportColorTool: Tool {
    let name = "report_color"
    let description = "Report the dominant color you observe"
    @Generable struct Arguments {
        @Guide(description: "The color, lowercase") var color: String
    }
    func call(arguments: Arguments) async throws -> String { "recorded" }
}

let model = /* MLXLanguageModel over mlx-community/Qwen3-VL-4B-Instruct-4bit,
               capabilities: [.toolCalling], loaded via VLMModelFactory */
let session = LanguageModelSession(model: model, tools: [ReportColorTool()])
_ = try await session.respond(to: "Use report_color to report: blue.")   // → fatal abort
```

Model: `mlx-community/Qwen3-VL-4B-Instruct-4bit` — the same model `VisionIntegrationTests` pins — loaded via `VLMModelFactory.shared.loadContainer(...)`.

Backtrace (originating frames are lost across the C boundary — this is where it bottoms out):

```
* thread #21, name = 'Task 7', queue = 'com.apple.root.default-qos.cooperative',
  stop reason = Fatal error: SmallVector out of range.
  at .../mlx-swift/Source/Cmlx/mlx-c/mlx/c/array.cpp:335
  frame #4: FMSmoke`closure #1 in ErrorHandler.dispatch(...,
    message="SmallVector out of range. at .../mlx-c/mlx/c/array.cpp:335")
    at ErrorHandler.swift:345:21 [opt] [inlined]
->  345   fatalError(message)
```

**Expected behavior**

Either the tool-calling turn works on a VLM-loaded model, or the unsupported combination is rejected with a typed, catchable error at session/respond time — not a fatal process abort.

 - OS Version: macOS 27.0 beta (26A5378n), Xcode 27 beta
 - Device: M4 Max MacBook/Studio-class, Apple silicon
 - Version: main @ f1573a93 (`MLXFoundationModels`)

**Additional context**

- Building the adapter on this beta seed needs `SamplingMode.Kind` case names `randomTopK`/`randomProbabilityThreshold` — the tree at f1573a93 matches `.top`/`.nucleus` from a newer seed; 2-line local rename, unrelated to this crash.
- `VisionIntegrationTests` only declares `[.vision]`, so this combination appears uncovered by the current test suite.
- On this revision, hitting this crash requires first working around the `$defs` envelope issue reported separately (#432) — with an unpatched converter, non-trivial tools die earlier with `constraintCompilationFailed`; with flat single-field tools (as above) the crash reproduces directly.
- Happy to run diagnostics on our setup — the repro is one command in our harness and fully deterministic.

This investigation was AI-assisted; every reproduction in this report was run and verified on hardware.

jyauxi:
Fixes #433.

## Root cause

The tool-calling path in `MLXFoundationModels` hand-built its model input from raw `applyChatTemplate` output:

```swift
let toolAwareInput = LMInput(tokens: MLXArray(toolAwareTokens))   // 1-D [N]
```

That 1-D token array happens to work for text models — the default `LLMModel.prepare` slices with `[.newAxis, ...]` and `GuidedGenerationLoop` batches the returned tokens — but every VLM `prepare` implementation consumes `input.text.tokens` as given and indexes `dim(1)` (e.g. `getRopeIndex` in Qwen3VL). On a 1-D array that call reaches `mlx_array_dim` → `shape.at(1)` → an uncatchable `SmallVector out of range` abort (`array.cpp:335` in mlx-swift 0.31.6 is the `mlx_error` call inside `mlx_array_dim`'s catch block, matching the reported crash line exactly).

The mechanism was confirmed in isolation: a random-weight micro `Qwen3VL` given `prepare` input shaped `[1, 8]` succeeds; the identical model given `[8]` reproduces the exact fatal from the issue. This also explains the issue's isolation matrix — the crash is per-model-family (VLM container), independent of whether an image is present, because it happens during prefill.

The same hand-built path had a second defect: it re-templated the transcript text-only, so any image/video content was silently dropped from tool-calling prompts.

## Fix

Route the tool-aware prompt through `context.processor.prepare(UserInput(chat:tools:additionalContext:))` — the same path the unconstrained and guided-generation flows already use. Each model family's processor then produces the token rank its `prepare` expects (LLM processors emit 1-D, VLM processors emit `[1, N]`), and vision content in the transcript reaches the model during tool calls.

The think-then-call Phase-2 continuation (tool-aware prompt + Phase-1 reasoning token IDs) is built by a new `continuationInput` helper that appends along the last axis, preserving the prompt's rank, dtype, and processed image/video — unit-tested in `ContinuationInputTests` (rank-1, rank-2, dtype preservation, media carry-through).

## Verification

- `MLXFoundationModelsTests`: green (17 suites, including the 4 new tests).
- Live repro from #433 (`mlx-community/Qwen3-VL-4B-Instruct-4bit`, `[.toolCalling]` and `[.vision, .toolCalling]`, with and without an image): previously 100%-fatal, now completes with the tool firing correctly; text-model tool calling unchanged.
- Production-shaped soak: 78 documents through a single session declaring `[.vision, .toolCalling, .guidedGeneration]` on the same VLM — zero crashes, and tool submissions carry verbatim image content (confirming the image-drop fix).
- Think-then-call Phase-2 exercised end-to-end on both ranks: Qwen3-8B (`[.toolCalling, .reasoning]`, rank-1) and Qwen3-VL-4B-Thinking (`[.toolCalling, .reasoning]`, rank-2) — reason → constrained tool call, correct results. The rank-2 shape was unreachable before this fix (fatal in prefill).

One note for reviewers: `VisionIntegrationTests` currently only declares `[.vision]`, so the VLM+toolCalling intersection has no upstream coverage; the new unit tests cover the input-construction contract, and I'm happy to add a device-only integration test for the intersection if you'd like it in this PR.

This investigation and fix were AI-assisted; every verification above was run on hardware (M4 Max, macOS 27.0 beta, Xcode 27 beta).

### DIFF
--- a/Libraries/MLXFoundationModels/MLXLanguageModel.swift
+++ b/Libraries/MLXFoundationModels/MLXLanguageModel.swift
@@ -1064,7 +1064,7 @@ public struct MLXLanguageModel: FoundationModels.LanguageModel, Sendable {
                         let allTools =
                             Array(request.enabledToolDefinitions) + [finalAnswerDef]
 
-                        // Re-tokenize using the model's native tool-aware chat
+                        // Re-render using the model's native tool-aware chat
                         // template (Qwen/Llama/Phi/Gemma all ship one in their
                         // tokenizer_config.json). This is what teaches the model
                         // *what* tools exist and how to decide between them; the
@@ -1072,8 +1072,6 @@ public struct MLXLanguageModel: FoundationModels.LanguageModel, Sendable {
                         // whatever tool call it emits.
                         let toolSpecs = try ToolCallingConversions.makeToolSpecs(
                             from: allTools)
-                        let tokenizerMessages = DefaultMessageGenerator().generate(
-                            messages: messages)
 
                         // Think-then-call is gated to the enable_thinking
                         // family (Qwen3/QwQ): their template both renders the tool
@@ -1098,12 +1096,18 @@ public struct MLXLanguageModel: FoundationModels.LanguageModel, Sendable {
                                 forThinkingEnabled: Self.thinkingEnabled(
                                     for: request.contextOptions.reasoningLevel))
                         }
-                        let toolAwareTokens = try context.tokenizer.applyChatTemplate(
-                            messages: tokenizerMessages,
-                            tools: toolSpecs,
-                            additionalContext: reasoningContext
-                        )
-                        let toolAwareInput = LMInput(tokens: MLXArray(toolAwareTokens))
+                        // Prepare through the model's UserInputProcessor (like the
+                        // unconstrained and guided paths) instead of hand-building
+                        // an LMInput from raw applyChatTemplate output: processors
+                        // produce the token rank their model family requires (LLM
+                        // processors emit [N]; VLM processors emit [1, N], and VLM
+                        // `prepare` fatally aborts on 1-D input), and they carry
+                        // image/video content through to the model.
+                        let toolAwareInput = try await context.processor.prepare(
+                            input: UserInput(
+                                chat: messages,
+                                tools: toolSpecs,
+                                additionalContext: reasoningContext))
 
                         let toolCallingGrammar =
                             try SchemaConverter.encodeToolCallingGrammar(
@@ -1195,8 +1199,8 @@ public struct MLXLanguageModel: FoundationModels.LanguageModel, Sendable {
                         let phase2Input =
                             reasoningTokenIDs.isEmpty
                             ? toolAwareInput
-                            : LMInput(
-                                tokens: MLXArray(toolAwareTokens + reasoningTokenIDs))
+                            : Self.continuationInput(
+                                from: toolAwareInput, appending: reasoningTokenIDs)
                         // Shared budget (match the unconstrained path): the
                         // envelope continues under the remaining budget, floored
                         // at the completion reserve so it always has room to close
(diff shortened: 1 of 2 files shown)


davidkoski:
@thechriswebb FYI

thechriswebb:
LGTM 🚢

davidkoski (approved):
Changes look good, thank you.

PR #436, opened 2026-07-16
Add cross-request LRU prompt cache (PromptTrie + byte-bounded eviction)

GoodOlClint:
## Proposed changes

Swift only reuses a KV cache within a single `ChatSession`. A new request whose prompt shares a prefix with an earlier one re-processes that prefix from scratch. Python `mlx_lm` has had a cross-request store for exactly this — `PromptTrie` + `LRUPromptCache` in `mlx_lm/models/cache.py`, used by `server.py` — and this ports it to Swift.

### What's added

- **`KVCache.nbytes`** (`Libraries/MLXLMCommon/KVCache.swift`) — byte accounting, the prerequisite for any byte-bounded policy. Every Python cache implements `nbytes`; no Swift cache did. The basis is the *allocated* (step-rounded) buffers from `innerState()`, not the offset-sliced `state`, matching Python's `KVCache.nbytes`.

  It is declared as an `open var` on `BaseKVCache` rather than left to the protocol-extension default, so subclasses get a dynamically-dispatched override point. (A protocol-extension default becomes the witness at `BaseKVCache`'s conformance and a subclass property would only shadow it statically.)
- **`PromptTrie`** — prefix trie returning the nearest stored sequence: `exact` / `shorter` / `longer`, plus the common-prefix length.
- **`LRUPromptCache`** — the store itself: `fetchNearestCache` (returns a deep copy plus the token remainder still to process), prefix trim-reuse, `maxSize` / `maxBytes` eviction, type-aware eviction ordering (`assistant` → `user` → `system`, so system prompts survive longest), and `statsByType()`.

### Two Python bugs fixed at port time

The port deliberately does not inherit two bugs still open in the Python implementation — reported in ml-explore/mlx-lm#1495 and fixed there by ml-explore/mlx-lm#1496:

1. A value stored at a single-token prefix never matched (`search` used `> 0` where `>= 0` is correct).
2. Eviction ignored fetch recency, making the store FIFO rather than truly LRU.

Both are pinned by regression tests (`PromptTrieTests`, `LRUPromptCacheTests`).

### Scope / notes

- **No existing behavior changes.** This adds new types plus one protocol requirement that ships with a default implementation. Nothing is wired into `ChatSession` / `Evaluate` — that is deliberate, to keep this reviewable on its own and to stay out of the way of #370, which approaches prefix reuse at the `ChatSession` level. The two are complementary rather than competing: #370 adds `supportsStaticPrefixReuse` and this adds `nbytes`; both touch only the `KVCache` protocol block, so they should merge cleanly in either order.
- `LRUPromptCache` is a non-`Sendable` `final class`, mirroring Python's single-server-thread assumption. Cross-task consumers wrap it in `SerialAccessContainer`, as `ChatSession` already does elsewhere.

## Checklist

### DIFF
--- a/Libraries/MLXLMCommon/PromptCache/LRUPromptCache.swift
+++ b/Libraries/MLXLMCommon/PromptCache/LRUPromptCache.swift
@@ -1,0 +1,219 @@
+// Copyright © 2025 Apple Inc.
+
+import Foundation
+
+/// Cross-request LRU prompt cache: keeps recently-used KV caches keyed by
+/// `(model, tokens)` and, for a new prompt, returns the nearest reusable cache
+/// plus the token remainder still to be processed.
+///
+/// Direct port of Python `mlx_lm`'s `LRUPromptCache` (`mlx_lm/models/cache.py`).
+/// It also carries fixes for two bugs still open in the Python implementation,
+/// reported in ml-explore/mlx-lm#1495 and fixed there by ml-explore/mlx-lm#1496:
+///
+/// - a value stored at a single-token prefix is now returned
+///   (``PromptTrie/search(model:tokens:)`` uses `lastIndex >= 0`).
+/// - fetching a cache refreshes its LRU recency, so the store is truly
+///   least-recently-*used* rather than FIFO. See ``fetchNearestCache(model:tokens:)``.
+///
+/// Eviction is type-aware: entries are ordered `assistant` → `user` → `system`,
+/// and the eviction policy sheds the most-populous higher-churn class first, so
+/// system prompts survive longest.
+///
+/// This is a plain non-`Sendable` `final class`, mirroring Python's
+/// single-server-thread assumption. Cross-task consumers must wrap an instance
+/// in a `SerialAccessContainer` (see `Utilities/SerialAccessContainer.swift`).
+public final class LRUPromptCache<Model: Hashable> {
+
+    /// One stored cache plus its byte size and type.
+    public struct CacheEntry {
+        public var promptCache: [KVCache]
+        public var nbytes: Int
+        public var cacheType: String
+    }
+
+    /// Per-type sequence statistics returned by ``statsByType()``.
+    public struct TypeStats {
+        public let sequences: Int
+        public let bytes: Int
+    }
+
+    /// Type-partitioned recency lists. Newest is appended; oldest is `first`.
+    private final class CacheOrder {
+        let ordering: [String]
+        var lrus: [String: [(Model, [Int])]]
+
+        init(ordering: [String] = ["assistant", "user", "system"]) {
+            self.ordering = ordering
+            self.lrus = Dictionary(uniqueKeysWithValues: ordering.map { ($0, []) })
+        }
+
+        var count: Int { lrus.values.reduce(0) { $0 + $1.count } }
+
+        func count(ofType cacheType: String) -> Int { lrus[cacheType]?.count ?? 0 }
+
+        func push(model: Model, tokens: [Int], cacheType: String = "assistant") {
+            lrus[cacheType, default: []].append((model, tokens))
+        }
+
+        func remove(model: Model, tokens: [Int]) {
+            for cacheType in ordering {
+                if let idx = lrus[cacheType]!.firstIndex(where: {
+                    $0.0 == model && $0.1 == tokens
+                }) {
+                    lrus[cacheType]!.remove(at: idx)
+                    return
+                }
+            }
+        }
+
+        /// Evict the least-recently-used entry from the most-populous
+        /// higher-churn class (assistant before user before system).
+        func pop() -> (Model, [Int]) {
+            var i = 0
+            while i + 1 < ordering.count {
+                let a = lrus[ordering[i]]!
+                let b = lrus[ordering[i + 1]]!
+                if !a.isEmpty && a.count >= b.count {
+                    return lrus[ordering[i]]!.removeFirst()
+                }
+                i += 1
+            }
+            return lrus[ordering[i]]!.removeFirst()
+        }
+    }
+
+    public let maxSize: Int
+    public let maxBytes: Int
+
+    private let trie = PromptTrie<Model, CacheEntry>()
+    private let lru = CacheOrder()
+    private var totalBytes = 0
+    private var bytesByType: [String: Int]
+
+    public init(maxSize: Int = 10, maxBytes: Int = Int.max) {
+        self.maxSize = maxSize
+        self.maxBytes = maxBytes
+        self.bytesByType = Dictionary(uniqueKeysWithValues: lru.ordering.map { ($0, 0) })
+    }
+
+    /// Number of stored sequences.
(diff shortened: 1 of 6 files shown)


Libraries/MLXLMCommon/PromptCache/LRUPromptCache.swift
```diff
@@ -0,0 +1,219 @@
...
+///   least-recently-*used* rather than FIFO. See ``fetchNearestCache(model:tokens:)``.
+///
+/// Eviction is type-aware: entries are ordered `assistant` → `user` → `system`,
+/// and the eviction policy sheds the most-populous higher-churn class first, so
+/// system prompts survive longest.
+///
+/// This is a plain non-`Sendable` `final class`, mirroring Python's
+/// single-server-thread assumption. Cross-task consumers must wrap an instance
+/// in a `SerialAccessContainer` (see `Utilities/SerialAccessContainer.swift`).
+public final class LRUPromptCache<Model: Hashable> {
```
davidkoski, line 25: What would be the type for `Model` in a typical system?  A String for the hugginface model id?

Libraries/MLXLMCommon/PromptCache/LRUPromptCache.swift
```diff
@@ -0,0 +1,219 @@
...
+        public let sequences: Int
+        public let bytes: Int
+    }
+
+    /// Type-partitioned recency lists. Newest is appended; oldest is `first`.
+    private final class CacheOrder {
+        let ordering: [String]
+        var lrus: [String: [(Model, [Int])]]
+
+        init(ordering: [String] = ["assistant", "user", "system"]) {
```
davidkoski, line 45: I wonder if the ordering should be an enum?  This is all private and nothing seems to supply a custom order.  Some functions, like `push()`, take a cacheType (which looks like a member of `ordering`).  Other functions, like `remove()`, enumerate the fixed `ordering` and would ignore other values (if present).

davidkoski: I think this is meant to match `Chat.Message.Role`, right?  `tool` is present there.

Libraries/MLXLMCommon/PromptCache/LRUPromptCache.swift
```diff
@@ -0,0 +1,219 @@
...
+            return (entry.promptCache.map { $0.copy() }, Array(tokens[shortLength...]))
+        }
+
+        return (nil, tokens)
+    }
+
+    /// Insert `cache` at `tokens` for `model`, updating byte accounting, purging
+    /// now-redundant prefixes of a trimmable cache, and evicting to satisfy the
+    /// size/byte limits.
+    public func insertCache(
```
davidkoski, line 146: I think this has to make a copy of the KVCache on the way in -- these are mutable reference types and will move forward as they are used (mutating what lives in the cache).

`main` has moved ahead quite a bit on KVCache -- I wonder if this would be better using `KVCacheStorage`?

Libraries/MLXLMCommon/PromptCache/LRUPromptCache.swift
```diff
@@ -0,0 +1,219 @@
...
+    /// Every hit refreshes the matched entry's LRU recency (upstream fix ml-explore/mlx-lm#1496).
+    public func fetchNearestCache(model: Model, tokens: [Int]) -> (
+        cache: [KVCache]?, remainder: [Int]
+    ) {
+        let result = trie.search(model: model, tokens: tokens)
+
+        if let exact = result.exact {
+            let entry = trie.get(model: result.model, tokens: exact)
+            refreshRecency(model: result.model, tokens: exact, cacheType: entry.cacheType)
+            return (entry.promptCache.map { $0.copy() }, [])
```
davidkoski, line 118: This may not work with LLMModel.prepare -- it will return `.tokens(y)` where y is this empty array.  The callers may not be able to handle that.  It could be an issue in the LLMModel/token code or perhaps an issue here.  The code below seems to leave one token in place: `prefix = min(tokens.count - 1, result.commonPrefix)`

Libraries/MLXLMCommon/PromptCache/LRUPromptCache.swift
```diff
@@ -0,0 +1,219 @@
...
+/// system prompts survive longest.
+///
+/// This is a plain non-`Sendable` `final class`, mirroring Python's
+/// single-server-thread assumption. Cross-task consumers must wrap an instance
+/// in a `SerialAccessContainer` (see `Utilities/SerialAccessContainer.swift`).
+public final class LRUPromptCache<Model: Hashable> {
+
+    /// One stored cache plus its byte size and type.
+    public struct CacheEntry {
+        public var promptCache: [KVCache]
```
davidkoski, line 29: This probably needs `LMOutput.State` as well -- quite a few models are using this for per-token state management and it is expected to be passed in for the next token.  If we broke this on turns I think we would be OK.

Libraries/MLXLMCommon/PromptCache/LRUPromptCache.swift
```diff
@@ -0,0 +1,219 @@
...
+    /// Total bytes held across all stored caches.
+    public var nbytes: Int { totalBytes }
+
+    /// Fetch the nearest reusable cache for `tokens`, returning a deep copy plus
+    /// the token remainder the caller still needs to process. Returns `(nil,
+    /// tokens)` on a total miss.
+    ///
+    /// Every hit refreshes the matched entry's LRU recency (upstream fix ml-explore/mlx-lm#1496).
+    public func fetchNearestCache(model: Model, tokens: [Int]) -> (
+        cache: [KVCache]?, remainder: [Int]
```
davidkoski, line 111: For VLMs we would need to be careful _which_ tokens we were processing.  Would it hook in right after the `applyChatTemplate`?  That is `[Int]`:

```swift
    public func prepare(input: UserInput) async throws -> LMInput {
        // Use structured content message generator for Gemma3's chat template
        let messages = Qwen2VLMessageGenerator().generate(from: input)

        var promptTokens = try tokenizer.applyChatTemplate(
            messages: messages, tools: input.tools,
            additionalContext: input.additionalContext)
```

Or would it consume `LMInput`, which has `<image>` tokens in it according to the image size, but not the contents.  The `processedImage` is separate from the tokens and is mixed in (potentially) here:

```swift
    func prepare(_ input: LMInput, cache: [KVCache], state: LMOutput.State?, windowSize: Int?)
        throws -> PrepareResult
```

This isn't integrated so I can't tell for sure how it fits in, but since it comes from mlx-lm (LLMs only) I wonder if it considers this case?

davidkoski (changes requested):
See comments.  I think there are a lot of edge cases that need covering here (and some of them are new since you made the PR -- my fault for not reviewing earlier).  The VLM is likely trouble and LMOutput.State might be an issue.  For sure this needs to store copies in the cache.

If you are interested in moving this forward, please rebase and take a look at KVCacheStorage and friends.

Thanks!

PR #437, opened 2026-07-16
Make BaseKVCache.ropeOffset overridable

GoodOlClint:
## Proposed changes

`BaseKVCache` does not declare `ropeOffset`, so the `KVCache` extension default (`.scalar(offset)`) is bound as the protocol witness at its conformance. A subclass that declares a property of the same name only shadows it statically — every access through a `KVCache` reference still resolves to `.scalar(offset)` and silently ignores the subclass.

That reference is the normal access path: models read the offset off a `KVCache`, e.g.

```swift
let offset = cache?.ropeOffset
```

Nothing in the tree overrides `ropeOffset` today, so **this changes no current behavior**. But `BaseKVCache` is `open`, and this is a trap for any cache that needs a non-scalar offset: a batched cache returning a per-row `.batch(...)` compiles cleanly, is then ignored, and every row silently gets the same scalar offset. That is a wrong-numbers bug with no compile-time signal, and it is the kind of cache several in-flight PRs need (e.g. #263's batched caches, and custom caches along the lines of #232 / #329).

### Fix

Declare `ropeOffset` as an `open` class property on `BaseKVCache` with the same `.scalar(offset)` default, so it is the real witness and subclasses get a dynamically-dispatched override point.

This is the same reason `state`, `isTrimmable`, and `update(keys:values:)` are already declared on `BaseKVCache` rather than left to extension defaults.

### Test

The test subclasses `BaseKVCache`, overrides `ropeOffset` with a per-row `.batch(...)`, and reads it back through a `KVCache` reference — the path models actually use.

It discriminates in both directions:

- **without** this change it does not compile — `error: property does not override any property from its superclass`, which is precisely the defect being fixed (you cannot override it at all);
- **with** it, the override is honored through the protocol reference and the test passes.

## Checklist

- [ ] I have updated the necessary documentation (if needed) — no doc changes needed; the new property is documented inline

### DIFF
--- a/Libraries/MLXLMCommon/KVCache.swift
+++ b/Libraries/MLXLMCommon/KVCache.swift
@@ -174,6 +174,10 @@ open class BaseKVCache: KVCache {
     public var offset: Int = 0
     public var maxSize: Int? { nil }
 
+    /// RoPE offset for this cache. `open` so subclasses can return a non-scalar
+    /// offset (e.g. a batched cache's per-row `.batch(...)`).
+    open var ropeOffset: RoPEOffset { .scalar(offset) }
+
     public func innerState() -> [MLXArray] { [] }
 
     open func update(keys: MLXArray, values: MLXArray) -> (MLXArray, MLXArray) {
--- a/Tests/MLXLMTests/KVCacheTests.swift
+++ b/Tests/MLXLMTests/KVCacheTests.swift
@@ -753,3 +753,28 @@ func preservesScoreDtype() {
         #expect(out.asType(.float32).sum().item(Float.self).isFinite)  // no -inf → NaN
     }
 }
+
+// MARK: - ropeOffset overridability
+
+/// A `BaseKVCache` subclass reporting a per-row RoPE offset, as a batched cache does.
+private final class BatchOffsetProbeCache: BaseKVCache {
+    override var ropeOffset: RoPEOffset { .batch(MLXArray([10, 20])) }
+
+    override func update(keys: MLXArray, values: MLXArray) -> (MLXArray, MLXArray) {
+        (keys, values)
+    }
+}
+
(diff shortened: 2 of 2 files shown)


Libraries/MLXLMCommon/KVCache.swift
```diff
@@ -174,6 +174,17 @@ open class BaseKVCache: KVCache {
...
 
+    /// RoPE offset for this cache. Declared as an `open` class property rather than
+    /// left to the ``KVCache`` extension default so that subclasses can override it
+    /// with dynamic dispatch — e.g. a batched cache returning a per-row `.batch(...)`.
+    ///
+    /// Without this declaration the extension default becomes the protocol witness at
+    /// `BaseKVCache`'s conformance, and a subclass property of the same name only
+    /// shadows it statically: any access through a ``KVCache`` reference (which is how
+    /// models read it) still resolves to `.scalar(offset)`, silently ignoring the
+    /// subclass.
```
davidkoski: Change looks good, but this comment might be better as a PR comment than a comment in the code -- I think this is exactly the property that people would expect.  See what you think.

GoodOlClint: Good call — trimmed it to a one-liner in fdd4fbd. Agreed it's the property people would expect; here's the "why `open`" for the record:

`BaseKVCache` didn't declare `ropeOffset` at all, so the `KVCache` extension default (`.scalar(offset)`) is bound as the witness at conformance. A subclass declaring its own `ropeOffset` only shadows it statically — reads through a `KVCache` reference (how models get it, `cache?.ropeOffset`) still resolve to `.scalar`, so the override is silently ignored. Declaring it `open` on `BaseKVCache` makes it the real, dynamically-dispatched witness. Same reason `state` / `isTrimmable` / `update` are already declared there rather than left to extension defaults.

davidkoski (approved):
Looks good, thank you!

PR #438, opened 2026-07-17
Fix FoundationModels API drift and the integration tests that no longer compiled

thechriswebb:
## Summary

The current FoundationModels SDK (macOS, iOS, and visionOS 27) changed its generation API. Our MLXFoundationModels integration tests were written against the older API and no longer compiled against it. This change finishes adapting to the new API and gets the integration tests building and running again, without changing how the model behaves in a shipping app.

## Background

Two parts of the FoundationModels generation API moved:

- Some type and case names were renamed. The cherry-picked groundwork this branch builds on (see below) already started tracking those.
- The values the framework uses to stream a response (generated text, tool calls, usage, and metadata) became opaque. Code that receives them can see that something was produced but can no longer read what it was.

The integration tests depended on the old names and on reading those streamed values directly. After the API moved, they failed to compile, and even once the names were fixed they had no way left to check the actual generated content.

## What this changes

1. A test-only way to observe generation. Every point where the model sends an event to the framework now goes through one of a few shared helpers. Each helper does two things: it sends to the framework exactly as before, and, if a test has asked to observe, it also hands over a plain, readable copy of the same event. Nothing observes in a normal build, so the framework receives the identical calls it always did and runtime behavior is unchanged.

2. The test stream now hands tests these readable events. It also quietly reads and discards the framework's opaque events, which is still necessary so that sending into the framework does not stall.

3. The integration test suites (text generation, reasoning, tool calling, and guided generation) now build against the new API and read these readable events instead of the framework's opaque values.

4. A stale documentation comment is corrected.

## Effect on shipping code

None. Nothing observes generation unless a test opts in, and the calls made to the framework are unchanged. This is test-only support: it makes the existing tests compile and run again on the current SDK, and does not alter how the model generates for real users.

## Dependency on an earlier PR, and rebasing

~~This work builds directly on Charlie Le's PR #431, which begins adapting to the drifted FoundationModels API (notably the renamed SamplingMode case names). I cherry-picked its three commits at the start of this work, so everything here is stacked on top of them.~~

~~PR #431 has not yet merged. Until it does, this branch should be reviewed and merged relative to #431, not main. Once #431 lands, this branch will most likely need to be rebased onto main before it can go in.~~

Given that #431 was just merged, I've since rebased. This should be good-to-go.

## Testing

I ran the MLXFoundationModels integration suites locally to validate.

### DIFF
--- a/IntegrationTesting/IntegrationTestingTests/MLXFoundationModelsIntegration/GuidedGeneration/GenerableRoundTripTests.swift
+++ b/IntegrationTesting/IntegrationTestingTests/MLXFoundationModelsIntegration/GuidedGeneration/GenerableRoundTripTests.swift
@@ -33,10 +33,8 @@ struct GenerableRoundTripTests {
         let stream = try await executeResponse(executor, request: request, model: model)
         var text = ""
         for try await event in stream {
-            if let response = event as? LanguageModelExecutorGenerationChannel.Response,
-                case .appendText(let delta) = response.action
-            {
-                text += delta.content
+            if case .appendText(let chunk, _, .response) = event {
+                text += chunk
             }
         }
         return text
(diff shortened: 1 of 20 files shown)


davidkoski (approved):
Changes look good, thank you!

PR #442, opened 2026-07-18
Optimize Qwen3.5 interleaved M-RoPE

aleroot:
## Proposed changes

In an Instruments profile of Qwen3.5 generation, `Qwen35Language.RotaryEmbedding.applyInterleavedMRope` accounted for approximately 1.84 billion cycles, or 11% of the generation worker’s CPU cycles. The implementation selected each frequency dimension using repeated `MLXArray` subscripting and then reconstructed the result with `stacked`, making `MLXArray.subscript` and `getItemND` significant hotspots.

This PR:

- Precomputes the M-RoPE modality indices once during `RotaryEmbedding` initialization.
- Replaces the per-dimension slicing loop and `stacked` operation with a single broadcasted `takeAlong` operation.
- Removes 32 individual frequency-column subscripts per invocation for the standard 64-dimensional Qwen3.5 configuration.
- Preserves the existing fallback behavior for missing M-RoPE section configuration.
- Adds regression tests comparing the optimized implementation against the previous slicing algorithm for standard, truncated, and fallback configurations.

The complete unit-test suite passes, including all existing Qwen3.5 continuation tests.

## Checklist

### DIFF
--- a/Libraries/MLXVLM/Models/Qwen35.swift
+++ b/Libraries/MLXVLM/Models/Qwen35.swift
@@ -222,36 +222,28 @@ enum Qwen35Language {
 
     final class RotaryEmbedding {
         private let invFreq: MLXArray
-        private let mropeSection: [Int]
+        private let mropeIndices: MLXArray
 
         init(dim: Int, base: Float, mropeSection: [Int]) {
             let safeDim = max(1, dim)
             var freq = MLXArray(stride(from: 0, to: safeDim, by: 2)).asType(.float32)
             freq = freq / Float(safeDim)
             self.invFreq = 1.0 / pow(MLXArray(base), freq)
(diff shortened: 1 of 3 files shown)


NivDvir (commented):
Went through both hunks — as far as I can tell the precomputed map reproduces the old selection exactly (the dim-1/dim-2 index classes are disjoint mod 3, so write order can't collide), and it stays off the state-handoff surface from #399/#420. One small thought: since `takeAlong` is a pure gather, the tests could probably assert exact equality rather than `atol: 1e-6` — a nice way to document that this can't perturb temp-0 decoding. Nice speedup!

davidkoski (approved):
Change looks good -- nice find!

PR #444, opened 2026-07-19
Add Poolside Laguna (XS 2.1) model

anupsv:
Adds MLXLLM support for the `laguna` model type: **Poolside Laguna XS 2.1**, a 33B-total / ~3B-active Mixture-of-Experts decoder for agentic coding. Verified against the [`poolside/Laguna-XS-2.1-NVFP4-mlx`](https://huggingface.co/poolside/Laguna-XS-2.1-NVFP4-mlx) checkpoint. Model math follows the mlx-lm reference (PRs [#1223](https://github.com/ml-explore/mlx-lm/pull/1223) / [#1334](https://github.com/ml-explore/mlx-lm/pull/1334)).

## Architecture (`Libraries/MLXLLM/Models/Laguna.swift`)

- **Attention** — GQA with per-head QK-RMSNorm and a per-head **softplus** attention-output gate (`g_proj`, computed in float32). Query-head count is per-layer (48 on full-attention layers, 64 on sliding-attention layers); 8 KV heads, `head_dim` 128.
- **Mixed attention** — interleaved sliding-window (512) and full attention, each with its own RoPE: full-attention layers use **YaRN** with a 0.5 partial-rotary factor; sliding layers use plain RoPE (theta 10000, full rotary).
- **MoE** — sigmoid top-k router with an `e_score_correction_bias` (selection uses the biased scores, weighting uses the unbiased scores), `norm_topk_prob`, a routed scaling factor, and an always-on **shared expert**. Layer 0 is a dense SwiGLU MLP (`mlp_only_layers`); the rest are sparse (256 experts, top-8) via `SwitchGLU`.
- Untied LM head; per-layer KV cache (`KVCacheSimple` for full attention, `RotatingKVCache` for sliding).

## Quantization (NVFP4)

The released checkpoints are **NVFP4** (group size 16, 4-bit) on **only** the MoE expert and shared-expert projections (`switch_mlp.*` / `shared_expert.*`); attention, the router gate, embeddings, the LM head and the dense layer-0 MLP stay full precision. This uses mlx-swift's existing `QuantizationMode.nvfp4` — **no new kernels**.

Because an entire decoder layer (the dense layer 0) carries no quantized submodule, the standard whole-model `quantize(model:)` pass in `loadWeights` throws `UpdateError.unexpectedStructure(key: "layers", …)` while descending the `layers` array (its replacement entry for layer 0 is `.none`). To keep this fully within the model file, `LagunaModel.init` **pre-quantizes only the expert / shared-expert modules per layer**, so the loader's pass finds them already `Quantized` and has nothing left to convert (dense layer 0 and attention remain BF16, matching the checkpoint).

## Registration

Registers `"laguna"` in `LLMTypeRegistry` (`Libraries/MLXLLM/LLMModelFactory.swift`).

## Testing

- Builds against the pinned `mlx-swift` 0.31.6.
- Loads `poolside/Laguna-XS-2.1-NVFP4-mlx` and produces coherent, instruction-following greedy generation on Apple Silicon (M4 Pro, macOS 26). Example:
  > *"Write a Go function that prints the first 10 Fibonacci numbers."* → a correct explanation + approach, terminating cleanly on EOS.
- This is a functional smoke test (loads + coherent generation), not a numeric teacher-forced parity check against the Python reference; a logit/argmax comparison vs. `mlx_lm` would be a good follow-up. Tokenizer / chat-template / tool-call-parser plumbing is out of scope for the model file.

### DIFF
--- a/Libraries/MLXLLM/LLMModelFactory.swift
+++ b/Libraries/MLXLLM/LLMModelFactory.swift
@@ -51,6 +51,7 @@ public enum LLMTypeRegistry {
         "openelm": create(OpenElmConfiguration.self, OpenELMModel.init),
         "internlm2": create(InternLM2Configuration.self, InternLM2Model.init),
         "deepseek_v3": create(DeepseekV3Configuration.self, DeepseekV3Model.init),
+        "laguna": create(LagunaConfiguration.self, LagunaModel.init),
         "granite": create(GraniteConfiguration.self, GraniteModel.init),
         "granitemoehybrid": create(
             GraniteMoeHybridConfiguration.self, GraniteMoeHybridModel.init),
--- a/Libraries/MLXLLM/Models/Laguna.swift
+++ b/Libraries/MLXLLM/Models/Laguna.swift
@@ -1,0 +1,583 @@
+// Copyright © 2025 Apple Inc.
+
+// port of https://github.com/ml-explore/mlx-lm/blob/main/mlx_lm/models/laguna.py
+//
+// Laguna (Poolside Laguna XS 2.1): a Mixture-of-Experts decoder with GQA,
+// per-head QK-norm, per-head softplus attention output gating, a mix of
+// sliding-window and full-attention layers (each with its own RoPE), a sigmoid
+// top-k router with an `e_score_correction_bias`, and a shared expert. The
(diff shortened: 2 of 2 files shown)


anupsv:
Closing — opened against the wrong repository by mistake. Re-targeting the Layr-Labs/mlx-swift-lm fork.

PR #445, opened 2026-07-19
Fix MTP integration test skip semantics; auto-download 31B checkpoint pair

CharlieTLe:
## Proposed changes

Several checkpoint-gated tests used Issue.record(...) + return when the 31B target/drafter pair was not in the local HF cache. Swift Testing reports Issue.record as a failure, not a skip, so these tests failed whenever the (very large) checkpoints were not pre-downloaded. Convert them to auto-download via the same Downloader mechanism the rest of the integration suite already uses, matching the established pattern (e.g. Gemma4EseriesLoadIntegrationTests) instead of skipping.

Also documents a known-stale-fixture gap in MTPRung4TokenParityTests: case_02_block4 and case_03_block6 fail because the recorded fixture output predates a checkpoint update, not because of a Swift bug -- confirmed by cross-checking against a live re-run of the pinned mlx-vlm reference, which reproduces Swift's output bit-for-bit.

## Checklist


davidkoski (approved):
Change looks good, thank you!

Issue #443, opened 2026-07-18
`savePromptCache` / `loadPromptCache` drop `LMOutput.State` — a disk-restored VLM prompt cache decodes at wrong M-RoPE positions


NivDvir:
Splitting this out of #399, where @spokvulcan flagged it as a known limitation and @davidkoski suggested an issue.

**What happens:** `ChatSession.saveCache(to:)` matches `.kvcache(let cache, _, _)` and passes only the KV arrays to `savePromptCache` — the third component (`LMOutput.State?`, the M-RoPE rope deltas from the last prefill) is dropped. The safetensors layout has no slot for it, `loadPromptCache` returns only `([KVCache], metadata)`, and both cache-accepting `ChatSession` initializers hard-code `state: nil`. So the in-memory turn loop threads the state correctly since #399/#419, but any disk round-trip silently loses it.

**Why it matters:** Qwen2/2.5/3-VL compute decode positions as cache offset + rope delta. A restored image-heavy prefix without its state decodes at text-only positions — the same drift class as #419/#420, entering through the persistence path instead. At temp 0 on dense grounding prompts this can flip bbox output silently.

**Possible directions:** (a) encode `LMOutput.State` into the existing string-metadata groups (needs a stable serialization + format version bump), or (b) narrower: have `loadPromptCache` return it and add a `state:` parameter to the cache-accepting `ChatSession` inits. @spokvulcan offered in #399 to follow up on the file format — happy to coordinate, and I can run the warm-vs-cold grounding comparison from #420 against whatever lands here as well.

spokvulcan:
Thanks for splitting this out @NivDvir, and sorry for the slow follow up on the file format.

We've been running a disk-persisted prefix cache for Qwen3.5/3.6 VL in production for a while now (Tesseract, an offline on-device assistant with an OpenAI-compatible server on top of mlx-swift-lm), so I have opinions that come from having hit this. Short version of what we do: we never persist `LMOutput.State`. We reconstruct it. And I think that's the direction upstream should go too, for a reason that only shows up once the cache stores more than one prefix.

### What we do today

[...]

Issue #420, opened 2026-07-14
Qwen2.5-VL / Qwen3-VL: M-RoPE state is dropped across ChatSession turns, so warm continuations run at wrong positions


spokvulcan:
`ChatSession` keeps the KV cache warm across turns, but `LMOutput.State`
(which carries the M-RoPE `positionIds`/`ropeDeltas` since #239/#283)
dies with each turn's `TokenIterator`. On the next turn the Qwen VLM
position branches see a warm cache with no rope deltas and recompute
positions from zero. Every warm continuation (multi-turn chat, tool-call
restart, restored prompt cache) runs the new tokens at wrong positions.
It's the cross-turn sibling of the prefill-to-decode drop #419 fixed.

#399 fixes this for Qwen3.5/3.6 by threading `state:` through
`LanguageModel.prepare` (mechanism, measurements, and tests there: the
tiny-model repro shows 0.43 max-abs logits divergence vs a 8.3e-07 noise
floor). The `prepare(state:)` rails are already in place for every model;
what's left per family is the per-model wiring in the position
computation:

- [ ] Qwen2.5-VL
- [ ] Qwen2-VL
- [ ] Qwen3-VL

Happy to do these as a follow-up PR. A grounding comparison on real
weights (@NivDvir offered one in #399) would make a good acceptance
check.

NivDvir:
Offer stands. I'll run the grounding comparison — Qwen2.5-VL real weights, dense-image set, temp 0, warm-continuation vs cold — against each per-model wiring PR as it lands; ping me on the first one.

NivDvir:
The Qwen2.5-VL wiring was blocking us downstream, so I went ahead and prepared it — #448 wires Qwen2.5-VL and Qwen2-VL per the checklist (Qwen3-VL needs deepstack-aware chunking, left unticked). I also ran the grounding comparison I offered: real weights, dense-image set, temp 0, warm-continuation vs cold — on current `main` warm turns are 0/10 usable, with the wiring 10/10; details in the PR. Happy to reshape if you'd planned a different structure for these.

NivDvir:
All three boxes here are wired now — Qwen2.5-VL and Qwen2-VL in #448, and Qwen3-VL in #475, which also added `Qwen3VLContinuationTests` covering warm text and image continuation, mid-image resume, and windowed vs single-shot.

[...]

PR #448, opened 2026-07-20
Qwen2.5-VL / Qwen2-VL: windowed prefill + state-threaded warm continuation (wires #420)

NivDvir:
# Qwen2.5-VL / Qwen2-VL: windowed prefill + state-threaded warm continuation (wires #420)

## Proposed changes

#399 added windowed prefill and state-threaded warm cross-turn continuation to
Qwen3.5/3.6, fixing multi-turn M-RoPE drift for that family. It also widened
`LanguageModel.prepare` to thread an optional `LMOutput.State`, but the
remaining Qwen VLMs listed in #420 were only updated to *accept and ignore* that
state (`state _:`). This PR wires the same mechanism through **Qwen2.5-VL** and
**Qwen2-VL**, so a warm KV cache is continued from its anchor instead of
recomputing M-RoPE positions from zero.

Without this, a warm continuation (multi-turn chat, tool-call restart, restored
prompt cache) silently recomputes positions from scratch: the post-image text
tail is placed at `cacheOffset + arange` instead of `cacheOffset + ropeDelta +
arange`, dropping the delta the image accumulated. On dense single-image
grounding at temperature 0 this surfaces as degenerate output.

Mirroring #399 exactly (not a bespoke path):

- **`getRopeIndex` gains an offset-aware `positionOffset`** (default `0`), applied
  to the text-only positions, the image `llmPositions`, and the returned rope
  delta — so a new image's diverging t/h/w indices are computed from the warm
  cache's anchor rather than zero. Same shape as Qwen3-VL's `getRopeIndex`.
- **`prepare` now consumes the incoming `state`** and delegates to a new
  `prepareContinuation` when the cache is already warmed (`offset > 0`) or the
  prompt is longer than `windowSize` (default `512`). The continuation runs the
  vision tower + image→token merge once over the remainder, computes the
  remainder's M-RoPE positions once from the seeded position anchor (cache offset
  plus the rope delta carried in `state`), then drives the language model forward
  in chunks of `windowSize` — bounding the full-attention scratch to
  `[heads, chunk, L]` instead of `[heads, L, L]`. The windowed loop uses
  `asyncEval` per chunk + a trailing `eval`, matching the Gemma3 chunked-prefill
  pattern.
- The returned resume state carries `getRopeIndex delta − cacheOffset`, so a
  caller threading state end-to-end continues the following turn through the
  existing decode reconstruction branch. Decode stays caller-owned.
- The single-shot `prepare` path now passes token ids (rather than a
  pre-embedded tensor) on the text-only branch, so a 2-D `[1, seq]` text
  continuation — what a warm text turn produces — is handled without the prior
  `[.newAxis, .ellipsis]` rank inflation. Image single-shot behaviour is
  unchanged.

The vision-embedding preamble is factored into small shared helpers
(`gatherVisionInputs`, `mergedVisionEmbeds`) used by both `prepare` and
`prepareContinuation`. Batched inputs keep the single-shot path.

**Qwen3-VL is intentionally left for a follow-up.** Its `prepare` threads
per-layer deepstack features and a visual mask into the language model; a
windowed continuation would have to slice those in lockstep with the embeddings
and positions, which is beyond the drop-in Qwen3.5/2.5/2 pattern. Wiring it would
be structural surgery, so it is out of scope here and its #420 box stays
unchecked.

## Tests

New `Tests/MLXLMTests/Qwen25VLContinuationTests.swift` mirrors
`Qwen35ContinuationTests`, run against tiny random-weight configs for **both**
wired models (no downloads, CI-safe). Per model:

- **Warm text continuation ≡ full prefill** — prefix in cache, remainder
  prefilled on top; bounded against a token-by-token decode noise floor.
- **Warm image continuation ≡ full prefill** — image in turn 1, rope delta
  threaded into turn 2's prefill via `state`.
- **Image-mid continuation resume state** — three turns, image in the middle
  turn; the post-image resume state must position turn 3 correctly.
- **Windowed prefill ≡ single-shot** (plain text) — `windowSize: 8` vs one shot.
- **Windowed image prefill ≡ single-shot** — image straddling the window
  boundary, the hard case for chunked slicing.

All 10 pass. The full `MLXLMTests` suite is green via
`xcodebuild test -scheme mlx-swift-lm-Package -destination 'platform=macOS' -skipPackagePluginValidation`,
and `swift-format` is clean.

## #420 checklist

- [x] Qwen2.5-VL — windowed prefill + warm continuation wired
- [x] Qwen2-VL — windowed prefill + warm continuation wired
- [ ] Qwen3-VL — deferred (deepstack + visual-mask threading needs lockstep
  chunking; not a drop-in of the #399 pattern)

## Grounding validation (real weights)

Ran `mlx-community/Qwen2.5-VL-7B-Instruct-4bit` (temp 0) on 10 GUI-grounding rows
(3 canonical + 7 dense screenshots), each asked twice: COLD (fresh session, single
turn) and WARM (turn 2 of a live session whose turn 1 grounded a different image,
state threaded between turns).

| config | warm turns with a parseable box | warm empty/degenerate |
|---|---|---|
| `main` (343cae3) | 0/10 | 10/10 — empty fences, or prose describing the *cached turn-1 image* |
| this branch | **10/10** | 0/10 |

Warm coordinates track the cold single-shot within a few pixels on clean rows;
where they diverge the model is legitimately attending to turn-1 conversational
context — the controlled random-weight tests above are the positional-correctness
check (warm ≡ full prefill on logits), and this run is the end-to-end usability
check on real weights.

### DIFF
--- a/Libraries/MLXVLM/Models/Qwen25VL.swift
+++ b/Libraries/MLXVLM/Models/Qwen25VL.swift
@@ -931,62 +931,8 @@ public class Qwen25VL: Module, VLMModel, KVCacheDimensionProvider {
         self._languageModel.wrappedValue = Language.LanguageModel(config.textConfiguration)
     }
 
-    /// Builds the multimodal input embedding for one prefill step.
-    ///
-    /// Returns the embeddings paired with the prefill-only MROPE state
-    /// (positionIds + ropeDeltas) — both nil on the no-image path. The
-    /// caller seeds `LMOutput.State` with these so subsequent decode
-    /// steps can reconstruct positions from `ropeDeltas + cacheOffset`
-    /// without mutating the model.
-    private func inputEmbeddings(inputIds: MLXArray, pixelValues: MLXArray?, frames: [THW]?)
-        -> (embeds: MLXArray, positionIds: MLXArray?, ropeDeltas: MLXArray?)
-    {
-        guard let pixelValues, let frames else {
-            return (
-                languageModel.model.embedTokens(inputIds[.newAxis, .ellipsis]),
-                nil, nil
-            )
-        }
-
-        // Get the input embeddings from the language model
-        let inputEmbeds = languageModel.model.embedTokens(inputIds)
-
-        // Get the output hidden states from the vision model
-        var hiddenStates = self.visionModel(pixelValues, frames: frames)
-
-        if hiddenStates.ndim == 2 {
-            hiddenStates = hiddenStates[.newAxis, 0..., 0...]
-        }
-
-        // Insert special image tokens in the input_ids
-        let mergedEmbeds = QwenVL.mergeInputIdsWithImageFeatures(
-            inputIds: inputIds, inputEmbeds: inputEmbeds, imageFeatures: hiddenStates,
-            imageTokenId: config.baseConfiguration.imageTokenId,
-            videoTokenId: config.baseConfiguration.videoTokenId)
-
-        // Compute MROPE 3D position IDs for spatial awareness
-        let spatialMergeSize = config.visionConfiguration.spatialMergeSize
-        let inputIds2D = inputIds.ndim == 1 ? inputIds[.newAxis, 0...] : inputIds
-        let (positionIds, ropeDeltas) = Qwen25VL.getRopeIndex(
-            inputIds: inputIds2D,
-            imageGridTHW: frames,
-            videoGridTHW: nil,
-            spatialMergeSize: spatialMergeSize,
-            imageTokenId: config.baseConfiguration.imageTokenId,
-            videoTokenId: config.baseConfiguration.videoTokenId,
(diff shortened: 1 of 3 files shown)


NivDvir:
One thing worth calling out from the diff: this removes the note in `Qwen2VL.swift` that explained chunked prefill was intentionally omitted, because chunking the embeddings without slicing `positionIds` in lockstep would feed wrong M-RoPE positions. That concern is addressed rather than ignored — `prepareContinuation` slices positions together with the embeddings (`positionIds[0..., 0..., start ..< end]`), which is what makes the chunked path safe for M-RoPE here. Happy to restore a short version of that note in the code if you'd prefer it stay.

davidkoski (approved):
Change looks good to me, thank you!

NivDvir:
Thanks @davidkoski! That helps a lot — together with #475 this closes #443 across the whole Qwen family. Happy to follow up on anything post-merge.

PR #449, opened 2026-07-20
Load EOS token IDs from nested text configs

aleroot:
## Proposed changes

Composite model configurations such as Qwen3.5 store `eos_token_id` under `text_config`. `mlx-swift-lm` only read the root-level value, so `<|endoftext|>` (`248044`) was not included in the stop-token set. When the model emitted it, generation continued into subsequent chat-template content, producing output such as `<|endoftext|><|im_start|>user`.

This change:

- Reads `eos_token_id` from `text_config` when it is absent at the model root.
- Preserves root-level and `generation_config.json` precedence.
- Uses the resolved IDs for both LLM and VLM loading.
- Adds focused tests for nested EOS decoding and root-level precedence.

This follows the same configuration precedence used by Hugging Face Transformers and mlx-vlm: generation attributes fall back to the nested text configuration when they are absent at the model root.

## Checklist

### DIFF
--- a/Libraries/MLXLLM/LLMModelFactory.swift
+++ b/Libraries/MLXLLM/LLMModelFactory.swift
@@ -573,7 +573,7 @@ public final class LLMModelFactory: GenericModelFactory {
         }
 
         // Load EOS token IDs from config.json, with optional override from generation_config.json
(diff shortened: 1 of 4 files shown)


davidkoski (approved):
Change looks good, thank you!

PR #451, opened 2026-07-22
Replace fatalError with throwing KVCacheError in recoverable KV cache paths

JuanColilla:
## Why

Several KV cache paths crashed the whole process with `fatalError` on
conditions that are recoverable, not programmer bugs:

- `KVCacheSimple.toQuantized` / `RotatingKVCache.toQuantized` crash when the
  requested group size doesn't evenly divide the model's key/value head
  dimensions, or (for `RotatingKVCache`) unconditionally, since rotating-cache
  quantization isn't implemented yet.
- `restoreCacheFromMetaState` (used by `loadPromptCache`) had no validation
  of the array/metaState counts it reads out of a saved prompt-cache file,
  so a truncated or corrupted file crashed the process instead of failing
  the load.

A library shouldn't `fatalError` on attacker- or disk-corruption-controlled
input, or on a caller passing a group size that happens not to fit a given
model's head dimension — both are inputs the caller can reasonably want to
handle (skip quantization, reject a bad file, log and continue).

## What

- `KVCacheSimple.toQuantized` and `RotatingKVCache.toQuantized` now `throws
  KVCacheError` instead of calling `fatalError`, matching the existing
  throwing surface of `loadPromptCache`/`savePromptCache`.
- `restoreCacheFromMetaState` now validates array/metaState counts for
  `KVCacheSimple`, `RotatingKVCache`, `QuantizedKVCache`, and `ChunkedKVCache`
  before constructing the cache, throwing `KVCacheError` with a message
  identifying which field was short.
- `maybeQuantizeKVCache`'s public, non-throwing signature is unchanged: it
  now falls back to leaving a cache unquantized (`try?`) rather than
  propagating, since silently keeping the unquantized cache was already its
  behavior for other ineligibility reasons.

## How tested

- `Tests/MLXLMTests/KVCacheTests.swift`: new regression tests constructing
  hand-crafted corrupt prompt-cache files (wrong array count for
  `KVCacheSimple`/`RotatingKVCache`, wrong metaState count for
  `QuantizedKVCache`/`ChunkedKVCache`) and asserting `loadPromptCache` throws
  `KVCacheError` instead of crashing; a test asserting
  `RotatingKVCache.toQuantized()` throws; a test asserting
  `KVCacheSimple.toQuantized(groupSize:bits:)` throws on a head dimension
  that divides none of the supported group sizes (32/64/128).
- `swift build` compiles clean across every target.
- This branch was rebased on top of `main`'s current state, including the
  `TurboQuant` KV-cache-compression path merged after this change was first
  written; its one call site to the now-throwing `toQuantized` is updated in
  a second, clearly separated commit so the diff shows exactly what had to
  adapt and why.

### DIFF
--- a/Libraries/MLXLMCommon/KVCache.swift
+++ b/Libraries/MLXLMCommon/KVCache.swift
@@ -465,7 +465,12 @@ public class KVCacheSimple: BaseKVCache, CustomDebugStringConvertible {
     /// Convert to quantized cache for maximum efficiency
     ///
     /// Use `updateQuantized()` and `quantizedScaledDotProductAttention()` for zero-overhead operation.
-    public func toQuantized(groupSize: Int = 64, bits: Int = 4) -> QuantizedKVCache {
+    ///
+    /// - Throws: ``KVCacheError`` if `groupSize` is not compatible with the key/value head
+    ///   dimensions of the model that produced this cache (no supported group size among
+    ///   32/64/128 evenly divides both). Callers should treat this as a recoverable
+    ///   configuration error (e.g. fall back to an unquantized cache) rather than a crash.
+    public func toQuantized(groupSize: Int = 64, bits: Int = 4) throws -> QuantizedKVCache {
         if let keys = self.keys, let values = self.values {
             // Quantize the current keys and values
             let currentKeys = keys[.ellipsis, ..<offset, 0...]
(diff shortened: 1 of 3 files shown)


aleroot:
@JuanColilla please see my PR: https://github.com/ml-explore/mlx-swift-lm/pull/453 
just removing fatal errors is not the right approach...
For quantization, simply changing the leaf conversion methods from fatalError to throws moves the failure to callers rather than preventing invalid configurations.

PR #453, opened 2026-07-22
Add typed KV cache configuration and runtime reporting

aleroot:
## Proposed changes

This PR is a follow-up to [#232](https://github.com/ml-explore/mlx-swift-lm/pull/232), which introduced TurboQuant KV-cache compression, Metal kernels, prompt-cache persistence, boundary-layer protection, and the initial `kvScheme` integration.

The underlying TurboQuant implementation is unchanged. This PR focuses on making KV-cache configuration safe, extensible, and consistently applied across the library.

Previously, configuration was spread across `maxKVSize`, `kvBits`, `kvGroupSize`, `quantizedKVStart`, and the stringly typed `kvScheme`. This caused several problems:

- Unknown scheme strings could be silently ignored.
- MTP speculative decoding did not forward `kvScheme`.
- Wired-memory measurement did not measure the requested TurboQuant configuration.
- `maxKVSize` could create an entirely rotating cache topology, making compression an all-FP16 no-op without exposing that result to callers.
- Cache traversal, conversion, and configuration forwarding were repeated across generation paths.
- Applications could not reliably report whether compression was active, pending, or skipped.

This PR introduces:

- A typed `KVCacheConfiguration` covering capacity, compression strategy, and compatibility policy.
- An opaque strategy representation, allowing future cache implementations to be added without exposing an exhaustive public enum.
- Typed affine and TurboQuant configurations, including quality-first, balanced, and memory-first presets.
- An immutable resolved cache plan shared by autoregressive, classic speculative, MTP, guided-generation, and wired-memory paths.
- Centralized traversal and classification of nested and hybrid cache topologies.
- Explicit compatibility policies for partially supported, mixed, and all-rotating cache configurations.
- Fail-closed behavior for typed compression requests that would otherwise affect no layers.
- Effective per-layer runtime reporting, including active, pending, skipped, rotating, and recurrent cache state.
- A compatibility adapter for the existing scalar and `kvScheme` parameters.
- Consistent typed capacity handling across the default, Falcon-H1, Qwen3.5, Pixtral, and Mistral3 cache factories.
- Updated KV-cache documentation and configuration tests.

A bounded compressed rotating cache is intentionally not introduced here. Rotating layers remain FP16, as in #232. Instead, callers can now explicitly allow partial compression or reject configurations where compression would not engage.

## Checklist

### DIFF
--- a/Libraries/MLXLMCommon/Evaluate.swift
+++ b/Libraries/MLXLMCommon/Evaluate.swift
@@ -64,6 +64,13 @@ public struct GenerateParameters: Sendable {
     /// When set, uses ``RotatingKVCache`` instead of ``KVCacheSimple``
     public var maxKVSize: Int?
 
+    /// Typed key-value cache configuration.
+    ///
+    /// When set, this is the canonical cache configuration. Do not combine it
+    /// with `maxKVSize`, `kvBits`, `kvGroupSize`, `quantizedKVStart`, or
+    /// `kvScheme`.
+    public var kvCache: KVCacheConfiguration?
+
     /// Number of bits to use for KV cache quantization. nil implies no cache quantization.
     public var kvBits: Int?
 
@@ -88,7 +95,8 @@ public struct GenerateParameters: Sendable {
     ///   maximum compression; K sensitivity varies by model family, so
     ///   validate on your model (asym is the recommended starting point).
     ///
-    /// Unrecognized schemes are ignored. See `resolveTurboScheme`.
+    /// Unrecognized schemes are rejected when generation starts. Prefer
+    /// ``kvCache`` for compile-time-safe configuration.
     public var kvScheme: String?
 
     /// Sampling temperature
@@ -131,6 +139,7 @@ public struct GenerateParameters: Sendable {
     public init(
         maxTokens: Int? = nil,
         maxKVSize: Int? = nil,
+        kvCache: KVCacheConfiguration? = nil,
         kvBits: Int? = nil,
         kvGroupSize: Int = 64,
         quantizedKVStart: Int = 0,
@@ -150,6 +159,7 @@ public struct GenerateParameters: Sendable {
     ) {
         self.maxTokens = maxTokens
         self.maxKVSize = maxKVSize
+        self.kvCache = kvCache
         self.kvBits = kvBits
         self.kvGroupSize = kvGroupSize
         self.quantizedKVStart = quantizedKVStart
@@ -607,18 +617,18 @@ public struct TokenIterator: TokenIteratorProtocol {
     public internal(set) var state: LMOutput.State?
 
     var y: LMInput.Text
-    var cache: [KVCache]
+    let cacheStorage: KVCacheStorage
+    var cache: [KVCache] {
+        get { cacheStorage.cache }
+        set { cacheStorage.replace(with: newValue) }
+    }
     var processor: LogitProcessor?
     let sampler: LogitSampler
 
     public var tokenCount = 0
     public let maxTokens: Int?
 
-    // Cache quantization parameters
-    let kvBits: Int?
-    let kvGroupSize: Int
-    let quantizedKVStart: Int
-    let kvScheme: String?
+    var kvCachePlan: KVCachePlan { cacheStorage.plan }
 
     // Internal metrics
     public var promptPrefillTime: TimeInterval = 0.0
@@ -637,22 +647,12 @@ public struct TokenIterator: TokenIteratorProtocol {
         prompt: MLXArray, model: any LanguageModel, cache: [KVCache]? = nil,
         parameters: GenerateParameters, components: GenerationComponents = .init()
     ) throws {
-        self.model = model
-        self.y = .init(tokens: prompt)
-        self.cache = cache ?? model.newCache(parameters: parameters)
-
-        self.processor = components.logitProcessor(parameters: parameters)
-        self.sampler = parameters.sampler()
-        self.maxTokens = parameters.maxTokens
-
-        self.kvBits = parameters.kvBits
-        self.kvGroupSize = parameters.kvGroupSize
-        self.quantizedKVStart = parameters.quantizedKVStart
-        self.kvScheme = parameters.kvScheme
-
-        self.promptPrefillTime = try measure {
-            try prepare(input: .init(text: y), windowSize: parameters.prefillStepSize)
-        }
+        let plan = try parameters.kvCachePlan()
+        try self.init(
+            input: .init(text: .init(tokens: prompt)), model: model,
+            cacheStorage: KVCacheStorage(
+                cache ?? model.newCache(parameters: parameters), plan: plan),
+            parameters: parameters, components: components)
     }
 
     /// Initialize a `TokenIterator` with the given input.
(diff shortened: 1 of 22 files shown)


aleroot:
@davidkoski this change addresses the cache-ownership concern discussed in [#487](https://github.com/ml-explore/mlx-swift-lm/pull/487#discussion_r3693071697) by making model-wide processed-token progress explicit in KVCacheStorage, rather than duplicating an offset across recurrent cache entries.
It also provides the ownership/copy foundation needed for the checkpoint optimization proposed after [#472](https://github.com/ml-explore/mlx-swift-lm/pull/472). It overlaps with the cache-container portion of [#358](https://github.com/ml-explore/mlx-swift-lm/pull/358), so that PR should be reevaluated or rebased after this lands rather than merging both ownership approaches independently.
Recurrent caches continue to own only their native state and metadata; shared timeline advancement, copying, and rollback happen once at the model-cache boundary.

davidkoski:
Thank you!

Libraries/MLXLMCommon/KVCacheTree.swift
```diff
@@ -0,0 +1,132 @@
...
+
+    static func turboQuantProtectedPaths(
+        in leaves: [KVCacheLeaf],
+        keyBits: Int,
+        valueBits: Int
+    ) -> Set<[Int]> {
+        let fragile = (keyBits > 0 && keyBits < 8) || valueBits <= 2
+        guard fragile else { return [] }
+
+        let attentionPaths = leaves.filter(\.isAttentionCache).map(\.path)
```
davidkoski: In the case of `Qwen3` it returns a mix of StandardKVCache and RotatingKVCache and this will apply it to all layers (since all are `isAttentionCache`, aka `!= .recurrent`).

Previously it did this:

```swift
    for i in 0 ..< cache.count {
        guard let simple = cache[i] as? KVCacheSimple, cache[i].offset > quantizedKVStart,
            !(cache[i] is TurboQuantKVCache)
        else { continue }
```

I don't understand the full impact of this, but the logic is different -- on purpose?

Libraries/MLXLMCommon/KVCachePlan.swift
```diff
@@ -0,0 +1,281 @@
...
+    }
+}
+
+/// Shared ownership for a realized cache and its dynamic application state.
+///
+/// Cache arrays have value semantics, but dynamic compression replaces array
+/// elements. Sharing this storage keeps sessions and iterators on the same
+/// realized array while the cache objects themselves remain reference types.
+/// Access is externally serialized by generation/session ownership.
+package final class KVCacheStorage: @unchecked Sendable {
```
davidkoski: Should this be `Sendable`?  Since the serialization is handled externally, I don't think we want to make any promises here -- it defeats whatever checks the compiler might give (though it may also force us to manually transfer between tasks).

Libraries/MLXLMCommon/Evaluate.swift
davidkoski, line 1: Do any of the `generate()` methods that take a `[KVCache]` need to be deprecated?

```swift
public func generate(
    input: LMInput,
    cache: [KVCache]? = nil,
    parameters: GenerateParameters,
    context: ModelContext,
    mtpDrafter: any MTPDrafterModel,
    blockSize: Int = 4,
    wiredMemoryTicket: WiredMemoryTicket? = nil
) throws -> AsyncStream<Generation> {
```

It doesn't work any worse than before, but if you are going to pass in a `cache` you should probably be dealing in a `KVCacheStorage`, right?  Otherwise you are sitting on this problem without warning.  Probably true of the `TokenIterator` inits.

Libraries/MLXLMCommon/Evaluate.swift
```diff
@@ -643,20 +643,32 @@ public struct TokenIterator: TokenIteratorProtocol {
...
     ) throws {
+        let plan = try parameters.kvCachePlan()
+        try self.init(
+            input: input, model: model,
+            cacheStorage: KVCacheStorage(
+                cache ?? model.newCache(parameters: parameters), plan: plan),
+            state: state, parameters: parameters)
+    }
+
+    package init(
```
davidkoski, line 686: Does this + KVCacheStorage need to be `public`?  Otherwise how would callers that are not `ChatSession` make use of this?

I think it would be gratifying if everybody used ChatSession as it encapsulates a lot of cache/session management + integrates tool calling, etc., but the lower level APIs still remain and I think they should be able to be used correctly.

Thoughts?  Was the `package` just to keep the API narrow?

davidkoski:
Summary to make sure I am understanding things:

OK, so `KVCacheStorage` is the container for caches and handles the need for reference semantics on cache mutation/upgrade.  The API change is on `TokenIterator` (and friends).  `LanguageModel` still sees `[KVCache]` as it doesn't need to replace the KVCache values -- that is done between tokens in the iterator.  One curiosity: the new types and methods are `package` scoped rather than `public`.  Should they be public?

`ChatSession` integrates with the new API:  it creates a `TokenIterator` directly and passes that to the generate loop.

`generate()` functions in `Evaluate.swift` that take `[KVCache]` maybe should be deprecated as they don't allow fixing of the cache replacement issue.

I think we could take this as-is -- it doesn't cause any new bugs.  We could discuss how to expose the API to let callers other than ChatSession participate in another PR.

(and also the configuration and runtime reporting stuff :-) )

davidkoski (changes requested):
Looks really good!

See my comments:

- turboQuantProtectedPaths -- this might be a bug, looks like a change in behavior at least
- Sendable on the KVCacheStorage -- might be promising too much

See what you think on the visibility too.  That could be a different PR.

NivDvir:
+1 on taking it as-is. On public vs package: as an out-of-package consumer building on the prompt-cache persistence line (#443), I'll eventually want callers outside ChatSession to participate — but the follow-up-PR route seems the right place to shape that surface rather than freezing it public now. I can bring a concrete consumer to that discussion: a validated append-only media-reuse branch that would rebase onto KVCacheStorage once this lands.

Libraries/MLXLMCommon/KVCacheTree.swift
aleroot: Good catch. It was unintended.
Fixed in 42284a2, with a regression test for mixed cache topologies.

Libraries/MLXLMCommon/Evaluate.swift
aleroot: Agreed in principle, but I’d defer deprecation until there’s a public `KVCacheStorage` replacement. Deprecating `[KVCache]` now would leave callers without a migration path; this should be handled in a focused follow-up.

Libraries/MLXLMCommon/Evaluate.swift
aleroot: Yes, `package` was intentional to avoid freezing the ownership and mutation contract prematurely. I agree lower-level callers need a public continuation API, but I’d design and expose that in a focused follow-up rather than making `KVCacheStorage` public as-is.

davidkoski:
I think this is ready to merge once the conflict with #401 is resolved.

davidkoski (approved):

aleroot:
> I think this is ready to merge once the conflict with #401 is resolved.

Should be resolved, thanks

PR #455, opened 2026-07-22
Qwen3VL vision: stop using a huge amount of memory to process images.

kklimuk:
Qwen3VL's vision prefill uses far more memory than the math requires:

1. **All images are merged into one attention sequence**. Attention memory grows with the square of the *total* pad count across images: two images totaling 8140 pads request a single 33.9 GB Metal buffer — past `maxBufferLength` on a 48 GB M4 Pro, which kills the generation. Doing the same work per-image completes in a 14.2 GB peak.
2. **The score tensor is materialized even for a single image**: one 6188-pad image makes a ~19.6 GB buffer, most of it a transient score buffer that a fused kernel would never allocate.

This is caused by:

1. The merge being enforced by a dense `[1, L, L]` additive `-1e9` mask over the concatenated sequence. However, the mask is block-diagonal — an image's patches never attend to another image's — which means it does `(Σ Lᵢ)²` of work instead of `Σ Lᵢ²`.
2. The Qwen3VL vision tower's head dim is 72 (1152 / 16 heads), outside the fused Metal kernel's supported 64/80/128, so `MLXFast.scaledDotProductAttention` silently falls back to composed ops that materialize the scores — `numHeads × L² × 2` bytes in one buffer.

Fixes:

1. **Attend each `cuSeqlens` segment independently with no mask**. This is mathematically identical to the block-diagonal mask, and the same shape as mlx-vlm's Python implementation (`mlx_vlm/models/qwen3_vl/vision.py` splits Q/K/V at `cu_seqlens` and passes `mask=None`).
2. **Zero-pad the head dim 72 → 80** so the fused O(L)-memory kernel dispatches — the same trick as `gemma4EnsureFusedSDPA` in this repo (and mlx-vlm's `ensure_fused_sdpa`). Padding is exact: the padded dims contribute nothing to the dot products and `scale` is passed explicitly.

Measured on Qwen3.5-9B (M4 Pro, 48 GB):

| case | before | after |
|---|---|---|
| single image, 6188 pads | 28.7 GB peak | 12.6 GB peak |
| two images, 8140 pads total | fatal (33.9 GB > maxBufferLength) | 14.2 GB peak |
| two-image prefill wall | 59.8 s | 36.3 s |
| greedy outputs | — | unchanged on every case tested |

Happy to hoist the padding helper somewhere shared (it now exists twice, for Gemma4 and here) if you'd prefer.

AI Disclosure: I worked with Claude Code to get this - but I tested it and looked at everything myself. I had to fork because the harness I'm working on explodes on Macbook Air 16Gb otherwise 😭


davidkoski (approved):
Really nice find and fix.  Thank you!

PR #456, opened 2026-07-22
Support multi-round tool calling in MLXFoundationModels

thechriswebb:
## Summary

This branch makes tool calling in `MLXFoundationModels` work end to end through Apple's Foundation Models APIs (`LanguageModelSession` / `ChatSession`). Before this, a session could issue at most one tool call and had no faithful notion of tool-choice modes; now the model can chain tool calls across multiple rounds, honor the caller's tool-choice mode, stream native tool output in order, and preserve structured results. The work also fixes several correctness issues surfaced along the way (sampling seeds, cancellation, LFM2 response prefixes, and a Qwen multi-turn regression).

The change is about 3,200 lines added across `MLXFoundationModels`, `MLXLMCommon`, and `MLXGuidedGeneration`, roughly half of which is test coverage.

## What changed

**Multi-round tool calling.** Prior tool calls and their results are replayed into the prompt the model sees, and the tool path runs on every round while tools are enabled. The model either chains another tool or answers from the replayed result; the turn ends when it answers, with no iteration cap, matching the behavior of `LanguageModelSession` and `ChatSession`.

**Tool-choice mode resolution.** Added `ToolCallingModeResolution` so the caller's intent (automatic, required, disallowed) is resolved and honored, including preventing a fallback to a plain response when a tool is required.

**Native allowed tool-output routing.** Added `AllowedToolOutputRouter` to route allowed tool use through native generation, stream ordered output at parse boundaries, and correctly handle EOS residuals, unfinished protocol tails, and ordering.

**Structured tool outputs.** Structured Foundation Models tool outputs are preserved rather than flattened to strings.

**Diagnostic sink.** Added a `GuidedGenerationDiagnosticSink` for tracing tool-call boundary signals during guided generation. It is dormant unless bound and preserved across the ambient generation context.

**Correctness fixes.**
- Sampling seeds are preserved and applied to Foundation Models generation.
- Guided-generation cancellation is propagated correctly.
- Fixed end-of-stream handling for multi-call tool output (duplicate response text and leaked protocol markers).
- Inline JSON strings are scanned safely in the tool-call parser.
- Fixed a Qwen degenerate tool-call output in multi-turn continuations.

**Removed.** The synthetic final-answer tool was removed in favor of letting the model answer directly.

## Shared changes in MLXLMCommon

The tool-calling layer in `MLXLMCommon` gained shared infrastructure that the Foundation Models routing builds on. None of it is Foundation Models-specific.

**Ordered tool-output event API.** The processor previously returned visible text and tool calls separately, which lost the relative order of prose and calls within a turn. A new public `Output` enum (`.response` / `.toolCall`) and methods (`processChunkOutputs`, `processEOSOutputs`, `drainToolCalls`) emit response and tool-call events in the order the model actually produced them, backed by an internal ordered queue.

**End-of-stream reconstruction per format.** When generation is cut off before a call's closing tag arrives, calls are now reconstructed from the raw buffer per format: Mistral's `[TOOL_CALLS] ... [ARGS] {...}` shape and LFM2's pythonic `[func()]` list shape (the latter via bracket balancing).

**Quote-aware JSON scanning.** Naive brace counting was replaced with a scanner that respects quoted strings and escapes, so braces inside string values (for example `{"path": "a}b"}`) no longer cause the processor to treat an object as complete prematurely.

**Protocol-marker sanitizing.** New helpers strip complete tag pairs, lone tags, and near-complete partial markers from text before it is emitted as user-visible response, so protocol tokens never leak into the UI.

**Qwen redundant-brace recovery.** `JSONToolCallParser` now recovers the exact shape some Qwen chat templates emit, an EOS-delimited call wrapped in a redundant extra brace on each side (`{{ ...valid call... }}`), and only when the inner content is one complete, valid tool call followed solely by the extra braces.

## Testing

Extensive unit and characterization coverage was added: mode resolution, allowed-tool output routing, transcript conversion, sampling/generation-options forwarding, structured tool output through `LanguageModelSession`, and tool-call grammar characterization. Behavioral checks span Gemma and Qwen to confirm a continuation answers from a tool result or issues another tool call rather than ignoring it. A grammar characterization test drives real models (Gemma-4 E2B and Qwen-3) through the required guided tool-call path and asserts that the generated buffer parses as a valid tool call.


davidkoski (approved):
Changes look good, thank you!

PR #457, opened 2026-07-22
DeepSeek-V3: populate kvHeads and update the KV cache once per attention call

GoodOlClint:
## Proposed changes

Two bugs on DeepSeek-V3's cached path. Together they mean **the model cannot generate** — only a `cache: nil` forward pass works.

This came out of @davidkoski's review on #379, which flagged the same double-update in the DeepSeek-V2 port and noted "DeepseekV3 would need the same fix". It does — plus a more basic one that hides it.

### 1. `kvHeads` was never populated (generation crashes)

```swift
public var kvHeads: [Int] = []   // never assigned
```

`newCache` derives the layer count from `kvHeads.count`, so it returned an **empty array**. The layer loop then does `cache?[i]`, which traps immediately:

```
Swift/ContiguousArrayBuffer.swift:692: Fatal error: Index out of range
```

Fixed by populating it as the other models do — one entry per layer:

```swift
self.kvHeads = Array(repeating: args.numKeyValueHeads, count: args.numHiddenLayers)
```

### 2. The KV cache was updated twice per attention call

Attention called `cache.update(...)` explicitly and then passed the resulting keys/values to `attentionWithCacheUpdate`, which performs the update itself. The second update appended the already-returned history back into the cache, so a prompt of L tokens left the cache at **2L** and every subsequent token attended over duplicated keys.

Dropped the explicit update and let the helper own the single one, matching `Llama` / `Qwen3` / `Cohere` and the Python reference. This also routes the quantized-cache path correctly, which the manual update bypassed.

### Test

Adds `DeepseekV3Tests` (there were none) with a cache-growth regression test — a tiny config, no downloads. A 5-token forward pass must leave every layer's cache at offset 5.

The two bugs isolate cleanly, which is how I verified the test actually discriminates:

| State | Result |
|---|---|
| `kvHeads` empty (current `main`) | crashes — `Index out of range`, test cannot run |
| `kvHeads` fixed, double-update present | fails — `cache advanced by 10 for 5 tokens` |
| both fixed | passes |

Full `MLXLMTests` run is green (223 XCTest, 0 failures). Note: `Gemma4ChunkedPrefillTests` fails on my machine both before and after this change — a pre-existing tight-tolerance numeric issue unrelated to DeepSeek.

## Checklist

- [ ] I have updated the necessary documentation (if needed) — no doc changes needed

### DIFF
--- a/Libraries/MLXLLM/Models/DeepseekV3.swift
+++ b/Libraries/MLXLLM/Models/DeepseekV3.swift
@@ -195,21 +195,14 @@ class DeepseekV3Attention: Module {
         kv = kv.reshaped(B, L, self.numHeads, -1).transposed(0, 2, 1, 3)
         let splitKv = split(kv, indices: [self.qkNopeHeadDim], axis: -1)
 
-        var (kNope, values) = (splitKv[0], splitKv[1])
+        let (kNope, values) = (splitKv[0], splitKv[1])
 
         let offset = cache?.ropeOffset
         qPe = applyRotaryPosition(rope, to: qPe, offset: offset)
         kPe = applyRotaryPosition(rope, to: kPe, offset: offset)
         kPe = repeated(kPe, count: numHeads, axis: 1)
 
-        var keys: MLXArray
-        if let cache = cache {
-            (keys, values) = cache.update(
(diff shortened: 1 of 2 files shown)


davidkoski (approved):
Fix looks great, thank you!

PR #458, opened 2026-07-22
Add GitHub workflow for IntegrationTesting tests

CharlieTLe:
## Proposed changes

Runs the IntegrationTestingTests target manually (workflow_dispatch) on the self-hosted macOS runner. Kept off the PR path since the tests download HF checkpoints, require Metal, and are long-running. Uses -parallel-testing-enabled NO to avoid concurrent workers racing on the shared on-disk Hugging Face cache.

## Checklist

### DIFF
--- a/.github/workflows/integration_tests.yml
+++ b/.github/workflows/integration_tests.yml
@@ -1,12 +1,9 @@
 name: Integration Tests
 
 # Heavy integration tests (Hugging Face model downloads, Metal GPU, long-running).
-# Kept out of the PR path so they never block merges; run manually or nightly.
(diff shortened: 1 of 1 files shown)


davidkoski:
Nice!  I was just talking with somebody about this!

I think it would be good to be able to manually trigger it.  Is nightly too often?  Maybe we start there and see how it goes?

CharlieTLe:
Yeah, I think this is a good balance of being able to trigger it any time and to have it run regularly. Happy to remove the nightly part if it's too noisy or often.

davidkoski (approved):
Looks good, thank you!

PR #460, opened 2026-07-22
Add Nanbeige4.2 (looped transformer) model support

spokvulcan:
Adds support for Nanbeige's Nanbeige4.2 models (`model_type: "nanbeige"`), e.g. [Nanbeige4.2-3B](https://huggingface.co/Nanbeige/Nanbeige4.2-3B).

The architecture is a looped transformer: the decoder stack of `num_hidden_layers` shared-weight layers runs `num_loops` times per forward. Each loop pass owns its own slice of the KV cache array, so `newCache` returns `num_loops * num_hidden_layers` caches and the forward indexes `cache[loopIndex * layers.count + i]`. The final RMSNorm is re-applied after each loop unless `skip_loop_final_norm` is set. Everything else is a standard Llama-style dense decoder (SwiGLU MLP, RoPE with the checkpoint's theta, no QK-norm).

This is a port of the Python mlx-lm implementation (MercuriusDream/mlx-lm `add-nanbeige-model`, not yet merged there either), checked against the official `modeling_nanbeige.py` on the HF repo. Config features the released checkpoints don't enable and that would need dedicated weights (`enable_double_loop_split`, `loop_share_kv`, `enable_depth_attention`, hyper-connections) are rejected at config validation rather than silently mis-running.

Also registers the model's chat conventions: XML function tool calls (the template's trained default) and Qwen3-style `<think>` reasoning with the `enable_thinking` template flag.

Verification:
- Greedy continuations match the Python implementation token-for-token on the same 8-bit weights ([MercuriusDream/Nanbeige4.2-3B-mlx-8bit](https://huggingface.co/MercuriusDream/Nanbeige4.2-3B-mlx-8bit)), long shared prefixes with single-token divergence/re-convergence at quantization noise level.
- 9 unit tests (no weights needed): config parsing/validation, cache slot layout, loop norm placement, warm continuation vs full prefill on a tiny random-weight model.
- `swift-format` clean, full package builds on current `main`.

### DIFF
--- a/Libraries/MLXLMCommon/Tool/ToolCallFormat.swift
+++ b/Libraries/MLXLMCommon/Tool/ToolCallFormat.swift
@@ -216,6 +216,12 @@ public enum ToolCallFormat: String, Sendable, Codable, CaseIterable {
             return .xmlFunction
         }
 
+        // Nanbeige4.2+ — chat template supports XML and JSON; XML is the trained
+        // default and the model card's recommendation for agentic use
+        if type.hasPrefix("nanbeige") {
+            return .xmlFunction
+        }
+
         // Mistral3 family (mistral3, mistral3_text, etc.)
         if type.hasPrefix("mistral3") {
             return .mistral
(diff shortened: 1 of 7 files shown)


Libraries/MLXLMCommon/Tool/ToolCallFormat.swift
```diff
@@ -216,6 +216,12 @@ public enum ToolCallFormat: String, Sendable, Codable, CaseIterable {
             return .xmlFunction
         }
 
+        // Nanbeige4.2+ — chat template supports XML and JSON; XML is the trained
+        // default and the model card's recommendation for agentic use
+        if type.hasPrefix("nanbeige") {
+            return .xmlFunction
+        }
```
davidkoski, line 223: Not for this PR, but this and the ReasoningConfig make me think that the model should be asked this question.  Better encapsulation.

davidkoski (approved):
Changes look good, thank you!

PR #462, opened 2026-07-23
Olmo3: fix newCache signature so the sliding-window cache is actually used

GoodOlClint:
Olmo3 declares `newCache(parameters:)` with a **non-optional** `GenerateParameters`:

```swift
public func newCache(parameters: GenerateParameters) -> [KVCache]
```

but the `LanguageModel` protocol requirement is `newCache(parameters: GenerateParameters?)`. Because the parameter type differs, this is an unrelated overload rather than an override of the protocol requirement. `Olmo3Model` also conforms to `KVCacheDimensionProvider`, which supplies the default `newCache` implementation (all `KVCacheSimple`), and that default is what satisfies the protocol. The result: the Olmo3-specific body **never runs**, and every `sliding_attention` layer gets an unbounded `KVCacheSimple` instead of `RotatingKVCache(maxSize: slidingWindow)`.

The prefill mask still applies the sliding window, so short prompts look correct, but during decode those layers attend past the window and the KV cache grows without bound — wrong outputs versus the reference model, plus unbounded KV memory on long contexts. The compiler emits no diagnostic because an unused overload alongside a satisfied protocol default is legal.

The fix is to make the parameter optional so the method satisfies the requirement. A comment is added so the signature is not "cleaned up" back to non-optional later.

## How this was found

Reviewing #379 (DeepSeek-V2) surfaced the KV-cache-wiring issues that became #437 (make `BaseKVCache.ropeOffset` overridable). That prompted a small audit of models whose `newCache` diverges from the Llama-style default. Olmo3 was the one real defect it turned up; the added tests also cover four other bespoke/hybrid-cache models (Jamba, GraniteMoeHybrid, Qwen3Next, GPTOSS, plus BaichuanM1 / MiMoV2Flash / AfMoE / Exaone4 / Mistral3Text) which were all found to be correct.

## Tests

New `Tests/MLXLMTests/CachedForwardSmokeTests.swift` builds each model from a tiny inline JSON config (no downloads), runs one prefill, and asserts every attention layer's cache advanced by exactly the token count. `testOlmo3SlidingLayersGetRotatingCache` additionally asserts each sliding layer receives a `RotatingKVCache`.

- Before the fix: the Olmo3 tests fail (`layer 0 is sliding_attention but got KVCacheSimple`).
- After the fix: all pass.

Run with:

```
xcodebuild test -scheme mlx-swift-lm-Package -destination 'platform=macOS' -skipPackagePluginValidation -only-testing:MLXLMTests/CachedForwardSmokeTests
```

## Checklist

- [ ] I have updated the necessary documentation (no public API or documented behavior changed; the fixed signature already matches the protocol)

### DIFF
--- a/Libraries/MLXLLM/Models/Olmo3.swift
+++ b/Libraries/MLXLLM/Models/Olmo3.swift
@@ -223,10 +223,6 @@ public class Olmo3Model: Module, LLMModel, KVCacheDimensionProvider {
         weights.filter { !$0.key.contains("self_attn.rotary_emb.inv_freq") }
     }
 
-    // The parameter must stay optional: `LanguageModel` declares
-    // `newCache(parameters: GenerateParameters?)`, and a non-optional parameter makes this
-    // an unused overload while `KVCacheDimensionProvider`'s all-`KVCacheSimple` default
-    // silently satisfies the protocol — dropping the sliding window entirely.
     public func newCache(parameters: GenerateParameters?) -> [KVCache] {
         var caches: [KVCache] = []
         for layerType in args.layerTypes {
--- a/Libraries/MLXLLM/Models/Olmo3.swift
+++ b/Libraries/MLXLLM/Models/Olmo3.swift
@@ -223,7 +223,11 @@ public class Olmo3Model: Module, LLMModel, KVCacheDimensionProvider {
         weights.filter { !$0.key.contains("self_attn.rotary_emb.inv_freq") }
     }
 
-    public func newCache(parameters: GenerateParameters) -> [KVCache] {
+    // The parameter must stay optional: `LanguageModel` declares
+    // `newCache(parameters: GenerateParameters?)`, and a non-optional parameter makes this
+    // an unused overload while `KVCacheDimensionProvider`'s all-`KVCacheSimple` default
+    // silently satisfies the protocol — dropping the sliding window entirely.
+    public func newCache(parameters: GenerateParameters?) -> [KVCache] {
         var caches: [KVCache] = []
         for layerType in args.layerTypes {
             if layerType == "full_attention" {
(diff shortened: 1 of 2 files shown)


Tests/MLXLMTests/CachedForwardSmokeTests.swift
```diff
@@ -0,0 +1,304 @@
...
+@testable import MLXLLM
+
+/// Cached-forward smoke tests for models with a bespoke `newCache` / hybrid cache layout.
+///
+/// Each test builds a tiny model from an inline JSON config (no downloads), runs one
+/// prefill through `newCache(parameters:)`, and asserts every attention layer's cache
+/// advanced by exactly the token count. This catches the two structural failures that
+/// shipped undetected in DeepSeek-V3: a cache array whose length disagrees with the
+/// layer count (index-out-of-range), and a double KV update (offset == 2 * tokens).
+final class CachedForwardSmokeTests: XCTestCase {
```
davidkoski, line 13: Nice

Libraries/MLXLLM/Models/Olmo3.swift
```diff
@@ -223,7 +223,11 @@ public class Olmo3Model: Module, LLMModel, KVCacheDimensionProvider {
         weights.filter { !$0.key.contains("self_attn.rotary_emb.inv_freq") }
     }
 
-    public func newCache(parameters: GenerateParameters) -> [KVCache] {
+    // The parameter must stay optional: `LanguageModel` declares
+    // `newCache(parameters: GenerateParameters?)`, and a non-optional parameter makes this
+    // an unused overload while `KVCacheDimensionProvider`'s all-`KVCacheSimple` default
+    // silently satisfies the protocol — dropping the sliding window entirely.
```
davidkoski: I think the comment is good for the PR but we probably don't need it in the code.  That said, this is kind of aggravating!  If we had an `override` it would have caught this (but it doesn't apply here).  Or no default implementation.

davidkoski:
Rebased.  CI not showing up (seen on another PR as well), so running tests & checks locally.

davidkoski (approved):
Change looks good, thank you!

PR #463, opened 2026-07-23
Add routing-correctness tests for GLM4MOE / GLM4MOELite MoE gates

GoodOlClint:
GLM4MOE and GLM4MOELite shipped without tests. Their `noaux_tc` grouped-router is the kind of code where a defect stays silent — a mis-wired gate still runs and produces plausible logits, so only a targeted test catches it. This adds `Tests/MLXLMTests/GLM4MoERoutingTests.swift` (tiny inline configs, no downloads). All pass — routing is correct in both models; no fix is bundled.

## What it pins

- **Group-limited routing** — with `n_group`/`topk_group`, selection is confined to the winning group(s).
- **`e_score_correction_bias`** — the bias steers *which* experts are selected, but the returned gate weights are gathered from the **unbiased** scores, not from `selectionScores`. A port that gathered the biased tensor would inflate every routed weight; a mutation to that effect fails the two correction-bias tests, and they pass on the current code.
- **`sanitize`** — per-expert `gate/up/down` weights stack into `switch_mlp`, and the trailing MTP layer is dropped.

Both gates are separate implementations (`GLM4MoEGate`, `GLM4MoELiteGate`) that share the `noaux_tc` contract, so each is pinned independently.

## Tests

```
xcodebuild test -scheme mlx-swift-lm-Package -destination 'platform=macOS' -skipPackagePluginValidation -only-testing:MLXLMTests/GLM4MoERoutingTests
```

5/5 pass.

## Checklist

- [ ] I have updated the necessary documentation (test-only; no API or behavior change)

### DIFF
--- a/Tests/MLXLMTests/GLM4MoERoutingTests.swift
+++ b/Tests/MLXLMTests/GLM4MoERoutingTests.swift
@@ -8,17 +8,11 @@ import XCTest
 
 /// Routing-correctness tests for the GLM-4 MoE gates (noaux_tc grouped top-k).
 ///
-/// These models had zero coverage. Unlike a cache offset bug, a mis-wired router
-/// still runs and produces plausible logits — the defect only shows as subtly wrong
-/// token probabilities, so it hides without a targeted test. Two structural properties
-/// are pinned here:
-///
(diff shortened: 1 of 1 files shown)


davidkoski (approved):
Looks great, thank you!

PR #464, opened 2026-07-24
Integration tests: build on both macOS 26 and 27 SDKs

CharlieTLe:
## Proposed changes

The IntegrationTesting job failed to compile on the Xcode 26.5 runner: the FoundationModels adapter (MLXFoundationModels) is gated behind canImport(FoundationModels, _version: 2) (macOS 27 SDK only), but the integration test files gated only on the always-set FoundationModelsIntegration trait, so they referenced symbols absent on the 26 SDK.

- Extend the 37 FoundationModels-gated test files' top-level guard to '#if FoundationModelsIntegration && canImport(FoundationModels, _version: 2)', mirroring the library so they compile out on the 26 SDK and stay active on 27.
- Workflow: prefer Xcode 27 (via DEVELOPER_DIR) when the runner has it so the full suite runs; otherwise fall back to the default toolchain and run the SDK-agnostic suites (MTP, Qwen3VL/Qwen3.5 vision, Coherence, Gemma4, tool calls).

## Checklist

### DIFF
--- a/.github/workflows/integration_tests.yml
+++ b/.github/workflows/integration_tests.yml
@@ -18,6 +18,29 @@ jobs:
         with:
           submodules: recursive
 
+      - name: Select Xcode
+        shell: bash
+        run: |
+          # The FoundationModels integration tests only build against the macOS 27
+          # SDK (canImport(FoundationModels, _version: 2)); on the macOS 26 SDK the
+          # MLXFoundationModels adapter compiles out and those tests are skipped.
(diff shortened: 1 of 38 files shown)


davidkoski:
Hrm, no easy way to run the action if it isn't in a branch on the main repo.  I think we can just iterate as needed with PRs.  The changes look right to me.

davidkoski (approved):

PR #465, opened 2026-07-25
Fix guided generation dropping required properties (streaming detokenizer)

thechriswebb:
## Summary

Guided generation (`LanguageModelSession(model:).respond(to:generating:)`) intermittently returned an object missing a **required** property, throwing `GeneratedContent does not contain a property '<name>'`. Reproduced reliably with Gemma-4; Qwen never hit it.

The root cause is **not** in the schema or the grammar — both were verified correct. It is in `NaiveStreamingDetokenizer.next()`, which dropped a token's text when the tokenizer's `decode()` is not append-only.

## Root cause

`next()` computed each incremental delta as a **character-count diff**:

```swift
let new = newSegment.suffix(newSegment.count - segment.count)
```

This assumes `decode(tokens)` grows append-only. SentencePiece decoding does not: appending a token can rewrite earlier whitespace. With Gemma's tokenizer:

- `decode([\n, "  "]) == "\n  "` (length 3)
- `decode([\n, "  ", ","]) == "\n ,"` (length 3 — a space collapses)

So `newSegment.count - segment.count == 0`, `suffix(0) == ""`, and the newly decoded character — here a **structural JSON comma** — is silently dropped. The grammar had committed the comma and stayed valid (`grammarTerminated == true`), but the *text streamed to FoundationModels* was invalid JSON, so parsing failed on a required property. Qwen uses byte-level BPE (append-only), so it never triggered the bug.

## Fix

Diff by **common prefix** instead of character count:

```swift
let common = newSegment.commonPrefix(with: segment)
let new = newSegment.dropFirst(common.count)
```

- Identical to the previous behavior for append-only tokenizers.
- For non-append-only decodes it never drops characters.
- Bonus: removes a latent scalar-vs-grapheme hazard, since `commonPrefix`/`count`/`dropFirst` all operate on grapheme clusters. A comment documents that correctness relies on the already-emitted prefix being stable (true for SentencePiece, which only collapses whitespace at the append boundary).

One file of production code changed: `Libraries/MLXLMCommon/Tokenizer.swift`.

## Testing

All runs on Apple silicon, macOS 27 SDK, Xcode 27. **No failure is attributable to this change.** The three pre-existing failures below fail identically on the baseline (fix stashed) and are unrelated; they're documented as follow-ups.

### New tests added
| Test | Kind | Result |
|---|---|---|
| `StreamingDetokenizerTests` — comma-drop regression, append-only guard, split-multibyte `\u{fffd}` path, newline-reset + non-append-only | model-free unit (`Tests/MLXLMTests`) | 4/4 pass (comma-drop test fails on old code, passes on new) |
| `RequiredPropertyStressTests` — 25× guided NPC generation on Gemma-4, env-gated `MLX_RUN_REQUIRED_PROPERTY_STRESS` | integration | **0/25 failures with the fix (was 3/15 before)** |

### Regression suites run to validate
| Suite / group | Tests | Result |
|---|---|---|
| `MLXLMTests` (model-free unit) | 238 | 0 failures |
| `MLXGuidedGenerationTests` (model-free unit) | 69 | 0 failures |
| `MLXFoundationModelsTests` (model-free unit) | 138 | 1 pre-existing failure (`grammarBuilderHoistsNestedDefsInBothArms`), unrelated |
| Guided-generation integration group (12 suites: StreamingDelta, GenerableRoundTrip, PlainChatGeneration, MultiModelGuidedGeneration, GuidedGenerationIntegration, GuidedGeneration, MaxTokenTruncation, StopTokenRegression, XGrammarBridge, LoopInvariantsOnXGrammar, FastForwardTokenizationDisagreement, GoldenReplay) | 60 | 3 pre-existing failures on the tiny `gemma-3-270m` model (truncated JSON / runaway integer overflow), unrelated |
| Tool-calling integration group (5 suites: StructuredToolOutputSession, ToolCallRoundTrip, FoundationModelsToolCalling, MultiTurnToolCalling, ToolCallGrammarCharacterization) | 20 | 0 failures |
| End-to-end stress loop (`RequiredPropertyStressTests`) | 25 generations | 0 failures |

**Total: ~525 individual tests across ~44 suites, plus the 25-generation stress loop.** Both Gemma (SentencePiece) and Qwen (byte-level BPE) paths, guided generation, free-form generation, tool calling, and grammar handling are all covered. The most sensitive check — `StreamingDeltaTests`, which asserts exact streaming deltas — passes.

### Pre-existing failures (not caused by this PR; each confirmed identical on the baseline with the fix stashed)
1. `ToolCallingSchemaTests.grammarBuilderHoistsNestedDefsInBothArms` — tool-calling schema-shape assertion mismatch.
2. `GuidedGenerationTests` "Single Activity schema…" — `gemma-3-270m` emits truncated JSON.
3. `MultiModelGuidedGenerationTests.intRoundTrip[gemma-3-270m]` — `gemma-3-270m` emits a runaway integer that overflows Swift's `Int`.

### DIFF
--- a/IntegrationTesting/IntegrationTestingTests/MLXFoundationModelsIntegration/GuidedGeneration/RequiredPropertyStressTests.swift
+++ b/IntegrationTesting/IntegrationTestingTests/MLXFoundationModelsIntegration/GuidedGeneration/RequiredPropertyStressTests.swift
@@ -1,0 +1,87 @@
+// Copyright © 2026 Apple Inc.
+
+#if FoundationModelsIntegration && canImport(FoundationModels, _version: 2)
+
+import Foundation
+import Testing
+import FoundationModels
+
+@testable import MLXFoundationModels
+
+// Types from the bug report: a top-level object with three required
+// properties, one of which is a nested object.
+@available(iOS 27.0, macOS 27.0, visionOS 27.0, *)
+@Generable
+private struct NPC: Equatable {
+    let name: String
+    let coffeeOrder: String
+    let picture: Picture
+}
+
+@available(iOS 27.0, macOS 27.0, visionOS 27.0, *)
+@Generable
+private struct Picture: Equatable {
+    @Guide(description: "Avoid human-like descriptions. Stick to animals, plants, or objects.")
+    let imageDescription: String
+}
+
+/// End-to-end regression for the bug where guided generation dropped a required
+/// object property. A Gemma-4 (SentencePiece) checkpoint could emit a token
+/// whose decode was not append-only (a space collapsing before a comma), and
+/// the streaming detokenizer's character-count diff silently dropped the comma,
+/// producing invalid JSON that failed `Generable` decoding. See the fix in
+/// `NaiveStreamingDetokenizer.next()` and the model-free regression in
+/// `StreamingDetokenizerTests`.
+///
+/// Heavyweight (downloads a Gemma-4 checkpoint and runs many generations), so
+/// gated behind `MLX_RUN_REQUIRED_PROPERTY_STRESS=1` — it must not run on every
+/// integration pass. Mirrors the `MLX_RUN_VLM_INTEGRATION` / `MLX_RUN_COLD_FETCH`
+/// gates elsewhere in this target.
+@Suite(
+    .serialized,
+    .timeLimit(.minutes(30)),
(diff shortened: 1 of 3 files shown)


davidkoski (approved):
Nice find, thanks!

Issue #466, opened 2026-07-25
Upstreaming a batch of measured inference perf wins (roughly 1.3x MoE decode, plus prefill, memory and load time)


spokvulcan:
We spent the last few weeks on a systematic perf pass over the Qwen3.5/3.6 hybrids (dense 4B and MoE 35B-A3B, both 4-bit PARO, M3 Max). One hard rule throughout: zero output change. Every accepted change was gated token-identical in interleaved A/B runs, and the kernel-level ones were additionally proven bitwise against the ops they replace. We'd like to upstream everything that survived, and this issue is the map so the individual PRs stay small and reviewable.

Planned PRs in this repo:

| # | change | measured effect |
|---|--------|-----------------|
| 1 | compiled decode schedule for Qwen3.5/3.6: per-layer decode traces plus whole-step traced segments split at the KV write; leak regression tests and bitwise contract tests included | MoE decode +7.5% short ctx, +3.7% at 8K; dense decode +1.7% |
| 2 | GDN decode conv1d rewritten as four fused multiply-adds (bitwise only with f32 accumulation, which the contract test pins) | MoE decode +1.8% at 8K |
| 3 | fused router top-k Metal kernel. Today every decode token fully sorts all 256 expert scores to pick 8, because argpartition delegates to sort | MoE decode +1.9% short ctx |
| 4 | PARO follow-up batch for #164: MoE model support, prepared checkpoints (35B load 41 s down to 9 s), pre-gather rotation, simdgroup-resident rotation kernel, decay-gate compile fuse | MoE prefill +3 to 4.5%, MoE decode +3%, dense decode +3.4 to 5% |
| 5 | balanced prompt chunking, equal chunk sizes instead of leaving a remainder | up to ~9% prefill at prompt lengths that end in a degenerate last chunk |

And on the mlx core side, filed against ml-explore/mlx but listed here for the whole picture:

| # | change | measured effect |
|---|--------|-----------------|
| A | rows-per-expert-aware gather_qmm_rhs tile geometry | MoE 32K prefill +6% (kernel itself +13 to 22% at production shapes) |
| B | command-buffer commit accounting split into input and output legs, plus coalesced buffer retention | MoE decode +2.6 to 4.5%, dense peak memory down 9 to 13% |
| C | eval-path CPU overhead: memoized custom-kernel source, flat degree map in eval_impl, identity-index cache for the gather ops | MoE decode +5 to 6%, MoE 8K prefill +3.6% |
[...]

PR #467, opened 2026-07-25
Qwen3.5/3.6: run the decode step through compiled traces

spokvulcan:
Decode on the Qwen3.5/3.6 hybrids rebuilds the whole graph op by op from Swift on every token. This PR runs the single-token decode step through MLX compiled traces instead.

**How it works**

- The step is tiled into a few traced segments (9 to 11 for the shipped configs), each running from just after one full-attention layer's SDPA to just before the next one's.
- Segments split at the KV write, the only part of a decode step that cannot live inside a trace (the cache grows every token). Everything else is static-shaped, so the segments compile concretely; no shapeless tracing.
- Per-layer and per-block traces cover the fallback paths (quantized/turbo caches, padded batches, a cold GDN state). Prefill is untouched: it is GEMM-bound and would pay a fresh trace per prompt shape.
- The traced bodies are shared with the general path (`callAsFunction` is rebuilt on them), so the compiled and uncompiled paths cannot drift apart.

**Correctness**

`compile` only fuses elementwise chains here, with per-op rounding preserved, so generation is unchanged. Every intermediate step was gated token-identical against the unfused build: 108 interleaved A/B pairs across the 4B dense and 35B-A3B MoE (4-bit, M3 Max), at 128 and 8K context.

**Decode throughput (M3 Max)**

- MoE: +7.5% at short context, +3.7% at 8K
- Dense: +1.7% to +3.5% depending on context

**Tests**

The compiled closures are stored on the modules they capture, so the captures must be `unowned`. Two lifecycle tests pin that releasing the model deallocates the blocks, for the plain-cache path and the quantized-cache fallback.

### DIFF
--- a/Libraries/MLXLLM/Models/Qwen35.swift
+++ b/Libraries/MLXLLM/Models/Qwen35.swift
@@ -444,6 +444,7 @@ final class Qwen35SparseMoeBlock: Module, UnaryLayer {
         if x.dim(1) != 1 {
             return forward(x)
         }
+        compileLock.lock()
         if compiledForward == nil {
             // [unowned self]: stored only on self, so it cannot outlive self;
             // a strong capture would cycle and leak the module, weights and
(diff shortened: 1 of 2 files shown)


davidkoski (approved):
Changes look good, thank you!

PR #468, opened 2026-07-25
Qwen3.5/3.6: fold the GDN decode conv into the compiled step

spokvulcan:
At S == 1 the GDN depthwise conv is a fixed 4-term dot per channel. Written as elementwise multiply-adds it folds into the compiled decode segment from #467, which deletes one dispatch and one hazard barrier per GDN layer per token. Prefill and any S > 1 call keep the conv kernel.

**Why it is bitwise, not approximately equal**

- The multiply-adds accumulate in f32 and round once at the end, which is exactly what MLX's Convolution kernel does for f16/bf16 input. Probed over 8192 channels: every channel matches bit for bit, while native-dtype accumulation differs in ~47% of them.
- f32 input is the exception: the kernel's own f32 path accumulates in a different order (102 of 256 channels differ), so f32 stays on the kernel.

**Tests**

The fused and general conv are two named bodies, `decodeConv` and `generalConv`, and the new test pins one against the other bit for bit in CI. If an MLX bump changes conv accumulation, a unit test fails instead of decode silently diverging from prefill.

**Decode throughput (M3 Max)**

- MoE 35B-A3B (4-bit): +1.77% median at 8K context, +0.60% at 128
- Dense flat; prefill and peak memory unchanged; parity gate 32/32 pairs token-identical

### DIFF
--- a/Libraries/MLXLLM/Models/Qwen35.swift
+++ b/Libraries/MLXLLM/Models/Qwen35.swift
@@ -372,10 +372,15 @@ final class Qwen35GatedDeltaNet: Module {
             qkv = MLX.where(mask[.ellipsis, .newAxis], qkv, 0)
         }
 
-        let convInput = concatenated([convState, qkv], axis: 1)
-        let newConvState = contiguous(convInput[0..., (-(convKernelSize - 1))..., 0...])
-
-        let convOut = silu(conv1d(convInput))
+        let convOut: MLXArray
+        let newConvState: MLXArray
+        if S == 1 {
+            (convOut, newConvState) = decodeConv(convState: convState, qkv: qkv)
+        } else {
+            let convInput = concatenated([convState, qkv], axis: 1)
+            newConvState = contiguous(convInput[0..., (-(convKernelSize - 1))..., 0...])
+            convOut = silu(conv1d(convInput))
+        }
 
         let convSplit = MLX.split(convOut, indices: [keyDim, 2 * keyDim], axis: -1)
         let q = convSplit[0].reshaped(B, S, numKHeads, headKDim)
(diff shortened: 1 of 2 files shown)


Libraries/MLXLLM/Models/Qwen35.swift
```diff
@@ -412,6 +535,40 @@ final class Qwen35SparseMoeBlock: Module, UnaryLayer {
...
+            // tape on every model release. The trace bakes the weights
+            // captured at first trace: swapping parameters on a live module
+            // would silently replay stale weights — recreate the module
+            // instead.
+            compiledForward = compile { [unowned self] x in forward(x) }
+        }
+        return compiledForward!(x)
+    }
+
+    private var compiledForward: ((MLXArray) -> MLXArray)?
```
davidkoski, line 606: This isn't thread safe -- it either needs a lock to guard it or use an initializer, e.g.

```swift
private let _addRMSNorm: @Sendable (MLXArray, MLXArray, MLXArray) -> MLXArray = compile(
    shapeless: true
) { residual, x, weight in
    residual + MLXFast.rmsNorm(x, weight: weight, eps: kRMSEps)
}
```

I _think_ the initializer is lazy, but using a lock may be clearer in this case (e.g. TurboQuantKernelOps)

davidkoski (changes requested):
Change looks good but see comment on the thread safety -- it is the same pattern for all the compiled functions.

Thanks!

Libraries/MLXLLM/Models/Qwen35.swift
davidkoski: Thinking about it, the lock is the only way to go.  A property on the class is only thread safe settable at init-time and the weights are not loaded yet.

spokvulcan: Good catch, thanks. Went with the lock: NSLock around each lazy compile site, with the call itself outside the hold, same idiom as TurboQuantKVCache. Put it on the #467 branch since the pattern comes from there, and rebased this one on top.

davidkoski: Perfect, let me review/merge that one first then.

davidkoski:
I think this is ready to go, just needs to be rebased to fix the conflict (with 467, I think because of the squash merge).  Thanks!

davidkoski (approved):
Looks great, just need to resolve conflicts.

spokvulcan:
Rebased on main, so this is just the conv change now

PR #469, opened 2026-07-25
Qwen3.5/3.6 MoE: fused router top-k kernel for decode (bit-identical, CI-pinned)

spokvulcan:
Independent of #467 and #468, applies directly on main.

ArgPartition::eval_gpu carries the comment "We direct arg partition to sort for now" and delegates to gpu_merge_sort. So every MoE layer, every decode token, the router fully sorts all 256 expert scores to pick the top 8, then does a gather, a sum and a divide on top. Three serial dispatches and three encoder-wide hazard barriers for a 256 to 8 selection. This PR replaces all three with one small Metal kernel on the single-row decode path.

The kernel is bit-identical to the chain by construction, with each piece derived from the MLX source it replaces rather than assumed:

- order: sort.h's LessThan compares values only, ThreadSort swaps on strict less-than, and merge_step takes from A on ties, so the sort is stable. Element i's ascending position is therefore exactly its rank count with index tie-break. Packing the index into the low 32 bits of a monotone bit key folds the value comparison and the tie-break into one 64-bit compare.
- sum: reduce.metal instantiates float16 with U = float16_t, and a row of 8 takes thread_reduce, meaning sequential accumulation in the output dtype from zero in slot order. The kernel does the same.
- divide: elementwise, same dtype.
- indices come out uint32, same as argPartition.

Two edge cases needed explicit handling: NaN maps above +inf and all NaNs tie (LessThan is explicitly NaN-aware), and the sign of zero is normalised because -0.0 and +0.0 compare equal under < but have different bit patterns.

A new test holds the fused kernel against the chain over softmax rows, heavy ties, all-equal rows, signed zeros and NaN rows, in f16, bf16 and f32, with dtypes and bit patterns asserted. So if a future MLX bump changes the sort's tie-breaking or the reduce's accumulation, CI fails a unit test instead of decode silently diverging from prefill.

Decode only. At prefill the block sort's O(E log^2 E) beats this kernel's O(E^2) rank count and there is no barrier to save, so many-row callers keep the original chain.

Measured on the 35B-A3B MoE hybrid (4-bit, M3 Max): decode +1.91% at 128 context (10 of 10 pairs) and +0.47% at 8K, prefill and peak memory flat, 32 of 32 A/B pairs token-identical. Dispatches per token went 1914 to 1827.

Two notes for reviewers. The kernel lives in the Qwen3.5 model file for now; it would generalise to every SwitchGLU-style router, happy to move it somewhere shared if you'd prefer that. And the MLXVLM copy of this MoE block still has the original chain, we did not touch it because we could not A/B it with our models.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -83,7 +83,6 @@ let package = Package(
             dependencies: [
                 "MLXLMCommon",
                 .product(name: "MLX", package: "mlx-swift"),
-                .product(name: "MLXFast", package: "mlx-swift"),
                 .product(name: "MLXNN", package: "mlx-swift"),
                 .product(name: "MLXOptimizers", package: "mlx-swift"),
             ],
--- a/Package.swift
+++ b/Package.swift
@@ -83,6 +83,7 @@ let package = Package(
             dependencies: [
                 "MLXLMCommon",
                 .product(name: "MLX", package: "mlx-swift"),
+                .product(name: "MLXFast", package: "mlx-swift"),
                 .product(name: "MLXNN", package: "mlx-swift"),
                 .product(name: "MLXOptimizers", package: "mlx-swift"),
             ],
--- a/Libraries/MLXLLM/Models/Qwen35.swift
+++ b/Libraries/MLXLLM/Models/Qwen35.swift
@@ -9,7 +9,6 @@
 
 import Foundation
 import MLX
-import MLXFast
 import MLXLMCommon
 import MLXNN
 
--- a/Libraries/MLXLLM/Models/Qwen35.swift
+++ b/Libraries/MLXLLM/Models/Qwen35.swift
@@ -14,37 +14,19 @@ import MLXNN
 
 // MARK: - Fused router top-k
 
-/// One-kernel replacement for the decode router tail:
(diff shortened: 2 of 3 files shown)


Package.swift
```diff
@@ -83,6 +83,7 @@ let package = Package(
             dependencies: [
                 "MLXLMCommon",
                 .product(name: "MLX", package: "mlx-swift"),
+                .product(name: "MLXFast", package: "mlx-swift"),
```
davidkoski: MLXFast as a package is deprecated.  It lives on as an enum in MLX (for callers who use it like a namespace):

```swift
public enum MLXFast {
```

and the functions are also at the top level in MLX.

You should be able to remove this link and the import of MLXFast.

davidkoski (changes requested):
Looks great, but see comment on MLXFast -- I think it should be removed (it will work without it and if not that is a bug we need to fix!)

Package.swift
spokvulcan: Right, removed both. Builds clean.

spokvulcan:
Ready for the review

davidkoski (approved):
Looks great, ready to merge as soon as CI finishes.  Thank you!

PR #470, opened 2026-07-25
Prefill: balanced chunking behind a new PrefillParameters (~9% at 32K)

spokvulcan:
`LLMModel.prepare` used to chunk the prompt in fixed strides of `prefillStepSize` and hand the TokenIterator the leftover, up to stepSize − 1 tokens, evaluated in one forward at the largest Lk of the whole prefill. That forward is disproportionately expensive. On Qwen3.6-35B-A3B at 32K (warm, Release, M3 Max), a 1024-token chunk in the pipelined loop costs 0.98 s while the 141-token remainder costs 2.94 s. The fix divides the prompt into the fewest equal chunks that respect the ceiling (32 × 997 instead of 31 × 1024 + 141), so no degenerate forward exists. Measured on full production prefill, in-process and ABBA-alternated so thermal drift loads neither arm: 51.0 s → 46.2 s, about 9%; all five ABBA blocks across two runs favor balanced (sign test p = 0.031). Splitting the remainder off instead (140 + 1) measured neutral — the tokens have to be absorbed into the interior of the pipeline.

On outputs: any chunking reorders floating-point accumulation, so both chunkings were probed against the unchunked single forward as reference. Neither is closer to the truth — at 3K and 8K each flips the same single token out of 64, at a reference top-two margin of 0.03 against a median margin of 12.4; at 5K neither flips anything. The old chunking is an approximation of exactly the same magnitude, not a ground truth.

## The API

Following the review discussion, the strategy is now per-request rather than a global, and the `windowSize: Int?` in `LanguageModel.prepare` is replaced by a `PrefillParameters` struct:

- `stepSize` — the per-forward ceiling; `nil` keeps each model's own default.
- `chunking` — `.balanced` (default), `.remainder` (the legacy boundaries, for anyone with pinned outputs), `.unchunked` (the reference computation; quadratic memory, for validation).
- `progress` — a `(processed, total)` callback with the same semantics as mlx-lm's `prompt_progress_callback`, including the terminal `(total, total)`.

Future prefill knobs become additive fields instead of protocol breaks.

The chunked-prefill loop itself lives on the struct: `forEachChunk(total:reserving:defaultStepSize:maximumStepSize:_:)` computes the balanced chunk length and owns the loop's shared obligations — cooperative cancellation between chunks, a per-chunk autorelease pool, and the progress report — so every `prepare()` drives it against its own defaults and caps. This is the honest version of "hoist it so all models benefit": the TokenIterator can't compute the schedule because the nil-`stepSize` default is per-model (512 generic; Gemma3Text's tuned `min(128, slidingWindow)`), and the M-RoPE models can't chunk at all. A side effect worth naming: the cancellation check and pooling previously existed only in the default LLM path; every chunked prefill now has them.

## Per-model adoption

Gemma3Text, Gemma3, Gemma4 (text-only), Idefics3, FastVLM, LFM2VL, Pixtral, Mistral3, and the Qwen3.5 windowed continuation now chunk balanced and report progress (the continuation additionally reserves the final position, so lm_head materializes `[1, 1, vocab]` rather than a full last chunk). Qwen2VL, Qwen25VL, Qwen3VL, GlmOcr, and Paligemma stay single-shot — M-RoPE position lockstep or fused multimodal masks — and fire the terminal progress call. The ~9% is measured on the default path; the other loops get the same arithmetic on the same shape argument, unmeasured.

## Compatibility

`GenerateParameters.prefillStepSize` (from #439) and the inits that took it still work as deprecated forwarders, as does a deprecated `prepare(windowSize:)` method for out-of-tree callers; their deprecation messages note that they forward into the balanced default (`.remainder` restores the legacy boundaries). Out-of-tree `LanguageModel` implementers get a compile error pointing at the new signature. The one silent corner: an out-of-tree `LLMModel` conformer that customized `prepare` with the old signature would fall back to the default implementation — Swift gives no way to diagnose that, hence this note.

## Tests

Chunk-length properties; pinned schedules at the measured operating point (31,885 @ 1024 → 32 near-equal chunks + 1 vs 31 strides + 141), collected from `forEachChunk` itself so the pin holds the shipping code; `.remainder` legacy-boundary reproduction; exact progress sequences through a real `TokenIterator`; and the `.logits`-model progress contract on a tiny Gemma4.

### DIFF
--- a/Libraries/MLXLLM/LLMModel.swift
+++ b/Libraries/MLXLLM/LLMModel.swift
@@ -65,11 +65,11 @@ extension LLMModel {
     /// 1024 plus a straggler. `prefillStepSize` becomes a pure loop-efficiency knob,
     /// bounded only by the peak memory of one chunk.
     public func prepare(
-        _ input: LMInput, cache: [KVCache], state: LMOutput.State?, windowSize: Int?
+        _ input: LMInput, cache: [KVCache], state: LMOutput.State?, prefill: PrefillParameters
     ) throws
         -> PrepareResult
     {
-        let prefillStepSize = windowSize ?? 512
+        let prefillStepSize = prefill.stepSize ?? 512
         var y = input.text
 
         // A prompt that fits in one chunk is handed to the iterator whole, exactly
--- a/Libraries/MLXLLM/LLMModel.swift
+++ b/Libraries/MLXLLM/LLMModel.swift
@@ -27,51 +27,38 @@ extension LLMModel {
     ) throws
         -> PrepareResult
     {
-        let stepSize = max(1, prefill.stepSize ?? 512)
-        var y = input.text
+        let stepSize = prefill.resolvedStepSize()
+        let y = input.text
         let total = y.tokens.size
 
-        // A prompt that fits in one chunk is handed to the iterator whole:
-        // chunking it would only add a second forward. `.unchunked` (a nil
-        // chunk length) takes the same path at any prompt length.
-        guard total > stepSize,
-            let chunkSize = prefill.chunkLength(forChunking: total - 1, defaultStepSize: stepSize)
-        else {
-            return .tokens(y)
-        }
-        let tail = prefill.chunking == .remainder ? stepSize : 1
+        // A prompt that fits in one chunk is handed to the iterator whole,
+        // keeping short prompts bitwise-identical to the pre-chunking path.
+        // `.unchunked` (forEachChunk processes nothing) takes the same route
+        // at any prompt length.
+        guard total > stepSize else { return .tokens(y) }
 
+        var processed = 0
         try withPreparedCache(cache, lengths: y.sequenceLengths) {
             // asyncEval lets the CPU build chunk N+1's graph while the GPU evaluates
-            // chunk N.
+            // chunk N. Under .remainder the reserved tail is the legacy leftover
+            // (up to a full step) rather than a single token.
             var state: LMOutput.State? = state
-            while y.tokens.size > tail {
-                // Cooperative cancellation between prefill windows. On iOS, GPU work
-                // submitted after the app moves to the background is rejected by the
-                // system ("Insufficient Permission"), and the resulting command-buffer
-                // error is thrown from a Metal completion handler where it cannot be
-                // caught, aborting the process. Without this check a long prompt's
-                // prefill cannot be interrupted, so apps cannot stop GPU submissions
-                // in time when entering the background. See ml-explore/mlx-swift-examples#230.
-                try Task.checkCancellation()
-                // Pool per chunk: long prompts run hundreds of chunk forwards
-                // before returning to any autorelease boundary.
-                autoreleasepool {
-                    let n = min(chunkSize, y.tokens.size - 1)
-                    let input = y[.newAxis, ..<n]
-                    let output = self(input, cache: cache.isEmpty ? nil : cache, state: state)
-                    state = output.state
-                    asyncEval(cache)
-                    y = y[n...]
-                    prefill.progress?(total - y.tokens.size, total)
-                }
+            processed = try prefill.forEachChunk(
+                total: total, reserving: prefill.chunking == .remainder ? stepSize : 1
+            ) { range in
+                let input = y[.newAxis, range]
+                let output = self(input, cache: cache.isEmpty ? nil : cache, state: state)
+                state = output.state
+                asyncEval(cache)
             }
 
             // Single sync after the loop to flush any remaining async work.
-            eval(cache)
+            if processed > 0 {
+                eval(cache)
+            }
         }
 
-        return .tokens(y)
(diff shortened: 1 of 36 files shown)


Libraries/MLXLLM/LLMModel.swift
```diff
@@ -4,6 +4,34 @@ import Foundation
...
+    ///
+    /// This is the *reference* computation — the one chunking approximates. It is
+    /// not usable in production (the `[1, H, Lq, Lk]` score matrix is quadratic in
+    /// the prompt), but at a few thousand tokens it fits, and it is the only way to
+    /// ask whether a chunking scheme moves logits *toward or away from* the true
+    /// answer. Comparing two chunkings to each other cannot answer that: neither is
+    /// a ground truth.
+    case unchunked
+
+    nonisolated(unsafe) public static var strategy: PrefillChunking = .balanced
```
davidkoski: We could just guard it with a lock to make it thread safe, but I think this is fine.

Is there any benefit to scoping this to a single request?  It seems like balanced is probably what you would want in all cases anyway.  Just wondering if we need to think about some kind of prefill parameters.

On a similar note @angeloskath mentioned that mlx-lm has observability on the prefill (prompt_progress_callback) so if more parameters are going to be passed in, perhaps we can inject this.

Libraries/MLXLLM/LLMModel.swift
```diff
@@ -27,12 +72,30 @@ extension LLMModel {
         let prefillStepSize = windowSize ?? 512
         var y = input.text
 
+        // A prompt that fits in one chunk is handed to the iterator whole, exactly
+        // as before: chunking it would only add a second forward.
+        let chunkSize: Int
+        let tail: Int
+        switch PrefillChunking.strategy {
```
davidkoski: I noticed there are 20-ish implementations of `prepare()` -- only models that use the default will get this behavior.

I wonder if this would be better hoisted into the TokenIterator?  That could compute a new `windowSize` and then (presumably) all the models would be able to use it, e.g. here:

```swift
    mutating func prepare(input: LMInput, windowSize: Int? = nil) throws {
        processor?.prompt(input.text.tokens)

        switch try model.prepare(input, cache: cache, state: state, windowSize: windowSize) {
```

Libraries/MLXLLM/LLMModel.swift
spokvulcan: Went the full way on this. The global is gone; there's now a PrefillParameters struct on GenerateParameters that rides through prepare(), so the strategy is per-request and the thread safety question disappears. It carries stepSize, the chunking case, and, since the seam was right there, Angelos's suggestion too: a progress callback with the same (processed, total) semantics as mlx-lm's prompt_progress_callback. prefillStepSize still works as a deprecated forwarder.

Libraries/MLXLLM/LLMModel.swift
spokvulcan: I tried the iterator version first and it can't quite reach the models it wants to. When prefillStepSize is nil the default step lives inside each model (512 generic, Gemma3Text uses min(128, slidingWindow)), so the iterator would have to guess what to balance against. And Qwen2VL/Qwen25VL ignore the step on purpose, chunking them would need positionIds sliced in lockstep with the embeddings. So the hoist landed one layer lower instead: the chunked-prefill loop itself is now PrefillParameters.forEachChunk, and every prepare() that chunks drives it against its own default and cap. That also gave all of those loops the cancellation check and autorelease pooling that only the default LLM path had. The models that can't chunk just report progress, and all of the implementations take the struct.

davidkoski: what if the LanguageModel had a default PrefillParameters property?  Would that help?

I will read through what you did -- it sounds good!

davidkoski: I think this is a great approach -- I noticed that the cancellation handling was only in the one spot also. This compositional approach solves that nicely.

spokvulcan: I tried that shape first, but the model side is really two things: a default step for when the request leaves stepSize nil, and a cap that clamps even an explicit stepSize (Gemma3Text caps at slidingWindow so a chunk never exceeds the rotating cache window). A property can express the default but not the clamp, and the single-shot models can't chunk at all. With Gemma3Text the only model differing from the generic 512, passing defaultStepSize/maximumStepSize at the call site felt like the smaller API. Happy to add a property if you'd want the defaults visible from outside.

spokvulcan:
Rebased onto main to clear the conflicts with #448

davidkoski:
Rebased & resolved conflicts.

davidkoski:
Frustrating -- CI doesn't want to appear.

davidkoski (approved):
Changes look good, thank you!

davidkoski:
Ran docs and unit tests locally because GitHub won't let me run the actions.  🤷

PR #471, opened 2026-07-25
ParoQuant: extend to MoE and make the whole path fast (#164 follow-up)

spokvulcan:
Follow-up to #164, which added ParoQuant (pairwise rotation quantization) for dense models. This PR extends it to MoE models and then makes the whole path fast. Eight commits in dependency order, each reviewable on its own.

**MoE support**

- `RotateSwitchGLU` gives the MoE expert block the same shared-rotation treatment the dense path got in #164, with the loader passes to match. This is what lets a PARO-quantized Qwen3.6-35B-A3B run at all.
- Two AWQ conversion fixes found while converting more checkpoints: convert every AWQ prefix, and cast scales to f16.

**Faster loading**

- The AWQ-to-PARO conversion matching was O(n^2) over tensor names and dominated load time. An O(1) rewrite plus a prepared-checkpoint artifact (convert once, reuse after) takes the 35B load from 41 s to 9 s.

**Faster inference**

- Rotate gate_up before the expert gather/sort. Rotation is row-independent and the gather only duplicates rows, so this is bitwise-identical at an eighth of the work. MoE prefill +3.2% at 8K, +4.5% at 32K.
- Compile-fuse the GatedDelta decay gate chain: six elementwise kernels (~180 launches per token on the 35B) become one. MoE decode +3.1% at short context, dense +1.4%.
- A simdgroup-resident rotation kernel. The old kernel spent its time in eight serialized threadgroup barrier rounds; one 32-lane simdgroup per tile makes that sync free. Kernel alone 1.7 to 2x; end to end, prefill +1.3 to 2.5% and dense decode +3.4 to 5%. Also fixes a latent bf16 compile failure. The old kernel stays as the fallback for groupSize != 128.

**Correctness**

Every numeric change was gated token-identical in interleaved A/B runs on both PARO models (4B dense and 35B-A3B MoE, M3 Max), and the three inference changes were additionally verified bit-identical against the paths they replace. All ParoQuantTests pass (24 tests, grown with this PR).

Numbers compound with the rest of the batch in #466; this PR alone is worth roughly +4 to 7% MoE prefill, +3% MoE decode, +4 to 7% dense decode, and the 4.6x load-time win on the 35B.


davidkoski:
Looks like tests are failing:

```
/Users/runner/actions-runner/_work/mlx-swift-lm/mlx-swift-lm/Tests/MLXLMTests/ParoQuantTests.swift:616: error: -[MLXLMTests.ParoQuantTests testPreparedCheckpointRoundTripIsBitExact] : XCTUnwrap failed: expected non-nil value of type "Dictionary<String, MLXArray>"
Test Case '-[MLXLMTests.ParoQuantTests testPreparedCheckpointRoundTripIsBitExact]' failed (1.048 seconds).
Test Case '-[MLXLMTests.ParoQuantTests testPreparedCheckpointStaleManifestSelfHeals]' started.
/Users/runner/actions-runner/_work/mlx-swift-lm/mlx-swift-lm/Tests/MLXLMTests/ParoQuantTests.swift:635: error: -[MLXLMTests.ParoQuantTests testPreparedCheckpointStaleManifestSelfHeals] : XCTAssertTrue failed
```

spokvulcan:
Found it. volumeAvailableCapacityForImportantUsage reports 0 on the runner, so the free space guard skipped the write. Now falls back to volumeAvailableCapacity.

Libraries/MLXLMCommon/ParoQuant/ParoQuantLoader.swift
```diff
@@ -469,12 +723,39 @@ public func loadParoQuantModel<T: LanguageModel>(
...
+    // 12b. Persist the Prepared Checkpoint in the background when this load
+    //      had to convert. Runs after the materialization above so the
+    //      model-bound tensors are already realized (the dict shares them
+    //      by reference); readiness never waits on the write, and a failed
+    //      or interrupted write only means the next load converts again.
+    if !usedPreparedCheckpoint {
+        // `[String: MLXArray]` is not Sendable; the boxed dict holds the
+        // immutable, already-evaluated checkpoint tensors shared with the
+        // live model, read-only on both sides.
+        let weightsBox = SendableBox(preSanitizeWeights)
```
davidkoski, line 795: This is _mostly_ true.  There could be weights that are in the safetensors file that are discarded, e.g. things in the vision tower might be dropped.  These were never loaded (lazy) and we may not need to write them back out.

davidkoski: @aleroot FYI, this is maybe a case to consider for the concurrent/synchronous weight loading:  it may load weights that the model doesn't want.  There could be a cost to the higher performance version in some cases.  Maybe :-)

Libraries/MLXLMCommon/ParoQuant/PairwiseRotation.swift
```diff
@@ -0,0 +1,492 @@
...
+    public init(dims: Int, groupSize: Int, krot: Int) {
+        self.theta = MLXArray.zeros([krot, dims / 2])
+        self.pairs = MLXArray.zeros([krot, dims], type: Int16.self)
+        // Assign through `.wrappedValue` so the `@ParameterInfo(key:)` metadata
+        // survives init — see the matching note in RotateQuantizedLinear.
+        self._channelScales.wrappedValue = MLXArray.ones([1, dims])
+        self.groupSize = groupSize
+        self._rotation = RotationDerivedState(dims: dims, krot: krot)
+
+        super.init()
```
davidkoski, line 487: Should this `freeze()` at the end?  `SwitchGLU.supportsDirectWeightedReduction` requires no trainable parameters.

Libraries/MLXLMCommon/ParoQuant/PairwiseRotation.swift
```diff
@@ -0,0 +1,492 @@
...
+    }
+
+    /// Recompute from the loaded checkpoint parameters. The results are
+    /// lazy graph nodes — materializing them is the caller's job (see
+    /// `RotationStatePreparing` for why).
+    mutating func prepare(
+        theta: MLXArray, pairs: MLXArray, channelScales: MLXArray, groupSize: Int
+    ) {
+        cosTheta = MLX.cos(theta)
+        sinTheta = MLX.sin(theta)
```
davidkoski, line 411: This will prepare the sin/cosTheta in float16 I think -- is that OK or should this be done in float32?

Libraries/MLXLMCommon/ParoQuant/ParoQuantLoader.swift
```diff
@@ -139,32 +172,109 @@ private func convertAutoAWQ(
...
+                continue
+            }
             if let val = weights[key], val.ndim == 1 {
                 weights[key] = val.reshaped(1, -1)
             }
+        } else if key.hasSuffix("scales") {
+            guard prefixes.contains(String(key.dropLast("scales".count))) else { continue }
+            // float16 to match `biases` in `quantizedMM` — f32 scales with
+            // f16 biases mismatch (upstream fix z-lab/paroquant#38).
+            weights[key] = weights[key]!.transposed().asType(.float16)
```
davidkoski: Would we ever see bfloat16?  If so, the float16 and bfloat16 mixed would produce float32.

Libraries/MLXLMCommon/ParoQuant/ParoQuantLoader.swift
```diff
@@ -399,44 +588,100 @@ public func loadParoQuantModel<T: LanguageModel>(
...
-    //    subdirectories, otherwise nested artefacts like an HF snapshot
-    //    cache under the checkpoint dir would be pulled in).
+    // Chat conventions. Same precedence as the model factories: an explicit
+    // value from the caller wins; then a registered resolver, which sees the
+    // model id the model cannot; then the model's own declaration.
+    if config.toolCallFormat == nil {
+        config.toolCallFormat =
+            ChatConventionsRegistry.shared.toolCallFormat(
+                modelId: config.name, modelType: baseConfig.modelType)
+            ?? model.toolCallFormat
```
davidkoski: This changed recently and should be:

```swift
                ?? ToolCallFormat.resolved(
                    forTokenizerDirectory: configuration.tokenizerDirectory,
                    modelFormat: model.toolCallFormat)
```

Libraries/MLXLMCommon/ParoQuant/PairwiseRotation.swift
```diff
@@ -0,0 +1,492 @@
...
+    let tile = batch <= 1 ? 1 : 4
+    let params = MLXArray([Int32(batch), Int32(dim), Int32(krot), Int32(groupSize)])
+
+    let kernel: MLXFast.MLXFastKernel
+    let threads: Int
+    if groupSize == 128 {
+        kernel = getRotationKernel(tile: tile, krot: krot, dtype: flat.dtype)
+        threads = 32
+    } else {
+        kernel = getGenericRotationKernel(tile: tile)
```
davidkoski: Consider groupSize > 128 -- the kernel inside this assumes a max size of 128.  Either make that dynamic or guard vs 128 and fatal if we can't handle it.

Libraries/MLXLMCommon/ParoQuant/PairwiseRotation.swift
```diff
@@ -0,0 +1,492 @@
...
+///   cross-simdgroup rendezvous to pay for. The tile layout is row-major
+///   (`tile[row * 128 + ch]`), so pair accesses are bank-conflict-free
+///   for any ROWS_PER_TILE (the old channel-major `tile[ch * R + row]`
+///   layout collapsed onto 8 banks for R = 4).
+/// - x/out IO is vectorized (`\(t4)` per lane covers the 128-channel
+///   group exactly); the f32 threadgroup tile is written/read as float4.
+///   `channel_scales` is loaded scalar + converted so its dtype may
+///   legitimately differ from the activation dtype.
+/// - The write-back casts explicitly (`\(t)(...)`): the old kernel's
+///   implicit float->element conversion did not compile for bfloat16.
```
davidkoski: Do we need a runtime check?  What if groupSize != 128 but we have a bfloat16?  Or is this saying it would now work?

Libraries/MLXLMCommon/ParoQuant/PairwiseRotation.swift
```diff
@@ -0,0 +1,492 @@
...
+/// generic kernel. Shared by `PairwiseRotation` and `RotateQuantizedLinear`.
+///
+/// Zero-row inputs pass straight through: gathered MoE activations can be
+/// legitimately empty, and a zero-sized grid dispatch is undefined. The
+/// guard lives here, with the grid math, so no caller needs one.
+nonisolated func dispatchPairwiseRotation(
+    _ flat: MLXArray, state: RotationDerivedState, groupSize: Int, krot: Int
+) -> MLXArray {
+    let batch = flat.dim(0)
+    if batch == 0 { return flat }
```
davidkoski, line 334: Does this need a dim/groupSize divisibility check as well?

Libraries/MLXLMCommon/ParoQuant/ParoQuantLoader.swift
spokvulcan: Right, it does. The snapshot is taken before sanitize on purpose so the same artifact feeds both the LLM and VLM containers, and the price is the vision tower riding along: 667 MB of the 4B artifact (17%), 893 MB of the 35B (4%). Those tensors are plain f16 and never go through conversion, so a follow-up could write only the converted keys and let the loader take the rest lazily from the source file. I'd rather do that separately than fold it in here, but happy to if you'd prefer it now.

Libraries/MLXLMCommon/ParoQuant/PairwiseRotation.swift
spokvulcan: Good catch, it didn't, added freeze() at the end of init

Libraries/MLXLMCommon/ParoQuant/PairwiseRotation.swift
spokvulcan: It does stay f16. theta is f16 in the checkpoints and the reference does mx.cos(theta) on it as is, so this matches z-lab bitwise. I measured it over the 200 theta tensors of the 4B: max |cos²+sin²-1| is 6.8e-4 in f16 vs 1.2e-7 in f32, well under the 4-bit weight noise. Going to f32 is a two line change since the kernel already upcasts, it just breaks parity with the reference and with every token-identical gate in this PR, so I left it

Libraries/MLXLMCommon/ParoQuant/ParoQuantLoader.swift
spokvulcan: You'd get an f32 matmul, quantized_matmul promotes to the widest of x, scales and biases. Nothing bf16 has shipped (every float tensor in both checkpoints is f16), but c39c560 now casts scales and biases to the checkpoint's float dtype, read from its rotation tensors, so a bf16 checkpoint stays bf16 end to end. Test covers both dtypes.

Libraries/MLXLMCommon/ParoQuant/ParoQuantLoader.swift
spokvulcan: Done, thanks

Libraries/MLXLMCommon/ParoQuant/PairwiseRotation.swift
spokvulcan: Went the dynamic route. The generic kernel is now templated on groupSize, krot and element type, so the tile and coefficient arrays are sized for the model instead of a fixed ceiling. What's left is structural (even groupSize, whole groups, half a group per CTA so at most 2048) and that's checked once: a typed error at load with the key path, and a precondition in the public inits.

Libraries/MLXLMCommon/ParoQuant/PairwiseRotation.swift
spokvulcan: It works now rather than needing a check. Same commit templates the element type into the generic kernel with an explicit cast on the store, so bf16 goes through both kernels. The CPU reference test now runs f16, bf16 and f32 on groupSize 8 and 128.

Libraries/MLXLMCommon/ParoQuant/PairwiseRotation.swift
spokvulcan: It didn't, no. The geometry check above covers dims % groupSize == 0 and dims >= groupSize. Before it an odd tail would just not get rotated, since numGroups floors.

Libraries/MLXLMCommon/ParoQuant/ParoQuantLoader.swift
davidkoski: That sounds fine -- it isn't a correctness issue.

davidkoski (approved):
Change looks good to me.  Thank you!

PR #472, opened 2026-07-26
Fix structured ChatSession continuations

aleroot:
## Proposed changes

This PR is a correctness follow-up to #313, which added support for continuing a cached `ChatSession` with structured messages such as tool results.

### Problem

`ChatSession` keeps a KV cache so the model does not need to process the entire conversation again on every turn. However, after #313, only the newly supplied messages were passed to the model’s chat template.

For example, after a tool call, the template could receive only:

```text
tool: <tool result>
```

The original user question and the assistant’s tool call existed in the model’s KV cache, but chat templates cannot inspect that cache. They can only inspect the structured messages provided to them.

Some conversation-aware templates, including Qwen templates, require the original user question to format a tool result correctly. When it was missing, generation failed with a Jinja error such as:

```text
No user query found in messages.
```

This made valid structured tool continuations fail in downstream agent applications, even though the tool itself completed successfully.

### Solution

`ChatSession` now retains the structured conversation together with its KV cache. For every continuation, it gives the chat template the complete conversation, including:

- The original user message
- The assistant response or tool call
- Tool results
- Later user and assistant messages

This allows conversation-aware templates to understand the relationship between a tool result and the user request that caused it.

The performance benefit introduced by #313 is preserved. Although the complete conversation is rendered, `ChatSession` compares the resulting tokens with those already represented by the KV cache. When the cached tokens are still an exact prefix, only the new token suffix is processed by the model.

If the template changes an earlier part of the prompt—for example, because the instructions changed—`ChatSession` safely rebuilds the cache instead of combining a new prompt with stale model state.

The implementation also:

- Preserves the existing behavior of low-level initializers that receive only a raw KV cache, because those initializers do not have enough information to reconstruct the original conversation.
- Handles standard and speculative decoding cache alignment.
- Rebuilds safely when a new turn introduces media that cannot be independently sliced from the prepared input.
- Records generated tokens inside the generation task using Swift 6 compiler-checked value semantics.

### Next step: message-boundary cache checkpoints

> This PR safely rebuilds the prompt whenever a rendered transcript changes an already-cached token region and the underlying cache cannot be trimmed.
>
> That is correct, but hybrid/recurrent models may lose cache-reuse performance because SSM/Mamba state cannot be rewound to an arbitrary longest-common-prefix offset. Wrapped rotating caches can have similar restrictions.
>
> A future optimization should retain bounded cache checkpoints at stable message boundaries particularly before assistant generation. When a chat template rewrites or removes part of the previous assistant response, `ChatSession` could restore the latest checkpoint that is still an exact prefix and prefill only the remaining suffix, without trimming recurrent state.
>
> The checkpoint should include:
>
> - Main and draft model caches.
> - Exact token prefix and offset.
> - Per-model continuation state such as M-RoPE state.
> - Compatibility information for media, masks, processing configuration, and cache parameters.
> - A bounded retention policy to control memory usage.
>
> Validation should compare checkpoint reuse against cold full-prefill logits for standard KV, rotating, and hybrid/SSM models, including Qwen templates that remove reasoning traces. Benchmarks should report cached-token count and TTFT at increasing conversation lengths.

## Checklist

  - `pre-commit` is not installed on this machine. The configured Swift formatter was run manually on the changed files, formatter lint passed, and `git diff --check` is clean.

### DIFF
--- a/IntegrationTesting/IntegrationTestingTests/ToolCallIntegrationTests.swift
+++ b/IntegrationTesting/IntegrationTestingTests/ToolCallIntegrationTests.swift
@@ -94,6 +94,12 @@ struct ToolCallIntegrationTests {
         try await ToolCallTests.qwen35EndToEndGeneration(container: container)
     }
 
+    @Test func qwen35StructuredToolContinuation() async throws {
+        let container = try await models.llmContainer(
+            for: .init(id: IntegrationTestModelIDs.qwen35))
+        try await ChatSessionTests.structuredToolContinuation(container: container)
+    }
+
     @Test func qwen35MultiTool() async throws {
         let container = try await models.llmContainer(
             for: .init(id: IntegrationTestModelIDs.qwen35))
--- a/Libraries/IntegrationTestHelpers/IntegrationTestHelpers.swift
+++ b/Libraries/IntegrationTestHelpers/IntegrationTestHelpers.swift
@@ -257,6 +257,69 @@ public enum ChatSessionTests {
         }
     }
 
+    /// Exercises the structured continuation path used by clients that execute
+    /// tool calls outside `ChatSession` and feed the results back afterward.
+    ///
+    /// Conversation-aware templates must receive the original user query and
+    /// assistant tool call again on the result turn. Rendering only the new
+    /// `.tool` message reproduces the Qwen Jinja "No user query found" failure.
+    public static func structuredToolContinuation(container: LLModelContainer) async throws {
+        let session = ChatSession(
+            container,
+            instructions:
+                "Use the weather tool whenever weather is requested. After receiving its result, answer the user briefly.",
+            generateParameters: GenerateParameters(maxTokens: 150, temperature: 0),
+            additionalContext: ["enable_thinking": false],
+            tools: [weatherToolSchema]
+        )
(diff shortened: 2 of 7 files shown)


aleroot:
Supersede #370 
Added 82caaa3 to make the API equivalence explicit. The tests now separately verify that both the legacy string API and structured-message API render the complete transcript while prefilling only the uncached suffix. The documentation also clarifies that both overload families share the same retained-conversation and cache-reuse path.


I benchmarked this on an Apple M2 using a quantized 8-layer Gemma test model, one generated token, and medians from five runs.
With approximately 729 / 2,909 / 5,821 tokens in the complete prompt, cached continuation through https://github.com/ml-explore/mlx-swift-lm/pull/472 took 17.10 / 18.83 / 18.82 ms, versus 38.45 / 130.14 / 243.81 ms when correctly rebuilding and prefilling the conversation: 2.25× / 6.91× / 12.96× faster.
These are controlled microbenchmark numbers rather than production-model latency, but they confirm the intended scaling behavior: model prefill depends on the uncached suffix, not the complete conversation length.

davidkoski (approved):
This change looks good to me!  Thank you!

PR #473, opened 2026-07-27
Add DeepSeek-OCR and Unlimited-OCR VLM support (MLXVLM)

### LINKED ISSUE #15: Support for Deepseek OCR
Hi @davidkoski thanks a ton for such useful package! 
I would like to request support for DeepSeek-OCR in the MLXVLM library. DeepSeek-OCR is a competitive open-source VLM that offers strong performance.

I'm willing to contribute to the implementation if guidance is provided 

Thanks!

jyauxi:
## Proposed changes

Adds two closely-related OCR VLMs to MLXVLM. Closes #15.

**Unlimited-OCR is a small delta on DeepSeek-OCR, which is why they are one PR.**
`UnlimitedOCR` is a 60-line file: two typealiases plus a `DeepseekOCR` subclass whose only
override is `newCache()`, returning a `RingSlidingKVCache` instead of the default. Reviewing
the two separately would mean reviewing the same SAM+CLIP vision stack, the same gundam
processor and the same MoE decoder twice in order to land a cache policy. If you would still
rather take them as two PRs, say so and I will split it — the history is already sequenced
so that DeepSeek-OCR ends at a clean commit with Unlimited-OCR stacked on top.

### DeepSeek-OCR

- Port from mlx-vlm (Python): SAM + CLIP dual vision encoder, 16x convolutional compressor,
  projector, and the DeepSeek-V2 MoE language decoder.
- `DeepseekOCRProcessor` with the `gundam` / `base` crop modes as first-class API
  (`DeepseekOCRProcessor.Mode`, `modeContext(_:)` / `mode(from:)`).
- Grounding / ref / det token surface: `DeepseekOCRSpecialTokens` (id resolution, prompt
  builders, `parseDetections(from:)`).
- Registry wiring: `deepseekocr` model type, `DeepseekOCRProcessor` / `DeepseekVLV2Processor`
  processor types.
- Adapted from [mzbac/deepseek-ocr.swift](https://github.com/mzbac/deepseek-ocr.swift) (MIT),
  refactored toward mlx-swift-lm patterns (`GlmOcr`, `DeepseekV3` sanitize / `@ModuleInfo`).
  Attribution is in `ACKNOWLEDGMENTS.md` and in the `DeepseekOCR.swift` header.

### Unlimited-OCR

- `RingSlidingKVCache` — Reference Sliding Window Attention (R-SWA). Prefill retains the full
  prompt and image KV; decode then overwrites a ring of `sliding_window_size` (default 128),
  so the cache stays constant-size regardless of output length.
- `UnlimitedOCR`: `newCache()` returns R-SWA caches. Everything else is inherited.
- Registry wiring for the `unlimited_ocr` / `unlimited-ocr` model types.
- Multipage fused prepare (`base` mode, `max_num=32`) on the shared processor.
- Optional sliding-window no-repeat n-gram logits processor, **off by default**, matching
  mlx-vlm.

### `RingSlidingKVCache` is not a substitute for `RotatingKVCache`

They differ in all three of prefill retention, rotation policy and RoPE offset, so the
existing cache cannot be reused here:

| | `RingSlidingKVCache` (R-SWA) | `RotatingKVCache` (existing) |
|--|------------------------------|------------------------------|
| Prefill | Retains **full** prompt + image KV | Truncates / rotates per rotating policy |
| Decode | After `prefill_length + window_size`, new KV overwrites a **ring** of size `window_size` | Keeps leading tokens; rotates from the start |
| RoPE | Absolute `ropeOffset` keeps increasing | Offset policy differs |

### Three things that touch shared code — flagging them up front

Adding two models should ideally touch only MLXVLM. Three changes reach further, each for a
concrete reason, and I am happy to narrow any of them. No new package dependencies are added.

**1. `VLMModelFactory.defaultHonorOrigModelType` defaults to `true`.** This teaches the shared
`_load` path about `_orig_model_type`. The motivation is real: Unlimited-OCR packs ship
`model_type: deepseekocr` with `_orig_model_type: unlimited-ocr` as a back-compat shim for
loaders that predate the Unlimited registration, so honoring the original type is what makes
such a pack load as Unlimited-OCR rather than silently as DeepSeek-OCR with an unbounded cache.
It is still a behavioral default change on a path shared by PaliGemma, Qwen, Gemma and the rest,
for a concept only these two packs use. If you prefer, `defaultHonorOrigModelType = false` with
the OCR call sites opting in explicitly is a small change — the per-call `honorOrigModelType:`
overloads already exist, so nothing else moves.

**2. `safetensorWeightURLs` falls back to `model.safetensors`** when a shard named in
`model.safetensors.index.json` is missing. Some published OCR packs index
`model-00001-of-000001.safetensors` while shipping `model.safetensors`, and without this they
fail to load at all. The cost is that a genuinely mispackaged repo now loads instead of erroring,
for every model, not just these. If you would rather not soften that path, I can gate the
fallback on the OCR model types or drop it and document the packs as unsupported.

**3. `SlidingWindowNoRepeatNGramProcessor` in MLXLMCommon (opt-in).** Ports mlx-vlm's
repetition guard for long OCR decodes. Nothing touches `GenerateParameters`: the guard attaches
per-request via `GenerationComponents(logitProcessorFactory:)` from #401. (An earlier revision
added three `GenerateParameters` properties for this; they were dropped after review feedback
in favor of the components route.)

## Checklist


Libraries/MLXLMCommon/BaseConfiguration.swift
```diff
@@ -207,6 +215,30 @@ public struct BaseConfiguration: Codable, Sendable {
...
+            let orig = origModelType?
+                .trimmingCharacters(in: .whitespacesAndNewlines),
+            !orig.isEmpty
+        else {
+            return modelType
+        }
+        let normalized = orig.lowercased().replacingOccurrences(of: "_", with: "-")
+        if normalized == "unlimited-ocr" {
+            return "unlimited-ocr"
+        }
```
davidkoski, line 238: This is not extensible outside of mlx-swift-lm -- developers cannot override this or add to it.  I think it is OK to merge but we should follow up with a way to extend it.

Ideas:

- a dictionary with with replacement text, e.g. `["unlimited_ocr": "unlimited-ocr"]`
- delegation -- register an array of closures, e.g. `(String) -> String?`

These could be stored in a locked array as a static property on BaseConfiguration.  It could be initialized with this (above) handling and callers could add to it if they had other needs.

Libraries/MLXLMCommon/Evaluate.swift
```diff
@@ -128,6 +128,19 @@ public struct GenerateParameters: Sendable {
...
+    /// (e.g. `35` in Unlimited-OCR examples), ``GenerateParameters/processor()``
+    /// attaches ``SlidingWindowNoRepeatNGramProcessor``.
+    public var noRepeatNgramSize: Int?
+
+    /// Search window for ``noRepeatNgramSize``. Unlimited-OCR examples use
+    /// `128` (single image) or `1024` (multi-page/PDF).
+    public var noRepeatNgramWindowSize: Int
+
+    /// Token IDs never banned by the n-gram processor (e.g. EOS).
+    public var noRepeatNgramWhitelistTokenIds: Set<Int>
```
davidkoski: I am a little leery of adding more to GenerateParameters.  We already have a repetition penalty, which sounds similar.  Is this new processor general purpose?  Or is it more likely to be used with these OCR models?

I wonder if this would fit better once #401 lands?

Libraries/MLXLMCommon/Evaluate.swift
```diff
@@ -231,7 +303,11 @@ public struct ArgMaxSampler: LogitSampler {
     public init() {}
 
     public func sample(logits: MLXArray) -> MLXArray {
-        argMax(logits, axis: -1)
+        // Cast to bfloat16 before argmax so ranking matches MLX Python models that
+        // keep decode logits in bf16. Float32 accumulation can invent a spurious
+        // ranking among tokens that are exact ties in bf16 (e.g. DeepseekOCR
+        // "4" vs "7" on the solid-red L3 fixture).
+        argMax(logits.asType(.bfloat16), axis: -1)
```
davidkoski: Neither mlx-lm or mlx-vlm do this -- both process the logits in whatever dtype the model produces.  This would affect _all_ models, not just these OCR models.  I wonder if the OCR models are producing float32 logits and perhaps _they_ should do the dtype conversion?

Libraries/MLXLMCommon/KVCache.swift
```diff
@@ -801,6 +801,174 @@ public class RotatingKVCache: BaseKVCache, CustomDebugStringConvertible {
...
+///
+/// Keeps the full prompt/prefill KV, then appends decode tokens until a ring of
+/// ``windowSize`` is full. Afterwards new decode keys/values overwrite that ring
+/// while `offset` (and therefore RoPE absolute positions) keeps increasing.
+///
+/// Do **not** substitute ``RotatingKVCache`` — that cache keeps leading tokens and
+/// rotates from the start, which is a different sliding-window semantics.
+///
+/// Port of Python `mlx_vlm.models.unlimited_ocr.language.RingSlidingKVCache`.
+public class RingSlidingKVCache: KVCacheSimple {
```
davidkoski: Looks good!

Libraries/MLXLMCommon/Evaluate.swift
xocialize: Author of #401 here — happy to confirm it fits, and it'd actually let this PR drop some code rather than add it.

`SlidingWindowNoRepeatNGramProcessor` is already a `LogitProcessor`, so once #401 lands it attaches through `GenerationComponents` with nothing added to `GenerateParameters`:

```swift
var components = GenerationComponents()
components.logitProcessorFactory = {
    SlidingWindowNoRepeatNGramProcessor(
        ngramSize: 35, windowSize: windowSize, whitelistTokenIds: whitelist)
}
```

That directly addresses the "leery of adding more to `GenerateParameters`" concern — the three `noRepeatNgram*` properties wouldn't need to live on the value type. Two things that line up nicely:

- The factory is invoked once per generation, so the processor's `prompt`/`didSample` state stays fresh each turn — which is what you want for a stateful n-gram tracker in a multi-turn session.
- #401 also adds a public `ChainedLogitProcessor` that's the same idea as the `SequentialLogitProcessor` this PR introduces (compose penalty + custom in order), so that helper would already be provided — no need for both.

The OCR-specific defaults (ngram size / window) are orthogonal to where the behavior lives — those could stay as a small preset that builds the `components`, without attaching to `GenerateParameters`. Glad to help wire it up whichever way you land once #401 is in.

Libraries/MLXLMCommon/Evaluate.swift
jyauxi: Good catch — you're right that this belongs in the model, not the shared sampler. Fixed in 2f857c5: `ArgMaxSampler` is back to upstream's plain `argMax(logits, axis: -1)`, and `DeepseekOCR.computeLogits` now emits bf16 logits itself (which `UnlimitedOCR` inherits).

Background on why the cast exists at all: Python mlx-vlm keeps decode logits in bf16, so greedy argmax there resolves ties at bf16 precision. The Swift port was ranking in higher precision, which broke exact-tie ordering on one of our parity fixtures (token "4" vs "7" on a synthetic solid-color image). Casting at the model boundary reproduces the Python ranking exactly, and no other model is affected now. Verified behavior-preserving: all three of our gold fixtures (single-page gundam ×2, three-page base with the n-gram processor active) reproduce byte-identically against their pre-change baselines.

Libraries/MLXLMCommon/Evaluate.swift
jyauxi: Agreed — and thanks @xocialize for confirming the fit. The processor itself is general purpose (it's a sliding-window no-repeat-n-gram ban), but today only these OCR models are known to need it, so I'd rather not grow `GenerateParameters` for it.

Happy to go whichever way you prefer:

1. **If #401 lands first** (my preference): I'll rebase, drop the three `noRepeatNgram*` properties and `SequentialLogitProcessor`, and attach `SlidingWindowNoRepeatNGramProcessor` through `GenerationComponents`, keeping the OCR-specific defaults (n-gram 35, window 128/1024) as a small preset that builds the components.
2. **If you'd rather merge this PR first**: I can drop the `GenerateParameters` additions now — the processor stays `public`, callers construct it directly — and wire it into `GenerationComponents` as a follow-up once #401 lands.

Either way the processor and its tests are unchanged; it's only the attachment point that moves.

Libraries/MLXLMCommon/BaseConfiguration.swift
jyauxi: Agreed on both counts — fine as a follow-up, and the replacement-dictionary shape seems right: a locked static `[String: String]` on `BaseConfiguration`, initialized with the current handling, that callers can add entries to (with the closure-delegation form as an escalation if someone needs pattern-based rewriting rather than exact matches). Happy to open that follow-up PR once this lands.

Libraries/MLXLMCommon/Evaluate.swift
jyauxi: Done in `b988c9f` — went with option 1 now that #401 has landed.

The three `noRepeatNgram*` properties and `SequentialLogitProcessor` are gone.
`SlidingWindowNoRepeatNGramProcessor` now attaches through `GenerationComponents`:

```swift
let components = GenerationComponents(
    logitProcessorFactory: { SlidingWindowNoRepeatNGramProcessor.unlimitedOCRSingleImage() }
)
```

`Evaluate.swift` is byte-for-byte identical to `main` again — this PR no longer touches
`GenerateParameters` at all. #401's public `ChainedLogitProcessor` does what the
`SequentialLogitProcessor` here was doing, so that helper is deleted rather than duplicated.
Net −71 lines.

One deviation from what I outlined above: instead of adding a preset that builds a
`GenerationComponents`, I left the OCR defaults where they already were — the existing
`SlidingWindowNoRepeatNGramProcessor.unlimitedOCRSingleImage()` / `.unlimitedOCRMultiPage()`
statics (n-gram 35, window 128/1024). Wrapping one of those in `GenerationComponents` is a
single line at the call site, so a separate preset didn't seem to earn its keep. Happy to add
one if you'd rather it be spelled out.

Thanks @xocialize — the factory-per-generation semantics are exactly right here. The processor
accumulates token history through `didSample`, so a fresh instance per generation is what you
want in a multi-turn session.

I also merged current `main` into the branch in the same push (this brings in #502's
`ChatConventionsProviding` migration). The one conflict was in `VLMModelFactory.swift`, where
the new `conventionsRegistry` landed adjacent to this PR's `_orig_model_type` remap docs — both
sides kept, and the tool-call/reasoning lookups now go through the registry rather than the
deleted `infer` chains. After the merge this PR deletes 7 lines total from pre-existing files
(5 in `Load.swift`, 2 in `VLMModelFactory.swift`); everything else is additive.

Verified locally: build clean, the eight `SlidingWindowNoRepeatNGramProcessor` tests pass,
`swift-format` lint clean, and DocC builds clean for `MLXLMCommon` and `MLXVLM`.

Libraries/MLXLMCommon/KVCache.swift
```diff
@@ -801,6 +801,174 @@ public class RotatingKVCache: BaseKVCache, CustomDebugStringConvertible {
...
+            // prefill length until the first single-token decode step.
+            if seqLen > 1 {
+                return super.update(keys: keys, values: values)
+            }
+            prefillLength = offset
+        }
+
+        let prefill = prefillLength!
+        if self.keys == nil || offset < prefill + windowSize {
+            let result = super.update(keys: keys, values: values)
```
davidkoski: I think this needs to consider seqLen -- super doesn't have any idea of the windowSize and seqLen can be > 1.

Libraries/MLXLMCommon/KVCache.swift
davidkoski: The port is faithful, but I wonder if this needs to be a subtype of `BaseKVCache` instead (like RotatingKVCache).  Consider:

```swift
    func isQuantizable(_ cache: KVCache) -> Bool {
        if let list = cache as? CacheList {
            return list.children.contains(where: isQuantizable)
        }
        return cache is KVCacheSimple
            && !(cache is QuantizedKVCache)
            && cache.offset > quantizedKVStart
    }
```

if we had a config where quantization was attempted, this cache would be quantized, likely with ill effect.

Either change the parent type or maybe we need an `isQuantizable` property on the KVCache that allows opt-in/opt-out.  KVCacheSimple could produce true and this would override to false.

Libraries/MLXLMCommon/KVCache.swift
jyauxi: Good catch — fixed in b546ee4. `update` no longer delegates blindly while the window is filling: a step that fits below `prefill + windowSize` grows in place, and a multi-token step that overlaps the ring now follows `RotatingKVCache`'s concat precedent — it returns [prefix | last `windowSize − 1` decode tokens | new tokens] in temporal order so every new token sees its full window, then retains only prefix + last `windowSize`. `makeMask` produces the matching array mask for that case (prefix always visible, decode columns limited to each query's trailing window). Also clamped `trim()`: trimming back to/below the marked prefix now unmarks it, so prompt-cache reuse can't leave a stale boundary (related to your `isTrimmable(after:)` idea on #506).

For the record, the Python reference has the same overshoot in its grow branch (and its multi-token ring branch writes slots sequentially without reordering), so this is a deliberate divergence-as-fix rather than drift — single-token decode, the only path Python exercises in practice, is numerically unchanged (our gold fixtures are byte-identical). Happy to upstream the same fix to mlx-vlm.

New tests: boundary-crossing step, wrapped-ring reorder, ring-step mask layout, trim-below-prefix. The branch is also current with `main` again (merged through #513, adopted the #470 `PrefillParameters` signature).

Libraries/MLXLMCommon/KVCache.swift
jyauxi: Went with your first suggestion: `RingSlidingKVCache` now extends `BaseKVCache` directly, mirroring `RotatingKVCache` — it owns its allocation path and no longer inherits `KVCacheSimple`'s `toQuantized()` surface, so the `is KVCacheSimple` eligibility check skips it for free. After merging current `main`, that composes nicely with the typed cache configuration from #453 too: the ring cache classifies as unsupported-for-compression in `KVCacheTree`, so both the legacy path and the new plan machinery leave it alone. That felt cleaner than adding an `isQuantizable` knob to the `KVCache` protocol for one cache — but if you'd like that opt-out on the protocol anyway (for future caches), happy to add it.

Test added: `maybeQuantizeKVCache` leaves the ring cache untouched while a `KVCacheSimple` control with identical contents converts.

jyauxi:
**Since your 2026-08-07 review** — three commits, in review order:

1. [`b546ee4`](https://github.com/jyauxi/mlx-swift-lm/commit/b546ee4) — the response to both comments: `RingSlidingKVCache` reparented to `BaseKVCache` + `seqLen > 1` handling ([isolated compare](https://github.com/jyauxi/mlx-swift-lm/compare/b988c9f...b546ee4): 2 files, `KVCache.swift` + its tests). This is the only commit that needs review.
2. [`13065f0`](https://github.com/jyauxi/mlx-swift-lm/commit/13065f0) — merge of current `main` (your #453/#462/#470/#146/#506/#513; zero conflicts, no manual edits).
3. [`ff98294`](https://github.com/jyauxi/mlx-swift-lm/commit/ff98294) — #470 `PrefillParameters` adoption (same shape as `GlmOcr`) + the root-README OCR section folded into `Libraries/MLXVLM/README.md`.

Workflows need approval to run on this push; the same matrix is green locally (build, 423 tests / 40 suites incl. the new #453/#470 suites, swift-format). Details in the two thread replies.

jyauxi:
**Catch-up merge with `main` — no change to the code under review.**

Since the last round this branch has merged `main` twice, most recently `a4a05e0`. Both are pure upstream catch-up:

- The OCR port is **byte-identical** to the head you last reviewed — `DeepseekOCR.swift`, `UnlimitedOCR.swift` and `SlidingWindowNoRepeatNGramProcessor.swift` all diff empty, and `RingSlidingKVCache` is untouched inside `KVCache.swift`.
- The only non-upstream content is a three-line conflict resolution. #523 (Muse-Glimmer) appends to the same three lists this PR does, so both sets of entries now sit side by side:
  - `Libraries/MLXVLM/VLMModelFactory.swift` (type registry, processor registry, `all()`)
  - `Libraries/MLXVLM/README.md`
  - `skills/mlx-swift-lm/references/supported-models.md`

A couple of notes on upstream changes that land near this PR, in case they save you a question:

- **#475** gave GLM-OCR an anchored warm-cache continuation path. The equivalent doesn't apply here: DeepSeek-OCR's decoder uses 1-D RoPE off `cache.ropeOffset`, so positions already equal the cache offset and there's no `rope_deltas` anchor to carry — `QwenVL.continuationAnchor` would be guarding an invariant this model satisfies by construction. Adopting `PrefillParameters.forEachChunk` chunked prefill in `DeepseekOCR.prepare` *is* worthwhile and the ring cache already tolerates it (`prefillLength` defers to the first single-token decode step, so N prefill chunks produce the same reference prefix as one forward). Happy to fold that into this PR if you'd rather have it here than as a follow-up — it needs a gold-fixture parity run since chunking changes float reduction order.
- **#514** doesn't reach `RingSlidingKVCache`; `keepCount` and `effectiveKVCacheCapacity` are package-scoped on `RotatingKVCache`.
- **Known gap, your call:** `restoreCacheFromMetaState` has no `RingSlidingKVCache` case, so saving and reloading a prompt cache for an Unlimited-OCR model throws `Unknown cache class`. The cache already implements `state`/`metaState`, so it's a small addition — say the word and I'll add it here, otherwise I'll keep it as a follow-up.

Local verification on the merge: `xcodebuild build-for-testing` green; `xcrun xctest MLXLMTests` → **581 tests / 49 suites, 0 failures**, 1 env-gated skip (`LAYER_PARITY_GOLDEN`).

davidkoski:
Don't worry about the xcode27 failure -- there is API drift in the beta SDK

Libraries/MLXLMCommon/KVCache.swift
```diff
@@ -899,6 +899,340 @@ public class RotatingKVCache: BaseKVCache, CustomDebugStringConvertible {
...
+///
+/// Multi-token steps before the first single-token step are treated as
+/// (possibly chunked) prefill and extend the retained reference prefix; the
+/// prefix boundary is marked by the first single-token decode step.
+///
+/// Do **not** substitute ``RotatingKVCache`` — that cache keeps leading tokens and
+/// rotates from the start, which is a different sliding-window semantics.
+///
+/// Port of Python `mlx_vlm.models.unlimited_ocr.language.RingSlidingKVCache`.
+public class RingSlidingKVCache: BaseKVCache, CustomDebugStringConvertible {
```
davidkoski, line 937: This needs the `isTrimmable(after:)` (newish API, probably after this was written) -- see RotatingKVCache

Libraries/MLXLMCommon/KVCache.swift
```diff
@@ -2071,6 +2406,16 @@ private func restoreCacheFromMetaState(
         cache.metaState = metaState
         return cache
 
+    case "RingSlidingKVCache":
+        guard metaState.count >= 4, let windowSize = Int(metaState[0]) else {
+            throw KVCacheError(
+                message: "Invalid RingSlidingKVCache metaState - expected 4 values")
+        }
+        let cache = RingSlidingKVCache(windowSize: max(windowSize, 1))
```
davidkoski, line 2426: I think this should probably have the guard like this other cases:

```swift
if !state.isEmpty {
    cache.state = state
}
```

Libraries/MLXVLM/Models/DeepseekOCR.swift
```diff
@@ -0,0 +1,2107 @@
...
+                preprocess(image: image, side: basePadSide),
+                [],
+                1,
+                1
+            )
+        }
+    }
+
+    @_spi(Testing)
+    public func prepareForTesting(input: UserInput) async throws -> PreparedImageInputs {
```
davidkoski: I think this name should be clarified -- it is also used in the production code path.  The "Testing" is misleading.  `internalPrepare()` or something along those lines perhaps?

Libraries/MLXVLM/Models/DeepseekOCR.swift
```diff
@@ -0,0 +1,2107 @@
...
+        guard !input.images.isEmpty else {
+            let promptTokens = tokenizer.encode(
+                text: userPromptText(from: input), addSpecialTokens: true)
```
davidkoski: This should probably use the message generator / chat template, e.g.:

```swift
        let messages = Qwen2VLMessageGenerator().generate(from: input)

        var promptTokens = try tokenizer.applyChatTemplate(
            messages: messages, tools: input.tools,
            additionalContext: input.additionalContext)
```

Libraries/MLXVLM/Models/DeepseekOCR.swift
```diff
@@ -0,0 +1,2107 @@
...
+                crop
+                    .normalized(mean: config.imageMeanTuple, std: config.imageStdTuple)
+                    .asMLXArray()
+                    .asType(.bfloat16))
+        }
+
+        return (processed, tilesWide, tilesHigh)
+    }
+
+    private func userPromptText(from input: UserInput) -> String {
```
davidkoski: This looks equivalent to `MessageGenerator`, e.g. see `Qwen2VLMessageGenerator`

Libraries/MLXVLM/VLMModelFactory.swift
```diff
@@ -355,9 +387,50 @@ public final class VLMModelFactory: GenericModelFactory {
...
+    ///     `model_type` is the DeepSeek shim. When `false`, the pack's own
+    ///     `model_type` is used (``DeepseekOCR``, plain KV cache).
+    ///
+    /// > Note: This is a per-call argument rather than process-global state on
+    /// > purpose: a caller loading two families concurrently would otherwise
+    /// > have to serialize the set-then-load pair itself.
+    public func _load(
+        configuration: ResolvedModelConfiguration,
+        tokenizerLoader: any TokenizerLoader,
+        honorOrigModelType: Bool
```
davidkoski: Rather than adding an overload, I think this could just have an optional with a default.

Libraries/MLXVLM/VLMModelFactory.swift
```diff
@@ -460,9 +535,13 @@ public final class VLMModelFactory: GenericModelFactory {
         let processorTypeOverrides: [String: String] = [
             "mistral3": "Mistral3Processor",
             "gemma4_unified": "Gemma4UnifiedProcessor",
+            // Hub packs often ship DeepseekVLV2Processor; native unlimited-ocr
+            // routing prefers UnlimitedOCRProcessor (same tokenize path).
+            "unlimited-ocr": "UnlimitedOCRProcessor",
+            "unlimited_ocr": "UnlimitedOCRProcessor",
```
davidkoski: See also https://github.com/ml-explore/mlx-swift-lm/pull/565 -- there may be some incoming changes in this area.

davidkoski (changes requested):
See questions/feedback regarding KVCache API and prepare/text handling.  Thanks!

Libraries/MLXLMCommon/KVCache.swift
jyauxi: Added `RingSlidingKVCache.isTrimmable(after:)` using the same predictive boundary as `RotatingKVCache`: the cache is trimmable only while `offset + positions < prefillLength + windowSize`. This rejects reuse before an append would enter ring-overwrite mode. Covered by `testIsTrimmablePredictsWhetherAppendWouldFillRing`.

Libraries/MLXLMCommon/KVCache.swift
jyauxi: Agreed. Restoration now assigns `cache.state` only when the serialized state is non-empty. `testEmptySerializationRoundTripDoesNotAssignEmptyState` covers the empty round trip.

Libraries/MLXVLM/Models/DeepseekOCR.swift
jyauxi: Renamed this to `internalPrepare`. Production `prepare` delegates to the same helper; the Testing SPI only exposes the intermediate result for assertions.

Libraries/MLXVLM/Models/DeepseekOCR.swift
jyauxi: Switched the preparation path to a `MessageGenerator` followed by `tokenizer.applyChatTemplate`.

One checkpoint-specific difference from the Qwen example: the shipped DeepSeek/Unlimited templates render `message.content` directly and do not translate Qwen-style structured image parts into `<image>`. `DeepseekOCRMessageGenerator` therefore emits the configured image placeholder as string content while preserving the full chat history and tool metadata.

The cached Unlimited production tokenizer now renders exactly one image token for a normal one-image prompt.

Libraries/MLXVLM/Models/DeepseekOCR.swift
jyauxi: Removed the last-user-prompt helper. Model-specific conversion now lives in `DeepseekOCRMessageGenerator`, which conforms to `MessageGenerator`. Tests cover string-content generation, explicit placeholders, multipage input, and preservation of the remaining chat-template output.

Libraries/MLXVLM/VLMModelFactory.swift
jyauxi: Removed the additional load overloads. `honorOrigModelType` is now a defaulted, immutable `VLMModelFactory` initializer option.

I kept `_load(configuration:tokenizerLoader:)` with its existing signature because it is the `GenericModelFactory` protocol witness; adding a defaulted argument does not satisfy that witness. This keeps the default behavior configurable without expanding the load overload family.

Libraries/MLXVLM/VLMModelFactory.swift
jyauxi: Adapted this to the #565-style processor-loading path. The Unlimited aliases now resolve through `VLMProcessorLoadingRegistry`, including checkpoints declaring `DeepseekVLV2Processor`.

I also changed the test to exercise `VLMProcessorLoadingRegistry.shared` directly instead of constructing a resolver with a duplicated copy of the production mapping.

jyauxi:
Thanks for the latest review. I merged current `main` (`9219508`) into the branch and addressed each open thread directly.

Verification:
- `MLXLMTests`: 693 tests / 54 suites, 0 failures
- strict `swift-format` check passed on the PR changes
- cached Unlimited production tokenizer rendered exactly one `<image>` token for a normal one-image prompt

The prompt test double now matches the checkpoints’ real string-only chat-template behavior, and the processor-registry test exercises the shared production registry instead of duplicating its mapping.

Ready for re-review when convenient.

jyauxi:
**Catch-up merge with `main` — no change to the code under review.**

Merged current `main` (`37688d2`) into the branch: one merge commit, zero conflicts, no manual edits. The OCR port is byte-identical to the head you last reviewed — `DeepseekOCR.swift`, `UnlimitedOCR.swift`, the processor, and the factory/registry files all diff empty, and `RingSlidingKVCache` is untouched inside `KVCache.swift` (the only delta in that file is #329's `VarianceNormalizedKVCache` arriving alongside it).

The main motivation is CI: the previous run's only failure was `integration_build_xcode27`, failing on the known beta-SDK drift in `MLXFoundationModels` — a target this PR doesn't touch, and the job #582 has since disabled on the PR path. This merge brings that in, so the rerun reflects only code this PR owns.

Local verification on the merge:
- `xcodebuild build-for-testing` green; `scripts/verify-docs.sh` passed
- `MLXLMTests`: 604 XCTest tests (0 failures, 2 env-gated skips) + 722 Swift Testing tests / 55 suites, all passed
- Local caveat: built with the `FoundationModelsIntegration` trait disabled, since `MLXFoundationModels` doesn't compile against the current Xcode 27 beta seed (same drift the retired CI job hit). That matches the stable-Xcode CI environment, where the adapter compiles out; `MLXLMTests` doesn't depend on it.

All seven threads from your 2026-08-24 review remain addressed as of the previous push — this merge changes none of that code. Ready for re-review when convenient.

PR #475, opened 2026-07-27
Prompt cache: persist model state, wire Qwen3-VL and GLM-OCR restoration, and fail closed

kklimuk:
## Problem

`savePromptCache` saved KV arrays but dropped `LMOutput.State`. Qwen vision models and GLM-OCR keep their M-RoPE continuation anchor (`ropeDeltas`) in that state, outside the KV cache. Restoring only KV therefore succeeded while assigning later text and images different positions from the uninterrupted session.

```text
Before save:  KV cache + ropeDeltas
Saved file:   KV cache
After load:   KV cache + no continuation anchor
```

Persistence is only half of the fix. A model must also consume the restored state when it prefills a warm suffix, and incomplete cache/state pairs must be rejected instead of silently assuming an anchor of zero.

## Scope

Following the split @NivDvir proposed in [the review thread](https://github.com/ml-explore/mlx-swift-lm/pull/475#issuecomment-5123562727), #448 took Qwen2.5-VL and Qwen2-VL and has since merged. This PR is rebased on it and owns the rest: the prompt-cache format and public snapshot API, Qwen3-VL and GLM-OCR consumption, and hardening of the existing Qwen3.5/3.6 path. It does not touch the Qwen2-family files.

### Model map

| Model | Anchor key | Consumption owner | This PR |
| --- | --- | --- | --- |
| Qwen3.5 / 3.6 | `qwen35.ropeDeltas` | existing #399 path | fail-closed and rank/batch hardening |
| Qwen3-VL | `qwen35vl.ropeDeltas` | this PR | warm continuation + windowed prefill + guard |
| GLM-OCR | `glmocr.ropeDeltas` | this PR | warm continuation + windowed prefill + guard |
| Qwen2.5-VL | `qwen25vl.ropeDeltas` | #448 (merged) | unchanged here |
| Qwen2-VL | `qwen2vl.ropeDeltas` | #448 (merged) | unchanged here |

### 1. Restore cache and state as one snapshot

`PromptCacheSnapshot` keeps the continuation pieces together:

```swift
public struct PromptCacheSnapshot {
    public let cache: [KVCache]
    public let metadata: [String: String]
    public let state: LMOutput.State?
}
```

- `savePromptCache` accepts optional model state.
- `loadPromptCacheSnapshot(url:)` restores KV, caller metadata, and state.
- `ChatSession.saveCache(to:)` persists its current model state with KV.
- Snapshot-based `ChatSession` initializers restore both together.
- `PromptCacheSnapshot` has a public initializer for in-process cache builders.

### 2. Store state as safetensors tensors and fail closed in old readers

Array-backed state entries are ordinary safetensors tensors under a reserved namespace. Metadata stores the format version and logical keys; keys are sorted for deterministic layout.

State-bearing cache class names receive a versioned compatibility marker. The snapshot loader understands the marker and restores both pieces. The legacy tuple loader, old Swift readers, and Python readers reject the unknown state-bearing cache class rather than silently loading KV alone.

State-free files retain their existing layout. Unsupported state values and malformed layouts fail before an incomplete destination file is written.

### 3. Consume restored state in Qwen3-VL and GLM-OCR

Both models now use the carried anchor for warm suffixes and support windowed prefill:

```text
positionOffset = cacheOffset + carriedRopeDelta
```

They compute explicit M-RoPE position IDs from that offset, prefill the suffix in bounded windows, and return a resume anchor for the following turn.

GLM-OCR additionally needed two underlying fixes:

- its decode-time rope-delta branch was incorrectly nested under the precomputed-position branch, so an anchor-only resume state fell through to unanchored sequential positions;
- its text-only path inflated already-batched tokens by another axis and could fail during reshape.

A small performance pass rode along on the touched paths, in case the diff raises eyebrows: Qwen3-VL's deepstack scatter indices are hoisted out of the per-layer loop (they are loop-invariant, and the old per-layer mask lookup forced a GPU→CPU sync per layer), a redundant `sum().item()` sync in the image merge is gone, and GLM-OCR's three identical per-token M-RoPE rows collapsed into one broadcast on the decode path. All behavior-preserving — the broadcast rows are identical by construction on that branch, and the continuation tests pin the outputs.

### 4. Fail loudly when continuation state is missing

For the stateful models owned by this branch:

- throwing `prepare` paths raise `ContinuationStateError.missingState` when a warm cache has no anchor;
- nonthrowing direct decode paths enforce the same invariant with a precondition (the protocol signature cannot throw);
- unsupported warm batched continuation is rejected rather than silently resetting positions;
- cold text-only prefill returns an explicit zero anchor so an ordinary live conversation remains valid.

The error directs callers to `loadPromptCacheSnapshot(url:)`, the snapshot-based `ChatSession` initializer, and state-bearing `savePromptCache`.

A text-only warm cache without state is rejected too: after KV has been detached from its history, the model cannot prove that the cached prefix contained no media. Existing KV-only files must be rebuilt from the original history.

### 5. Let continuation state ride through speculative decoding

An earlier revision of this branch declared stateful models ineligible for speculative decoding (`requiresContinuationState`). That gate is gone — it was solving a problem these models don't have. The state they carry is a position anchor, not an accumulator: decode reads `ropeDeltas` and positions from `cacheOffset + delta`, never rewriting the delta. A rejected proposal trims the KV rows *and* the offset, so the anchor stays correct with no rollback. The state that genuinely cannot be rewound — the MTP drafter's per-position hidden states — rides its own `MTPSpeculativeTokenIterator` and is untouched here.

Concretely:

- `SpeculativeTokenIterator` accepts `mainState:`, threads it through prefill and every verification pass, and exposes the current value as `state` — the same contract `TokenIterator` already had.
- `ChatSession` seeds carried state into the speculative iterator and reads the post-prefill state back, so a stateful model keeps its anchor across speculative turns and `saveCache` persists it.
- Speculation is skipped — falling back to normal generation and abandoning the draft cache — when the turn carries media, or when a warm main cache has no matching draft cache (a restored snapshot, or any turn after a fallback): the draft cannot catch up without re-prefilling history it doesn't have. Rebuilding the draft cache from the session's token history is a follow-up; see Boundaries.
- The five cache-taking `generate` / `generateTokens` helpers now accept `state:`, so a snapshot restored outside `ChatSession` can be continued through the free functions instead of silently dropping its anchor — the exact failure this branch exists to fix.

### 6. Steer callers away from splitting snapshots

`ChatSession(ctx, cache: snapshot.cache)` still compiles, so documentation now identifies it as the unsafe split and points callers to `promptCache:`. The legacy `loadPromptCache(url:)` API is documented as state-free and rejects state-bearing files rather than dropping their state.

## Tests

All model tests use tiny inline random-weight configurations; no model downloads are required.

- **Continuation tests** for Qwen3-VL, GLM-OCR, and Qwen3.5 assert that a warm continuation matches a full cold prefill, that windowed prefill matches single-shot (including an image straddling a window boundary, and Qwen3-VL's deepstack slicing), that a cold prefill carries a zero anchor, that state survives a prompt-cache round trip, and that a warm cache arriving without its anchor throws rather than drifting.
- **KV cache tests** cover state tensor round trips, deterministic layout, the compatibility marker in both directions, legacy files, malformed layouts, and every supported cache class.
- **ChatSession tests** cover snapshot initialization and save/restore, carried state running under speculative decoding, and the two fallbacks — a media turn, and a warm prebuilt cache with no draft cache.

Full package suite: **319 XCTest tests and 401 Swift Testing tests, 0 failures**.

## Snapshot size

State size depends on what the model returns. A continuation state is generally only the anchor. A cold Qwen3.5 or GLM-OCR prefill may also contain precomputed position IDs, approximately `3 × tokens × 4` bytes—small relative to the KV cache, though still worth trimming in a follow-up.

## Boundaries

- **Qwen2.5-VL and Qwen2-VL still fail open.** #448 gave them warm continuation but treats a missing anchor as zero rather than throwing, so a KV-only restore drifts silently there where the models in this PR now refuse. Porting the `ContinuationStateError` guards across is a follow-up.
- **The free `generate` helpers seed state in but do not hand the post-turn state back.** The `Generation` stream carries no state, so a caller who needs the next snapshot should drive `TokenIterator` directly (its `state` property is public) or use `ChatSession`. Returning state through the stream's completion info is planned as a follow-up.
- **Speculation does not yet resume after a media turn or a snapshot restore.** The draft cache is abandoned on fallback and never rebuilt, so those sessions stay on normal generation. `ChatSession` already tracks the token history needed to re-prefill a draft; wiring that up (and persisting prompt tokens in snapshots so restores can do the same) is a follow-up.
- **Batched warm continuation is rejected**, not supported: the models capture per-row anchors but consume a single scalar. Uniform-offset batched resume (one snapshot expanded to N rows) is a natural follow-up.
- **A restored session appends turns instead of re-rendering the transcript.** A snapshot carries KV and model state but not the structured conversation, so later turns append the new message where a live session (since #472) re-renders the whole thing. On an image-bearing turn a live session rebuilds its cache while a restored one continues warm: positions stay correct, but the prompts differ — and on vision encoders that attend across image boundaries (Qwen2-VL passes `mask: .none` over the joint patch sequence, and the processor concatenates every image into one tower call) so do the new image's features, since it is encoded alone rather than beside the cached one. Qwen2.5-VL, Qwen3-VL, and GLM-OCR isolate each image and are unaffected. Persisting the transcript would close this, and would also supply the token ids the draft rebuild above needs. Thanks to @NivDvir for flagging [the divergence](https://github.com/ml-explore/mlx-swift-lm/pull/475#issuecomment-5137801225) and [the vision-encoder narrowing](https://github.com/ml-explore/mlx-swift-lm/pull/475#issuecomment-5151665297).
- **Provenance-based state reconstruction is not included.** Per-image grid provenance is required to rebuild state at an arbitrary radix-cache offset or rescue legacy KV-only files.

## Issues and related work

- Addresses [#443](https://github.com/ml-explore/mlx-swift-lm/issues/443).
- Advances the remaining Qwen3-VL and GLM-OCR items in [#420](https://github.com/ml-explore/mlx-swift-lm/issues/420).
- Builds on [#448](https://github.com/ml-explore/mlx-swift-lm/pull/448), now merged, which wired Qwen2.5-VL and Qwen2-VL.


NivDvir:
Reviewed the full diff, and ran it against real Qwen2.5-VL grounding. Numbers at the bottom.

The compatibility marker is the strongest part of this design.
I built the same fix locally with a metadata `format_version` and realized it fails open: old readers ignore unknown metadata, skip the state tensors, and silently restore KV-without-state — recreating the exact bug being fixed.
Rewriting the cache class name closes that in old Swift and Python readers both.
The empty-state → legacy-format behavior and failing before the file is written are also right.

One scoping note.
On current main, Qwen2.5-VL and Qwen2-VL `prepare` still take `state _:` and ignore it.
For those models a snapshot restore succeeds and the first resumed turn still decodes at wrong positions.
#448 wires the consumption side; the two together close #443 for that family.
Until then "QwenVL" here is effectively Qwen3.5.

Minor: `ChatSession(ctx, cache: snapshot.cache)` still compiles without the snapshot's state — in-process unbundling the file format can't catch.
Maybe a doc note on the `cache:` init pointing at `promptCache:` as the preferred path.

Measured — this PR merged locally onto #448 (clean merge, disjoint file sets), Qwen2.5-VL-7B-Instruct-4bit, temp 0, grounding prompts:
- Byte-identity: seed turn → `saveCache` → `loadPromptCacheSnapshot` → identical continuation turn. Restored output byte-equal to the never-saved session on 2/2 areas. Restore cost ~5 ms for a ~155 MB cache.
- Warm viability (the #420 methodology, 10 rows): 0/10 warm-empty after the disk round trip. Same substrate without #448 underneath: 10/10 empty.
- Warm-vs-cold edge deltas: max 114 px, mean 53 — at or better than our frozen in-memory warm reference (423 / 93).

So on this family the PRs compose: #475 persists, #448 consumes, and neither alone closes #443.

kklimuk:
@NivDvir working on getting the fixes for the other models in here as well.

NivDvir:
Heads-up before you write the Qwen2.5-VL / Qwen2-VL side — that's #448, already built and measured (the numbers above ran on #475+#448 merged; disjoint file sets, clean merge).
It's more than minimal `state:` consumption: windowed prefill with positionIds sliced in lockstep, and continuations whose remainder contains a new image.
The full map on main: only Qwen3.5/3.6 consume restored state. Qwen2.5-VL, Qwen2-VL, Qwen3-VL and GlmOcr all take `state _:` and ignore it — Qwen3-VL per #420.
Natural split so we don't collide: #448 covers the Qwen2 pair; Qwen3-VL + GlmOcr are unclaimed.
Or carry it all and cherry-pick #448 with attribution — either way I'll re-run the measured comparison on the combined shape.

kklimuk:
@NivDvir it grew **a bit** 😅 . 
I followed your advice to let it work together with #448.
A lot of it's is just tests - can remove those if you prefer.

davidkoski:
Should I review #448 next so we can get it merged -- would that help?

NivDvir:
@davidkoski Yes — and you already did, thank you. #448-first is the clean order: #475 doesn't touch the Qwen2 model files, so it rebases clean on top.

@kklimuk Keep the tests.
They're the acceptance bar for exactly the silent-position failure class this is about — a warm continuation that compiles, runs, and decodes at wrong positions only fails in tests like these.
The model map in the new description is exactly the split. Thanks for formalizing it.

Once both land I'll re-run the #420 warm/cold grounding comparison on the combined shape against real main — the closing check for #443.

NivDvir:
Re-ran the comparison on the current head (c980c27 = main with #448 + this PR).
Bottom line: the disk cache works on the shipping shape — all 10 grounding rows parse after a save/restore round trip, and the restore output is byte-identical to my 07-29 run.

One heads-up, not a bug here: since #472, an in-memory session and a restored one diverge on any later image-bearing turn (the in-memory side re-templates the full transcript; a restored session keeps fragment continuation). Might be worth a line in the snapshot docs.

Details or harness available if useful.

kklimuk:
@NivDvir got the comment in!

NivDvir:
Thanks for adding that — the fragment-continuation note is exactly the right place for it.

One narrowing worth considering for the wording: "positions stay correct; the prompt differs"
holds for Qwen2.5-VL, which builds a real per-frame vision attention mask
(`Qwen25VL.swift:641-656`). Qwen2-VL has no per-image mask (`Qwen2VL.swift:594-601` passes
`mask: .none`, and `:575-577` gives `B == 1` for images), so every patch attends to every
other — meaning a restored session that encodes the new image *alone* gets different vision
features than one encoding it alongside the cached image, at matched positions. So for
Qwen2-VL the two paths differ in more than prompt content.

Tiny random-weight measurement only, and this shouldn't hold the PR — happy to share the
harness if it's useful, or leave it as a follow-up.

kklimuk:
@NivDvir i’ll fix the comment and merge main in to resolve conflicts! Anything else needed to make it mergeable?

kklimuk:
@NivDvir fixed!

NivDvir:
Nothing else — you've done both, and the narrowed note reads exactly right. Thanks for turning those around so fast.

I measured the restore path on 07-31 at the then-current head: 10/10 grounding rows parse through the save/load round trip, byte-identical to the 07-29 run. Since then the branch has taken the doc narrowing and main's ChatConventions migration (#502) — neither touches `ChatSession`, `KVCache`, `Evaluate` or the Qwen2 model files, so that measurement still describes this head. CI is green on `693de51d8`. Approving, since the repo asks for at least one review and this is ready for one.

NivDvir (approved):
Measured firsthand on 07-31 — save/restore round trip byte-identical, 10/10 grounding rows parse after restore. Changes since land outside the restore path (`ChatSession`, `KVCache`, `Evaluate` and the Qwen2 model files are untouched). CI green on `693de51d8`. No blockers from me.

kklimuk:
@NivDvir hmm, looks like merging is still sadly blocked:
Screenshot 2026-08-03 at 3 04 03 PM

NivDvir:
Turns out mine doesn't tick that box — reviewDecision still shows REVIEW_REQUIRED, so the required review has to come from someone with write access. @davidkoski, the restore-path validation is already done and on the record above, so hopefully this is a quick pass whenever you get to it.

kklimuk:
@davidkoski would love your help!
been merging latest about once per day to make sure this PR is free of conflicts and its getting to be a lot of work. =)

NivDvir:
Correction to my note above. The head has moved since I approved (`693de51d8` → `9bf9ec576`), and the upstream merge brings `26f3a48` ("Add custom LogitProcessor injection to GenerateParameters / ChatSession"), which touches `ChatSession.swift` and `Evaluate.swift` — two of the four files I said were untouched. That argument doesn't hold as written.

I still expect the 07-31 restore measurement to describe this head: the change is additive plumbing at the sampling layer (`components` on `ChatSession`, `ChainedLogitProcessor` in `Evaluate`), while the save/load round trip I measured exercises cache and position handling. Happy to re-run it on `9bf9ec576` if you'd rather have that on the record before review.

davidkoski:
> @davidkoski would love your help! been merging latest about once per day to make sure this PR is free of conflicts and its getting to be a lot of work. =)

Yeah, sorry about that -- a lot of work is landing in the KVCache, Evaluate and ChatSession areas.  I will review this one in the morning and that will make it easier to rebase and land.

Libraries/MLXVLM/Models/GlmOcr.swift
```diff
@@ -1016,24 +1031,45 @@ public class GlmOcr: Module, VLMModel, KVCacheDimensionProvider {
...
-        let dtype = visionModel.patchEmbed.proj.weight.dtype
+        let inputIds = input.text.tokens
+        let cacheOffset = cache.first?.offset ?? 0
 
-        var allPixels: MLXArray?
-        var allFrames: [THW] = []
+        guard cacheOffset == 0 || state?[ropeDeltasKey] != nil else {
+            throw ContinuationStateError.missingState(
+                model: "GlmOcr", key: ropeDeltasKey.id)
+        }
```
davidkoski: The other two models have code like this:

```swift
    public func prepare(
        _ input: LMInput,
        cache: [any KVCache],
        state: LMOutput.State?,
        windowSize: Int?
    ) throws -> PrepareResult {
        let inputIds = input.text.tokens
        let inputIds2D = inputIds.ndim == 1 ? inputIds[.newAxis, 0...] : inputIds
        let cacheOffset = faCacheOffset(cache)
        guard cacheOffset == 0 || inputIds2D.dim(0) == 1 else {
            throw ContinuationStateError.unsupportedBatchContinuation(model: "Qwen35")
        }
```

Is this needed here? or does the missingState cover it?

Libraries/MLXLMCommon/ChatSession.swift
```diff
@@ -572,6 +592,85 @@ public final class ChatSession {
...
+    /// > re-rendering earlier messages. Positions stay correct, but the prompt differs — and on
+    /// > vision encoders that attend across image boundaries (Qwen2-VL today) so do the new
+    /// > image's features, since it is encoded alone rather than beside the cached one.
+    /// > Use a history initializer when later turns must re-render the full conversation.
+    public convenience init(
+        _ model: ModelContainer,
+        instructions: String? = nil,
+        promptCache: consuming PromptCacheSnapshot,
+        speculativeDecoding: SpeculativeDecodingConfig? = nil,
+        generateParameters: GenerateParameters = .init(),
```
davidkoski, line 625: There is a new parameter for ChatSession that should be forwarded here and in the other init: `components: GenerationComponents`

Libraries/MLXLMCommon/ChatSession.swift
```diff
@@ -985,12 +1084,13 @@ public final class ChatSession {
...
-                            // Seed the iterator with the carried state; read
-                            // back the post-prefill state (prefill runs in the
-                            // iterator's init, and the rope delta does not
-                            // change during decode) so the next turn — or the
-                            // next tool restart — anchors correctly.
+                        /// Every caller below reaches this because speculation was refused, so the
+                        /// draft cache is dropped: retaining it would leave it at an offset the
+                        /// main cache never visits.
+                        func defaultGenerationWithoutDraft() throws -> GenerationRun {
+                            draftKVCache = nil
```
davidkoski: I think this blocks the potential restart as the code that would set reusedMainCacheWithoutDraft is gated on `draftKVCache != nil`

Libraries/MLXLMCommon/ChatSession.swift
```diff
@@ -1007,7 +1107,17 @@ public final class ChatSession {
...
+                        // Carried state is safe to speculate on: the anchors position from the
+                        // cache offset, which a rejected proposal rewinds along with the KV rows.
+                        // Media is not — the draft model would have to prefill the same images.
+                        // `input` may have been narrowed to a token-only suffix above, so test
+                        // the prepared input for media.
+                        if let speculativeDecoding,
+                            preparedInput.image == nil,
+                            preparedInput.video == nil,
+                            preparedInput.audio == nil,
+                            draftKVCache != nil || kvCache.processedTokenCount == 0
```
davidkoski: The line that blocks the reset of the drafting

kklimuk:
@davidkoski i think i've done it!

davidkoski (approved):
Changes look good, thank you!

kklimuk:
@davidkoski @NivDvir thanks for all your help getting over the line

davidkoski:
Thank you for the contribution!  The community makes this work!

PR #477, opened 2026-07-27
Integration tests CI: run on Xcode 26 only

CharlieTLe:
## Proposed changes

Support only Xcode 26 and run just the tests that compile there. The MLXFoundationModels adapter and its integration tests are gated on canImport(FoundationModels, _version: 2) (macOS 27 SDK), so on Xcode 26 they compile out automatically and the target builds only the SDK-agnostic suites (MTP, Qwen3VL/Qwen3.5 vision, Coherence, Gemma4, tool calls, masks) — no test filtering needed.

Replace the prefer-Xcode-27 selection with an explicit Xcode 26 pick that fails loudly if Xcode 26 is absent, so the job never lands on a 27 beta whose moving FoundationModels SDK breaks the build.
## Checklist

- [ ] I have read the [CONTRIBUTING](https://github.com/ml-explore/mlx-swift-lm/blob/main/CONTRIBUTING.md) document

### DIFF
--- a/.github/workflows/integration_tests.yml
+++ b/.github/workflows/integration_tests.yml
@@ -18,28 +18,27 @@ jobs:
         with:
           submodules: recursive
 
-      - name: Select Xcode
+      - name: Select Xcode 26
         shell: bash
         run: |
-          # The FoundationModels integration tests only build against the macOS 27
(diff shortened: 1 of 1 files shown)


davidkoski (approved):
Looks good, thank you!

PR #479, opened 2026-07-28
Fix Gemma 3n Boolean attention masks

codesworth:
## Proposed changes

Fixes a fatal error when running Gemma 3n with a prompt longer than the model's sliding window. There is no corresponding issue filed.

`createAttentionMask` returns Boolean causal and sliding-window masks. `Gemma3nDecoderLayer` re-applied the sliding window to that mask and passed its Boolean dtype to `MLXArray.maskFill`, which traps during prompt prefill once the prompt exceeds the sliding window (512 tokens):

    Fatal error: maskFill requires a floating-point dtype, got bool

The decoder-layer rewrite was also redundant: the sliding window is already encoded in the mask the language model builds via `createAttentionMask(h:cache:windowSize:)`. Separately, the attention layer converted Boolean masks to the query's floating-point dtype, turning an allowed/blocked mask into an additive 1/0 mask that blocks nothing.

Changes:

- Remove the decoder-layer sliding-mask rewrite and its `caches:`/`cachePosition:` plumbing. The decoder layer now passes the already-complete mask straight to attention, and its signature matches Python mlx-lm's `(x, mask, cache, per_layer_input)`.
- Stop converting Boolean masks to the query dtype in `Gemma3nAttention`. Boolean masks now reach `MLXFast.scaledDotProductAttention` unchanged, as in Python mlx-lm and the other models in this repo.
- Keep the key-length mask slice as a small dtype-preserving guard (`gemma3nAdjustedAttentionMask`) for caller-supplied masks; masks created internally already match the key length. Python mlx-lm has no equivalent because its language model takes no caller-supplied mask.
- Add `Gemma3nMaskTests`: a regression test that traps on `main` and passes with this change, an equivalence check between the Boolean mask and the corresponding additive float mask, and unit coverage of the dtype-preserving slice (including positions only the sliding window blocks).

## Checklist


davidkoski (approved):
Changes look good, thank you!

PR #480, opened 2026-07-28
Integration tests: scope down heavy suites, fix flakes and memory pressure

CharlieTLe:
## Proposed changes

The Xcode-26 job compiled and ran but was cancelled at the 2h cap and showed instability. Root causes and fixes:

Timeout: MTPIteratorEndToEndDiagnosticTests auto-downloads a ~35GB 31B target+drafter pair with no gate; a cold run spent ~83 min on one test. Gate the ten 31B-dependent tests on both snapshots already being cached (has31BPair, mirroring MTPAcceptanceRateTests / MTPQuantizationOnsetTests). The E4B/E2B tests keep their own env gates.

Scope: split CoherenceIntegrationTests into a fast default subset (llama3.2-1B, qwen3-1.7B, gemma3-1B-qat, phi3.5, granite3.3-2B) and an opt-in CoherenceFullIntegrationTests suite (MLX_RUN_FULL_COHERENCE=1) for the rest.

Memory/crash: loading ~20 multi-GB models in one serialized run accumulated resident weights until the process was jetsammed (Metal compiler XPC crashes, e.g. the gemma3n_E2B hang). Route every coherence test through a helper that evicts the model (IntegrationTestModels.evictLLM) and calls GPU.clearCache() after each, bounding residency to one model.

Failures: disable olmoe_1B_7B (GPTNeoXTokenizer unsupported by the loader); add transient-network retry to IntegrationTestModels.llmContainer for the granite -1005 'network connection lost' flake.

Validated: build-for-testing succeeds on Xcode 27; the default coherence subset runs green (5/5) locally with the per-model release path.
## Checklist

### DIFF
--- a/IntegrationTesting/IntegrationTestingTests/CoherenceIntegrationTests.swift
+++ b/IntegrationTesting/IntegrationTestingTests/CoherenceIntegrationTests.swift
@@ -3,6 +3,7 @@
 import Foundation
 import HuggingFace
 import IntegrationTestHelpers
+import MLX
 import MLXHuggingFace
 import MLXLLM
 import MLXLMCommon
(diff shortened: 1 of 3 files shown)


davidkoski (approved):
These improvements look good!

PR #481, opened 2026-07-29
 Fix lfm2 toolcall parser and glm4 gate GLM-4-0414

CharlieTLe:
## Proposed changes

The integration ToolCallIntegrationTests surfaced two failures once the suite ran in CI. Investigating the raw model output showed both were masked by non-deterministic sampling, so first pin generation to temperature 0 (greedy), matching the coherence/MTP tests, which made the failures reproducible.

LFM2 (real fix): the model emits its arguments wrapped in a schema-style object, e.g. get_weather(properties={"location": "Tokyo", "unit": "celsius"}). PythonicToolCallParser mis-parsed this two ways: the key=value regex truncated the object value at the first comma ("{\"location\": \"Tokyo\""), and it never recovered the real location/unit args. Rework parseArguments to split arguments bracket/brace/quote-aware (no comma truncation), parse object/array values as JSON, and unwrap a single recognized wrapper key (properties/parameters/arguments/args/ kwargs) that is not itself a declared parameter. Quoted scalars still go through schema-based type conversion. Adds unit tests for the wrapper-unwrap and for a preserved (non-wrapper) object-valued argument.

GLM-4-9B-0414: emits a markerless funcname\n{json-args} tool call — its chat template only lists tools and defines no <tool_call>/<arg_key> delimiters, so the .glm4 parser (GLM-4.5/4.7 <arg_key> format) never detects it. Supporting this shape needs a dedicated markerless parser plus schema-anchored streaming detection; disable glm4EndToEnd with a documented reason and leave it as a follow-up.

Validated: all 62 MLXLMTests/ToolTests pass; lfm2EndToEnd passes end-to-end against the real model; mistral3/nemotron/qwen35 tool-call tests unaffected.

## Checklist

### DIFF
--- a/IntegrationTesting/IntegrationTestingTests/ToolCallIntegrationTests.swift
+++ b/IntegrationTesting/IntegrationTestingTests/ToolCallIntegrationTests.swift
@@ -35,7 +35,17 @@ struct ToolCallIntegrationTests {
         try await ToolCallTests.glm4FormatAutoDetection(container: container)
     }
 
-    @Test func glm4EndToEnd() async throws {
+    // GLM-4-9B-0414 emits a markerless `get_weather\n{json-args}` tool call — its
+    // chat template only lists the available tools and defines no <tool_call> /
+    // <arg_key> delimiters. The .glm4 parser targets the GLM-4.5/4.7 <arg_key>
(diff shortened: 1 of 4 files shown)


davidkoski (approved):
Great finds!  Thank you!

PR #482, opened 2026-07-29
Let models declare chat conventions via ChatConventionsProviding

CharlieTLe:
## Proposed changes

Introduce an opt-in ChatConventionsProviding protocol so a model declares its own tool-call format and reasoning config next to its definition, rather than via the centralized model_type hasPrefix chains in ToolCallFormat.infer / ReasoningConfig.infer. The model factories ask the model first (through the region-safe LanguageModel.declared* accessors) and fall back to the existing inference for models not yet migrated.

Migrate Nanbeige as the first slice: it now declares .xmlFunction and its <think>/enable_thinking reasoning config, and the nanbeige branches are removed from both infer methods. Tests relocated onto the model conformance.

Addresses davidkoski's review note on PR #460.

## Checklist

### DIFF
--- a/Libraries/MLXLMCommon/ChatConventions.swift
+++ b/Libraries/MLXLMCommon/ChatConventions.swift
@@ -1,0 +1,49 @@
+// Copyright © 2025 Apple Inc.
+
+import Foundation
+
+// MARK: - ChatConventionsProviding
+
+/// A model's chat conventions: how it encodes tool calls and how it reasons.
+///
+/// Adopted by the model so this knowledge lives with the model definition rather
+/// than in centralized `model_type` string tables (``ToolCallFormat/infer(from:configData:)``
+/// and ``ReasoningConfig/infer(from:modelId:configData:)``). The model factories
+/// probe this protocol before falling back to those inference chains.
+///
+/// Opt-in: the protocol extension defaults both properties to `nil`, so a
+/// conforming type overrides only the property that applies to it. This mirrors
+/// the ``ModelConfigurationValidating`` opt-in pattern.
+public protocol ChatConventionsProviding {
+    /// The tool-call format this model emits, or `nil` for the JSON default.
+    var toolCallFormat: ToolCallFormat? { get }
+
+    /// The model's reasoning protocol, or `nil` for non-reasoning models.
+    var reasoningConfig: ReasoningConfig? { get }
+}
+
+extension ChatConventionsProviding {
+    public var toolCallFormat: ToolCallFormat? { nil }
+    public var reasoningConfig: ReasoningConfig? { nil }
+}
+
+// MARK: - LanguageModel probe
+
+extension LanguageModel {
+    /// The tool-call format this model declares via ``ChatConventionsProviding``,
+    /// or `nil` if it does not conform / declares no format.
+    ///
+    /// Exposed as a method on the model so factories read it by borrowing `self`
+    /// and receiving a `Sendable` result, rather than downcasting the model
+    /// inline (which would taint the model's region and block returning it as a
+    /// `sending` result).
+    public var declaredToolCallFormat: ToolCallFormat? {
+        (self as? ChatConventionsProviding)?.toolCallFormat
+    }
+
+    /// The reasoning config this model declares via ``ChatConventionsProviding``,
+    /// or `nil` if it does not conform / declares none.
+    public var declaredReasoningConfig: ReasoningConfig? {
+        (self as? ChatConventionsProviding)?.reasoningConfig
+    }
+}
--- a/Libraries/MLXLMCommon/ChatConventions.swift
+++ b/Libraries/MLXLMCommon/ChatConventions.swift
@@ -6,14 +6,13 @@ import Foundation
 
 /// A model's chat conventions: how it encodes tool calls and how it reasons.
 ///
-/// Adopted by the model so this knowledge lives with the model definition rather
-/// than in centralized `model_type` string tables (``ToolCallFormat/infer(from:configData:)``
-/// and ``ReasoningConfig/infer(from:modelId:configData:)``). The model factories
-/// probe this protocol before falling back to those inference chains.
-///
-/// Opt-in: the protocol extension defaults both properties to `nil`, so a
-/// conforming type overrides only the property that applies to it. This mirrors
-/// the ``ModelConfigurationValidating`` opt-in pattern.
+/// This knowledge lives with the model definition rather than in centralized
+/// `model_type` string tables (``ToolCallFormat/infer(from:configData:)`` and
+/// ``ReasoningConfig/infer(from:modelId:configData:)``). ``LanguageModel``
+/// conforms with `nil` defaults, so the model factories read `model.toolCallFormat`
+/// / `model.reasoningConfig` directly and fall back to the inference chains when a
+/// model declares nothing. A model opts in by overriding either property in an
+/// extension.
 public protocol ChatConventionsProviding {
     /// The tool-call format this model emits, or `nil` for the JSON default.
     var toolCallFormat: ToolCallFormat? { get }
@@ -25,25 +24,4 @@ public protocol ChatConventionsProviding {
 extension ChatConventionsProviding {
     public var toolCallFormat: ToolCallFormat? { nil }
     public var reasoningConfig: ReasoningConfig? { nil }
-}
-
-// MARK: - LanguageModel probe
-
-extension LanguageModel {
-    /// The tool-call format this model declares via ``ChatConventionsProviding``,
-    /// or `nil` if it does not conform / declares no format.
-    ///
-    /// Exposed as a method on the model so factories read it by borrowing `self`
-    /// and receiving a `Sendable` result, rather than downcasting the model
-    /// inline (which would taint the model's region and block returning it as a
-    /// `sending` result).
-    public var declaredToolCallFormat: ToolCallFormat? {
-        (self as? ChatConventionsProviding)?.toolCallFormat
-    }
-
-    /// The reasoning config this model declares via ``ChatConventionsProviding``,
-    /// or `nil` if it does not conform / declares none.
-    public var declaredReasoningConfig: ReasoningConfig? {
-        (self as? ChatConventionsProviding)?.reasoningConfig
-    }
 }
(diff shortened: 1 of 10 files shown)


Libraries/MLXLMCommon/ChatConventions.swift
```diff
@@ -0,0 +1,49 @@
...
+    public var declaredToolCallFormat: ToolCallFormat? {
+        (self as? ChatConventionsProviding)?.toolCallFormat
+    }
+
+    /// The reasoning config this model declares via ``ChatConventionsProviding``,
+    /// or `nil` if it does not conform / declares none.
+    public var declaredReasoningConfig: ReasoningConfig? {
+        (self as? ChatConventionsProviding)?.reasoningConfig
+    }
+}
```
davidkoski: Rather than add another layer I would probably do this in the calling code (the load methods).  Either `LanguageModel` can conform to `ChatConventionsProviding` (which seems fine since there are default implementations) or for models that want to conform:

```swift
if let provider = model as? ChatConventionsProviding {
    ...
}
```

Just making LanguageModel conform seems the simplest.

Libraries/MLXLMCommon/ChatConventions.swift
```diff
@@ -0,0 +1,49 @@
...
+/// than in centralized `model_type` string tables (``ToolCallFormat/infer(from:configData:)``
+/// and ``ReasoningConfig/infer(from:modelId:configData:)``). The model factories
+/// probe this protocol before falling back to those inference chains.
+///
+/// Opt-in: the protocol extension defaults both properties to `nil`, so a
+/// conforming type overrides only the property that applies to it. This mirrors
+/// the ``ModelConfigurationValidating`` opt-in pattern.
+public protocol ChatConventionsProviding {
+    /// The tool-call format this model emits, or `nil` for the JSON default.
+    var toolCallFormat: ToolCallFormat? { get }
```
davidkoski, line 18: The current method takes the configuration data as a parameter:

```swift
    public static func infer(from modelType: String, configData: Data? = nil) -> ToolCallFormat? {
...
        // Llama family (need secondary signal for Llama 3 vs 1/2)
        if type == "llama" {
            guard let data = configData,
                let json = try? JSONSerialization.jsonObject(with: data) as? [String: Any]
            else { return nil }

            // Secondary signal 1: vocab_size >= 128000 (Llama 3 uses 128256, Llama 2 uses 32000)
            if let vocabSize = json["vocab_size"] as? Int, vocabSize >= 128000 {
                return .llama3
            }

            // Secondary signal 2: rope_scaling with rope_type == "llama3"
            if let ropeScaling = json["rope_scaling"] as? [String: Any],
                let ropeType = ropeScaling["rope_type"] as? String,
                ropeType == "llama3"
            {
                return .llama3
            }

            return nil
        }
```

I don't think we need it here because the model has already been initialized with the configuration -- any values it wants it could just keep from that (if it doesn't retain the whole thing).

Nothing to do, just musing.

Libraries/MLXVLM/VLMModelFactory.swift
```diff
@@ -381,9 +381,11 @@ public final class VLMModelFactory: GenericModelFactory {
         mutableConfiguration.eosTokenIds = eosTokenIds
         mutableConfiguration.stopStrings.formUnion(generationConfig?.stopStrings ?? [])
 
-        // Auto-detect tool call format from model type if not explicitly set
+        // Auto-detect tool call format: ask the model, then fall back to model type
         if mutableConfiguration.toolCallFormat == nil {
-            mutableConfiguration.toolCallFormat = ToolCallFormat.infer(from: baseConfig.modelType)
+            mutableConfiguration.toolCallFormat =
+                model.declaredToolCallFormat
+                ?? ToolCallFormat.infer(from: baseConfig.modelType)
```
davidkoski, line 388: It doesn't need to be done here, but perhaps a followup -- all the models should have this logic migrated and we can get rid of the `infer` method.  Or can we?  This does allow callers without a model instance to get this data.  Do they need it?  It looks like the only uses in that form are in the tests.

Tests/MLXLMTests/ToolTests.swift
```diff
@@ -1065,10 +1072,6 @@ struct ToolTests {
         #expect(ToolCallFormat.infer(from: "qwen3_next_moe") == .xmlFunction)
         #expect(ToolCallFormat.infer(from: "QWEN3_NEXT") == .xmlFunction)
 
-        // Nanbeige models (prefix matching)
-        #expect(ToolCallFormat.infer(from: "nanbeige") == .xmlFunction)
-        #expect(ToolCallFormat.infer(from: "NANBEIGE") == .xmlFunction)
```
davidkoski, line 1116: I wonder if these should all be replaced by something like the llama call has below:

```swift
        let llama3RopeConfig = """
            {
                "model_type": "llama",
                "rope_scaling": {
                    "rope_type": "llama3"
                }
            }
            """.data(using: .utf8)!
        #expect(ToolCallFormat.infer(from: "llama", configData: llama3RopeConfig) == .llama3)
```

A call like this:

```swift
        let model: LanguageModel
        do {
            model = try await typeRegistry.createModel(
                configuration: configData, modelType: baseConfig.modelType)
        } catch let error as DecodingError {
            throw ModelFactoryError.configurationDecodingError(
                configurationURL.lastPathComponent, configuration.name, error)
        }
```

that then checked the tool / reasoning type?

Libraries/MLXLMCommon/ChatConventions.swift
davidkoski: But having them explicitly adopt it is maybe the cleanest -- that allows writing an extension.  In that case maybe we don't want default implementations of the properties.  If you declare it you _have_ to provide it.

CharlieTLe: LanguageModel now inherits ChatConventionsProviding (nil defaults), so the factories read model.toolCallFormat / model.reasoningConfig directly and the declared* accessors are gone.

Libraries/MLXVLM/VLMModelFactory.swift
CharlieTLe: Thanks for the follow-up! I'd love to get there eventually. 

The main catch with completely removing `infer` right now is that we have two cases we can't resolve from a single model instance:

1. **Llama 1, 2, vs. 3:** We need the vocabulary size or rope-scaling to tell these apart. A `LlamaModel` could check its own config for this, so we can definitely migrate it.
2. **DeepSeek-R1-Distill:** This reports a base model type (like `qwen2` or `llama`) and we can only really tell it apart by the repo ID. That feels like it belongs in a registry override rather than on the model itself.

So the game plan would be to migrate the pure `model_type` models over to the protocol first, move the Llama logic into `LlamaModel`, route R1-Distill through a registry override, and then we can finally drop `infer`.

Also, regarding callers without a model instance, the static form is only used in tests right now, so dropping it is really just a test-migration task rather than an API breaking change.

Tests/MLXLMTests/ToolTests.swift
CharlieTLe: This makes sense and aligns well with making the model the single source of truth. I’d prefer to handle this in a focused follow-up PR instead of expanding this one. We can swap the `infer(from: "...")` string assertions for `typeRegistry.createModel(...)` and assert `model.toolCallFormat` and `model.reasoningConfig`. This also gives us a clear path to remove the static `infer` API once every model is migrated. I’ve already kept the Nanbeige coverage on the model instance (`NanbeigeTests`), so this PR at least demonstrates the target shape.

Libraries/MLXVLM/VLMModelFactory.swift
davidkoski: The DeepSeek-R1-Distill is interesting.  Right now we have a cross-cutting factory for the ReasoningConfig.  The two drawbacks there are:

- it can't be extended outside of mlx-swift-lm
- it is not encapsulated

As you note, we could do it via the registry -- ModelConfiguration carries an overriding toolCallFormat and reasoningConfig.  This only works if we have a registry entry for it or the caller constructs the ModelConfiguration, but it does fit in with the current design.

When we convert the deepseek models we would want to register them (some of them anyway) in the appropriate AbstractModelRegistry.

Another approach might be to have a registry of `ChatConventionsProviding` (along the lines of the AbstractModelRegistry) where callers outside mlx-swift-lm could register additional providers.  I think we can decide on this specific case later, but it does suggest that maybe the properties we have here _should be_ functions that take `modelId` to allow for this.

Libraries/MLXVLM/VLMModelFactory.swift
```diff
@@ -381,9 +381,11 @@ public final class VLMModelFactory: GenericModelFactory {
...
         mutableConfiguration.stopStrings.formUnion(generationConfig?.stopStrings ?? [])
 
-        // Auto-detect tool call format from model type if not explicitly set
+        // Auto-detect tool call format: ask the model, then fall back to model type
         if mutableConfiguration.toolCallFormat == nil {
-            mutableConfiguration.toolCallFormat = ToolCallFormat.infer(from: baseConfig.modelType)
+            mutableConfiguration.toolCallFormat =
+                model.toolCallFormat
+                ?? ToolCallFormat.infer(from: baseConfig.modelType)
         }
```
davidkoski, line 389: This is a pre-existing gap, but we should probably close it here:  the VLMModelFactory doesn't call the reasoning config and thus there is no call to the `ChatConventionsProviding` for that here.

- [x] add call to `reasoningConfig` here

Libraries/MLXVLM/VLMModelFactory.swift
davidkoski: Thinking about the function vs property more, I think we should leave it as a property.  That is correct for the model to provide it.

We can add a cross cutting protocol (interceptors) that consumes modelId (maybe among other things) later.

Libraries/MLXVLM/VLMModelFactory.swift
CharlieTLe: VLMModelFactory now handles reasoningConfig just like the LLM factory. It first checks model.reasoningConfig, then falls back to ReasoningConfig.infer(from:modelId:configData:). It also properly threads this into the built ModelConfiguration -- fixing both the resolution and the plumbing, since it was previously being dropped.

davidkoski (approved):
Changes look great, thank you!

PR #321, opened 2026-05-29
Make MLXLLM compilable on Linux

gmondada:
## Proposed changes

Make this package compilable on Linux.

Not all modules (targets) can be compiled on Linux. MLXVLM, for instance, has no Linux implementation yet.

Please note that, from a multi-platform perspective, `UserInput.Image` and `UserInput.Video` are very good. However, `UserInput.VideoFrame` is not suitable for a multi-platform environment. Therefore, I'm proposing to deprecate the `UserInput.VideoFrame.frame` property and replace it by `UserInput.VideoFrame.image`.

This will allow, in the future, to create a MediaProcessor which uses UserInput.Image instead of CIImage, making abstraction of the native API being used.
  
## Checklist

- [x] I have read the [CONTRIBUTING](https://github.com/ml-explore/mlx/blob/main/CONTRIBUTING.md) document
- [x] I have run `pre-commit run --all-files` to format my code / installed pre-commit prior to committing changes
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] I have updated the necessary documentation (if needed)

PR #483, opened 2026-07-30
Fix remaining Linux build breaks in MLXLMCommon and MLXGuidedGeneration

GoodOlClint:
## Proposed changes

Four small gaps that stop the package building on Linux. All are follow-ups to #321 ("Make MLXLLM compilable on Linux"), which established the `Libraries/MLXLMCommon/Linux/` shim directory and the `canImport` guard style this PR follows. Found while bringing the package up on a Linux/CUDA host (Swift 6.3.3, aarch64, Ubuntu 24.04); none of them involve the CUDA backend itself.

| | Break | Fix |
|---|---|---|
| 1 | `autoreleasepool` is unavailable without the Objective-C runtime. Six call sites: `MLXLMCommon/Evaluate.swift` (×4), `MLXLMCommon/MTPSpeculativeTokenIterator.swift`, `MLXLLM/LLMModel.swift` | New `Libraries/MLXLMCommon/Linux/Autoreleasepool.swift` behind `#if !canImport(ObjectiveC)` |
| 2 | `MLXGuidedGeneration/GuidedGenerationLoop.swift` imports `os` unguarded. The existing `Logger` shim is `internal` to MLXLMCommon, so sibling modules can't reach it | Per-module `MLXGuidedGeneration/Linux/Logger.swift`, matching how the shim is already scoped; drop the unguarded import |
| 3 | The Linux `Logger` shim declared only `info` and `error` | Both copies now declare the full `os.Logger` level set (`debug`, `notice`, `warning`, `fault`) |
| 4 | `MLXCXGrammar` links `-lc++` unconditionally; Linux links libstdc++ via the Swift toolchain and has no libc++ | `.when(platforms: [.macOS, .iOS, .visionOS, .tvOS])`, as the adjacent `-w` `unsafeFlags` line already is |

Two notes on scope, so the diff reads honestly:

- The `public` on the `autoreleasepool` shim is load-bearing: the sixth call site is in **MLXLLM**, a separate module. Its signature matches Foundation's, so both `autoreleasepool { }` and `autoreleasepool(invoking:)` keep resolving.
- On (3), MLXLMCommon itself only logs at `.info` today, so completing that copy's levels is prophylactic. The level that is needed *now* is `warning`, for the new MLXGuidedGeneration copy — `GuidedGenerationLoop.swift` logs at `.warning`.

`MLXFoundationModels`' `import os` is deliberately untouched: it sits behind `#if FoundationModelsIntegration` and compiles out on Linux, so it was never a live break.

**Out of scope:** MLXVLM is still un-ported on Linux (unguarded `AVFoundation`/`CoreImage`/`CoreMedia` across ~17 files). #87 is the existing vehicle for that and this PR does not touch any file it changes.

### Verification

Apple platforms are unaffected by construction — both shims sit behind `canImport` checks that are false on Apple platforms, and the linker setting still applies there. Confirmed on macOS (Xcode 26.6, Swift 6.3.3):

- `pre-commit run --all-files` — clean.
- `xcodebuild test -scheme mlx-swift-lm-Package -destination 'platform=macOS' -skipPackagePluginValidation` — 312 tests, 31 suites. The only failures are 3–4 tolerance issues in `Gemma4ChunkedPrefillTests/"Chunked and single-pass prefill agree on logits and cache offsets"`, which **reproduce on unmodified `main` (b207f60)** and vary in count between runs (4 issues on `main`, 3 on this branch). Pre-existing and flaky, not from this change.
- `scripts/verify-docs.sh` — all documentation builds passed.

On Linux, these four breaks no longer occur. The wider Linux/CUDA bring-up they were found in reached a clean build (466.89s) and decoded Llama-3.2-1B-Instruct-4bit at 196.4 tok/s and gemma-4-12B-it-4bit at 23.0 tok/s on an sm_121 device (CUDA 13.0.88, Swift 6.3.3) — with MLXVLM excluded, per the scope note above.

## Checklist

- [ ] I have added tests that prove my fix is effective or that my feature works — **no test added:** "does this platform compile" is a build-configuration property, not something a unit test can assert, and CI has no Linux *build* job (the `lint` job is a style-only container). Happy to add one if you'd like a Linux build lane in CI as part of this.
- [x] I have updated the necessary documentation (if needed) — no user-facing API change; `scripts/verify-docs.sh` passes.


davidkoski (commented):
Changes look good, thank you!

davidkoski (approved):

PR #486, opened 2026-07-31
Centralize tool-call authorization and output routing

aleroot:
## Proposed changes

Tool-call syntax is now parsed independently from authorization, undeclared calls are consumed at the processor boundary, and generation uses ordered outputs so malformed or unfinished tool protocol is not emitted as assistant text or fed back into conversation history.

Rejects undeclared tool function names as `example_function_name` that the model sometimes may try to call by mistake.

I kept this PR focused on centralising declared-tool filtering and ordered output routing, but I can add an observable rejection outcome here or track it as a follow-up, depending on the preferred API direction.

*I am also aware that the stronger long-term approach is constrained generation, using XGrammar to restrict tool names and arguments while decoding. Even with constrained decoding, however, the processor would still need defensive handling for truncation, malformed output, unsupported models, and legacy string-based formats. I therefore see this PR as a necessary compatibility and safety layer, not as a replacement for constrained decoding.*

## Checklist

### DIFF
--- a/Libraries/MLXFoundationModels/ToolCalling/AllowedToolOutputRouter.swift
+++ b/Libraries/MLXFoundationModels/ToolCalling/AllowedToolOutputRouter.swift
@@ -14,7 +14,6 @@ struct AllowedToolOutputRouter {
 
     private var reasoningEmitter: ReasoningEventEmitter?
     private let toolProcessor: ToolCallProcessor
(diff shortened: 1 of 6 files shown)


aleroot:
Superseded by https://github.com/ml-explore/mlx-swift-lm/pull/512

PR #487, opened 2026-07-31
Fix Qwen3.5 hybrid cache offset accounting

ngutech21:
## Proposed changes

Qwen3.5/Qwen3.6 hybrid models use both recurrent and KV caches. In the compiled prefill and decode paths, the recurrent cache state was advanced manually, but its token offset was not updated. This caused the recurrent and KV cache offsets to drift apart, breaking cache validation and potentially causing failures when continuing generation from a warmed cache.
The fix increments the recurrent cache offset by the number of processed tokens whenever its state is advanced, covering prefill, compiled decode, fallback decode, and the corresponding vision-language path.
This keeps all cache layers synchronized, makes cached continuation reliable, and adds regression tests for text and vision-language models to prevent future offset-accounting regressions.
## Checklist

### DIFF
--- a/Libraries/MLXLLM/Models/Qwen35.swift
+++ b/Libraries/MLXLLM/Models/Qwen35.swift
@@ -341,6 +341,7 @@ final class Qwen35GatedDeltaNet: Module {
             cache[0] = newConvState
             cache[1] = newRecState
             cache.advance(inputs.dim(1))
+            cache.offset += inputs.dim(1)
         }
         return out
     }
@@ -744,6 +745,7 @@ final class Qwen35DecoderLayer: Module {
         cache[0] = out[1]
         cache[1] = out[2]
         cache.advance(1)
+        cache.offset += 1
         return out[0]
     }
 
@@ -1020,6 +1022,7 @@ public class Qwen35TextModelInner: Module {
                 mambaCache[0] = outputs[1 + 2 * i]
                 mambaCache[1] = outputs[2 + 2 * i]
                 mambaCache.advance(1)
+                mambaCache.offset += 1
             }
 
             pendingAttention = []
--- a/Libraries/MLXVLM/Models/Qwen35.swift
+++ b/Libraries/MLXVLM/Models/Qwen35.swift
@@ -564,6 +564,7 @@ enum Qwen35Language {
             if let cache {
                 cache[1] = state
                 cache.advance(S)
+                cache.offset += S
             }
 
             out = norm(out, gate: z)
--- a/Tests/MLXLMTests/Qwen35CompiledDecodeLifecycleTests.swift
+++ b/Tests/MLXLMTests/Qwen35CompiledDecodeLifecycleTests.swift
@@ -46,6 +46,18 @@ final class Qwen35CompiledDecodeLifecycleTests: XCTestCase {
             Qwen35TextConfiguration.self, from: Data(json.utf8))
     }
 
+    private func assertCacheOffsets(
+        _ cache: [KVCache], equal expected: Int,
+        file: StaticString = #filePath, line: UInt = #line
+    ) {
+        XCTAssertTrue(
+            cache.allSatisfy { $0.offset == expected },
+            "expected every cache offset to equal \(expected), got \(cache.map(\.offset))",
+            file: file,
+            line: line
+        )
+    }
+
     /// Runs two decode steps (the first installs the compiled closures, the
     /// second replays them), releases the model, and asserts the
     /// closure-owning blocks deallocate.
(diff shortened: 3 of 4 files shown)


Libraries/MLXLLM/Models/Qwen35.swift
```diff
@@ -341,6 +341,7 @@ final class Qwen35GatedDeltaNet: Module {
             cache[0] = newConvState
             cache[1] = newRecState
             cache.advance(inputs.dim(1))
+            cache.offset += inputs.dim(1)
```
davidkoski, line 344: Nice find!

It looks like all the models that use MambaCache probably have this same bug: Qwen3Next, Jamba, GraniteMoeHybrid, NemotronH, LFM2, LFM2MoE, Mamba2

That suggests that maybe the answer isn't to update all the call sites, but rather encapsulate this inside the `advance()` method itself.  There is _some_ coupling between the KVCache implementations and the models, but this seems like maybe it should be internal bookkeeping.

That said, we have a unit test that guards against this explicitly:

```swift
@Test func testArraysCacheAdvanceUpdatesSequenceMetadataOnly() throws {
    let cache = ArraysCache(size: 2, leftPadding: [3, 5])
    cache.offset = 7
    cache.prepare(lengths: [4, 6])

    cache.advance(2)

    #expect(cache.offset == 7)
    #expect(cache.leftPaddingValues == [1, 3])
    #expect(cache.lengthsValues == [2, 4])
}
```

Perhaps MambaCache needs an update that encapsulates all of this code?

```swift
            cache[0] = newConvState
            cache[1] = newRecState
            cache.advance(inputs.dim(1))
            cache.offset += inputs.dim(1)
```

But perhaps the real issue is that the python KVCache doesn't work this way.  In particular the `offset` is not on the base KVCache type. 

@aleroot what do you think -- I think the ownership of the caches is spread among many but perhaps you have a well founded opinion since you wrote the test.

I am inclined to make the swift KVCache implementations more closely match the python versions -- that will make porting easier and the python implementations probably get more use (so are well tested).

davidkoski (changes requested):
Fixes a real issue but perhaps exposes a bigger fault -- see comments.  Thank you!

Libraries/MLXLLM/Models/Qwen35.swift
ngutech21: I agree that the cleanest minimal fix is to specialize `MambaCache.advance()`:

```swift
public class MambaCache: ArraysCache {
    public override func advance(_ count: Int) {
        super.advance(count)
        offset += count
    }
}
```

This preserves the existing metadata-only behavior of `ArraysCache.advance()`, while making `MambaCache` responsible for its own sequence bookkeeping. It should also fix the other affected models automatically, since they already call `advance()`.

FalconH1 already updates the Mamba cache offset manually:

```swift
cache.advance(y.dim(1))
cache.offset += y.dim(1)
```

Therefore, FalconH1 does not currently have the same missing-offset bug at this call site. However, if offset advancement is moved into `MambaCache.advance()`, its explicit offset update—and the corresponding updates in Qwen3.5—must be removed to avoid incrementing the offset twice.

I would add tests verifying that:

- `ArraysCache.advance()` does not change its offset.
- `MambaCache.advance()` updates the offset, lengths, and padding.
- Mamba2 cache offsets remain aligned across prefill and continuation.
- A serialized and restored `MambaCache` preserves its offset.

The serialization test is important because `copy()` already preserves the offset, but prompt-cache serialization currently does not. `MambaCache.metaState` may therefore need a versioned offset field; otherwise, a restored cache will report offset `0` despite containing valid recurrent state.

This proposal preserves Swift’s current cache contract by centralizing the offset update in MambaCache. It does not fully align the cache hierarchy with Python, where offset is not part of the base cache type. Matching Python would require a broader refactor of KVCache, CacheList, and ChatSession, which may be better handled separately.

aleroot: I agree this is a broader ownership issue, but I would avoid making Swift a Python-shaped copy here.
offset is primarily an attention/KV-storage and RoPE-position concept; ArraysCache.advance(_:) is recurrent batch-metadata bookkeeping. Making them mutate together creates a convenient model-wide invariant, but encodes that invariant in the wrong place by duplicating one shared timeline across every child cache.
A cleaner Swift design would make that timeline explicit at the cache-bundle/session boundary, while each child cache owns only its native state:

- the bundle owns processedTokenCount;
- attention caches own KV offset/storage and RoPE-related position state;
- recurrent caches own recurrent state and padding metadata only.

That avoids independent per-layer counters drifting out of sync. It also gives the normal reuse path a single O(1) progress check, while cache-wide scans remain useful for debugging, deserialization, or consistency validation.
As an incremental fix, a separate progress property is fine, but I would keep ArraysCache.advance(_:) unchanged and avoid adding offset += N at all Mamba call sites. If the current API must be preserved before a bundle abstraction lands, I would centralize that compatibility progress bookkeeping in MambaCache, rather than repeat it throughout every model and execution path.

For example : 

Keep `offset` as-is, but add an explicitly different semantic:

```swift
public protocol KVCache: Evaluatable {
    var offset: Int { get }                 // KV / RoPE position
    var processedTokenCount: Int { get }    // logical model timeline
    // existing requirements...
}

open class BaseKVCache: KVCache {
    public var offset: Int = 0

    // Attention caches naturally use their KV position as progress.
    open var processedTokenCount: Int { offset }
}
```

`MambaCache` retains its metadata-only `advance`, but tracks logical progress separately:

```swift
public class ArraysCache: BaseKVCache {
    open func advance(_ tokenCount: Int) {
        if let lengths {
            self.lengths = lengths - tokenCount
        }
        if let leftPadding {
            self.leftPadding = leftPadding - tokenCount
        }
    }
}

public final class MambaCache: ArraysCache {
    private var processedTokens = 0

    public override var processedTokenCount: Int {
        processedTokens
    }

    public override func advance(_ tokenCount: Int) {
        super.advance(tokenCount)  // lengths / padding only
        processedTokens += tokenCount
    }
}
```

Then cache reuse validates the correct invariant:

```swift
let expected = cachedTokenIds.count

let cacheProgress = kvCache.first?.processedTokenCount
let allCachesAreAligned =
    !kvCache.isEmpty &&
    kvCache.allSatisfy { $0.processedTokenCount == expected }

let extendsCachedPrefix =
    cacheProgress == expected &&
    allCachesAreAligned &&
    fullTokenIds.starts(with: cachedTokenIds)
```

`CacheList` should expose the common child value and fail loudly in debug if its children disagree:

```swift
public override var processedTokenCount: Int {
    let counts = children.map(\.processedTokenCount)
    assert(Set(counts).count <= 1, "Composite cache children diverged")
    return counts.first ?? 0
}
```

This is a practical transition: correct semantics, very low risk, and no repeated `offset += N` changes.

aleroot: Just to clarify the intended ownership: the `processedTokenCount` property in my previous example would be a compatibility bridge for the current `[KVCache]` API, not the final source of truth.

The shared timeline should ultimately exist once on the model-cache container:

```swift
public protocol CacheStorage: AnyObject, Evaluatable {
    func copy() -> any CacheStorage
    // state, serialization, batching, etc.
}

public final class ModelCache {
    public let entries: [any CacheStorage]
    public private(set) var processedTokenCount: Int

    public init(
        entries: [any CacheStorage],
        processedTokenCount: Int = 0
    ) {
        self.entries = entries
        self.processedTokenCount = processedTokenCount
    }

    public func commitProcessedTokens(_ count: Int) {
        precondition(count >= 0)
        processedTokenCount += count
    }
}
```

The model would update that value once at the model boundary, rather than duplicating the same mutable counter across every layer cache. `ChatSession` could then validate reuse against one authoritative value:

```swift
guard cache.processedTokenCount == cachedTokenIds.count else {
    rebuildCache()
    return
}
```

Copying, serialization, trimming and speculative rollback would preserve or update that same container-owned position.

So I see the per-cache property from my previous example as the low-risk transition for the current API. The intended direction would be a model-cache container owning the shared timeline, while attention and recurrent caches own only their native storage and metadata.

davidkoski: Makes sense, I think.  So revisit post #453 / #358.

ngutech21:
superseded by #453

PR #488, opened 2026-07-31
Fix gated delta recurrence precision

aleroot:
## Proposed changes

Use Kahan compensated summation for the recurrent k-state dot product so low-order terms do not drift at long context. Disable Metal reassociation and contraction only inside this reduction to preserve the algorithm under default fast math.

Add an adversarial BF16-path regression that checks the FP32 recurrent state against the correctly rounded reference result.

## Checklist


davidkoski (approved):
Looks great, thank you!

PR #490, opened 2026-07-31
Fix tied quantized Qwen3 MoE head sanitization

JamesPriceZV:
When `tie_word_embeddings` is enabled, converted quantized checkpoints can contain `lm_head.weight`, `lm_head.scales`, and `lm_head.biases`. Qwen3MoE currently removes only `lm_head.weight`, leaving the quantization metadata to load into the absent optional module and failing with `Unable to set lm_head on Qwen3MoEModel`.

Remove every `lm_head.*` tensor for tied embeddings, matching the model's use of `embed_tokens` as the output projection. Preserve all head tensors for untied configurations.

Adds focused tests for both tied and untied quantized heads.

Validation: `swift test --filter Qwen3MoESanitizeTests` (2 passed). Real model that exposed this: `justneedsomeavailableusername/Qwen3-MOE-4x0.6B-2.4B-Writing-Thunder-V1.2-mlx-4Bit`.

### DIFF
--- a/Libraries/MLXLLM/Models/Qwen3MoE.swift
+++ b/Libraries/MLXLLM/Models/Qwen3MoE.swift
@@ -255,7 +255,9 @@ public class Qwen3MoEModel: Module, LLMModel, KVCacheDimensionProvider {
         var sanitizedWeights = weights
 
         if configuration.tieWordEmbeddings {
-            sanitizedWeights["lm_head.weight"] = nil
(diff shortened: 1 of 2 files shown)


davidkoski (approved):
Change looks good.  Filed #525 -- a lot of models have a similar setup.  Maybe we want to apply the same fix to all of them.

PR #502, opened 2026-07-31
Migrate all models to ChatConventionsProviding; delete the infer chains

### LINKED ISSUE #501: Cross-cutting extension point for chat conventions keyed on model id
Follow-up split out of the review discussion on #482.

`ChatConventionsProviding` (added in #482) covers everything a model knows about itself. Two drawbacks @davidkoski identified in the original cross-cutting factory for `ReasoningConfig` were:

> - it can't be extended outside of mlx-swift-lm
> - it is not encapsulated

#482 addressed encapsulation for model-intrinsic conventions. This issue tracks the remaining piece: conventions keyed on something *other* than the model instance, notably the Hugging Face repo id. The motivating case is DeepSeek-R1-Distill, whose checkpoints report a base `model_type` (`qwen2` / `llama`) and so cannot be recognized by any model declaration.

@davidkoski's proposal:

> Another approach might be to have a registry of `ChatConventionsProviding` (along the lines of the `AbstractModelRegistry`) where callers outside mlx-swift-lm could register additional providers.

and, on the shape of the model's own hooks:

> Thinking about the function vs property more, I think we should leave it as a property. That is correct for the model to provide it. We can add a cross cutting protocol (interceptors) that consumes `modelId` (maybe among other things) later.

**Design decisions taken when implementing:**

- The model's own hooks stay plain properties. `modelId`-keyed logic lives in the separate resolver layer rather than being pushed into `ChatConventionsProviding`.
- Resolvers receive `modelId` + `modelType` only. Passing raw `configData` was rejected as re-introducing the raw-JSON poking #482 moved away from; passing the model instance was rejected as coupling the seam to model internals.
- Resolvers sit *above* the model in the resolution chain, because repo-id knowledge is more specific than model-intrinsic knowledge, and so a downstream package can correct a built-in declaration without constructing a `ModelConfiguration`.
- Most recently registered resolver wins, so downstream registrations can override built-ins.

Resulting precedence at load time:

1. an explicit value already on the `ModelConfiguration` (registry entry or caller)
2. a registered `ChatConventionsResolving`
3. the model's own `ChatConventionsProviding` declaration

**Status:** implemented rather than deferred, since removing `ReasoningConfig.infer` entirely depended on having somewhere for the id-keyed R1-Distill case to live. Filed to capture the framing and the decisions above.

CharlieTLe:
## Proposed changes

Completes the migration #482 started. Every model that previously relied on a centralized `model_type` prefix match now declares its own tool-call format and reasoning config, and both `ToolCallFormat.infer` and `ReasoningConfig.infer` are gone.

Conventions that are not model-intrinsic move to a new extension seam: `ChatConventionsResolving` + `ChatConventionsRegistry`, modeled on AbstractModelRegistry. The built-in DeepSeekR1ConventionsResolver recognizes R1-Distill by repo id, which no model instance can do (those checkpoints report a base `qwen2`/`llama` model_type). This also lets code outside mlx-swift-lm supply or override conventions, which addresses the "can't be extended outside of mlx-swift-lm" half of the original review.

Factory precedence is now: explicit ModelConfiguration value, then a registered resolver, then the model's own declaration.

The design of that seam is recorded in #501, which captures @davidkoski's original framing plus the decisions taken here: the model's own hooks stay plain properties, resolvers receive `modelId` + `modelType` only (not raw `configData` or the model instance), resolvers sit above the model because repo-id knowledge is more specific than model-intrinsic knowledge, and the most recently registered resolver wins so downstream registrations can override built-ins. Implementing it was a prerequisite for deleting `ReasoningConfig.infer`, since the id-keyed R1-Distill case needed somewhere to live.

Closes #501.

Llama needed care: LlamaModel backs both the `llama` and `mistral` registry entries and LlamaConfiguration did not decode `model_type`, so a vocabulary-only check would have newly resolved .llama3 for Mistral-Nemo (vocab 131072). The config now carries model_type and the declaration gates on it, with a regression test.

Intentional tightening:
- plain DeepSeek-V3 no longer reports always-on reasoning. `deepseek_v3` is an architecture shared with R1, while always-on reasoning is checkpoint and chat-template behavior, so R1 is recognized by repo id in the resolver instead. The old `ReasoningConfig.infer` matched the bare type and treated plain V3 as a reasoning model.
- nemotron_labs_diffusion no longer reports .xmlFunction
- unregistered types the prefixes would have matched (lfm2_5, glm4_5, qwen3_next_moe, mistral3_text) resolve to nil, since declarations only cover types with a registry creator
- case-insensitive model_type matching is gone; the type registry lookup is exact-match, so such configs could never load anyway

Qwen35MoEModel and the VLM Qwen35MoE subclass their base and inherit its declaration; a test pins that Swift semantic.

## Checklist


aleroot:
Nice direction. One thing I’d reconsider is declaring .alwaysOnThinking directly on DeepseekV3Model: deepseek_v3 describes an architecture shared by V3 and R1, while always-on reasoning is checkpoint/chat-template behavior. Since this PR already introduces the resolver layer, keeping R1-specific behavior there would preserve that abstraction boundary more cleanly.

Libraries/MLXLLM/LLMModelFactory.swift
```diff
@@ -616,21 +616,21 @@ public final class LLMModelFactory: GenericModelFactory {
...
-        // otherwise ask the model, then fall back to model_type inference.
+        // Chat conventions. Precedence: an explicit value on the configuration
+        // (registry entry or caller) wins; then a registered resolver, which sees
+        // the repo id the model cannot; then the model's own declaration.
+        let modelId = configuration.name
         if mutableConfiguration.toolCallFormat == nil {
             mutableConfiguration.toolCallFormat =
-                model.toolCallFormat
-                ?? ToolCallFormat.infer(from: baseConfig.modelType, configData: configData)
+                ChatConventionsRegistry.shared.toolCallFormat(
```
aleroot: inject the registry into the factories rather than hard-coding .shared

Libraries/MLXLMCommon/ChatConventionsRegistry.swift
```diff
@@ -0,0 +1,102 @@
...
+    }
+
+    /// Add a resolver. It takes precedence over everything registered before it.
+    public func register(_ resolver: any ChatConventionsResolving) {
+        lock.withLock { resolvers.append(resolver) }
+    }
+
+    /// First tool-call format offered by a resolver, in reverse registration order.
+    public func toolCallFormat(modelId: String, modelType: String) -> ToolCallFormat? {
+        lock.withLock {
```
aleroot: Could we snapshot resolvers while holding the lock, then invoke them after releasing it? Since resolvers can be implemented externally, calling them under the non-recursive lock could deadlock on re-entry and keeps the registry locked during arbitrary resolver work. The same applies to reasoningConfig.

Libraries/MLXLMCommon/ChatConventionsRegistry.swift
```diff
@@ -0,0 +1,102 @@
...
+
+// MARK: - Built-in resolvers
+
+/// Recognizes DeepSeek-R1 and its distills by repo id.
+///
+/// R1-Distill checkpoints report a base `model_type` (`qwen2` / `llama`), so no
+/// model instance can declare their always-on reasoning. Plain DeepSeek-V3 shares
+/// R1's `deepseek_v3` model type and declares reasoning itself, so this resolver
+/// keys purely on the id.
+public struct DeepSeekR1ConventionsResolver: ChatConventionsResolving {
```
davidkoski, line 102: @aleroot is this the R1/V3 cross cutting piece that you were thinking of?

aleroot: Yes

Libraries/MLXLLM/LLMModelFactory.swift
CharlieTLe: Done. Both factories now take `conventionsRegistry: ChatConventionsRegistry = .shared` alongside the other registries, and the call sites use it instead of reaching for `.shared`. Default means nothing existing changes. Added a test for the injected and defaulted cases.

Libraries/MLXLMCommon/ChatConventionsRegistry.swift
CharlieTLe: This was a bug, not just a style issue. Fixed by snapshotting the resolver list under the lock and iterating outside it, so resolvers never run while we are holding it. That covers both the re-entrancy deadlock and keeping the registry closed during arbitrary resolver work.

There is a test with a resolver that calls back into the same registry. Heads up that it hangs rather than fails if this regresses, since that is what a deadlock does.

Libraries/MLXLMCommon/ChatConventionsRegistry.swift
CharlieTLe: Yes. That exchange is what pushed me to take R1 off `DeepseekV3Model` completely (more in the other thread). The resolver is the only thing that knows about R1 now, keyed on repo id, and its doc comment covers both reasons it cannot live on a model: R1-Distill reports a base `qwen2`/`llama` type, and R1 proper shares `deepseek_v3` with plain V3.

Design notes are in #501, which this closes.

CharlieTLe:
Moved it. `DeepseekV3Model` declares nothing now and R1 is handled by id in `DeepSeekR1ConventionsResolver`. The architecture and the reasoning behavior are separate concerns, so the resolver is the better home.

One consequence worth being explicit about: plain DeepSeek-V3 now gets no reasoning config, where before it got always-on. That is a change against `main`, not just against the first commit here, because the old `infer` matched on the bare `deepseek_v3` type. The old comment more or less admitted it was wrong:

> Plain DeepSeek-V3 shares R1's "deepseek_v3" model_type; this type is treated as reasoning, refined by registry overrides.

I have treated it as a fix and added it to the tightening list in the description. R1 itself still resolves by id, and there is a test covering both. If you would rather V3 keep advertising reasoning for now, we could add a resolver rule for the `deepseek_v3` type instead, which keeps today's behavior without putting it back on the model. Happy either way.

davidkoski (approved):

PR #504, opened 2026-08-01
Support Gemma 4's labeled reasoning channels

abra-code:
ReasoningConfig describes a reasoning protocol as a delimiter pair, which Gemma 4 does not use. Its delimiters are special tokens (soc_token id 100 <|channel>, eoc_token id 101 <channel|>) and the opening token is followed by a role label, terminated by a newline, that names the span:

    <|channel>thought\n ... <channel|>the answer

A pair cannot carry that. Anchored on "<|channel>" it emits the label as the first word of the thinking; anchored on "<|channel>thought\n" it stops matching the moment the label is anything else and leaks the raw opener into the answer.

Add ReasoningChannel, carried as an optional ReasoningConfig.channel, and give ReasoningEventEmitter a label state. nil keeps the pair protocol exactly as it was, so Qwen3 and DeepSeek-R1 are unchanged and the existing ReasoningEventEmitterTests are untouched.

The label is consumed as metadata rather than emitted; a label not listed as a response label routes to reasoning, so an unknown channel stays readable and cannot leak markers into the answer; hold-back covers the longest watched delimiter, since both Gemma 4 tokens are 10 characters and a splitter sized for </think> tears them; an opener whose label never terminates settles into a visible body instead of buffering without bound; and a stray end delimiter with no opener is swallowed, because a special token arriving alone means the prompt opened the channel and it cannot be prose the model typed.

ReasoningChannel.gemma4 lists "content" as a response label defensively rather than from observation: no shipped Gemma 4 template emits such a channel, and Google's own x-regex response schema names the answer group "content" while leaving it outside any channel. Listing it costs nothing and means a channel-wrapped answer would not be hidden in the thought stream if one ever did appear.

The four Gemma 4 model types declare the protocol themselves through ChatConventionsProviding, as NanbeigeModel already does.

Also add a token-id overload of promptEndsInsideReasoning so the prompt-tail rule has one implementation rather than two.

Co-Authored-By: Claude

## Proposed changes

## Checklist


abra-code:
Without the fix the channel "thought" is displayed with tags in the message:
thought channel

abra-code:
conflicts resolved

abra-code:
Re-applied changes on top of the latest main - the merge conflicts with the latest work in the same area required more changes than just straightforward resolutions.

aleroot:
@abra-code sorry to be an intruder in your PR, but I have been seeing this issue as well in my app. 
The `ReasoningChannel` abstraction and bounded streaming tests are solid, but I don’t think this is mergeable yet because routing is limited to the Foundation Models path. `ChatSession`/standard generation still sends Gemma output directly to `ToolCallProcessor`, so the same stream can be classified differently and tool syntax inside an open thought channel may be promoted or missed depending on the API. I ask for @davidkoski point of view here, but I don't think we should misalign capabilities between ChatSession and the new API... 

I think to make this PR really complete I would:

- Move Gemma 4 reasoning/tool routing into a shared, token-aware `TokenStreamDecoder`.
- Match the structural special-token IDs before detokenization.
- Treat `<|tool_call>` as an explicit transition from reasoning into tool parsing.
- Reuse that decoder from both standard generation and `MLXLanguageModel`.
- Add integration tests covering both paths.
- Rebase onto current `main`; no dependency on #548 is required, though its validation layer should remain downstream of structural parsing.

The configuration model and most scanner tests are good, the main change needed is one protocol state machine for every generation surface...

abra-code:
@aleroot: your comments are most welcome - no intrusion. I should have added the info in the notes that the changes in this PR are meant to be the part one of the two PRs. I decided to split them into 2 separate changes because the second one, implementing what you noticed is missing, touches many more files and has more potential for conflicts (or, in military terms used by Claude AI: has a bigger blast radius...). I naively thought that the changes can be reviewed and committed relatively quickly one after another. But sitting on a PR for 3 weeks while the ground beneath moves makes it hard to maintain (and ironically makes a stronger case for keeping the changes separate). The second proposed change is here: https://github.com/abra-code/mlx-swift-lm/commit/33e617c22c9d1f914e3ed1a3598f5643ceefea39

@davidkoski: do you think both PRs could be presented, reviewed and merged as one? The second branch is built on top of the first PR's branch so contains all intended changes together.

As aside: my project [mlx-agent](https://github.com/abra-code/mlx-agent) works around and post-processes the gemma 4 thinking tags but I wanted to propose the changes here because it's the right place that would not require a bandaid downstream.

davidkoski:
Normally it is easier to review smaller PRs, but in this case it might be better to review both parts at once.

abra-code:
@davidkoski @aleroot: this is the combined PR with both parts: https://github.com/ml-explore/mlx-swift-lm/pull/563
We can close this one and move the conversation to the new PR.

abra-code:
Closing this old PR  to focus on the new, combined one:
https://github.com/ml-explore/mlx-swift-lm/pull/563

PR #506, opened 2026-08-01
Stand down from MTP speculation before the sliding cache wraps

### LINKED ISSUE #505: [BUG] MTP speculative decoding crashes once generation crosses the sliding window
## Describe the bug

`MTPSpeculativeTokenIterator` caps usable context at the model's sliding window.
With an `mtpDrafter`, any stream whose prompt plus generation crosses
`sliding_window` (512 by default on Gemma 4) dies with a precondition failure in
`Gemma4AssistantDraftModel.forwardHidden`:

```
sliding KV length 515 exceeds slidingWindow 512 — the production invariant that
lets the bidirectional sliding-window mask early-exit to all-zeros has been
violated
```

A long prompt dies on the second speculation round; a short prompt dies when
generation crosses the window. `precondition` is not compiled out in release, so
this is a hard crash, not a debug-only assert.

## To Reproduce

```swift
// Any Gemma 4 target + assistant drafter pair, e.g.
//   mlx-community/gemma-4-e4b-it-4bit
//   mlx-community/gemma-4-E4B-it-assistant-bf16
let stream = try generate(
    input: lmInput,          // prompt longer than text_config.sliding_window
    parameters: GenerateParameters(maxTokens: 64, temperature: 0),
    context: context,
    mtpDrafter: drafter,
    blockSize: 4
)
for await event in stream { /* crashes on the first draftBlock after the wrap */ }
```

A short prompt with `maxTokens` large enough to carry total context past the
window reproduces it equally.

## Expected behavior

Generation completes. Speculative decoding is an optimization; when the
conditions for it stop holding, the iterator should fall back to ordinary
single-token generation rather than terminating the process.

## Additional context

The chain, in order:

MTP's accept/reject step rewinds the main cache with `trimPromptCache`, which
requires a trimmable cache. `RotatingKVCache.isTrimmable` is
`offset < maxCacheSize`, so a sliding-window cache is untrimmable forever once it
wraps — the rotation has already overwritten the entries a rewind would restore.
The iterator checks `canTrimPromptCache` exactly once, in `init`, against a fresh
zero-offset cache where the check is vacuously true, and never re-checks. Every
verify pass is multi-token and therefore goes through `RotatingKVCache.updateConcat`,
whose documented allowance returns sliding K/V of length up to
`maxCacheSize + S - 1`; the emit hook hands exactly those arrays to `sharedKV`,
and the next round's `draftBlock → forwardHidden` reads that length and trips the
precondition above.

There is a second effect before the crash lands: on the round that crosses the
window, `trimPromptCache` returns 0 and the rewind silently does nothing. Because
its guard is all-or-nothing over the cache array, even the trimmable
full-attention entries go untrimmed, so that round's rejected drafts stay in the
cache and the logical offset runs ahead of the true sequence.

The precondition itself is correct and should stay — it converted what would
otherwise have been a silent attention-mask bug into a precise report. The gap is
that nothing stands the iterator down when speculation stops being viable.

## Proposed fix

Stand down to the iterator's existing sticky-passthrough regime instead of
crashing, checked at prefill and again at the top of each speculative round, so a
user who asks for MTP on a long context gets a working stream plus a
`passthroughReason` in the completion stats. The check tests whether the cache can
absorb the coming round and still be rewound, rather than whether it has already
wrapped, so the boundary round never strands rejected drafts and the stream stays
token-identical to plain autoregressive generation.

PR to follow.

angelsbrood:
## Proposed changes

cc @davidkoski

Fixes #505. Follow-up to #308.

With an `mtpDrafter`, total context was effectively capped at the model's sliding
window: once prompt plus generation crossed it, `draftBlock → forwardHidden` hit
`precondition(slidingKvLen <= textCfg.slidingWindow)` and the process died.

MTP's accept/reject step rewinds the main cache with `trimPromptCache`, and
`RotatingKVCache.isTrimmable` is `offset < maxCacheSize` — a sliding-window cache
is untrimmable forever once it wraps. The iterator checked `canTrimPromptCache`
only in `init`, against a fresh zero-offset cache where it is vacuously true, and
never again; meanwhile every verify pass goes through `updateConcat`, whose
allowance returns sliding K/V up to `maxCacheSize + S - 1`, and the emit hook
hands exactly those arrays to the drafter on the following round.

This adds `KVCache.isTrimmable(after:)` as a cache-owned look-ahead query. The
private helper, `standDownIfNoTrimHeadroom(positions:reason:)`, asks each
main-cache entry whether it remains trimmable after the coming round. It is called
at two points: at the end of `init` once the prompt is in, and at the top of each
speculative round before any `draftBlock`. When the cache cannot absorb the
coming round and still be rewound, the iterator switches to its existing
sticky-passthrough regime — the same mechanism already used when the target stops
emitting drafter state — with a one-time log line and a `passthroughReason` on
`GenerateCompletionInfo`. Nothing throws, and generation completes.

`RotatingKVCache` owns the strict sliding-window calculation,
`offset + positions < maxCacheSize`.
The predicate asks whether the cache has *headroom* for the round rather than
whether it has *already wrapped*, which matters for correctness rather than just
crash-avoidance: the wrap happens inside the verify pass, so an after-the-fact
check would let the boundary round run, and that round's `trimPromptCache` returns
0 (its guard is all-or-nothing over the array, so even the trimmable
full-attention entries go untrimmed). Its rejected drafts would stay in the cache
and every later token would be conditioned on tokens the target had rejected.
Standing down while headroom remains keeps every rewind valid.

One ordering change comes with it: `next()` now drains `pendingTokens` before
testing `passthrough`. The prefill bonus is queued there by `prepare`, and a
stand-down that fires at the end of `init` would otherwise skip it and start the
stream a position ahead of an equivalent greedy run. Every pre-existing
passthrough transition happens inside `speculateRound`, which `next()` reaches
only after clearing that buffer, so the reordering is a no-op for them.

The precondition, `updateConcat`'s overflow allowance, the emit hook, and the
classic `SpeculativeDecoding` path are all untouched. The precondition is correct
and doing its job — it turned what would have been a silently wrong attention
pattern into a precise report.

Two notes for reviewers. Speculative-decoding telemetry now describes only the
pre-stand-down prefix of such a stream while `emittedTokenCount` keeps climbing,
so `acceptanceRate` skews on runs that cross the window; happy to gate or annotate
that if you'd prefer. And the classic `SpeculativeTokenIterator` has the same
init-only trimmability check and degrades silently rather than crashing — left
alone here to keep this diff to the reported bug, but I can follow up.

## Verification

Run on macOS (Apple silicon), `swift-format` via pre-commit, against
`main` @ `cd1ab3d`.

```
xcodebuild build-for-testing -scheme mlx-swift-lm-Package -destination 'platform=macOS'
xcrun xctest .../Build/Products/Debug/MLXLMTests.xctest
  → Test run with 317 tests in 31 suites passed after 44.302 seconds

pre-commit run --all-files
  → swift-format ... Passed
```

Three regression tests were added to `MTPSpeculativeTokenIteratorTests`, all
CI-scoped against the existing synthetic mocks (no checkpoint required). Each was
confirmed to fail with the production change reverted and pass with it:

- `testMTPIteratorStandsDownWhenPromptFillsSlidingWindow` — a prompt past the
  window stands down during `init`; the stream still completes to `maxTokens`, the
  drafter is never invoked, and the prefill bonus is still the first token emitted.
- `testMTPIteratorStandsDownWhenSlidingCacheWrapsMidStream` — a stream that
  speculates a full round first and only then runs out of headroom stands down
  mid-stream with `acceptedCount > 0`.
- `testMTPIteratorMatchesGreedyAcrossMidStreamStandDown` — the wrap-crossing
  stream emits exactly the tokens a plain `TokenIterator` emits from the same
  model.

`CountingKVCache` gained an opt-in `wrapAt:` parameter and mirrors
`RotatingKVCache.isTrimmable(after:)` through the same strict
`offset + positions < wrapAt` boundary. `maxSize` still reports the window. It
defaults to nil — unbounded and always trimmable — so every pre-existing test
is unaffected.

Not run: the `IntegrationTesting` suites, which need checkpoints from the HF cache
and do not run in CI.

## Checklist

- [x] I have updated the necessary documentation (if needed) — the window bound on speculation is now documented on both public `generate(mtpDrafter:)` overloads.

### DIFF
--- a/Libraries/MLXLMCommon/MTPSpeculativeTokenIterator.swift
+++ b/Libraries/MLXLMCommon/MTPSpeculativeTokenIterator.swift
@@ -408,24 +408,23 @@ public struct MTPSpeculativeTokenIterator: TokenIteratorProtocol {
     /// Stand down to passthrough if the main cache cannot absorb `positions`
     /// more tokens and still be rewound.
     ///
-    /// MTP's accept/reject step depends on `trimPromptCache`, and
-    /// ``RotatingKVCache/isTrimmable`` is `offset < maxCacheSize` — a sliding
-    /// window cache is untrimmable forever once it wraps. The wrap happens
-    /// *inside* the verify pass, so a check that merely asks "has it wrapped?"
-    /// fires a round too late: that round's rewind silently returns 0 (the
-    /// guard in `trimPromptCache` is all-or-nothing over the array, so even the
-    /// trimmable full-attention entries are skipped) and its rejected drafts
-    /// stay welded into every layer's cache. Standing down while headroom
-    /// remains keeps every rewind valid, so a passthrough-crossing stream
-    /// stays bit-exact to a plain autoregressive run.
+    /// MTP's accept/reject step depends on `trimPromptCache`, so each cache
+    /// predicts whether it will remain trimmable after the verify positions
+    /// are appended. The append happens *inside* the verify pass, so a check
+    /// that merely asks "is it trimmable now?" fires a round too late: that
+    /// round's rewind silently returns 0 (the guard in `trimPromptCache` is
+    /// all-or-nothing over the array, so even the trimmable full-attention
+    /// entries are skipped) and its rejected drafts stay welded into every
+    /// layer's cache. Standing down while headroom remains keeps every rewind
+    /// valid, so a passthrough-crossing stream stays bit-exact to a plain
+    /// autoregressive run.
     @discardableResult
     private mutating func standDownIfNoTrimHeadroom(
         positions: Int, reason: String
     ) -> Bool {
         guard !passthrough else { return false }
-        let hasHeadroom = mainCache.allSatisfy { cache in
-            guard let maxSize = cache.maxSize else { return true }
-            return cache.offset + positions < maxSize
+        let hasHeadroom = mainCache.allSatisfy {
+            $0.isTrimmable(after: positions)
         }
         guard !hasHeadroom else { return false }
         switchToPassthrough(reason: reason)
--- a/Libraries/MLXLMCommon/MTPSpeculativeTokenIterator.swift
+++ b/Libraries/MLXLMCommon/MTPSpeculativeTokenIterator.swift
@@ -135,6 +135,17 @@ public struct MTPSpeculativeTokenIterator: TokenIteratorProtocol {
         let prefillStart = Date.timeIntervalSinceReferenceDate
         try prepare(input: input, windowSize: parameters.prefillStepSize)
         self.promptPrefillTime = Date.timeIntervalSinceReferenceDate - prefillStart
+
+        // The guard above ran against a fresh, zero-offset cache, where a
+        // rotating cache is trimmable vacuously. Re-check now that the prompt
+        // is in: one whose length leaves no room for a speculative round can
+        // never be rewound, so speculation is unavailable for this stream.
+        // `blockSize` bounds the first round's width.
+        standDownIfNoTrimHeadroom(
+            positions: blockSize,
+            reason:
+                "prompt fills the sliding window — MTP rewind unavailable; generating without speculation"
+        )
     }
 
     /// Prefill the main model with the prompt. The drafter has no cache to
@@ -241,6 +252,18 @@ public struct MTPSpeculativeTokenIterator: TokenIteratorProtocol {
             let sharedKV = state[mtpSharedKVStatesKey]
         else {
             switchToPassthrough(reason: "main model did not emit drafter state")
+            return
+        }
+
+        // This round's verify pass commits `numDraft + 1` positions. If that
+        // would carry the sliding cache past its window, the rewind below
+        // would no-op and strand this round's rejected drafts in the cache —
+        // stand down now, while the cache is still rewindable.
+        if standDownIfNoTrimHeadroom(
+            positions: numDraft + 1,
+            reason:
+                "sliding cache wrapped mid-stream — MTP rewind unavailable; continuing without speculation"
+        ) {
             return
         }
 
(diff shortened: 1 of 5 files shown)


angelsbrood:
Pushed a doc note on the window bound prompted by reporter's commentary.

Libraries/MLXLMCommon/MTPSpeculativeTokenIterator.swift
```diff
@@ -382,6 +405,37 @@ public struct MTPSpeculativeTokenIterator: TokenIteratorProtocol {
...
+    /// stays bit-exact to a plain autoregressive run.
+    @discardableResult
+    private mutating func standDownIfNoTrimHeadroom(
+        positions: Int, reason: String
+    ) -> Bool {
+        guard !passthrough else { return false }
+        let hasHeadroom = mainCache.allSatisfy { cache in
+            guard let maxSize = cache.maxSize else { return true }
+            return cache.offset + positions < maxSize
+        }
```
davidkoski, line 428: I think this logic should belong to the KVCache itself -- that way different variants can handle it appropriately.  perhaps:

```swift
func isTrimmable(after positions: Int) -> Bool
```

davidkoski (changes requested):
Looks good but see my question on moving the logic of size/trimming inside KVCache itself.

Libraries/MLXLMCommon/MTPSpeculativeTokenIterator.swift
angelsbrood: Moved the look-ahead decision into `KVCache` as `isTrimmable(after:)`. `RotatingKVCache` now owns the strict window boundary, `CacheList` delegates to its children, and the iterator only asks each cache whether the upcoming verify positions remain rewindable. I also added cache-level boundary and existential-dispatch tests; the existing MTP stand-down and greedy-parity tests still pass.

Verification: `pre-commit run --all-files` passes; all six cache-policy tests and all three MTP stand-down regressions pass, and `MLXLMTests` completed 323 tests in 31 suites. The all-target run still has the unrelated `ToolCallingSchemaTests.grammarBuilderHoistsNestedDefsInBothArms` failure.

davidkoski (approved):
Looks good, thank you!

PR #507, opened 2026-08-02
Add opt-in q4_0 lattice calibration to model conversion

GoodOlClint:
## Proposed changes

Adds an opt-in **calibration** to model conversion that reproduces a checkpoint's existing `Q4_0` quantization grid instead of deriving a new one.

### The problem

Some checkpoints ship weights that were already trained against a specific quantization grid — Google's Gemma 4 QAT releases are the motivating case. `MLX.quantized` derives a fresh grid from those weights by min/max, which discards the alignment they were trained for. The result is a converted model that sits further from the original than it needs to, at the same bit width and the same file size.

The grid is recoverable: it is 32-element groups scaled by the **signed** element of largest magnitude, `d = extremum / -8`, which is expressible as ordinary affine `scales = d`, `biases = -8 * d`.

### What this changes

```swift
public enum ModelConversionQuantizationCalibration: Sendable, Equatable {
    case standard   // today's behaviour, unchanged and still the default
    case q4Zero     // reproduce the Q4_0 grid
}
```

`calibration` is added to `ModelConversionQuantization` and surfaced on `ModelConversionOptions`. `.q4Zero` resolves to 4 bits / group size 32 / affine, and **rejects** a conflicting explicit value rather than silently overriding it. Per-layer predicate overrides are validated on the same terms, before the output directory is created or removed.

**This is a derivation, not a storage format.** The output is ordinary affine 4-bit, so converted models load and run through the existing path with no decode-side change anywhere. The diff touches only `ModelConversion.swift` and its tests.

### Evidence

Each converted model was compared against its own bfloat16 reference over 200 chat-formatted prompts, teacher-forced, measuring mean KL divergence per token position (lower is better) and exact top-1 agreement. Stock group-32 conversion vs this calibration, **at identical file size**:

| model | KL, stock | KL, calibrated | top-1, stock | top-1, calibrated |
|---|---|---|---|---|
| Gemma 4 E2B | 0.0220 | **0.0175** | 95.7% | **96.2%** |
| Gemma 4 E4B | 0.0160 | **0.0112** | 96.6% | **97.2%** |
| Gemma 4 12B | 0.1219 | **0.0631** | 92.8% | **94.3%** |
| Gemma 4 31B | 0.0321 | **0.0205** | 96.5% | **96.9%** |

Correctness was established independently of those measurements: the derivation reproduces Google's own shipped `Q4_0` GGUF **exactly** — 100% of codes and scales matched across 721M weights on 12B — and this Swift implementation was then checked tensor-by-tensor against that verified artifact on E2B: **275 tensors, 0 mismatches**, bit-identical packed codes, scales and biases.

### Scope and limits, stated plainly

- **Only `Linear` is calibrated.** Other quantizable layers still land on the same 4-bit/group-32 affine grid but derive it the standard way, because `QuantizedEmbedding` has no initializer accepting precomputed scales and biases. That affects accuracy, not loadability, and is documented on the API. Happy to follow up if an upstream initializer would be welcome.
- **No benefit on mixture-of-experts.** The same measurement on Gemma 4 26B-A4B showed no significant change (0.0707 → 0.0680, overlapping confidence intervals), so no claim is made there.
- **Opt-in only.** No model-name or fingerprint auto-detection; callers ask for it explicitly.
- Measurements are single-run on one prompt corpus with a fixed seed.

### Notes for review

- The case is named `q4Zero` rather than `q4_0` to fit Swift naming; the format name `Q4_0` is kept in documentation.
- A mirrored change for `mlx-lm` is prepared, so the two conversion paths stay in step.

## Checklist

Tests: 14 added to `ModelConversionTests`, covering hand-computable codes/scales/biases, positive and negative extrema, truncation and clipping at both ends, degenerate all-zero groups, per-group scaling across multiple rows and groups, dequantization onto the intended lattice, rejection of incompatible settings and per-layer overrides, and one pinning the signed-extremum rule apart from `-absMax/8` — the error that silently produces a different grid on roughly half of all groups. Full `xcodebuild` suite passes.

---

**Published weights using this path** — 4-bit MLX conversions of the Gemma 4 QAT checkpoints, which load with released `mlx-lm` (they are ordinary affine 4-bit; none of this PR is needed to *use* them):

- [gemma-4-E2B-it-qat-q4_0-mlx](https://huggingface.co/goodolclint/gemma-4-E2B-it-qat-q4_0-mlx)
- [gemma-4-E4B-it-qat-q4_0-mlx](https://huggingface.co/goodolclint/gemma-4-E4B-it-qat-q4_0-mlx)
- [gemma-4-12B-it-qat-q4_0-mlx](https://huggingface.co/goodolclint/gemma-4-12B-it-qat-q4_0-mlx)
- [gemma-4-31B-it-qat-q4_0-mlx](https://huggingface.co/goodolclint/gemma-4-31B-it-qat-q4_0-mlx)

Each was produced by this branch and verified bit-identical to Google's shipped `Q4_0` grid before upload.

### DIFF
--- a/Libraries/MLXLMCommon/ModelConversion.swift
+++ b/Libraries/MLXLMCommon/ModelConversion.swift
@@ -24,7 +24,7 @@ public enum ModelConversionQuantizationCalibration: Sendable, Equatable {
     /// Re-deriving a grid from such weights discards the alignment they were trained for;
     /// reproducing it keeps them on their intended lattice.
     ///
-    /// Requires 4 bits, a group size of 32, and ``QuantizationMode/affine``. Supplying a
+    /// Requires 4 bits, a group size of 32, and `QuantizationMode.affine`. Supplying a
     /// conflicting value throws ``ModelConversionError/incompatibleCalibration(_:_:)``
     /// rather than silently overriding it.
     ///
--- a/Libraries/MLXLMCommon/ModelConversion.swift
+++ b/Libraries/MLXLMCommon/ModelConversion.swift
@@ -9,6 +9,32 @@ package let modelConversionSidecarPatterns = [
 ]
 package let modelConversionDownloadPatterns = ["*.safetensors"] + modelConversionSidecarPatterns
 
+/// How the quantization grid is derived from the source weights.
+///
+/// This selects a *derivation*, not a storage format: every calibration produces ordinary
+/// quantized layers that load and run without any decode-side support.
+public enum ModelConversionQuantizationCalibration: Sendable, Equatable {
+    /// Derive the grid from the weights, as MLX does by default.
+    case standard
+
+    /// Reproduce the `Q4_0` grid: symmetric, 32-element groups, scaled by the signed
+    /// element of largest magnitude.
+    ///
+    /// Use this for checkpoints whose weights were already trained against that grid.
+    /// Re-deriving a grid from such weights discards the alignment they were trained for;
+    /// reproducing it keeps them on their intended lattice.
+    ///
+    /// Requires 4 bits, a group size of 32, and ``QuantizationMode/affine``. Supplying a
+    /// conflicting value throws ``ModelConversionError/incompatibleCalibration(_:_:)``
+    /// rather than silently overriding it.
+    ///
+    /// Only `Linear` layers are calibrated. Other quantizable layers are still quantized to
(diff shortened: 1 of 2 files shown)


davidkoski (approved):
Change looks good, thank you!

PR #509, opened 2026-08-04
Qwen2.5-VL: accumulate vision cuSeqlens across temporal slices

### LINKED ISSUE #503: `Qwen25VL` vision attention mask mis-accumulates `cuSeqlens` for frames with `t > 1`
`Libraries/MLXVLM/Models/Qwen25VL.swift:667-674`:

```swift
var cuSeqlens = [0]
for frame in frames {
    let seqLen = frame.h * frame.w
    cuSeqlens.append(
        contentsOf: Array(repeating: seqLen, count: frame.t).map {
            cuSeqlens.last! + $0
        })
}
```

`map` builds its entire result before `append(contentsOf:)` runs, so `cuSeqlens.last!` is the
same value for every element of the map rather than a running total. A frame with `t = 3`,
`h*w = 16` yields `[16, 16, 16]` where the cumulative offsets should be `[16, 32, 48]`.

`cuSeqlens` feeds `attentionMask(sequenceLength:cuSeqlens:)` (`:641-656`), which marks
`[start..<end, start..<end]` blocks as attendable. With the wrong cumulative values the
full-attention mask is not block-diagonal per frame for any `t > 1` input — i.e. video. With
`frames = [THW(3,4,4), THW(2,2,4)]` the cumulative list stops at 24 while the buffer has 64
rows, leaving rows 24-63 with an all-`false` mask row.

Images are unaffected: `t == 1` makes the map single-element, so `cuSeqlens.last! + seqLen` is
correct.

This is reachable — `Qwen25VL.swift:856-905` is a full video path (`<|video_pad|>`, real
`t > 1` frames). It appears to be uncovered: `git grep "THW(" Tests/` returns only
`THW(1, …)`, so nothing in the test suite exercises `t > 1`.

The fix is a one-liner — accumulate explicitly rather than reading `last!` inside the `map`:

```swift
var running = cuSeqlens.last!
for _ in 0 ..< frame.t {
    running += seqLen
    cuSeqlens.append(running)
}
```

Happy to open a PR with a `t > 1` regression test if that's useful.

Found while checking whether a warm-cache continuation could safely narrow a video payload;
it is independent of that work and reproduces on `main` today.

NivDvir:
## Proposed changes

Fixes #503.

### Problem

`Qwen25VL.swift` builds the vision full-attention mask from cumulative per-frame
boundaries:

```swift
cuSeqlens.append(
    contentsOf: Array(repeating: seqLen, count: frame.t).map {
        cuSeqlens.last! + $0
    })
```

`map` builds its entire result before `append(contentsOf:)` runs, so `cuSeqlens.last!`
is the same value for every element of a frame's chunk rather than a running total.

For `frames = [THW(3, 4, 4), THW(2, 2, 4)]` — 64 patch rows — the boundaries come out
`[0, 16, 16, 16, 24, 24]` instead of `[0, 16, 32, 48, 56, 64]`. `attentionMask(sequenceLength:cuSeqlens:)`
marks `[start..<end, start..<end]` attendable per boundary pair, so rows 24...63 fall
outside every block.

Worth being precise about the consequence: the boolean mask becomes an additive
`-10000` bias, not `-inf`, so those rows do not produce NaN — their softmax converges
to a uniform distribution over every patch in the payload, across frame boundaries.
The failure is silent.

Still images are unaffected: `t == 1` makes the chunk single-element, so
`cuSeqlens.last! + seqLen` is already correct.

### Fix

Carry the accumulator explicitly, one block per temporal slice.

The computation is extracted to `visionCuSeqlens(_:)` at internal visibility for one
reason: the original sits inside `private enum Vision`, which tests cannot reach even
with `@testable`, and `git grep "THW(" Tests/` showed only `THW(1, …)` — the `t > 1`
path had no coverage at all. Because the mask bias is finite, a downstream numeric
assertion does not reliably surface the bug; the boundaries themselves do. Happy to
re-house the helper wherever you'd prefer.

### Tests

Four cases in `Qwen25VLVisionMaskTests`: single image, multiple images, a multi-frame
video grid, and a strict-monotonicity check (a repeated boundary is an empty block,
which is exactly how the bug manifests).

Against the previous logic the two temporal tests fail with
`[0, 16, 16, 16, 24, 24]` vs `[0, 16, 32, 48, 56, 64]`; the still-image tests pass
either way.

Full `MLXLMTests` passes apart from the pre-existing `TurboQuantIntegrationTests`
flake (`cos 0.944` against a `0.97` bar), which is unrelated to this change.

### DIFF
--- a/Libraries/MLXVLM/Models/Qwen25VL.swift
+++ b/Libraries/MLXVLM/Models/Qwen25VL.swift
@@ -326,6 +326,25 @@ private enum Language {
 
 // MARK: - Vision
 
+/// Cumulative patch-row boundaries, one per temporal slice, used to build the
+/// block-diagonal vision attention mask.
+///
+/// Each frame contributes `t` blocks of `h * w` rows, so the boundaries must be a
+/// running total across every slice of every frame. Internal rather than nested in
+/// `Vision` so that `t > 1` (video) grids can be covered directly by tests.
+func visionCuSeqlens(_ frames: [THW]) -> [Int] {
+    var cuSeqlens = [0]
+    for frame in frames {
+        let seqLen = frame.h * frame.w
+        var running = cuSeqlens.last!
+        for _ in 0 ..< frame.t {
+            running += seqLen
+            cuSeqlens.append(running)
+        }
+    }
+    return cuSeqlens
+}
+
 private enum Vision {
 
     static fileprivate func applyMultimodalRotaryPositionEmbedding(
@@ -664,15 +683,7 @@ private enum Vision {
 
             // prepare attention masks
             let seqLen = hiddenStates.dim(0)
-            var cuSeqlens = [0]
-            for frame in frames {
-                let seqLen = frame.h * frame.w
-                cuSeqlens.append(
-                    contentsOf: Array(repeating: seqLen, count: frame.t).map {
-                        cuSeqlens.last! + $0
-                    })
-            }
-            let cuSeqlensArray = MLXArray(cuSeqlens)
+            let cuSeqlensArray = MLXArray(visionCuSeqlens(frames))
 
             let fullAttentionMaskBool = attentionMask(
                 sequenceLength: seqLen, cuSeqlens: cuSeqlensArray)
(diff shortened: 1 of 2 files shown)


davidkoski (approved):
Change looks good, thank you.

PR #510, opened 2026-08-04
Make MambaCache rewindable so speculative decoding can run on hybrid SSM/attention models

agerjura:
## The problem

`MTPSpeculativeTokenIterator` refuses to initialise without a trimmable main cache — `MLXLMCommon/MTPSpeculativeTokenIterator.swift:115`:

```swift
guard canTrimPromptCache(self.mainCache) else {
    throw KVCacheError(
        message: "MTP speculative decoding requires a trimmable main KV cache.")
}
```

`canTrimPromptCache` is an `allSatisfy` over every layer's cache (`KVCache.swift:1883`), and `MambaCache` never overrides `isTrimmable`, so it inherits `BaseKVCache`'s `false` (`KVCache.swift:205`). Neither does its parent `ArraysCache`.

So **one linear-attention layer disables speculative decoding for the entire model.** For hybrid Qwen3.5/3.6 GDN, `layer_types` with `full_attention_interval: 4` gives 30 `MambaCache` instances out of 40 layers. `Evaluate.swift:880` and `ChatSession.swift:851` gate the same way.

This is not Qwen-specific. Ten models in this repo construct a `MambaCache`: Qwen35, Mamba2, Jamba, NemotronH, GraniteMoeHybrid, BaichuanM1, LFM2, LFM2MoE, LFM2VL. None of them can speculate today.

There is also no way to fix this from outside the module: `MambaCache` is `public`, not `open`, so it cannot be subclassed downstream — and substituting a different `KVCache` fails *silently*, because `Qwen35.swift:670` and `:729` use `cache as? MambaCache`, which yields `nil` and makes the GDN layer run with no cache at all.

## The approach

A recurrent state has no rows to drop — token `t`'s contribution is already folded into `t+1`'s state — so `trim(_:)` cannot be implemented the way `KVCacheSimple.trim` is. What it *can* mean is "restore the state as it was `n` tokens ago", which needs history.

`rewindDepth` opts in to keeping a bounded history:

- **Default `0` means nothing changes.** Nothing is recorded, nothing is allocated, `isTrimmable` stays `false`, and `MTPSpeculativeTokenIterator` still throws for callers who have not opted in. This is deliberately not on by default — see the performance note below.
- `checkpoint(advancing:)` advances the position and snapshots. Models call it once per forward pass; with `rewindDepth == 0` it is a single branch.
- Memory is `rewindDepth + 1` copies of the slots — for Qwen3.6 GDN, ~2.06 MiB per linear layer (recurrent `[1,32,128,128]` f32 + conv `[1,4,8192]` bf16), so ~61.9 MiB per retained step across 30 layers.

Three details that are easy to get wrong, and are the reason the diff looks the way it does:

1. **Snapshots store the raw slots, not `state`.** `ArraysCache.state` drops empty slots, so a model with `convKernelSize == 1` (slot 0 never populated) has `state == [ssm]`, and restoring through the `state` setter would resize the cache from two slots to one — the next `cache[1]` write then traps.

2. **Position is tracked inside the cache, not via `offset`.** `ArraysCache.advance(_:)` only decrements `lengths`/`leftPadding` and never touches `offset` (`KVCache.swift:1301`), and no model using `MambaCache` maintains it. Keying checkpoints on `offset` would silently record position 0 every step and make `trim` a no-op. Keeping position local also means this change cannot perturb anything that reads `offset`.

3. **`trim` is exact or refused, never approximate.** Over-rewinding is not a safe fallback: if the attention caches rewind `n` and the SSM state rewinds `n+2`, the next forward runs at the attention position with a state from two tokens earlier and produces wrong output with no error. Since `trimPromptCache` does not treat a `0` return as failure, `isTrimmable` is `true` only when a snapshot actually exists to honour.

The floor is the oldest retained snapshot. A rewind past it — including to before a prefill, which nothing snapshots — refuses and leaves state untouched.

## Performance: this is a key, not a speedup

I want to be straight about this, because the obvious reading of "trimmable SSM state" is a throughput claim and that claim is weak:

- vLLM's own batch-1 column reports **0.99×** for Qwen3.5-4B GDN.
- SGLang reports "slightly slower at batch 1".
- On my own per-token byte accounting the SSM-state term is 125.83 MB of 1,795.39 MB/token = **7.0%**, so a **1.07× ceiling even if the SSM kernel were free**.

All three are right and this PR does not dispute any of them. The value here is that MTP speculative decoding currently **cannot start at all** on these models, and this is what unblocks measuring it.

My own projection for MTP on a 24 GB M4 (103 GB/s measured streaming, 90.09 ms/token) brackets **0.93× to 1.48×** depending on two unknowns I cannot resolve without this change landing: which acceptance metric applies (published MTP numbers for this family report the same runs as both ~79–85% exact-match and ~46–47% probabilistic), and how much the routed-expert set overlaps between adjacent positions at `blockSize = 2`. **The bracket includes a loss.** That is precisely why `rewindDepth` defaults to off, and why I have framed this as unblocking a measurement rather than delivering a speedup.

## Tests

`Tests/MLXLMTests/MambaCacheRewindTests.swift`, 12 tests, no model download:

- the default path stays untrimmable, and `trim` refuses rather than guessing
- a hybrid list is untrimmable until the `MambaCache` opts in — the actual gate
- rewind restores state exactly, is repeatable, and the cache is usable after
- out-of-history, before-the-start, and non-positive trims all refuse and leave state untouched
- a rewind cannot land inside a prefill
- history is bounded by `rewindDepth`
- empty slots survive a rewind (detail 1 above)
- `copy()` preserves rewindability, independently of the original
- end to end through `trimPromptCache`

Existing suites pass unchanged. On an M4 / macOS, via the path CONTRIBUTING documents:

```
xcodebuild test -scheme mlx-swift-lm-Package -destination 'platform=macOS' \
  -skipPackagePluginValidation -only-testing:MLXLMTests
→ ** TEST SUCCEEDED **   Executed 291 tests, with 0 failures (0 unexpected)
```

That includes the pre-existing `MambaCache` round-trip and batch-metadata tests, and the Qwen3.5 chunked-prefill and cache suites.

## Open question for maintainers

I chose **exact-or-refuse** for `trim`. The alternative is to return the actual amount rewound and require callers to honour the return value — `trimPromptCache` already propagates it, but the MTP iterator's use of it would need auditing, and a mismatch there is silent. If you would rather have that shape, I am happy to switch.

I also left `MambaCache` as `public` rather than `open`. `open` does not compile here — `error: superclass 'ArraysCache' of open class must be open` — so opening it up for downstream subclassing means opening `ArraysCache` too. Fixing `MambaCache` itself makes that unnecessary for this use case, but say the word if you want both opened as a follow-up.

### DIFF
--- a/Libraries/MLXLMCommon/KVCache.swift
+++ b/Libraries/MLXLMCommon/KVCache.swift
@@ -1400,7 +1400,56 @@ public class ArraysCache: BaseKVCache {
 }
 
 /// Simple cache for Mamba-style state space models
+///
+/// ## Rewindability
+///
+/// A recurrent state has no rows to drop: token `t`'s contribution has already
+/// been folded into token `t+1`'s state, so there is nothing to slice off the
+/// end. `trim(_:)` therefore cannot be implemented the way `KVCacheSimple`
+/// implements it, and `MambaCache` reports `isTrimmable == false` by default.
+///
+/// That default is what blocks speculative decoding on hybrid SSM/attention
+/// models: ``MTPSpeculativeTokenIterator`` requires
+/// ``canTrimPromptCache(_:)``, which is an `allSatisfy` over every layer's
+/// cache, so a single non-trimmable entry disables speculation for the whole
+/// model.
+///
+/// Opting in to ``rewindDepth`` makes the cache rewindable by keeping a bounded
+/// history of past states and restoring one on `trim`. With `rewindDepth == 0`
+/// — the default — nothing is recorded, nothing is allocated, and behaviour is
+/// identical to before this capability existed.
+///
+/// Callers must set `rewindDepth` to at least the deepest rewind they will
+/// request (for speculative decoding, `blockSize - 1`).
 public class MambaCache: ArraysCache {
+
+    /// Bounded history of past cache slots, oldest first, tagged with the token
+    /// position each snapshot represents.
+    ///
+    /// The slots are stored rather than ``state`` because `state` drops empty
+    /// slots: a `MambaCache` whose conv slot is unused (`convKernelSize == 1`)
+    /// has `state == [ssm]`, and restoring that through the `state` setter would
+    /// resize the cache from two slots to one.
+    private var rewindHistory: [(position: Int, slots: [MLXArray?])] = []
+
+    /// Token position this cache has advanced to, maintained independently of
+    /// ``BaseKVCache/offset``.
+    ///
+    /// `ArraysCache` never advances `offset` — `advance(_:)` only decrements
+    /// `lengths`/`leftPadding` — and models using this cache do not maintain it
+    /// either. Tracking position here keeps rewind self-contained instead of
+    /// making it depend on a field nothing currently sets.
+    private var rewindPosition: Int = 0
+
+    /// How many tokens of rewind history to retain. `0` (the default) disables
+    /// rewinding entirely.
+    ///
+    /// Memory cost is `rewindDepth + 1` copies of the cache slots. For
+    /// Qwen3.5/3.6 GDN that is ~2.06 MiB per linear-attention layer
+    /// (recurrent `[1, 32, 128, 128]` f32 plus conv `[1, 4, 8192]` bf16), so
+    /// ~61.9 MiB per retained step across 30 such layers.
+    public var rewindDepth: Int = 0
+
     public init(leftPadding: [Int]? = nil) {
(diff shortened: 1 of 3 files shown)


aleroot:
@davidkoski I think this should be closed, because it is effectively superseded by #351 
In addition I think that if generalized multi-token Mamba rewind is revisited, it should not be implemented as ordinary cache history...

PR #511, opened 2026-08-04
Allow downstream specialization of the Qwen3.5 GDN/MoE blocks and SwitchGLU

agerjura:
## What is currently unreachable

The Qwen3.5 language stack in `Libraries/MLXVLM/Models/Qwen35.swift` cannot be specialized from another module:

| Line | Declaration | Blocker |
|---|---|---|
| `221` | `enum Qwen35Language {` | no access modifier ⇒ internal, so everything nested inside is unreachable regardless of its own modifier |
| `441` | `final class GatedDeltaNet: Module {` | `final` **and** internal |
| `574` | `final class SparseMoeBlock: Module, UnaryLayer {` | `final` **and** internal |
| `626` | `final class DecoderLayer: Module {` | `final` **and** internal |
| `681` | `final class Model: Module {` | `final` **and** internal |

And in `Libraries/MLXLMCommon/SwitchLayers.swift`, `SwitchGLU` (`:42`) and `SwitchLinear` (`:222`) are `public`, not `open` — subclassable only inside this module.

## Why I care

I serve Qwen3.6-35B-A3B on a 24 GB Apple M4 with the routed experts **streamed from SSD** into a bounded RAM cache rather than held resident. That needs one thing: substitute my own expert-MLP implementation for `SparseMoeBlock`'s, keeping the rest of the vendored model — and critically its weight keys — untouched.

Because none of the above is subclassable, I reimplemented instead. That is ~445 lines of `StreamingSwitchGLU` plus a `ResidentSwitchGLU` and a `ResidentMoeBlock`, and the large majority of it re-derives what `SwitchGLU` already does correctly: the three `@ModuleInfo(key:)` weight bindings, the gather/scatter sort, the quantized-vs-dense dispatch. Every upstream fix to `SwitchGLU` is a fix I have to notice and hand-port. Being able to subclass would delete most of that and keep me on your code path.

## The change

Mechanical access widening, no behaviour change:

- `Qwen35Language` → `public`
- `GatedDeltaNet`, `SparseMoeBlock`, `DecoderLayer`, `Model` → `open` (dropping `final`)
- `SwitchGLU`, `SwitchLinear` → `open`
- the three forward methods a subclass would actually want to override — `SwitchGLU.callAsFunction`, `SwitchLinear.callAsFunction`, `SparseMoeBlock.callAsFunction` → `open`

`Module` is already `open` (`MLXNN/Module.swift:97`), so nothing needed to change on your side of the dependency.

### One cascade worth flagging

`SparseMoeBlock` conforms to `UnaryLayer`, which is a public protocol, so once the class became `open` the compiler required its `callAsFunction` to be public too:

```
error: method 'callAsFunction' must be declared public because it matches
       a requirement in public protocol 'UnaryLayer'
```

I made it `open` rather than `public`, since overriding that method is precisely the point of the change. Mentioning it because it is the one place where "just widen the class" was not sufficient, and it is not obvious from reading.

## Deliberately not included

The internal notes this came from bundled a second, unrelated change: porting the merged GDN decode conv-fold from the MLXLLM copy of this model to the MLXVLM twin. That is a real optimization with its own bitwise test, but it is a behaviour change and it belongs in its own PR against its own review. I left it out to keep this one purely mechanical. Happy to send it separately if you want it.

I also left `FusedGateUpSwitchGLU` alone — it is `public` like the others, but I have no use case for subclassing it and did not want to widen API surface speculatively. Say the word if you would rather these move together.

## Evidence

- `swift build --target MLXVLM` → clean, exit 0.
- Full `swift test` → **296 tests, 0 failures**, unchanged.
- `swift-format` applied with the repo's `.swift-format`.
- Line citations verified against `main` at `3d292ec`.

Access control is the whole diff — 9 insertions, 9 deletions — so "no behaviour change" is checkable by reading it.

### DIFF
--- a/Libraries/MLXLMCommon/SwitchLayers.swift
+++ b/Libraries/MLXLMCommon/SwitchLayers.swift
@@ -39,7 +39,7 @@ public func scatterUnsort(x: MLXArray, invOrder: MLXArray, shape: [Int]? = nil) 
 
 // MARK: - SwitchGLU
 
-public class SwitchGLU: Module {
+open class SwitchGLU: Module {
     @ModuleInfo(key: "gate_proj") var gateProj: SwitchLinear
     @ModuleInfo(key: "up_proj") var upProj: SwitchLinear
     @ModuleInfo(key: "down_proj") var downProj: SwitchLinear
@@ -95,7 +95,7 @@ public class SwitchGLU: Module {
         super.init()
     }
 
-    public func callAsFunction(_ x: MLXArray, _ indices: MLXArray) -> MLXArray {
+    open func callAsFunction(_ x: MLXArray, _ indices: MLXArray) -> MLXArray {
         var x = MLX.expandedDimensions(x, axes: [-2, -3])
 
         let doSort = indices.size >= 64
@@ -219,7 +219,7 @@ public class FusedGateUpSwitchGLU: Module {
 
 // MARK: - SwitchLinear
 
-public class SwitchLinear: Module, Quantizable {
+open class SwitchLinear: Module, Quantizable {
     @ModuleInfo(key: "weight") var weight: MLXArray
     @ModuleInfo(key: "bias") var bias: MLXArray?
 
@@ -262,7 +262,7 @@ public class SwitchLinear: Module, Quantizable {
         self._bias.wrappedValue = bias
     }
 
-    public func callAsFunction(
+    open func callAsFunction(
         _ x: MLXArray, _ indices: MLXArray, sortedIndices: Bool = false
     ) -> MLXArray {
         let weightT = self.weight.swappedAxes(-1, -2)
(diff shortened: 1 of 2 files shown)


davidkoski (changes requested):
The change generally looks good, but there are a few issues:

- types like `SwitchGLU` still have a `public` init -- maybe this should also be `open`?
- several types are `open` but have `internal` init: GatedDeltaNet, SparseMoeBlock, DecoderLayer, Model

The lack of init has two effects: you can't override the init (maybe fine) and you also won't be able to instantiate it (probably defeats the purpose of the PR).

aleroot:
Looking at this I think the intent is right. 
I’d open  `SwitchGLU`  /  `SwitchLinear`  /  `FusedGateUpSwitchGLU`  (inits + forwards) and the Qwen MoE/decoder construction chain. I’d leave  `GatedDeltaNet`   final  unless there’s a concrete override for it.

davidkoski:
Rebased for conflicts & adjusted public inits.  Note: I thought init required `open` but that is not for inits -- public is equivalent to open on an open class.

davidkoski (approved):
Change looks good.  Just waiting for CI.  Thanks!

PR #512, opened 2026-08-05
Surface rejected tool calls as generation events

aleroot:
## Proposed changes

Tool-call-shaped model output can be malformed, incomplete, contain invalid arguments, or reference a function that was not declared by the application. Previously, this output could be suppressed or surfaced as ordinary assistant text without an observable semantic result. Agent loops could therefore finish without receiving either a valid tool call or an actionable failure, and malformed protocol output could be persisted into conversation history.

This PR supersedes #486. It retains its centralized tool authorization and ordered output routing while adding a typed, observable rejection API.

A grouped API was considered:

```swift
case tool(ToolEvent)
```

with `ToolEvent` containing accepted and rejected outcomes. However, replacing the existing public `toolCall(ToolCall)` case would break every caller that constructs or matches `.toolCall`. Keeping both cases would not provide a transparent compatibility alias because Swift enum cases are distinct runtime values: emitting both could execute tools twice, while emitting only the new case would silently stop existing consumers from receiving calls.

The additive design:

```swift
case toolCall(ToolCall)
case rejectedToolCall(RejectedToolCall)
```

therefore provides the safest compatibility trade-off. Existing non-exhaustive `.toolCall` consumers continue to work, while exhaustive `Generation` switches must explicitly handle the new rejection outcome.

## Checklist


davidkoski (approved):
This looks good to me.  Thank you!

PR #514, opened 2026-08-05
Enforce and expose effective maxKVSize cache policy

aleroot:
## Proposed changes

`GenerateParameters.maxKVSize` can currently be silently ignored by models that provide custom `newCache(parameters:)` implementations. This is particularly problematic for hybrid architectures, where attention and state-space layers use different cache types.

A caller may request a bounded KV cache while the model continues allocating an unbounded attention cache, with no public API for detecting that the requested policy was not applied. This makes memory usage unpredictable during long conversations and forces downstream applications to inspect private cache implementation details.

This PR makes the requested and effective KV-cache policies explicit and reliable.

### What

- Enforces `maxKVSize` across the audited model-specific cache implementations.
- Applies the smaller of the architecture's sliding-window size and the requested `maxKVSize`.
- Handles full-attention, sliding-window, hybrid attention/state-space, and pure state-space models.
- Adds a typed `EffectiveCacheConfiguration` API that describes:
  - the requested maximum KV-cache size;
  - the effective policy for every model layer;
  - bounded and unbounded attention caches;
  - rotating/sliding-window caches;
  - state-space caches;
  - composite caches used by hybrid models;
  - whether the request was fully applied, partially applied, unsupported, or ignored.
- Exposes the effective policy through `LanguageModel`, `ModelContainer`, and `ChatSession`.
- Makes `ChatSession` report the realized cache configuration rather than allocating and inspecting probe caches.
- Rebuilds a structured conversation cache when `maxKVSize` changes, ensuring the new policy is actually applied without losing the conversation transcript.
- Preserves externally restored raw caches when their original prefix cannot be reconstructed, while still reporting their actual effective policy.
- Safely handles very small limits such as `maxKVSize = 1`.
- Adds documentation describing cache-policy implementation and inspection.

### Why

`maxKVSize` is a memory-management contract. Silently ignoring it can cause unexpectedly high memory pressure, especially during long-running conversations and on memory-constrained Apple Silicon devices.

The new API also lets applications distinguish between:

- a limit that was fully enforced;
- a limit applied only to the attention layers of a hybrid model;
- an architecture where token-window limits do not apply;
- and an implementation that failed to apply the requested policy.

This distinction is important because state-space caches do not grow per token in the same way as attention KV caches and should not be reported as bounded attention caches.

With this change, an App using the library can:

- rely on `maxKVSize` being enforced for supported attention layers;
- display accurate user-facing diagnostics about the active cache policy;
- distinguish fully bounded models from partially bounded hybrid models;
- avoid allocating probe caches or casting to internal cache implementations;
- safely change the cache limit between conversation turns;
- make better model-selection and memory-management decisions before long generations;
- avoid presenting a configured memory limit as active when the model cannot apply it.

## Checklist

### DIFF
--- a/Libraries/MLXLMCommon/CacheConfiguration.swift
+++ b/Libraries/MLXLMCommon/CacheConfiguration.swift
@@ -1,0 +1,364 @@
+// Copyright © 2026 Apple Inc.
+
+import Foundation
+
+// MARK: - Layer kinds
+
+/// How a single model layer retains state across generation steps.
+///
+/// This is the stable, descriptive counterpart to concrete ``KVCache`` instances.
+/// Obtain it through ``KVCacheStatus`` rather than inspecting or
+/// casting runtime cache objects.
+public enum CacheLayerKind: Sendable, Hashable {
+    /// Standard or implementation-defined attention KV.
+    ///
+    /// A `nil` `maxSize` is unbounded. A non-nil value describes a bounded
+    /// custom cache that is not a ``RotatingKVCache``.
+    case attention(maxSize: Int?)
+
+    /// Sliding / rotating attention window (``RotatingKVCache``).
+    ///
+    /// Used both for architecture-imposed sliding-window attention and for
+    /// caller-requested capacity limits on full-attention layers.
+    case rotatingAttention(maxSize: Int, keep: Int)
+
+    /// Fixed-size state-space / Mamba / GDN recurrent state (``MambaCache``).
+    ///
+    /// These caches are not token-windowed; a ``GenerateParameters/maxKVSize``
+    /// policy does not apply to them.
+    case stateSpace
+
+    /// Composite per-layer cache (``CacheList``), e.g. Mamba + attention.
+    case composite([CacheLayerKind])
+
+    /// Whether this kind stores attention keys/values that can be size-limited.
+    public var isAttention: Bool {
+        switch self {
+        case .attention, .rotatingAttention:
+            return true
+        case .composite(let children):
+            return children.contains(where: \.isAttention)
+        case .stateSpace:
+            return false
+        }
+    }
+
+    /// Whether this kind is a bounded attention window.
+    public var isBoundedAttention: Bool {
+        switch self {
+        case .attention(let maxSize):
+            return maxSize != nil
+        case .rotatingAttention:
+            return true
+        case .composite(let children):
+            return children.contains(where: \.isBoundedAttention)
+        case .stateSpace:
+            return false
+        }
+    }
+
+    /// Effective attention token budget, if any.
+    public var attentionMaxSize: Int? {
+        switch self {
+        case .attention(let maxSize):
+            return maxSize
+        case .rotatingAttention(let maxSize, _):
+            return maxSize
+        case .composite(let children):
+            return children.compactMap(\.attentionMaxSize).max()
+        case .stateSpace:
+            return nil
+        }
+    }
+
+    fileprivate var containsStateSpace: Bool {
+        switch self {
+        case .stateSpace:
+            return true
+        case .composite(let children):
+            return children.contains(where: \.containsStateSpace)
+        case .attention, .rotatingAttention:
+            return false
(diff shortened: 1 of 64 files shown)


Libraries/MLXLMCommon/CacheConfiguration.swift
```diff
@@ -0,0 +1,350 @@
...
+        }
+    }
+}
+
+// MARK: - Construction helpers
+
+private func checkedCacheSize(_ size: Int, name: StaticString) -> Int {
+    precondition(size > 0, "\(name) must be greater than zero")
+    return size
+}
```
davidkoski: Any concern that this might produce fatal errors?  It looks like the size typically comes from model config, e.g. `let slidingWindow = config.slidingWindow`.  Given the effort in #453 to make bad config an Error rather than a fatal, is that a concern here?  A negative window sounds pretty unusable anyway, so perhaps not.

The biggest issue is the signature at the top level is not throwing:

```swift
public func newCache(parameters: GenerateParameters? = nil) -> [KVCache]
```

but given where it is typically called from, I think we _could_ change it if needed.

Libraries/MLXLMCommon/CacheConfiguration.swift
```diff
@@ -0,0 +1,350 @@
...
+
+/// The smaller of the architecture window and legacy `maxKVSize`.
+///
+/// Typed cache capacity deliberately does not participate: model-native sliding
+/// windows retain their architecture-defined window under `KVCacheConfiguration`.
+private func effectiveSlidingWindow(parameters: GenerateParameters?, window: Int) -> Int {
+    let window = checkedCacheSize(window, name: "sliding window")
+    if let maxKVSize = parameters?.maxKVSize {
+        return min(window, checkedCacheSize(maxKVSize, name: "maxKVSize"))
```
davidkoski: This also exposes a potential for fatals on a programmer controlled property with no validation (or documentation).

This seems more reasonable for a fatal -- this would be a programmer error.  The model config is input, which probably ought not be able to trigger a fatal (though I am sure that (checkedCacheSize) is not the only path to do so).

davidkoski: I wonder if it would make sense / be safe to validate the GenerateParameters here:

```swift
    package func kvCachePlan() throws -> KVCachePlan {
```

then this code could safely assert that it is > 0 as it already ran through an error producing path.  If you got here, you are doing something strange.

Libraries/MLXLMCommon/ChatSession.swift
```diff
@@ -814,11 +834,42 @@ public final class ChatSession {
...
+                            restored.uncommittedTokens.removeAll()
+                            kvCache = KVCacheStorage(
+                                model.newCache(parameters: generateParameters),
+                                plan: kvCachePlan)
+                            draftKVCache = nil
+                            lmState = nil
+                            conversation = restored
+                        } else {
+                            // A restored raw cache has no transcript to rebuild from;
+                            // refuse to continue under a changed configuration.
```
davidkoski: Is this true?  Well, of course it is true, but I given the conditions in the `if` I think this covers the case where there _is_ a conversation but nothing in the plans changed, right?

davidkoski (changes requested):
Looks good!  See what you think about the error case -- given the recent work in turning some fatals into throws this might be worth fixing.

Libraries/MLXLMCommon/CacheConfiguration.swift
aleroot: Good catch. I made it throwing KVCacheConfigurationError

Libraries/MLXLMCommon/ChatSession.swift
aleroot: You are correct. Also covers an existing conversation when the realized plan is unchanged. I updated the comment to describe both cases.

davidkoski (approved):
I think this is ready to go -- just needs a rebase to fix a conflict.  Thank you!

PR #515, opened 2026-08-06
`ChatSession`: reuse the KV cache for append-only media turns

NivDvir:
## Proposed changes

A turn that adds a new image to an unchanged transcript extends the cached prefix exactly as a text turn does, but #472 excludes every media turn from reuse, so it re-prefills the whole transcript and re-runs the vision tower over every cached image. (The figure in the comments below shows it.)

- `PreparedInputSplitting`: an opt-in protocol for models that can split their prepared input at a token boundary, keeping only the suffix's media.
- `AppendOnlyMediaRule` in `PromptCacheReusePolicy`: fires when the turn carries new media, extends the cached prefix, and no speculative decoding is configured. `ChatSession` asks the model to split, verifies the returned tokens against the boundary, and downgrades to `.rebuild` otherwise. The trim path is untouched, so the #472 guard still holds.
- `Qwen25VL` conforms: its vision mask isolates each frame, so dropping a cached image leaves the remaining features unchanged -- `testQwen25VLAppendOnlyMediaSplitMatchesFullPrefill` checks warm split against cold prefill on a tiny random-weight model. `Qwen2VL` attends across images and does not conform; `testQwen2VLAppendOnlyMediaSplitDiverges` shows why.
- Refused, and rebuild as before: video, temporal grids, audio, precomputed position ids, non-uniform masks, a boundary inside a media block. No other model changes.

Overlaps #581 in the reuse switch; happy to rebase across it if that lands first. No speedup is claimed; cost is stated in prompt tokens and encoder passes only.

## Checklist

## AI usage

- AI usage disclosure: Yes, I used Claude Code for parts of this. I reviewed and understand all of it.

### DIFF
--- a/Libraries/MLXLMCommon/ChatSession.swift
+++ b/Libraries/MLXLMCommon/ChatSession.swift
@@ -1133,10 +1133,21 @@ public final class ChatSession {
                             // and downgrade to a rebuild before prefilling.
                             var mediaSuffixInput: LMInput?
                             if case .appendMediaSuffix(let suffixStart, _) = decision {
-                                mediaSuffixInput = (model as? PreparedInputSplitting)?
+                                let splitInput = (model as? PreparedInputSplitting)?
                                     .splitPreparedInput(
                                         preparedInput, droppingFirst: suffixStart)
-                                if mediaSuffixInput == nil {
+                                // Returning *a* value is not enough: the ledger below
+                                // advances to `representedTokens` on the strength of the
+                                // requested boundary, so a suffix carrying any other
+                                // tokens would be prefilled verbatim and leave the record
+                                // describing a cache that was never built. Accept only the
+                                // exact tokens the boundary names.
+                                if let splitInput,
+                                    splitInput.text.tokens.asArray(Int.self)
+                                        == Array(promptTokenIds[suffixStart...])
+                                {
+                                    mediaSuffixInput = splitInput
+                                } else {
                                     decision = .rebuild
                                 }
                             }
--- a/Libraries/MLXLMCommon/ChatSession.swift
+++ b/Libraries/MLXLMCommon/ChatSession.swift
@@ -1091,7 +1091,8 @@ public final class ChatSession {
                                 previousGenerationUncommittedTokens:
                                     currentConversation.uncommittedTokens,
                                 structuredToolCallCount: structuredToolCallCount,
-                                usesSpeculativeDecoding: speculativeDecoding != nil)
+                                usesSpeculativeDecoding: speculativeDecoding != nil,
+                                canSplitPreparedMedia: model is PreparedInputSplitting)
                             let cacheState = PromptCacheState(
                                 cachedTokens: cachedTokenIds,
                                 processedTokenCount: kvCache.processedTokenCount,
@@ -1125,6 +1126,21 @@ public final class ChatSession {
                                 }
                             }
 
+                            // Splitting a prepared input is the other decision that
+                            // can fail while being applied: only the model can carve
+                            // a media-carrying suffix, and it declines any boundary
+                            // it cannot prove equivalent to a cold prefill. Verify
+                            // and downgrade to a rebuild before prefilling.
+                            var mediaSuffixInput: LMInput?
+                            if case .appendMediaSuffix(let suffixStart, _) = decision {
+                                mediaSuffixInput = (model as? PreparedInputSplitting)?
+                                    .splitPreparedInput(
+                                        preparedInput, droppingFirst: suffixStart)
+                                if mediaSuffixInput == nil {
+                                    decision = .rebuild
+                                }
+                            }
+
                             switch decision {
                             case .prefillAll:
                                 break
@@ -1141,6 +1157,13 @@ public final class ChatSession {
                                 // cache and use it alone for this continuation.
                                 draftKVCache = nil
                                 requiresMainOnlyContinuation = true
+
+                            case .appendMediaSuffix:
+                                // A declined split was downgraded to `.rebuild`
+                                // above, so this is always populated here.
+                                if let mediaSuffixInput {
+                                    input = mediaSuffixInput
+                                }
 
                             case .trimToCommonPrefix(let commonPrefixLength, _):
                                 input = LMInput(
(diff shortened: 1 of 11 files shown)


NivDvir:
Rebased onto main — though the `PromptCacheReusePolicy` refactor replaced the inline block this patched, so it is a redesign rather than a rebase.

It now fits the new shape: an `appendMediaSuffix` decision case, an `AppendOnlyMediaRule`, and a `canSplitPreparedMedia` fact on `PromptCacheTurn` that keeps the policy free of MLX and model types. The split can be declined per input, so it is verified and downgraded to `.rebuild` before prefilling — the same treatment `trimToCommonPrefix` gets.

One deliberate difference from `ExtendCachedPrefixRule`: the media path does not require the absence of an attention mask. A VL processor pairs media with an all-ones mask, so that guard would disable the path for the inputs it exists to serve; the splitter rejects anything that is not all-ones.

Libraries/MLXVLM/Models/QwenVL.swift
```diff
@@ -294,4 +294,152 @@ public struct QwenVL {
...
+        for (index, id) in ids.enumerated() where id == padTokenId {
+            totalPadCount += 1
+            if index < prefixTokenCount { prefixPadCount += 1 }
+        }
+
+        var padCounts: [Int] = []
+        var rowCounts: [Int] = []
+        padCounts.reserveCapacity(frames.count)
+        rowCounts.reserveCapacity(frames.count)
+        for frame in frames {
```
aleroot, line 405: This accepts image frames with t > 1, although the safety analysis above excludes temporal grids because their vision rows are not independently reusable. I reproduced this with THW(2, 4, 4) followed by THW(1, 2, 4): the split incorrectly returns non-nil. Please require frame.t == 1 here and add a refusal test.

Libraries/MLXLMCommon/ChatSession.swift
```diff
@@ -1063,6 +1064,21 @@ public final class ChatSession {
                                 }
                             }
 
+                            // Splitting a prepared input is the other decision that
+                            // can fail while being applied: only the model can carve
+                            // a media-carrying suffix, and it declines any boundary
+                            // it cannot prove equivalent to a cold prefill. Verify
+                            // and downgrade to a rebuild before prefilling.
+                            var mediaSuffixInput: LMInput?
+                            if case .appendMediaSuffix(let suffixStart, _) = decision {
```
aleroot, line 1138: It only verifies that the splitter returned a value. A conformer could return the wrong token suffix, after which the cache ledger would still be advanced to representedTokens and become inconsistent with the actual cache. Please verify that the returned tokens equal promptTokenIds[suffixStart...]; otherwise fall back to .rebuild.

aleroot (commented):
The core approach looks sound, I would suggest to rebase against main again as there have been a few changes last few days and look at the comments, for the rest it seems ok.

NivDvir:
Context for the reviewer, from running this branch head (`3845c86`) on real weights
(Qwen3-VL-4B-4bit, 10-row grounding protocol, decision instrumented per turn):

Two coverage observations, both by design of the current diff, recorded here so the review
can weigh them deliberately rather than discover them later:

1. **Qwen3-VL is outside the conformance set** (only `Qwen25VL` conforms; `Qwen2VL` is
   deliberately excluded — its unmasked vision attention makes a split unsound, as the
   in-code comment explains). For Qwen3-VL the exclusion is only conservatism: its vision
   attention is block-diagonal over `cuSeqlens` with correct per-frame accumulation, so
   items are feature-isolated, and deepstack needs no split handling at all — it never
   appears in a prepared input; the model recomputes it from the pixels each forward. I
   have the conformance ready on a branch
   ([`feat/qwen3vl-prepared-input-splitting`](https://github.com/NivDvir/mlx-swift-lm/tree/feat/qwen3vl-prepared-input-splitting):
   extension + position-equivalence tests against Qwen3-VL's own `getRopeIndex` + an
   end-to-end split-continuation test for both conformers), validated on real weights:
   10/10 warm turns take `appendMediaSuffix`, grounding parseable on every row, cold
   outputs byte-identical.

2. **Text-only turns after a media turn still rebuild, for every VL model including the
   conforming one.** Observed turn facts: `newMedia=false preparedMedia=true mask=true
   extendsPrefix=true → rebuild`. The transcript's earlier image is re-prepared each turn
   with its all-ones mask; `ExtendCachedPrefixRule` refuses on the mask, and
   `AppendOnlyMediaRule` requires `carriesNewMedia=true`, so neither serves the shape —
   although a conforming model's splitter would carve a media-less suffix trivially.
   Fixable by letting `AppendOnlyMediaRule` (or a sibling rule) also accept
   `carriesNewMedia=false` when the model can split; not done in this diff.

Neither observation regresses anything (`rebuild` is today's behavior); they bound what the
diff speeds up. Happy to open the Qwen3-VL follow-up once this lands.

Libraries/MLXVLM/Models/QwenVL.swift
NivDvir: Fixed in 707bf93: `splitVisionPayload` now requires `t == 1` per frame, with a refusal test for a temporal grid carried in the image payload — your THW(2, 4, 4) → THW(1, 2, 4) reproduction now returns nil. The rebase is also done; head is on top of 7871b09.

Libraries/MLXLMCommon/ChatSession.swift
NivDvir: Fixed in 707bf93: the returned suffix is now compared against `promptTokenIds[suffixStart...]`, and any mismatch downgrades to `.rebuild` before the ledger advances — the same treatment a declined split gets.

NivDvir:
A picture of the change, for anyone reading the description cold. Nothing in the PR moved; this is the same before/after the text above describes, drawn.

What a media turn costs — before and after #515

Green is the prefix kept in the KV cache, orange is what the turn prefills. Before: a new screenshot re-prefills every earlier one and re-runs the vision tower over it. After: the cached prefix stays and only the suffix is prefilled — one encoder pass per turn at any conversation length, Qwen2.5-VL only, with every unsafe shape still rebuilding as listed above. The figures at the bottom are the ones already in the description.

aleroot:
@NivDvir nice picture 😉

NivDvir:
@aleroot rebased on main and squashed to one commit; added the vision-tower test you'd have wanted (`testQwen25VLAppendOnlyMediaSplitMatchesFullPrefill`, and the Qwen2-VL counterexample). Aware this overlaps #581 in the reuse switch -- happy to rebase across it if that lands first.

aleroot:
@NivDvir you have done a nice work here. If I was to decide I would probably merge this before #581 and then rebase #581 on top the new main ... Thanks for this, I think it is a quality pull request from my point of view.

davidkoski (approved):
Changes look good, thank you!

NivDvir:
Merged! 🐐 Thank you, @davidkoski.

PR #516, opened 2026-08-07
Speculate past the sliding window in MTP

### LINKED ISSUE #474: MLXVLM/Gemma4Assistant.swift:288: Precondition failed: sliding KV length 515 exceeds slidingWindow 512
Trying to get Gemma4 + MTP working

Unless I'm missing something with the supported prompt size, perhaps I am loading the model + drafter incorrectly?

```
// MLXService.swift

import Foundation
import HuggingFace
import MLX
import MLXHuggingFace
import MLXLLM
import MLXLMCommon
import MLXVLM
import Tokenizers

extension LLMRegistry {
    static public let gemma4_e4b_it_bf16 = ModelConfiguration(
        id: "mlx-community/gemma-4-e4b-it-bf16",
        defaultPrompt: "What is the difference between a fruit and a vegetable?",
        extraEOSTokens: ["<turn|>"]
    )
}

@Observable
class MLXService {
    static let availableModels: [LMModel] = [
        LMModel(
            name: "gemma4:E4B",
            configuration: LLMRegistry.gemma4_e4b_it_4bit,
            type: .vlm
        ),
    ]

    private static let drafterConfigs: [String: ModelConfiguration] = [
        "gemma4:E4B": ModelConfiguration(
            id: "mlx-community/gemma-4-E4B-it-assistant-bf16"
        ),
    ]

    private let modelCache = NSCache<NSString, ModelContainer>()
    private let drafterCache = NSCache<NSString, MTPDrafterContainer>()

    @MainActor
    private(set) var modelDownloadProgress: Progress?

    init() {
        Task {
            print("registering gemma4 assistant")
            await Gemma4AssistantRegistration.register()
            print("registration done")
        }
    }

    private func load(model: LMModel) async throws -> ModelContainer {
        Memory.cacheLimit = 20 * 1024 * 1024

        if let container = modelCache.object(forKey: model.name as NSString) {
            print("target cache hit: \(model.name)")
            return container
        }

        print("loading target: \(model.name)")

        let factory: ModelFactory =
            switch model.type {
            case .llm: LLMModelFactory.shared
            case .vlm: VLMModelFactory.shared
            }

        let downloader = #hubDownloader()
        let loader = #huggingFaceTokenizerLoader()

        let container = try await factory.loadContainer(
            from: downloader,
            using: loader,
            configuration: model.configuration
        ) { progress in
            Task { @MainActor in
                self.modelDownloadProgress = progress
                let pct = Int(progress.fractionCompleted * 100)
                print("target download \(model.name): \(pct)%")
            }
        }

        modelCache.setObject(container, forKey: model.name as NSString)
        print("target loaded: \(model.name)")
        return container
    }

    private func loadDrafter(for model: LMModel) async throws -> any MTPDrafterModel {
        guard let config = Self.drafterConfigs[model.name] else {
            throw NSError(
                domain: "MLXService",
                code: 1,
                userInfo: [
                    NSLocalizedDescriptionKey:
                        "No MTP drafter config for \(model.name)"
                ]
            )
        }

        if let cached = drafterCache.object(forKey: model.name as NSString) {
            print("drafter cache hit: \(model.name)")
            return await cached.perform { $0.model }
        }

        print("loading drafter: \(config.id)")

        let downloader = #hubDownloader()
        let container = try await MTPDrafterModelFactory.shared.loadContainer(
            from: downloader,
            using: #huggingFaceTokenizerLoader(),
            configuration: config
        ) { progress in
            Task { @MainActor in
                self.modelDownloadProgress = progress
                let pct = Int(progress.fractionCompleted * 100)
                print("drafter download: \(pct)%")
            }
        }

        drafterCache.setObject(container, forKey: model.name as NSString)
        print("drafter loaded: \(model.name)")
        return await container.perform { $0.model }
    }

    func generate(messages: [ChatMessage], model: LMModel) async throws
        -> AsyncStream<Generation>
    {
        let modelContainer = try await load(model: model)

        let mtpDrafter: (any MTPDrafterModel)?
        do {
            mtpDrafter = try await loadDrafter(for: model)
            print("using MTP speculative decoding")
        } catch {
            print("drafter failed, falling back to normal generation: \(error)")
            mtpDrafter = nil
        }

        let chat = messages.map { message -> Chat.Message in
            let role: Chat.Message.Role =
                switch message.role {
                case .assistant: .assistant
                case .user: .user
                case .system: .system
                }
            let images = message.images.map { UserInput.Image.url($0) }
            let videos = message.videos.map { UserInput.Video.url($0) }

            let enrichedContent: String
            if message.id == -1 {
                enrichedContent =
                    "[room_id:\(message.roomId)] \(message.content)"
            } else {
                enrichedContent =
                    "[message_id:\(message.id)] [gossip_id:\(message.gossipId.map(String.init) ?? "")] [user_id:\(message.userId)] [timestamp:\(message.createdAt)] [reply_to:\(message.replyToIds)] \(message.content)"
            }

            return Chat.Message(
                role: role,
                content: enrichedContent,
                images: images,
                videos: videos
            )
        }

        let userInput = UserInput(
            chat: chat,
            processing: .init(resize: .init(width: 1024, height: 1024))
        )

        return try await modelContainer.perform { (context: ModelContext) in
            let lmInput = try await context.processor.prepare(input: userInput)
            let parameters = GenerateParameters(temperature: 0.7)

            if let drafter = mtpDrafter {
                return try MLXLMCommon.generate(
                    input: lmInput,
                    parameters: parameters,
                    context: context,
                    mtpDrafter: drafter,
                    blockSize: 4
                )
            } else {
                return try MLXLMCommon.generate(
                    input: lmInput,
                    parameters: parameters,
                    context: context
                )
            }
        }
    }
}

```

angelsbrood:
**Status:** Follow-up to [#506](https://github.com/ml-explore/mlx-swift-lm/pull/506), implementing the design agreed on [#474](https://github.com/ml-explore/mlx-swift-lm/issues/474#issuecomment-5209017252) and shaped by @aleroot's [review](https://github.com/ml-explore/mlx-swift-lm/issues/474#issuecomment-5212511569). Addresses the limitation #506 disclosed rather than the crash it fixed. Fixes #474. cc @davidkoski, @aleroot.

## Proposed changes

Removes the sliding-window bound on Gemma 4 MTP speculative decoding.

MTP's accept/reject step rewound the main KV cache with `trimPromptCache`, and a `RotatingKVCache` cannot be rewound once its ring has wrapped — the write a rewind would undo has already evicted the entries it would need to restore. #506 made that safe by standing the drafter down as soon as the stream's total context crossed the window. On E2B/E4B the window is 512, which a chat template and any real system prompt clear at prefill, so in practice the drafter stood down at `init` and never fired at all. That is what #474 reports.

The fix is the one described on-thread — append to an overlay, mix it with the live cache on use, commit whatever is accepted at the end — with @aleroot's placement: the transaction lives on `KVCacheStorage`, where the timeline lives, and each cache leaf chooses how it takes part. Rounds run against staged caches; commit appends only the bonus plus the accepted drafts through the normal write path and advances the timeline by exactly that much in the same operation; rejected drafts are dropped rather than written and undone. Nothing is trimmed on the MTP path any more, so the wrap stops mattering.

Speculation is an optimization, so the emitted stream is unchanged — what changes is only *where the speedup applies*, from "inside the window" to "everywhere".

## Design

`RotatingKVCache` gains one read-only `package` method. `logicalView(tail:)` returns the trailing `tail` entries in chronological order without touching `keys`, `values`, `idx` or `offset`. It is built from the same two steps the multi-token write path already uses — `temporalOrder(_:)` to linearize, then the front-trim that preserves the pinned `keep` prefix — so a view of length `n` holds exactly the entries a write that front-trimmed to `n` rows would have presented. When the ring is already chronological, which is where every multi-token write leaves it, both steps degrade to slices and nothing is copied.

The pinned prefix is a floor on the result, not just a splice point: a `tail` below `keep` still returns it, because those entries are not evictable and a view that dropped them would be a context the ring can never present. So the length is `max(min(tail, count), min(keep, count))` — plain `min(tail, count)` for the `keep == 0` case every sliding-window model uses, which is also the only case the staged round reaches.

**The transaction lives on `KVCacheStorage`.** That is where `processedTokenCount` lives, and `storage.trim` already pairs cache mutation with timeline rollback atomically; a commit that moved leaf K/V without advancing the timeline by exactly the retained count in the same operation would recreate that split brain one level up. `beginRound(maximumPositions:)` returns a handle carrying the caches to run the model against, and the handle deliberately has no `commit` of its own — commit is `storage.commit(round, retaining:)`, so the two cannot be separated or reordered. The transaction is entirely `package`; the only public API this PR adds is one `LMOutput.Key`, described below.

**Each leaf picks its own implementation** behind `KVCacheRoundStrategy`, with selection fail-closed. An append-only cache takes a provisional tail and a cursor commit — deliberately cheaper than staging, since a global layer's presentation is its whole history. A rotating cache stages K/V beside the ring. Anything else is refused *before* a single entry is touched, and the caller falls back to unstaged decoding. Append-only leaves are admitted on `isTrimmable(after: maximumPositions)` rather than `isTrimmable`: the entry that matters is one that is trimmable now and stops being trimmable *during* the round, which the narrower question cannot see. That restores the predictive check #506 introduced.

The part worth reviewing closely is that the rotating strategy does not reconstruct what the cache would have done. It presents `logicalView(tail: maxSize - 1)` followed by the staged rows, and it delegates `makeMask` to the live cache, whose offset has not moved. So the model sees the same arrays in the same order under the same mask it would have seen with no round at all, and correctness rests on the existing write path rather than on this type imitating it. The `maxSize - 1` bound is forced rather than chosen: it is both `updateConcat`'s front-trim target and `makeMask`'s `cappedOffset`, so any other value desynchronizes the mask from the presentation — with the bound set to `maxSize` the parity test fails and attention then broadcast-errors on the shape mismatch.

**The emitted shared-K/V state is reconciled by the same commit.** Otherwise the cache can be right while the next `draftBlock` still attends over a rejected tail. Doing it exactly needs to know which cache entry each emitted tuple came from — a sliding layer's snapshot is bounded by its ring and a global layer's is not, and at the crossing the two are indistinguishable by length. The target already has that mapping in scope where its capture loop runs, so it reports it alongside the tuples under a new `LMOutput.Key`; a snapshot that arrives without one is refused rather than reconciled against a guessed bound. Reconciliation drops the rejected tail first and clamps the head second, and runs after every emitting forward, prefill included — and a refusal stands the drafter down at whichever of those it happens at, prefill included.

That key is `public`, like the three MTP state keys beside it. It is part of the contract a target implements rather than an implementation detail of this one: emitting shared K/V without it now means having the snapshot refused, so a target outside this package could not take part at all if it were `package`.

The shape is Ollama's, which ships it as the default Gemma 4 MTP path in its MLX runner: [ollama/ollama#15980](https://github.com/ollama/ollama/pull/15980), `x/mlxrunner/cache/cache.go` — `Speculation`, `BeginSpeculation`, `speculativeKVCache`, `speculativeRotatingKVCache`, `RotatingKVCache.View`/`logicalTail` — MIT. Design port with credit, not a code translation: the implementation here is derived from this repository's own write path and mask construction. One point of agreement is worth stating because it was reached independently — their `View` is `logicalTail(maxSize - 1)`, and the same bound falls out here of `updateConcat`'s front-trim and `makeMask`'s `cappedOffset`, which is why it is forced rather than chosen.

### Answering the round-larger-than-the-window question

Not reachable at a sane block size — a round is `blockSize` positions against a 512- or 1024-entry window — but the API takes an arbitrary `blockSize`, so it is now clamped to `window + 1` rather than trapping. Wider rounds still decode correctly: masks are position-relative, the staged view is clamped, and commit rotates. What degrades is the meaning of the statistics, because the deepest drafts in such a round attend over less context than the verifier gave them. The clamp exists for that reason, and there is a test that a round at `window + 1` still matches greedy exactly.

## Behavior notes

**`passthroughReason` no longer reports either sliding-window string.** Both retire. Anything watching for them will now see `nil`, because the drafter no longer stands down for that reason — which is the point, but it is a silent change for a consumer keying off the string rather than off the behavior. The remaining triggers are a target that stops emitting drafter state (mid-stream KV quantization), a cache that cannot stage a round, and a target that emits shared K/V it does not say the source of.

**Speculation can now stand down before it starts, not only mid-stream.** That last trigger is reachable from `prepare`, so a target with an incomplete emit hook reports a `passthroughReason` from token zero rather than only after the first round. The stream is unaffected — `next()` already drained the prefill bonus ahead of the passthrough branch for exactly this case — but a consumer that reads `passthroughReason` as "speculation stopped partway" should read it as "speculation stopped, or never started".

**Speculation no longer commits anything the consumer did not see.** A round commits its accepted prefix before that prefix is drained, so a stop token partway through used to leave positions in the cache that were never emitted — and `finalizeGeneration`'s plain trim is all-or-nothing over the array, so past a sliding window one wrapped ring made it return zero for *every* leaf, including the trimmable global ones. The commit now leaves behind what it takes to undo itself, and only where a trim would not do. Capture is cheap (a full-range slice reuses the underlying node, which is how `RotatingKVCache.copy()` already works) and measured peak GPU memory is unchanged.

What it leaves behind names cache *slots*, not the objects the commit happened to hold, and the rewind resolves each leaf from the storage and requires its trim to come back exact. So replacing an entry between the commit and the rewind — which is what dynamic compression does — invalidates the record, and the rewind falls back to the plain trim rather than restoring into a leaf the storage no longer owns. Past a wrap that fallback can take nothing back, which is the pre-existing behavior; it is the right side to fail on.

## Telemetry

@aleroot — with the window trigger gone, acceptance statistics describe the whole stream again, so the prefix skew #506 had to disclose (`roundCount > 0` while `emittedTokenCount` kept climbing through passthrough) no longer arises on this path. A round's transients are bounded and small: at most `blockSize` staged rows per sliding layer for the duration of one round, plus at most one linearization of the ring, which in the steady state is a slice rather than a gather because every multi-token write leaves the ring chronological. The restore point is the one addition worth naming to you — it holds a wrapped ring's pre-commit contents for a single round, so its analytic worst case is one extra K+V ring per sliding layer, and measured peak GPU memory came out identical with and without it. Nothing here needs gate awareness.

## What is untouched

The live cache write paths, plain generation, prefill, the classic `SpeculativeTokenIterator`, `#506`'s `isTrimmable(after:)` API and its passthrough machinery, and all public API other than the one added `LMOutput.Key` and the behavior notes above. The classic path's identical init-only trimmability defect remains the separate follow-up #506 promised; folding it in here would mix two arguments.

One thing found while reading and deliberately not fixed: `gemma4AdjustAttentionMask` in MLXVLM suffix-trims an over-long array mask, while its MLXLLM twin in `Gemma4Text.swift` prefix-trims — opposite alignments for the same situation. MTP runs the MLXVLM path, so it is not reachable from this change, but it is a live trap for anyone who wires the text-only backbone to a windowed cache.

## Verification

E4B-it, 4-bit target and bf16 assistant, `blockSize` 4, temperature 0, on an M5 Pro / 64 GB. Each row is one A/B: the same harness, the same prompts, `main` at `c97539d` versus this branch. The reported figure is MTP tok/s over plain-decode tok/s **measured within the same run**, so machine drift cancels — absolute rates moved up to 15% between runs and should not be compared across rows.

| Shape | Total context | `main` | This branch |
|---|---|---|---|
| short — 31-token prompt, 200 generated | 231, inside the window | **1.39×** | **1.43×** |
| crossing — 266-token prompt, 600 generated | 866, crosses at ~250 generated | **0.84×**, stands down | **1.17×** |
| long — 266-token prompt, ~1500 generated | ~1770 | **0.83×**, stands down | **1.06×** |

The short row is the control: identical within noise, and acceptance is 0.342 on both arms, which is what output identity should look like from the outside.

The two rows that matter are the ones that cross the window. On `main` the drafter stands down there and `passthroughReason` reports the mid-stream wrap; on this branch it never does. Note that `main`'s crossing row is **slower than not using MTP at all** — 0.84× on an exactly-equal 600 generated tokens on both arms — because the stream pays for ~110 speculative rounds, stands down, and finishes without the speedup. I did not isolate why post-stand-down decoding lands below plain rather than level with it; it is behavior of the code being replaced, and it makes the case for replacing it rather than against.

Two caveats. The long row's arms generated different token counts (1501 versus 1398 here, 1532 versus 1398 on `main`) because greedy MTP and greedy plain decoding diverge on real weights — SDPA kernel ordering differs between a multi-token verify forward and a single-token step, and a near-tie argmax eventually flips. The synthetic suite is what proves algorithmic identity; this row cannot. And tok/s decays with context on a model with unbounded global-attention layers, so unequal lengths bias that row somewhat. The crossing row has exactly 600 generated tokens on all four arms and carries the argument on its own.

An earlier revision of this branch put the transaction in a free-standing type and clamped the shared K/V in the emit hook. Measuring the two back-to-back isolates what the current shape costs: throughput is neutral (ratios move within the 3–7% run-to-run band) and **peak GPU memory is identical at every stage**, so hosting the transaction on the storage and carrying a restore point are free. One expectation of mine did not survive contact with the numbers, so it is not claimed here: correcting the shared-K/V clamp order did *not* measurably raise acceptance (0.467 → 0.461 on the comparable row), because at a 512-entry window the old order cost the drafter at most three rows — 0.6% of its context. It is a correctness fix that would matter at a small window or a large block size, not a throughput win.

Unit suite: `Test run with 480 tests in 40 suites passed` (423 before this change), run at **each of the nine commits**, not just the tip. XCTest side is unchanged from a pristine `upstream/main` checkout on this machine — 13 pre-existing tolerance failures in the Qwen/Nanbeige warm-continuation oracles, same set before and after, none near this code. `pre-commit run --all-files` passes. `scripts/verify-docs.sh` exits 1 identically on this branch and on a clean `upstream/main` with the same build cache, so it is not attributable here.

The four review-fix commits are additive, and the numbers above are the tip's. They do not move the measurements: the view's floor is integer arithmetic on a path that already ran, the slot change touches only what happens after the last token, and the prefill refusal is a branch not taken by any target that emits a source map.

## Tests

Nothing in the repository drove a `RotatingKVCache` ring past its wrap before this, so the rotation in `updateInPlace`, the linearization in `temporalOrder`, and the windowed mask `makeMask` builds over a multi-token presentation were all uncovered — and they are what a staged round rests on. They are pinned here too.

`KVCacheTests` covers the view: chronology past a double wrap, tail clamping, non-mutation of the ring across reads, pinned-prefix splicing at `keep > 0` and the prefix floor at `tail` of zero, below `keep` and at `keep` (including a ring holding fewer entries than `keep` reserves), the property that `logicalView(tail: maxSize - 1)` plus new rows equals the live write path's presentation *and* its mask span, and post-wrap attention parity against a logical-order reference that never rotated.

`KVCacheRoundTests` (new) covers the transaction against real position-encoded K/V at tiny windows, where a permutation error changes the numbers: the live ring is bit-identical across a staged round and after a rollback; the presentation and mask match a cache that really wrote the same rows; attention through a staged round matches the logical-order reference past a wrap; and after accept-none, accept-all, partial prefixes across the wrap, a pinned prefix, an empty start, and rounds wider than the window, the live caches hold exactly what a cache fed only the accepted tokens holds. Also that every layer advances by exactly the accepted count, which is what `KVCacheStorage`'s processed-token timeline audits.

Nine of those are checks against the *other* implementation of the same idea rather than against hand-computed expectations, which is the correctness oracle @aleroot asked for: snapshot the storage with `KVCacheStorage.copy()` before a round, run the round, replay only the accepted tokens into the snapshot through the ordinary write path, and require the two to agree — down to `metaState`, so an `idx` or ring-rotation divergence that the logical view hides is still caught. The same oracle, run after a rewind against a replay of exactly the emitted tokens, is what pins the finalize fix. Removing the restore point makes all of them fail; forcing the rewind back onto the plain trim path fails those *and* the iterator-level check below.

Two more use that oracle to pin which *object* a rewind reaches: two sliding layers at different widths, driven to a timeline where one ring has wrapped and the other has not, so a single commit produces a restore point for the first and a plain trim record for the second; and a rewind attempted after an entry has been replaced, which has to fall back to the plain trim rather than restore into a leaf the storage no longer owns. The first traps on the alignment assert if the record names the round's staged wrapper instead of the slot — at unit and at iterator level — and the exact-match trim precondition catches that wrapper's no-op directly.

`MTPSpeculativeTokenIteratorTests` — the three stand-down tests #506 added are rewritten rather than deleted, since the behavior they pinned is the behavior being removed. They now pin that speculation survives a prompt longer than the window and a mid-stream crossing, and that the stream stays token-identical to a plain `TokenIterator` with prefill below, at, and above the window and generation several multiples past it. That last one needed a target mock scripted by absolute position rather than by forward-call index, because the old mock only stays in step with a greedy run when every draft is accepted — too weak to test an accept/reject path. Plus: the drafter receives exactly the window it can attend over — `min(timeline, window)`, not merely no more than it — across a long run, an unstageable cache is refused at `init`, one that stops being stageable mid-stream falls back stickily and still completes the stream, and the block-size clamp is exact.

And a target that emits shared K/V without saying where it came from stands the drafter down during `init` — once for each shape `prepare` can return, since each reconciles at a different site — with the stream still token-identical to plain greedy decoding, and without the re-prime forward that only exists to fetch state this stream has stopped having a use for.

Also at iterator level: after every `next()` the leaves and the timeline stay in step, and after a consumer stops partway through a round the timeline describes what the consumer actually saw. That bound is two-sided on purpose — the familiar `- 1` holds only when the last emitted token was a verifier sample, whose K/V is deliberately not cached in its own round, and stopping mid-buffer ends on an accepted draft instead.

Coverage boundary, stated in the test file so it is not overread: the iterator tests pin bookkeeping — offsets, spans, commit counts, token identity — not K/V content, because that mock scripts by position rather than by values. Content is `KVCacheRoundTests`' job.

## Ceilings worth naming

`KVCacheRoundStrategy` cannot be a requirement on `KVCache` itself: that protocol is public, so a `package` requirement cannot be added to it and a public one would be new API. Strategy selection is therefore a type switch, and a `KVCache` conformer from outside this package can never opt into staging — it gets write-through if it can prove it stays trimmable, and refusal otherwise. Fail-closed, but permanent until someone wants to lift it.

Hosting the transaction on `KVCacheStorage` fixes no *observed* bug today, because nothing shares the storage: the MTP iterator builds its own and it never escapes, and `ChatSession` has no MTP path. The value is structural — the timeline update cannot be forgotten or reordered — and it is what makes an MTP `ChatSession` path safe to add later. Worth saying plainly rather than overselling.

Nested `CacheList` topologies stay refused. `KVCacheTree.rewrite` mutates the live list in place, so supporting them means a transaction inside a transaction; no shipped model needs it.

Snapshot/replay is not built as a strategy, per @aleroot's framing — it is the fallback for a future leaf that cannot be staged, and it is already earning its keep as the correctness oracle.

Thanks again @sihayas for giving this a kick.

### DIFF
--- a/Libraries/MLXLMCommon/KVCache.swift
+++ b/Libraries/MLXLMCommon/KVCache.swift
@@ -602,6 +602,49 @@ public class RotatingKVCache: BaseKVCache, CustomDebugStringConvertible {
         }
     }
 
+    /// The trailing `tail` cache entries in chronological order, without mutating the cache.
+    ///
+    /// `update(keys:values:)` is the only other way to read a rotating cache's contents, and it
+    /// necessarily writes. This is the read-only counterpart: `keys`, `values`, `idx` and
+    /// `offset` are all left exactly as they were, so a caller can present the ring's history
+    /// alongside K/V it has not committed yet.
+    ///
+    /// The result is built by the same two steps the multi-token write path uses --
+    /// ``temporalOrder(_:)`` to linearize, then the front-trim that preserves the pinned `keep`
+    /// prefix -- so a view of length `n` holds exactly the entries a write that front-trimmed to
+    /// `n` rows would have presented. When the ring is already chronological (`idx` at the end of
+    /// the buffer, which is where every multi-token write leaves it) both steps degrade to
+    /// slices and nothing is copied.
+    ///
+    /// - Parameter tail: Maximum number of entries to return. Clamped to what the cache holds;
+    ///   a negative value returns an empty view.
+    /// - Returns: `(keys, values)` shaped `[B, kvHeads, min(tail, count), headDim]`, or `nil`
+    ///   before the first write.
+    public func logicalView(tail: Int) -> (MLXArray, MLXArray)? {
+        guard let keys = self.keys, let values = self.values else { return nil }
+
+        let orderedKeys = temporalOrder(keys)
+        let orderedValues = temporalOrder(values)
+
+        let available = orderedKeys.dim(2)
+        let bound = Swift.min(Swift.max(tail, 0), available)
+        let trimSize = available - bound
+        guard trimSize > 0 else { return (orderedKeys, orderedValues) }
+
+        // `keep == 0` is the sliding-window case (Gemma 3/3n/4, GPT-OSS, Exaone4): no pinned
+        // prefix to splice around, so the trailing window is one slice per array.
+        if keep == 0 {
+            return (
+                orderedKeys[.ellipsis, trimSize..., 0...],
+                orderedValues[.ellipsis, trimSize..., 0...]
+            )
+        }
+        return (
+            trim(trimSize: trimSize, orderedKeys),
+            trim(trimSize: trimSize, orderedValues)
+        )
+    }
+
     private func updateConcat(keys: MLXArray, values: MLXArray) -> (MLXArray, MLXArray) {
         if self.keys == nil {
             self.keys = keys
--- a/Libraries/MLXLMCommon/KVCache.swift
+++ b/Libraries/MLXLMCommon/KVCache.swift
@@ -616,18 +616,27 @@ public class RotatingKVCache: BaseKVCache, CustomDebugStringConvertible {
     /// the buffer, which is where every multi-token write leaves it) both steps degrade to
     /// slices and nothing is copied.
     ///
-    /// - Parameter tail: Maximum number of entries to return. Clamped to what the cache holds;
-    ///   a negative value returns an empty view.
-    /// - Returns: `(keys, values)` shaped `[B, kvHeads, min(tail, count), headDim]`, or `nil`
-    ///   before the first write.
-    public func logicalView(tail: Int) -> (MLXArray, MLXArray)? {
+    /// The pinned `keep` prefix is a floor, not just a splice point: a `tail` below it still
+    /// comes back, because those entries are not evictable and a view that dropped them would be
+    /// a context this ring can never present. With `keep == 0` -- every sliding-window model --
+    /// the floor is zero and the length is exactly `min(tail, count)`.
+    ///
+    /// - Parameter tail: Requested number of trailing entries. Clamped to what the cache holds;
+    ///   a negative value is read as zero.
+    /// - Returns: `(keys, values)` shaped `[B, kvHeads, n, headDim]` where
+    ///   `n == max(min(tail, count), min(keep, count))`, or `nil` before the first write.
+    package func logicalView(tail: Int) -> (MLXArray, MLXArray)? {
         guard let keys = self.keys, let values = self.values else { return nil }
 
         let orderedKeys = temporalOrder(keys)
         let orderedValues = temporalOrder(values)
 
         let available = orderedKeys.dim(2)
-        let bound = Swift.min(Swift.max(tail, 0), available)
+        // Raising the bound to the pinned prefix is what keeps the front-trim's second slice in
+        // range; stated here rather than left to slice clamping, since the length it produces is
+        // the documented contract.
+        let requested = Swift.min(Swift.max(tail, 0), available)
+        let bound = Swift.max(requested, Swift.min(keep, available))
         let trimSize = available - bound
         guard trimSize > 0 else { return (orderedKeys, orderedValues) }
 
--- a/Libraries/MLXLMCommon/KVCachePlan.swift
+++ b/Libraries/MLXLMCommon/KVCachePlan.swift
@@ -22,6 +22,7 @@ package struct KVCachePlan: Sendable, Equatable {
 
     package func validated(_ storage: KVCacheStorage) throws -> KVCacheStorage {
         precondition(storage.plan == self, "KVCacheStorage used with a different plan")
+        precondition(!storage.roundIsOpen, "validated(_:) ran inside a staged round")
         storage.cache = try validated(storage.cache)
         return storage
     }
@@ -37,6 +38,7 @@ package struct KVCachePlan: Sendable, Equatable {
     /// unsupported state, later decode steps only compare the completed plan.
     package func apply(to storage: KVCacheStorage) {
         precondition(storage.plan == self, "KVCacheStorage used with a different plan")
+        precondition(!storage.roundIsOpen, "apply(to:) ran inside a staged round")
         guard !storage.isApplicationTerminal else { return }
         guard let configuration else {
             storage.isApplicationTerminal = true
@@ -63,6 +65,7 @@ package struct KVCachePlan: Sendable, Equatable {
         to storage: KVCacheStorage
     ) throws -> KVCacheApplicationResult? {
         precondition(storage.plan == self, "KVCacheStorage used with a different plan")
+        precondition(!storage.roundIsOpen, "applyAndValidate(to:) ran inside a staged round")
         guard !storage.isApplicationTerminal else { return nil }
         guard let configuration else {
             storage.isApplicationTerminal = true
@@ -103,10 +106,23 @@ package struct KVCachePlan: Sendable, Equatable {
 /// permit unsynchronized mutation of both the cache and its progress state.
 package final class KVCacheStorage {
     package var cache: [KVCache] {
-        didSet { isApplicationTerminal = false }
+        didSet {
+            precondition(
+                !roundIsOpen,
+                "the realized cache was replaced while a staged round was open")
+            isApplicationTerminal = false
+        }
     }
     package let plan: KVCachePlan
     package fileprivate(set) var isApplicationTerminal = false
+
+    /// Whether a staged round is presently writing provisionally against these entries.
+    ///
+    /// A round hands the model substitute caches and defers part of its writes, so the entries
+    /// and the timeline are deliberately out of step until it commits. Operations that assume
+    /// they are in step -- replacing entries, trimming, snapshotting -- are refused while one is
+    /// open rather than silently reading a half-written position.
+    package internal(set) var roundIsOpen = false
     /// Authoritative logical position represented by this model cache.
     ///
     /// Individual cache entries retain their native offsets and metadata, but
@@ -142,6 +158,7 @@ package final class KVCacheStorage {
     @discardableResult
     package func trim(_ count: Int) -> Int {
         precondition(count >= 0, "Trim count cannot be negative")
+        precondition(!roundIsOpen, "cannot trim while a staged round is open")
         let trimmed = trimPromptCache(cache, numTokens: count)
         precondition(
             trimmed <= processedTokenCount,
@@ -152,6 +169,7 @@ package final class KVCacheStorage {
 
     /// Create an independent snapshot while preserving plan and progress.
     package func copy() -> KVCacheStorage {
+        precondition(!roundIsOpen, "cannot snapshot while a staged round is open")
         let copy = KVCacheStorage(
             cache.map { $0.copy() },
             plan: plan,
--- a/Libraries/MLXLMCommon/KVCachePlan.swift
+++ b/Libraries/MLXLMCommon/KVCachePlan.swift
@@ -116,13 +116,15 @@ package final class KVCacheStorage {
     package let plan: KVCachePlan
     package fileprivate(set) var isApplicationTerminal = false
 
+    private var openRound: KVCacheRound?
+
     /// Whether a staged round is presently writing provisionally against these entries.
     ///
     /// A round hands the model substitute caches and defers part of its writes, so the entries
     /// and the timeline are deliberately out of step until it commits. Operations that assume
     /// they are in step -- replacing entries, trimming, snapshotting -- are refused while one is
     /// open rather than silently reading a half-written position.
-    package internal(set) var roundIsOpen = false
+    package var roundIsOpen: Bool { openRound != nil }
     /// Authoritative logical position represented by this model cache.
     ///
     /// Individual cache entries retain their native offsets and metadata, but
@@ -165,6 +167,84 @@ package final class KVCacheStorage {
             "Cache trimmed beyond its processed-token timeline")
         processedTokenCount -= trimmed
         return trimmed
+    }
+
+    /// Open a staged round over these entries, or `nil` when one of them cannot take part.
+    ///
+    /// Every leaf is classified before anything is constructed, so a refusal leaves the entries
+    /// exactly as they were and the caller can fall back to unstaged decoding. Nested
+    /// ``CacheList`` topologies are refused too: they expand to more leaves than array slots, and
+    /// the caches handed back have to line up with the live array positionally.
+    ///
+    /// - Parameter maximumPositions: the widest write the round will make. Append-only leaves are
+    ///   admitted only if they can still be rewound after absorbing it.
+    package func beginRound(maximumPositions: Int) -> KVCacheRound? {
+        precondition(!roundIsOpen, "a staged round is already open on this storage")
+
+        let leaves = KVCacheTree.leaves(in: cache)
+        guard leaves.count == cache.count, leaves.allSatisfy({ $0.path.count == 1 }) else {
+            return nil
+        }
+
+        var strategies = [any KVCacheRoundStrategy]()
+        strategies.reserveCapacity(leaves.count)
+        for leaf in leaves {
+            guard
+                let strategy = KVCacheRoundStrategyFactory.make(
+                    for: leaf, maximumPositions: maximumPositions)
+            else { return nil }
+            strategies.append(strategy)
+        }
+
+        let round = KVCacheRound(strategies: strategies, maximumPositions: maximumPositions)
+        openRound = round
+        return round
+    }
+
+    /// Keep the first `retaining` positions the round wrote and drop the rest, advancing the
+    /// timeline by exactly that much.
+    ///
+    /// Leaf commit and timeline advance are one operation on purpose. A commit that moved leaf
+    /// K/V without moving the timeline -- or the reverse -- would recreate, one level up, the
+    /// split brain this storage exists to prevent.
+    @discardableResult
+    package func commit(_ round: KVCacheRound, retaining: Int) -> KVCacheRoundCommit {
+        precondition(openRound === round, "committing a round that is not open on this storage")
+        let written = round.writtenPositions
+        precondition(
+            retaining >= 0 && retaining <= written,
+            "cannot retain \(retaining) of \(written) written positions")
+
+        for strategy in round.strategies {
+            strategy.commit(retaining: retaining)
+        }
+        processedTokenCount += retaining
+        openRound = nil
+
+        assert(
+            nativeAttentionOffsetsAreAligned,
+            "committing a staged round left the leaves and the timeline out of step")
+
+        return KVCacheRoundCommit(
+            committedPositions: retaining,
+            discardedPositions: written - retaining,
+            emittedLengths: round.strategies.map(\.emittedLength))
+    }
+
+    /// Drop everything the round wrote.
+    @discardableResult
+    package func rollback(_ round: KVCacheRound) -> KVCacheRoundCommit {
+        commit(round, retaining: 0)
+    }
+
+    /// How much of the emitted sequence the given leaf can still describe, outside a round.
+    ///
+    /// The whole stream for a global layer; the trailing window for a sliding one. This is the
+    /// same quantity ``KVCacheRoundCommit/emittedLengths`` reports, for callers that need it
+    /// without a round in hand -- prefill, for instance.
+    package func emittedLength(forLeaf index: Int) -> Int {
+        guard cache.indices.contains(index) else { return processedTokenCount }
+        return Swift.min(processedTokenCount, cache[index].maxSize ?? .max)
     }
 
     /// Create an independent snapshot while preserving plan and progress.
--- a/Libraries/MLXLMCommon/KVCachePlan.swift
+++ b/Libraries/MLXLMCommon/KVCachePlan.swift
@@ -110,6 +110,9 @@ package final class KVCacheStorage {
             precondition(
                 !roundIsOpen,
                 "the realized cache was replaced while a staged round was open")
+            // A rewind record names slots, and a replacement changes what those slots hold --
+            // so it invalidates the record rather than being restored into blindly.
+            lastRound = nil
             isApplicationTerminal = false
         }
     }
@@ -118,11 +121,19 @@ package final class KVCacheStorage {
 
     private var openRound: KVCacheRound?
 
+    /// How to take one leaf's share of a commit back, addressed by slot rather than by object.
+    ///
+    /// `point == nil` means a plain `trim(_:)` undoes it exactly, which is true for every leaf
+    /// that had not wrapped when the commit ran.
+    private struct LeafRewind {
+        let slot: Int
+        let point: (any KVCacheLeafRestorePoint)?
+    }
+
     /// What it would take to undo the most recent commit, kept only until the next write.
     private struct CompletedRound {
         let committedPositions: Int
-        let restorePoints: [any KVCacheLeafRestorePoint]
-        let trimmableLeaves: [KVCache]
+        let leaves: [LeafRewind]
     }
     private var lastRound: CompletedRound?
 
@@ -201,7 +212,7 @@ package final class KVCacheStorage {
         for leaf in leaves {
             guard
                 let strategy = KVCacheRoundStrategyFactory.make(
-                    for: leaf, maximumPositions: maximumPositions)
+                    for: leaf, slot: leaf.path[0], maximumPositions: maximumPositions)
             else { return nil }
             strategies.append(strategy)
         }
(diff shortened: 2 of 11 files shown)


Libraries/MLXLMCommon/KVCache.swift
```diff
@@ -602,6 +602,49 @@ public class RotatingKVCache: BaseKVCache, CustomDebugStringConvertible {
...
+    /// prefix -- so a view of length `n` holds exactly the entries a write that front-trimmed to
+    /// `n` rows would have presented. When the ring is already chronological (`idx` at the end of
+    /// the buffer, which is where every multi-token write leaves it) both steps degrade to
+    /// slices and nothing is copied.
+    ///
+    /// - Parameter tail: Maximum number of entries to return. Clamped to what the cache holds;
+    ///   a negative value returns an empty view.
+    /// - Returns: `(keys, values)` shaped `[B, kvHeads, min(tail, count), headDim]`, or `nil`
+    ///   before the first write.
+    public func logicalView(tail: Int) -> (MLXArray, MLXArray)? {
```
aleroot: With `keep > 0`, `tail < keep` can return more than `tail`, so the documented `min(tail, count)` shape is not really true. I would keep this `package` and add explicit coverage for `tail == 0`, `tail < keep`, and `tail == keep`.

Libraries/MLXLMCommon/KVCachePlan.swift
```diff
@@ -136,22 +162,155 @@ package final class KVCacheStorage {
...
+
+        // Built before the commit, because that is the last moment the pre-commit contents of a
+        // wrapped ring still exist.
+        var restorePoints = [any KVCacheLeafRestorePoint]()
+        var trimmableLeaves = [KVCache]()
+        for strategy in round.strategies {
+            if let point = strategy.makeRestorePoint(retaining: retaining) {
+                restorePoints.append(point)
+            } else {
+                trimmableLeaves.append(strategy.presented)
```
aleroot: I think this should retain a stable cache slot/path, not `strategy.presented`.

For a rotating strategy, `presented` is the temporary staged wrapper, whose `trim` is intentionally a no-op. The same identity issue appears if dynamic compression replaces an append-only leaf after commit: `rewindLastRound` would trim the old object rather than the cache currently owned by `storage`.

That can move the timeline and restored leaves backwards while leaving another live leaf ahead. I would resolve the current leaf from `KVCacheStorage` at rewind time and require the trim result to match exactly.

Libraries/MLXLMCommon/MTPSpeculativeTokenIterator.swift
```diff
@@ -183,6 +205,12 @@ public struct MTPSpeculativeTokenIterator: TokenIteratorProtocol {
             processor?.didSample(token: token)
             y = .init(tokens: token)
             mainState = result.state
+            // Prefill emits too, and a multi-token prefill presents a sliding layer more than its
+            // window just as a verify pass does. Nothing was rejected here, so only the head
+            // clamp applies.
+            reconcileSharedKVState(
```
aleroot: The `false` result is ignored at all three prefill/prime sites.

Missing source metadata therefore traps later through an `assert` in debug, but in release the assertion disappears and the drafter can receive an unreconciled sliding snapshot. I would switch to passthrough immediately here...

Libraries/MLXLMCommon/MTPDrafterModel.swift
```diff
@@ -131,6 +131,18 @@ public let mtpSharedKVStatesKey =
...
+/// Which cache entry each ``mtpSharedKVStatesKey`` tuple was read from, keyed the same way.
+///
+/// A speculative round keeps only part of what its verify pass wrote, so the emitted snapshot has
+/// to be reconciled against the cache afterwards. That reconciliation is exact only if the
+/// consumer knows the entry behind each tuple: a sliding layer's snapshot is bounded by its ring
+/// and a global layer's is not, and at the crossing the two are indistinguishable by length.
+///
+/// `package` rather than `public`: unlike the keys above, both the writer and the reader live
+/// inside this package, so this adds no API.
+package let mtpSharedKVSourceIndicesKey =
```
aleroot: This is now required for safe MTP reconciliation, but it is `package` while the other target/iterator state keys are public...

Libraries/MLXLMCommon/KVCache.swift
angelsbrood: Right about the contract. Plainly, the behavior was already correct: both new tests passed before any code change, because MLX clamps the out-of-range slice, so the front-trim already returned the pinned prefix. Wrong doc and over-wide surface, not a silent bug. The documented length is now `max(min(tail, count), min(keep, count))` (a pinned entry isn't evictable, so a view that dropped it would be a context the ring can never present) with the bound raised explicitly rather than arriving through a slicing detail (a no-op refactor, so not claimed as mutation-covered). `keep == 0`, every sliding-window model, is untouched: exactly `min(tail, count)`. Narrowed to `package` as asked: both call sites are `RotatingStagedKVCache`. Tests pin the floor at `tail` 0/1/2 with `keep: 2` (plus 3 as the first tail that selects anything) and the floor bounded by what the cache holds when fewer entries exist than `keep` reserves. Also unreachable from a staged round: its tail is `min(live.offset, maxSize − 1)`, which below `keep` takes the `trimSize == 0` early return. 5e8077a.

Libraries/MLXLMCommon/KVCachePlan.swift
angelsbrood: Live bug, conceded directly: a rotating leaf below its window went into the trimmable set as `presented` (the `RotatingStagedKVCache`, whose `trim(_:)` is `{ 0 }`) so the rewind moved the timeline and left the ring where it was. Reachability is narrow: the record only exists when another leaf produced a restore point, so it takes two sliding layers at different widths with the timeline between them. Shipped Gemma 4's uniform window doesn't hit it, `makePromptCacheWithLayerCount` does. Both of your scenarios land in the same place and one move fixes both: a strategy carries its slot, records name slots rather than objects, rewind resolves each leaf from the storage, every slot is validated before any is touched, and each trim must return exactly what was asked. `RotatingRestorePoint` holds no cache reference at all now. One divergence from your suggestion: `cache`'s `didSet` drops the record, so a rewind after a compression swap degrades to the plain all-or-nothing trim rather than trimming the swapped-in leaf. Past a wrap that can take back nothing, the pre-existing behavior, but it beats restoring a snapshot into a leaf the storage no longer owns; happy to revisit if you'd rather trim the current object. Evidence: fail-first SIGTRAPs at the alignment assert (`KVCachePlan.swift:295`); after the fix, removing the `didSet` invalidation or restoring the object capture each turns it red again, the latter on the new precondition, `entry 1 took back 0 of 2 positions`. Tests: rewind reaches the live ring when only some layers wrapped; rewind refused after replacement; and a mixed-width early stop through `finalizeGeneration` that asserts its own premise so it can't pass vacuously. 04f7a8b.

Libraries/MLXLMCommon/MTPSpeculativeTokenIterator.swift
angelsbrood: Got it— all three sites now switch to the same sticky passthrough, same reason, at the point of refusal. Your debug/release framing is exactly what the fail-first run showed: it trapped at the round's freshness assert (`MTPSpeculativeTokenIterator.swift:340`), the very assert that vanishes in release. Reachable from `init`, and `next()` already drains the prefill bonus ahead of the passthrough branch, so the emitted stream is unchanged— pinned token-identical to a plain `TokenIterator` for each of `prepare`'s three emitting shapes. One call you might want to push on: the `.logits` re-prime forward is skipped once prefill's own state has been refused. It exists only to obtain drafter state, which the stream no longer has a use for, so refusal takes the same single-bonus path the already-had-state branch takes, saving the forward. `finalizeGeneration` still discards the result and now says why in a comment. The body note reads "stopped, or never started." 4f79fa6.

Libraries/MLXLMCommon/MTPDrafterModel.swift
angelsbrood: Read as make it `public`— done, and if that misreads you it's one line back. The `package` justification ("writer and reader both live here") was true of the current writer but the wrong test: the other three keys are public because they describe what a target must emit to take part, and this one now belongs to that contract. A target emitting shared K/V without it has its snapshot refused, so left `package`, no target outside this package could satisfy the contract at all. One public constant added, the type was already public, nothing else moved. e2733fd.

davidkoski:
General comment, not a merge blocker: each token will call `temporalOrder(_ array: MLXArray) -> MLXArray` twice, once for `presentation()` and once for `updateConcat()`.  Maybe this is cheap or fixed cost (since it is a rotating cache).  Might be worth measuring to see if doing that once would be helpful or not.

Libraries/MLXLMCommon/RotatingStagedKVCache.swift
```diff
@@ -0,0 +1,136 @@
...
+            return (stagedKeys, stagedValues)
+        }
+        return (
+            concatenated([view.0, stagedKeys], axis: 2),
+            concatenated([view.1, stagedValues], axis: 2)
+        )
+    }
+
+    private func emptyPresentation() -> (MLXArray, MLXArray) {
+        let empty = MLXArray.zeros([1, 1, 0, 1])
```
davidkoski, line 90: Maybe this should consider a dtype -- if we reach this code (and I think we cannot) we could introduce float32s by accident.  Maybe we can get the dtype from live?

Not a PR blocker, but maybe a safety issue for a followup.

davidkoski (approved):
Change looks good to me, thank you!

angelsbrood:
@davidkoski I measured this after merge. The experiment retained the chronological live prefix already produced by presentation() and reused it for the following multi-position commit, avoiding the second temporalOrder(_:) pass that updateConcat() would otherwise perform. Zero- and one-position commits were unchanged. The existing accepted-only replay and wrap oracles all passed.

Five alternating baseline/experiment pairs on the stable 600-token window-crossing case produced these MTP/plain speedups:
- 1.1827× → 1.2198×
- 1.1882× → 1.2342×
- 1.1790× → 1.2124×
- 1.2425× → 1.2114×
- 1.1864× → 1.1806×

The median paired improvement was +2.833%, with the experiment ahead in 3/5 pairs. That missed the thresholds I set beforehand of at least 3% median improvement and 4/5 favorable pairs, so I’m leaving the implementation unchanged.

Correctness and memory were exact across every run: identical output hashes, 751 proposed / 346 accepted draft tokens, and 4,878 MB peak GPU memory. There may be a small throughput signal here, but I didn't think it was robust enough to justify another PR. Happy to revisit at your discretion, thanks for flagging it.

PR #520, opened 2026-08-09
Add single-dispatch TurboFlash for short contexts

aleroot:
## Proposed changes

This PR improves generation speed for models using a TurboQuant-compressed KV cache.

For short contexts of up to 128 tokens, TurboFlash attention now completes its work in a single GPU operation instead of using separate calculation and reduction operations. Longer contexts and unsupported configurations continue using the existing implementation.

The optimized path is used while the combined prompt, conversation history, and generated response contain no more than 128 tokens.

Typical examples include:

- A short question followed by a concise answer
- Rewriting a short email or message
- Classification or information extraction from a few sentences
- Brief summaries
- Short code completions
- The first turn of a lightweight chat session

If generation grows beyond 128 tokens, it automatically switches to the existing implementation. Therefore, the beginning of a longer response may still benefit, while long documents, large system prompts, and established multi-turn conversations generally will not.

### Benchmark

Tested on an Apple M2 with 16 GB unified memory using:

- `mlx-community/Qwen3-4B-4bit`
- TurboQuant K4/V4 cache
- 30 prompt tokens
- 96 generated tokens
- 2 warm-up runs and 7 measured runs

| Metric | Before | After | Improvement |
|---|---:|---:|---:|
| Median generation speed | 21.341 tok/s | 22.491 tok/s | **+5.39%** |
| Mean generation speed | 21.313 tok/s | 22.535 tok/s | **+5.74%** |

This is approximately one additional generated token per second on the tested machine. Prefill and longer-context performance are not affected by this optimization.

## Checklist

  - The documented manual formatter was run on every changed file using the repository’s `.swift-format` configuration and produced no diff.
- [x] I have updated the necessary documentation (if needed) — no documentation changes were required

### DIFF
--- a/Libraries/MLXLMCommon/TurboQuantKernels.swift
+++ b/Libraries/MLXLMCommon/TurboQuantKernels.swift
@@ -541,6 +541,123 @@ enum TurboQuantMetalKernels {
         }
         """
 
+    /// TurboFlashAttention bounded decode path: score, online softmax, value
+    /// accumulation, normalization, and optional inverse value rotation in one
+    /// dispatch. One SIMD group owns one query row and walks the cache in token
+    /// order. This avoids allocating pass-1 partials and launching pass 2 when
+    /// the context is short enough that cross-block parallelism is unnecessary.
+    ///
+    /// Grid: (32, totalQueries, 1)
+    /// Threadgroup: (32, 1, 1)
+    ///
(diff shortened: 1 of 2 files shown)


davidkoski (approved):
Changes look good to me.  Thank you!

PR #521, opened 2026-08-10
Add protocol-aware thinking budget enforcement

aleroot:
## Proposed changes

Reasoning models can spend most or all of their output limit “thinking,” leaving too little space for the answer shown to the user. This PR adds a configurable thinking budget that limits reasoning while reserving space for the final response.

The implementation is model-aware and does not assume that every model using `<think>` tags supports the same early-stop behavior like mlx-vlm does...

## Checklist

### DIFF
--- a/Libraries/MLXLLM/Models/Nanbeige.swift
+++ b/Libraries/MLXLLM/Models/Nanbeige.swift
@@ -358,7 +358,7 @@ extension NanbeigeModel {
     // the chat template also supports JSON.
     public var toolCallFormat: ToolCallFormat? { .xmlFunction }
 
-    // <think>/</think>, toggled via `enable_thinking` (template default true),
-    // same contract as the Qwen3 family.
-    public var reasoningConfig: ReasoningConfig? { .thinkTagsWithEnableThinking }
+    // Nanbeige shares Qwen's thinking tags and tool-call boundary, but not the
+    // original Qwen3 family's documented hard-budget transition.
+    public var reasoningConfig: ReasoningConfig? { QwenReasoningProtocol.tagged }
 }
--- a/Libraries/MLXVLM/Models/Qwen3VL.swift
+++ b/Libraries/MLXVLM/Models/Qwen3VL.swift
@@ -1917,5 +1917,7 @@ public struct Qwen3VLMessageGenerator: MessageGenerator {
 // MARK: - Chat conventions
 
 extension Qwen3VL {
-    public var reasoningConfig: ReasoningConfig? { .thinkTagsWithEnableThinking }
+    // Qwen3-VL shares Qwen's tags and tool-call boundary, but does not declare
+    // the original Qwen3 family's model-specific hard-budget transition.
+    public var reasoningConfig: ReasoningConfig? { QwenReasoningProtocol.tagged }
 }
--- a/Libraries/MLXVLM/Models/Qwen3VLMoE.swift
+++ b/Libraries/MLXVLM/Models/Qwen3VLMoE.swift
@@ -831,3 +831,11 @@ public final class Qwen3VLMoE: Module, VLMModel, KVCacheDimensionProvider {
         return false
     }
 }
+
+// MARK: - Chat conventions
+
+extension Qwen3VLMoE {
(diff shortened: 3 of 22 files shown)


Libraries/MLXLLM/Models/Nanbeige.swift
```diff
@@ -358,7 +358,7 @@ extension NanbeigeModel {
     // the chat template also supports JSON.
     public var toolCallFormat: ToolCallFormat? { .xmlFunction }
 
-    // <think>/</think>, toggled via `enable_thinking` (template default true),
-    // same contract as the Qwen3 family.
+    // <think>/</think>, toggled via `enable_thinking` (template default true).
+    // Nanbeige shares Qwen's surface syntax, not its documented budget protocol.
     public var reasoningConfig: ReasoningConfig? { .thinkTagsWithEnableThinking }
```
davidkoski: I am not sure I understand the impact of this.  Would this benefit from a custom `nanbeige` config that added the `implicitEndDelimiters: ["<tool_call>"],` that seems to be the big difference?

Would nanbeige be susceptible to the same thing that would have qwen models start a tool call without closing the thinking?

Libraries/MLXVLM/Models/Qwen3VL.swift
```diff
@@ -2037,5 +1917,5 @@ public struct Qwen3VLMessageGenerator: MessageGenerator {
 // MARK: - Chat conventions
 
 extension Qwen3VL {
-    public var reasoningConfig: ReasoningConfig? { .thinkTagsWithEnableThinking }
+    public var reasoningConfig: ReasoningConfig? { QwenReasoningProtocol.tagged }
```
davidkoski, line 1922: Should this match Qwen3 (LLM), which is `.qwen3`?

Libraries/MLXVLM/Models/Qwen3VLMoE.swift
```diff
@@ -831,3 +831,9 @@ public final class Qwen3VLMoE: Module, VLMModel, KVCacheDimensionProvider {
         return false
     }
 }
+
+// MARK: - Chat conventions
+
+extension Qwen3VLMoE {
+    public var reasoningConfig: ReasoningConfig? { QwenReasoningProtocol.tagged }
```
davidkoski, line 840: Should this match Qwen3MoE (LLM), which is `.qwen3`?

davidkoski (changes requested):
Looks good, just a couple questions then I think we are ready to merge!

Libraries/MLXLLM/Models/Nanbeige.swift
aleroot: Good catch. Nanbeige uses the same <think> and <tool_call> surface protocol, so an unclosed thinking span would otherwise have the same parsing problem if <tool_call> appeared before </think>. I switched it to QwenReasoningProtocol.tagged, which adds implicitEndDelimiters: ["<tool_call>"] while deliberately keeping budgetTransition nil.

Libraries/MLXVLM/Models/Qwen3VL.swift
aleroot: I don’t think this should use .qwen3. That preset does more than declare the tags: it opts the model into the original Qwen3 family’s published, model-specific hard-budget transition. Qwen3-VL documents the <think> and <tool_call> protocol, but not that early-stop transition, so .tagged is intentional: it preserves the tags and tool-call boundary without claiming an unvalidated forced transition.

Libraries/MLXVLM/Models/Qwen3VLMoE.swift
aleroot: Same reasoning as the dense Qwen3-VL model.

Libraries/MLXVLM/Models/Qwen3VL.swift
davidkoski: Got it, thanks for the clarification!

davidkoski (approved):
Changes look good, thanks!

PR #523, opened 2026-08-11
Add Muse-Glimmer, a 30B agentic multimodal model, with ATEM tool calling

CharlieTLe:
## Proposed changes

Adds `muse_glimmer` support, porting [`mlx_vlm.models.muse_glimmer`](https://github.com/Blaizzy/mlx-vlm/tree/main/mlx_vlm/models/muse_glimmer). [`meta-models/Muse-Glimmer-30B`](https://huggingface.co/meta-models/Muse-Glimmer-30B) (Apache-2.0) is a 52-layer dense text transformer paired with a 50-layer native-resolution vision tower. Nothing in the Swift MLX ecosystem could load it before this; `mlx-community/Muse-Glimmer-30B-4bit` runs at ~26 tok/s decode and ~19 GB resident on an M4 Max.

**Scope note.** This started as a self-contained model port and now also carries tool-calling support, so it touches shared infrastructure. Reviewers may want to read it as two parts:

1. **The model** (commits 1–2) — one new file `Libraries/MLXVLM/Models/MuseGlimmer.swift`, registry entries, tests.
2. **Agentic protocol support** (commit 3, by @aleroot, originally #524 — which supersedes it) — the ATEM tool-call format plus the Onyx response protocol, which modifies `MLXLanguageModel`, `Evaluate`, `ChatSession`, `TokenStreamDecoder`, `ToolCallFormat` and the Harmony parsers.

### What's non-standard in the model

Most of the review risk is here, so each item is commented at its definition:

- **NoPE on global layers.** `layer_rope_theta` is 0 on the full-attention layers and 500000 on the sliding ones, so RoPE is skipped entirely every fourth layer.
- **Gated attention output.** `output * sigmoid(gate_proj(x))`, where the gate reads the *layer input*, not the attention result. Its checkpoint key collides with `mlp.gate_proj` at a different path.
- **Additional QK scale.** Queries are multiplied by `qk_scale_factor` (3.87) *on top of* the usual `1/sqrt(head_dim)`, after a parameterless QK norm — effective scale `3.87/sqrt(128)`, not a replacement for the softmax scale.
- **Norms scaled by `1 + weight`** (the checkpoint stores a scale centered at zero), with two eps values per block: `rms_norm_eps` on the two "pre" norms, the much smaller `post_norm_eps` on the two "post" ones. The final `model.norm` is a plain `RMSNorm` instead.
- **Three parameterless norms** — `embed_norm`, `qk_norm`, `perception_emb_norm` — have no learnable scale and so are absent from the checkpoint. They're modelled as parameter-free modules so the module tree matches the weight keys exactly.
- **Window attention without a mask.** The ViT permutes tokens into contiguous windows and runs chunked SDPA, so no banded mask is materialized (at 4096 merged tokens that mask would have been ~268M elements). The window side reuses `pos_emb_height` (32), which looks accidental upstream but is what every published checkpoint was trained with, so it's reproduced rather than "fixed".
- **Learned position table resampled** onto the actual patch grid as a four-corner bilinear gather, including the validity masks that *zero* out-of-range corners rather than clamping them.
- **2D RoPE and the 2×2 patch merge are hand-rolled.** `MLXFast.RoPE(freqs:)` can't express two positions per token applied to different dimension blocks. The merge places the channel axis *before* the two merge axes, so each merged vector is channel-major.
- **ViT head dim is 96**, off the fused SDPA path, so q/k/v are padded to 128 and sliced back. Zero padding contributes nothing to any dot product, so this is exact rather than approximate.
- **The patch embedder is a `Linear`, not a convolution.** The processor emits pre-flattened patches in raster order, laid out `(temporal, channels, ph, pw)` — both differ from `QwenVL.patchify`, and both matter because the vision tower does its own merge grouping downstream.

The 4-bit checkpoint is only *partially* quantized — the entire vision tower stays bf16 — which needs no special handling, since `Load.swift` already keys quantization off the presence of `.scales`.

### Image token budget

The image token budget dominates time-to-first-token, and not only through ViT cost: **each merged token becomes a prompt token that must be prefilled through all 52 text layers**, and prefill runs at ~200 tok/s. Measured on an M4 Max with a 2400×2400 image:

| image budget | prompt tokens | time to first token |
|---|---|---|
| 4096 (checkpoint default) | 4,158 | 26.6 s |
| 1024 | 1,086 | 7.2 s |

Text-only control runs confirm the split — 965 text tokens take 4.3 s and 4,205 take 19 s with no image at all, so the text stack is the dominant term rather than the ViT.

As first written the budget was unreachable from a caller: `preprocess` overwrote `UserInput.Processing.resize` and the processor configuration had no public initializer. Two ways to lower it now:

- `UserInput.Processing.maxPixels`, the existing model-agnostic per-call resize budget. One merged token covers `patch_size * merge_size` squared = 784 pixels, so the budgets convert directly.
- `MuseGlimmerProcessorConfiguration(imageProcessor: .init(maxImageTokens:))` for a request-independent default.

A caller-supplied `resize` now bounds the source dimensions instead of being discarded; it cannot set the output size directly, since that has to land on the patch/merge grid, and it will not upscale. Defaults are unchanged — with neither override the same image yields an identical grid and token count.

### Known caveat: reasoning is not observable from app code

`TextToolTokenLoopHandler` deliberately drops reasoning from the public `Generation` stream, and `TokenStreamDecoder` — the documented alternative — is `package`-level, so an app outside this library cannot consume reasoning at all. Since this model reasons at length before answering, the effect is a silent window *after* prefill: on a 1034-token image prompt, prefill completes at 5.8 s but the first visible token arrives at 10.5 s.

The upside is that the protocol's control tokens are now stripped and the answer text arrives clean. Discussion in the thread; a `Generation.reasoning(String)` case or a public `TokenStreamDecoder` would resolve it, and I'm happy to do that work either here or as a follow-up.

### Verification

A wrong permutation in a model like this produces fluent, confident, wrong text, so correctness was established against the Python reference rather than by reading output.

Unit tests (no weights, no downloads) compare every permutation and index computation to reference values as **exact integers**: patch merge, patchify layout, one-indexed `(w, h)` position ids, window index and boundaries, the position-table gather, and `smart_resize` across 20 aspect ratios. Plus config decode/derive, the sliding/full cache mix, text-only and image prefill shapes, the image-budget behaviour, and the agentic-protocol and Onyx cache-continuation suites. 463 tests pass.

On the real 4-bit weights, replaying the reference's own token ids and post-resize `pixel_values`:

- final-position logits: cosine **0.9999937**, max abs 0.31 on softcapped ±20 logits
- **top-1 and top-5 agree exactly**
- greedy decode matches token for token to step 34, then splits on `,` vs `.` — a genuine near-tie, top-2 margin 0.125 against a 0.3125 worst-case logit disagreement, so that step is undetermined in bf16

This was re-run after merging the tool-calling commit and is unchanged, so nothing in `Evaluate.swift` or `TokenStreamDecoder` perturbs the model path.

Per-position cosine looks loose deep in the ViT (activations reach ~1e3 before `ln_post`, which then amplifies residual differences). That's a property of the model, not the port — running the *reference* in fp32 against its own bf16 output diverges further:

| stage | reference fp32 vs bf16 | this port vs bf16 |
|---|---|---|
| `vision_layer_3` | min cos 0.999699 | 0.999979 |
| `vision_layer_49` | min cos 0.641, 328 positions < 0.999 | 0.791, 154 |
| `vision_ln_post` | min cos 0.100, 1177 positions < 0.999 | 0.046, 530 |

Separately, the real processor path produces a token stream identical to the reference's apart from a duplicate BOS that the reference harness introduced by re-tokenizing an already-rendered template — so `<|patch|>` expansion, the system message and `strftime_now` date rendering all match exactly.

### Out of scope

Video input is rejected explicitly rather than silently mishandled. Multi-image prompts are not implemented.

## Checklist


aleroot:
Great implementation, thanks.

Please see this https://github.com/ml-explore/mlx-swift-lm/pull/524 and evaluate the merging into your branch, muse glimmer is an agentic model so it has to use tools 😉

CharlieTLe:
Thanks — this is good work, and I've merged it into `muse-glimmer` so #523 now carries both the model and your tool support. Your commit keeps its authorship. I think #524 can be closed as superseded, but say the word if you'd rather sequence them the other way.

Two small notes on the merge itself: there was one additive conflict in `MuseGlimmerTests.swift` (we'd both appended suites — all seven are present now), your tests needed `@testable import MLXLMCommon` which I carried over, and I ran `swift-format` since the pre-commit box was unchecked. 463 tests pass, and I re-ran the numerical parity check against the Python reference afterwards — cosine 0.9999937 on the final-position logits, top-1/top-5 exact — so nothing in `Evaluate.swift` or `TokenStreamDecoder` perturbed the model path.

Also worth calling out that your `guard input.videos.isEmpty` is a genuine bug fix in my code, not just plumbing. I'd declared video out of scope but never rejected it, so a rendered `<|video|>` placeholder could reach the model with no corresponding features. Thanks for catching that.

**One thing I'd like your view on: reasoning is now unobservable to app code.**

`TextToolTokenLoopHandler` drops it from the public stream:

```swift
case .reasoning:
    // The public Generation stream intentionally exposes only response
    // text and tool calls.
    return .more
```

I understand the intent, and the payoff is real — the control tokens are stripped and the answer text arrives clean, which is much nicer than what I had. But `TokenStreamDecoder`, which the comment points to as the alternative, is `package`-level, so an app outside this library has no way to consume reasoning at all.

For a model that reasons at length before answering, the user-visible effect is a second silent window after prefill. Measured on an M4 Max, 1034-token prompt with an image:

| | before | after |
|---|---|---|
| prefill completes | 5.9 s | 5.8 s |
| first visible token | 5.9 s | 10.5 s |

So ~4.7 s where the app has nothing to show and no way to know why. I worked around it in my demo app by naming the phase, but that's inferring state rather than observing it.

Would you be open to either surfacing reasoning as a `Generation` case (`.reasoning(String)`, which clients can ignore), or making `TokenStreamDecoder` public? Either would let a UI show thinking-in-progress. Happy to do that work if you'd prefer.

**And a possible gap worth a second look:** `response_template` sets `"repeats": true` on `tool_calls`, so the model may emit several `<atem:invoke>` blocks. Reading `ATEMToolCallParser.parse`, it extracts one invoke and then requires the remainder to be whitespace-only, which I *think* means a second invoke makes the whole parse return `nil` and the tool call is dropped silently. I may well be misreading it, and `ToolCall` being singular may make this a framework constraint rather than something to fix here — but if it is the behaviour, failing closed on a valid multi-call response seems worth a test either way.

aleroot:
@CharlieTLe thanks, I am closing my other PR. 
I have just opened another one on your branch with some fixes you may want to include. I agree reasoning should be exposed as `Generation.reasoning(String)` rather than making the decoder public.

davidkoski (approved):
This looks good to me!  Thank you!

PR #526, opened 2026-08-11
Remove the untyped staged-cache fallback

angelsbrood:
Follow-up to #516 and [David's review comment](https://github.com/ml-explore/mlx-swift-lm/pull/516#discussion_r3760049169).

`RotatingStagedKVCache.update(keys:values:)` always stages its inputs before constructing a presentation. Its private `presentation()` helper nevertheless had a no-staged-input branch that manufactured empty `float32` arrays. That branch was unreachable, but it left a latent dtype hazard if the helper's call order ever changed.

This removes the fallback by making the invariant structural:

- pass the staged K/V arrays into `presentation` as non-optionals;
- delete `emptyPresentation()`; and
- pin the first empty-live-cache presentation to its staged `bfloat16` dtype and shape.

The new test passes before the refactor because the fallback was already unreachable. This is hazard removal, not a reproduced behavior bug.

No public API changes.

## Testing

- Full `MLXLMTests.xctest` bundle: all 494 Swift Testing tests pass; no new XCTest failures relative to the fresh `upstream/main@2af378b` baseline.
- `pre-commit run --all-files`

### DIFF
--- a/Libraries/MLXLMCommon/RotatingStagedKVCache.swift
+++ b/Libraries/MLXLMCommon/RotatingStagedKVCache.swift
@@ -38,9 +38,11 @@ final class RotatingStagedKVCache: KVCache {
     var maxSize: Int? { live.maxSize }
 
     func update(keys: MLXArray, values: MLXArray) -> (MLXArray, MLXArray) {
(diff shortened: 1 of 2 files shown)


davidkoski (approved):
Nice fix!  Thank you!

PR #527, opened 2026-08-11
Fix nested tool grammar schema test

angelsbrood:
## Proposed changes

`SchemaConverter.encodeToolCallingGrammar(tools:)` currently emits per-tool dispatch in both the wrapped and bare structural-tag arms, while the existing nested-definition test still reads the former direct schema shape.

This test-only correction follows both current dispatch arms, verifies that `Traveler` and `Passport` remain in each local `$defs`, and checks that every nested `$ref` stays rooted at `#/$defs/`.

There is no public API or runtime behavior change.

## Validation

- Current `main` focused baseline: 11/12 `ToolCallingSchemaTests`, with only the stale structural lookup failing
- This branch focused suite: 12/12 `ToolCallingSchemaTests`
- Complete `xcodebuild test` comparison: zero branch-only failing cases or crash suites; this branch matches the current `main` failure set, while the baseline additionally failed `TurboQuantIntegrationTests.testRawKeyModeBFloat16MatchesReference()`
- `pre-commit run --all-files`: passed
- `scripts/verify-docs.sh`: passed for every documentation target

## Checklist

- [x] I have run `pre-commit run --all-files`
- [x] I have added tests that prove the correction is effective
- [x] Documentation changes are not needed for this test-only correction

### DIFF
--- a/Tests/MLXFoundationModelsTests/ToolCallingSchemaTests.swift
+++ b/Tests/MLXFoundationModelsTests/ToolCallingSchemaTests.swift
@@ -262,7 +262,7 @@ struct ToolCallingSchemaTests {
     }
 
     @Test
-    func grammarBuilderHoistsNestedDefsInBothArms() throws {
+    func grammarBuilderRetainsNestedDefsInBothPerToolArms() throws {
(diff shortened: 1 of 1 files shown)


davidkoski (approved):
Looks good, thank you!

PR #529, opened 2026-08-13
Add Qwen 3.5 JSON tool-call fallback

aleroot:
## Proposed changes

Qwen 3.5 is instructed (via its chat template) to format tool/function calls using a specific XML-like syntax, e.g.:

```
<tool_call><function=get_weather><parameter=city>Tokyo</parameter></function></tool_call>
```

In practice, the model sometimes ignores this instruction and emits an older JSON-style format instead (the same style used by earlier Qwen and Llama models):

```
<tool_call>{"name": "get_weather", "arguments": {"city": "Tokyo"}}</tool_call>
```

Previously, this library only understood the XML format for Qwen 3.5, so whenever the model slipped into the JSON style, the tool call would fail to parse and the app would treat it as plain text instead of actually calling the tool.

This PR teaches the Qwen 3.5 parser to recognize and accept **both** formats. It still requires the call to be wrapped in the standard `<tool_call>...</tool_call>` tags, and it only accepts tools that were actually declared to the model, so this does not loosen validation — it just widens what "a correctly formatted call" is allowed to look like. Malformed or mixed-up payloads (e.g. XML tags with JSON content jammed inside, or trailing garbage after the call) are still rejected rather than guessed at.

## Checklist

### DIFF
--- a/Libraries/IntegrationTestHelpers/IntegrationTestHelpers.swift
+++ b/Libraries/IntegrationTestHelpers/IntegrationTestHelpers.swift
@@ -979,8 +979,8 @@ public enum ToolCallTests {
     public static func qwen35FormatAutoDetection(container: LLModelContainer) async throws {
         let config = await container.configuration
         try check(
-            config.toolCallFormat == ToolCallFormat.xmlFunction,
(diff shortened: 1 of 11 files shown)


davidkoski (approved):
Change looks good, thank you!

PR #530, opened 2026-08-13
Expose Gemma 4 encoder surface at @_spi(GemmaEncoder)

xocialize:
Mirrors #387 (`6608a35`, Gemma 3) for Gemma 4: exposes the decoder stack at `@_spi(GemmaEncoder)` scope so client code can implement an **encoder-style all-hidden-states tap** — using Gemma 4 as a frozen text encoder whose per-layer states condition a downstream model, rather than only consuming the final hidden state.

This is the same use case #387 enabled for Gemma 3. The SPI opt-in keeps the surface off the advertised public API of `MLXLLM` while making it reachable from another module.

### Exposed

| Symbol | Was |
|---|---|
| `Gemma4TextConfiguration.hiddenSize` / `.numHiddenLayers` | internal |
| `Gemma4DecoderLayer` + its `callAsFunction` | `private` |
| `Gemma4TextModelInner` + `embedTokens` / `layers` / `norm` / `embedScale` / `config` | `private` |
| `Gemma4TextModel.model` | `fileprivate` |

### Two things that differ from the Gemma 3 change

**1. `Gemma4SharedKVState` moves `package` → `public`** (`MLXLMCommon`). It appears in the signature of `Gemma4DecoderLayer.callAsFunction`, and an SPI method can't be called from outside the package while a type in its signature is package-scoped. Clients that don't use KV sharing pass `nil` and ignore the returned value. If you'd prefer to avoid widening it, the alternative is an SPI overload that hides the shared-KV parameter/return entirely — happy to switch if that's more to taste.

**2. Four `Gemma4DecoderLayer` stored properties become explicitly `fileprivate`** (`selfAttn`, `mlp`, `router`, `experts`). Making the class public changes its members' default access from internal to something that can't reference a private type. `fileprivate` is the narrowest fix; nothing widens that didn't have to.

### Sufficiency test

`Gemma4EncoderAccessTests` follows `Gemma3EncoderAccessTests`: it implements the tap **in test code using only the exposed surface**, importing `MLXLLM`/`MLXLMCommon` *without* `@testable` and opting in via `@_spi(GemmaEncoder)` — so it fails to compile if the exposure is insufficient, which is the property actually worth testing.

It checks state count (`numHiddenLayers + 1`), `(B, T, hiddenSize)` shapes, first state == scaled embedding, determinism, and **that the layer stack actually transforms the state** — the last so the suite can't pass on a tap that never runs a layer. Tiny randomly-initialized model, no weights downloaded.

The test config sets `hidden_size_per_layer_input: 0`, `enable_moe_block: false`, `num_kv_shared_layers: 0` — the configuration an encoder tap sees. Every layer is self-contained, so the client passes `nil` for `perLayerInput`/`sharedKV`/`positionOffset`, and the sliding-vs-global attention split is resolved internally from the layer index, so a client tap never needs to know about it.

### Verification

- `swift build --target MLXLLM` clean.
- The tap has been implemented in a real downstream client against this surface (not just the test) and compiles with no `@testable` and no internal access.
- ⚠️ I could **not** run the test suite locally: `swift test` builds `MLXFoundationModels`, which currently fails against the macOS 27 SDK on `main` independently of this change — `extraneous argument label 'capabilities:'` at `MLXLanguageModel.swift:565` and a `ConvertibleToGeneratedContent` conversion error at `:718`. That looks like Apple FoundationModels API drift rather than anything here, and I did not touch it. Flagging in case it's news; CI on a supported SDK should exercise the new test normally.

### DIFF
--- a/Libraries/MLXLLM/Models/Gemma4.swift
+++ b/Libraries/MLXLLM/Models/Gemma4.swift
@@ -49,7 +49,12 @@ public class Gemma4Model: Module, LLMModel, KVCacheDimensionProvider {
     public var vocabularySize: Int { languageModel.vocabularySize }
     public var kvHeads: [Int] { languageModel.kvHeads }
 
-    @ModuleInfo(key: "language_model") fileprivate var languageModel: Gemma4TextModel
+    /// The text model, exposed at `@_spi(GemmaEncoder)` scope.
+    ///
+    /// This is the type the model factory actually produces for `gemma4_unified`, so an
+    /// encoder-style client tap cannot reach `Gemma4TextModel` without it — exposing only
+    /// the inner types is not sufficient for a real load path.
+    @ModuleInfo(key: "language_model") @_spi(GemmaEncoder) public var languageModel: Gemma4TextModel
 
     public init(_ config: Gemma4Configuration) {
         self._languageModel.wrappedValue = Gemma4TextModel(config.textConfig)
--- a/Libraries/MLXLLM/Models/Gemma4Text.swift
+++ b/Libraries/MLXLLM/Models/Gemma4Text.swift
@@ -45,8 +45,8 @@ private let _geluMul: @Sendable (MLXArray, MLXArray) -> MLXArray = compile(
 
 public struct Gemma4TextConfiguration: Codable, Sendable {
     var modelType: String = "gemma4_text"
-    var hiddenSize: Int = 1536
-    var numHiddenLayers: Int = 35
+    @_spi(GemmaEncoder) public var hiddenSize: Int = 1536
+    @_spi(GemmaEncoder) public var numHiddenLayers: Int = 35
     var intermediateSize: Int = 6144
     var numAttentionHeads: Int = 8
     var headDim: Int = 256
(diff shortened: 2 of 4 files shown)


davidkoski (approved):
Change looks good, thanks!

xocialize:
Pushed a follow-up commit: `Gemma4Model.languageModel` also needs `@_spi(GemmaEncoder)` exposure.

`gemma4_unified` resolves through `LLMModelFactory` to `Gemma4Model`, and its `languageModel` was `fileprivate` — so exposing only `Gemma4TextModelInner` / `Gemma4DecoderLayer` left a client able to reach the inner types but with no way to obtain the `Gemma4TextModel` that owns them. Insufficient for any real load path.

Worth flagging how it slipped through, since it is a property of the test rather than the code: `Gemma4EncoderAccessTests` was constructing `Gemma4TextModel(config)` directly, bypassing the wrapper the factory produces. It proved the inner types were reachable while the actual entry point was still closed. The test now decodes a wrapped `gemma4_unified` config, builds `Gemma4Model`, and reaches the text model through `.languageModel` — walking the same path production walks.

davidkoski:
It looks like it needs a swift-format run (make sure you are on the right version (603.0.0) as they produce slightly different output).

xocialize:
Done — ran `swift-format` **603.0.0**, built from source rather than using the Xcode-bundled binary (which reports `main`, so exactly the skew you flagged). Only the test file needed changes; the three source files were already clean under 603.0.0. Rebuilt `MLXLLM` after to confirm the reformat was inert.

Thanks for the review.

PR #531, opened 2026-08-13
Tool call parser hardening

### LINKED ISSUE #493: Pythonic tool-call outer extraction is not nesting/quote-aware
## Context
`PythonicToolCallParser.parse()` (`Libraries/MLXLMCommon/Tool/Parsers/PythonicToolCallParser.swift`) extracts the call with the lazy regex `#"\[(\w+)\((.*?)\)\]"#`. The inner `parseArguments` splitter added in #481 is bracket/brace/quote-aware, but the outer extraction is not.

## Problem
An argument value that literally contains `)]` — for example `foo(note={"x": "a)]b"})` — can be truncated before `parseArguments` ever runs, so the argument dictionary is silently wrong.

## Proposed
Extract the call body with a balanced, quote-aware scan (the same approach as the private `splitTopLevel` / `topLevelIndex` helpers) instead of the lazy regex, so both layers are consistent.

## Acceptance
- A unit test with a `)]`-containing argument value parses correctly.
- All existing pythonic parser tests still pass.

Refs: #481

### LINKED ISSUE #495: Add pythonic tool-call parser unit tests (array values, no-unwrap guard, parseEOS)
## Context
#481 added two pythonic-parser tests to `Tests/MLXLMTests/ToolTests.swift`: the LFM2 object-wrapper unwrap, and a preserved (non-wrapper) object-valued argument. A few paths around the new nesting-aware argument splitting remain uncovered.

## Proposed / Acceptance
Add unit tests for:
- an **array-valued** argument containing inner commas (e.g. `search_many(queries=["swift", "mlx"])`) — `splitTopLevel` handles `[`, but only `{` is currently tested;
- a **2-argument** call where one value is an object, asserting it is **not** unwrapped (guards the `arguments.count == 1` condition in `unwrapArgumentWrapper`);
- the **`parseEOS` / `parseMultiple`** path with a wrapper-object call, covering multi-call streaming rather than just single `parse()`.

Refs: #481

aleroot:
## Proposed changes

Let's continue after #529.
This PR fixes three bugs in how tool calls are read back from models, and adds a fallback so models that aren't explicitly recognized still get their tool calls parsed. It's one PR because the first two bugs share a single root cause and are fixed by one new shared component.

**1. Argument values were being silently cut off (closes #493)**

Two formats — Pythonic (used by LFM2) and Gemma — write arguments separated by punctuation, like `notify(note='hello', count=2)`. Both used a simple text search to find those separators, so any value that happened to *contain* one was truncated:

- Pythonic: `notify(note='a)]b')` produced `note = "'a"` — the value was cut at `)]` and a stray quote was left behind.
- Gemma: `call:search{filters:{"city":"Paris","limit":1}}` cut the value at the first comma **and turned the leftover text into an invented second argument** that the model never asked for.

Both now use a new shared `StructuredTextScanner`, which understands that text inside quotes, inside Gemma's escape markers, or inside nested brackets is part of a value and not punctuation. This also replaces two other hand-written copies of the same logic elsewhere in the codebase, so there is now one implementation instead of three.

**2. The numbers 0 and 1 were being turned into `false` and `true`**

Found while testing the above. When an argument value was decoded as JSON, `{"limit": 1}` came out as `{"limit": true}`. The cause is a Foundation detail: a boxed number holding 0 or 1 can be read as a boolean, and the conversion tried that first. This affected **nine** of the tool-call formats and broke tool execution whenever a tool declared a whole-number parameter.

**3. Missing test coverage (closes #495)**

Adds the three Pythonic parser tests requested in the issue: an array value containing commas, a two-argument call asserting the object is *not* unwrapped, and the multi-call `parseEOS` path.

**4. Unrecognized models now get their tool calls parsed**

Previously, if a model's architecture didn't explicitly declare which tool-call format it uses, the library assumed plain JSON — so any model using a different format had every tool call silently pass through as ordinary text. The chat template shipped with a model is what teaches it to write calls in a particular format, so it can be used to identify that format. The library now reads it as a **last** resort, only after the existing checks have all come up empty, so no currently-working model changes behavior. This mirrors the equivalent logic in the Python mlx-lm library, and also honors a `tool_parser_type` entry when a model ships one.

@CharlieTLe FYI

## Checklist

### DIFF
--- a/Libraries/MLXLMCommon/Tool/ToolCallFormatInference.swift
+++ b/Libraries/MLXLMCommon/Tool/ToolCallFormatInference.swift
@@ -81,18 +81,50 @@ extension ToolCallFormat {
 
 extension ToolCallFormat {
 
-    /// The dialect a checkpoint declares or implies, read from its tokenizer files.
+    /// Resolves a model declaration against the dialect selected by its checkpoint.
     ///
     /// An explicit `tool_parser_type` wins, matching mlx-lm, because it is the
-    /// checkpoint author correcting what the template alone would imply.
-    package static func resolved(forTokenizerDirectory directory: URL) -> ToolCallFormat? {
+    /// checkpoint author correcting what the template alone would imply. Otherwise,
+    /// the selected tool template refines a model declaration that cannot parse the
+    /// dialect it prompts. This matters for architecture heuristics such as Llama 3:
+    /// Hermes checkpoints use that architecture while prompting framed JSON instead
+    /// of Llama's inline JSON.
+    ///
+    /// A model-declared superset remains selected. For example, Qwen 3.5 prompts XML
+    /// but its parser deliberately accepts both XML and framed Hermes JSON.
+    package static func resolved(
+        forTokenizerDirectory directory: URL,
+        modelFormat: ToolCallFormat? = nil
+    ) -> ToolCallFormat? {
         let configuration = TokenizerToolCallConfiguration(directory: directory)
 
         if let declared = configuration.toolParserType.flatMap(ToolCallFormat.init(toolParserType:))
         {
             return declared
         }
-        return configuration.chatTemplate.flatMap(ToolCallFormat.inferred(fromChatTemplate:))
+
+        guard
+            let templateFormat = configuration.chatTemplate.flatMap(
+                ToolCallFormat.inferred(fromChatTemplate:))
+        else { return modelFormat }
+        guard let modelFormat else { return templateFormat }
+
+        return modelFormat.accepts(templateFormat: templateFormat) ? modelFormat : templateFormat
+    }
+
+    /// Whether a model-selected parser is a superset of the dialect prompted by
+    /// the selected chat template.
+    private func accepts(templateFormat: ToolCallFormat) -> Bool {
+        switch self {
+        case .qwen35:
+            return templateFormat == .xmlFunction || templateFormat == .json
+        case .gptOSS, .atem:
+            // These select complete token-stream protocols, not interchangeable
+            // payload parsers. Incidental textual markers must not reroute them.
+            return true
+        default:
+            return self == templateFormat
+        }
     }
 }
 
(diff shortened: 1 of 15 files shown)


davidkoski:
@thechriswebb FYI

Libraries/MLXLMCommon/Tool/ToolCallFormatInference.swift
```diff
@@ -0,0 +1,163 @@
...
+                self = .named(try container.decode([Named].self))
+            }
+        }
+
+        var text: String? {
+            switch self {
+            case .single(let template):
+                return template
+            case .named(let templates):
+                return (templates.first { $0.name == "default" } ?? templates.first)?.template
```
davidkoski: This might defeat the purpose of point 4.

Consider: https://huggingface.co/NousResearch/Hermes-3-Llama-3.1-8B/blob/main/tokenizer_config.json:

```
  "chat_template": [
    {
      "name": "default",
      "template": "{{bos_token}}...{% endif %}"
    },
    {
      "name": "tool_use",
      "template": "{%- macro json_to_python_type(json_spec) %}...{%- endif %}\n"
    }
  ],
```

This would chose the `default` to look at, but swift-transformers (if presented with tools) would pick `tool_use`:

```swift
                } else if let tools, !tools.isEmpty, let toolUseTemplate = templateDict["tool_use"] {
                    // Use tool use chat template from config
                    selectedChatTemplate = toolUseTemplate
```

davidkoski (changes requested):
Looks good to me, but see question regarding point 4 before we merge.  Thanks!

Libraries/MLXLMCommon/Tool/ToolCallFormatInference.swift
aleroot: Good catch, fixed in 3b4e83e. Thanks.
Named-template selection now mirrors swift-transformers for tool use: tool_use is preferred when present, with default as the fallback.
I also handled the Hermes precedence case end-to-end. Although Hermes is initially classified as .llama3, the selected tool template can now refine that declaration when no explicit tool_parser_type exists and the declared parser cannot decode the prompted dialect.

aleroot:
> Looks good to me, but see question regarding point 4 before we merge. Thanks!

Thanks @davidkoski , when this lands, I have another branch almost ready for a PR with some refinement on some additional weak points observed during my weekend sessions.

davidkoski (approved):
Looks good to me.  Thank you!

PR #533, opened 2026-08-14
Enforce copy semantics on LogitProcessor for speculative decoding (#508)

### LINKED ISSUE #508: [BUG] LogitSampler and LogitProcessor are not restricted to structs
SpeculativeTokenIterator and MTPSpeculativeTokenIterator both assume value type semantics:

```swift
        // Draft generation: autoregressive loop with draft model
        var draftProcessor = processor  // Copy to discard later

...

        let mainTokens: MLXArray
        if var verifyProcessor = processor {
            // Process each position sequentially so that the processor sees tokens sampled at earlier positions

...

        // Local copy: process() may mutate processor state via Swift struct
        // value semantics, but those mutations stay scoped to this verify
        // loop. Canonical processor state at `self.processor` is only
        // updated by the accept loop below, which evolves it across the
        // actually-emitted tokens. This keeps rejected-draft sampling from
        // polluting cross-round state.
        if var verifyProcessorCopy = processor {

```

This is currently true in `RepetitionContext`, `PresencePenaltyContext`, `FrequencyPenaltyContext` and `PenaltyProcessor` but is not generally true given the types.  Note that these are stateful on purpose, so this is a real issue.

There is no way to restrict the types to structs, so this is pending trouble.

Maybe something like `func copy() -> Self` on LogitProcessor with an override for AnyObject that fatals?

aleroot:
## Proposed changes

Add copy() requirement to LogitProcessor with default value semantics for structs and a fail-fast trap for reference types. Scope draft and verify passes in SpeculativeTokenIterator using copied processor state, preventing unaccepted draft and verification tokens from polluting canonical state. 

Fix #508 

## Checklist

### DIFF
--- a/Libraries/MLXGuidedGeneration/CompositeLogitProcessor.swift
+++ b/Libraries/MLXGuidedGeneration/CompositeLogitProcessor.swift
@@ -19,6 +19,10 @@ public struct CompositeLogitProcessor: LogitProcessor {
         self.processors = processors
     }
 
+    public func copy() -> Self {
+        CompositeLogitProcessor(processors.map { $0.copy() })
+    }
+
     public mutating func prompt(_ prompt: MLXArray) {
         for i in processors.indices {
             processors[i].prompt(prompt)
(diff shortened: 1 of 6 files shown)


davidkoski (approved):
Great, I think this is what we need.

I guess we could go with no default implementation on the protocol and then wouldn't need the fatal for classes as it wouldn't compile if not implemented.  But, if a person looked at one of the others and saw return self they could easily copy that without knowing the details behind it.  Tradeoffs either way.

Anyway, thank you for the fix!

PR #534, opened 2026-08-14
Support video processing options in UserInput and MediaProcessing

### LINKED ISSUE #429: UserInput.Processing could use video options
UserInput.Processing could use video options, e.g. 

```swift
structure VideoProcessing: Sendable {
    public var targetFrames: Int?
    public var targetFramesPerSecond: Double?
}

public var video = VideoProcessing() 
```

and support in MediaProcessing for consuming these.  This could replace (or supply) some of the parameters, e.g. `targetFPS:`

```swift
    static public func asProcessedSequence(
        _ video: UserInput.Video,
        targetFPS: (CMTime) -> Double,
```

aleroot:
## Proposed changes

Add UserInput.VideoProcessing supporting targetFrames and framesPerSecond sampling. Integrate video processing configuration into MediaProcessing.asProcessedSequence and VLM pipelines.

Fixes #429 

Another clean up before release. 

## Checklist

### DIFF
--- a/Libraries/MLXVLM/MediaProcessing.swift
+++ b/Libraries/MLXVLM/MediaProcessing.swift
@@ -6,6 +6,7 @@ import MLX
 import MLXLMCommon
 
 public typealias VideoFrame = UserInput.VideoFrame
+public typealias VideoProcessing = UserInput.VideoProcessing
 
 public struct ProcessedFrames {
     public let frames: [MLXArray]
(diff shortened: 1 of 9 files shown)


Libraries/MLXVLM/MediaProcessing.swift
```diff
@@ -347,70 +348,188 @@ public enum MediaProcessing {
...
+        let estimatedCount = Int(round(duration.seconds * fps))
+        var count = max(min(estimatedCount, maxFrames), 1)
+        if let totalAvailableFrames {
+            count = min(count, totalAvailableFrames)
+        }
+
+        return (0 ..< count).map { i in
+            let tick = Int64(round(Double(i) * Double(timescale) / fps))
+            return CMTime(value: tick, timescale: timescale)
+        }
```
davidkoski, line 388: A _slight_ difference from the previous:

```swift
        let sampledTimeValues = MLXArray.linspace(
            0, duration.value, count: Int(finalFrameCount)
        ).asArray(Int64.self)
```

New: sample frames at exactly the framerate
Old: sample frames at approximately the framerate but include the last frame

It looks like the python code also matches the "old" behavior.

I am not sure which one is desired -- the new one seems like it matches the expressed intent while the old one seems like it might be a convenience given the tools at hand.

I am ok with either approach.  Any opinion here @aleroot ?

davidkoski (changes requested):
Take a look at the question and see what you think.  Either way, this looks good.

Libraries/MLXVLM/MediaProcessing.swift
aleroot: This is a great catch! This is bug I have introduced. For example with this new code a 120 second video at 1 fps with `maxFrames = 32` would sample only timestamps 0…31s, dropping the remaining 89 seconds. The old approach samples 32 frames across the full 120 seconds. Because the pre-fix code clamps the number of samples without recomputing their spacing.

Updated

davidkoski: Ah, I forgot the interaction with that!

davidkoski (approved):
Changes look good, thank you!

PR #408, opened 2026-07-08
fix: Honor safetensors index when loading weights

Zrxrxrx:
## Proposed changes

`loadWeights` currently loads every `.safetensors` file found in the model directory. This can break checkpoints such as [`mlx-community/Qwen3.5-4B-OptiQ-4bit`](https://huggingface.co/mlx-community/Qwen3.5-4B-OptiQ-4bit/tree/main), whose directory contains the primary `model.safetensors` plus auxiliary safetensors files, while `model.safetensors.index.json` maps the model weights only to `model.safetensors`.

Loading the auxiliary files into the main model weight dictionary can pass unrelated tensors through model sanitization and corrupt generation. In the Qwen3.5 OptiQ case, this shows up as incoherent text generation even though the same checkpoint loads correctly through Python `mlx-lm`.

This PR treats `model.safetensors.index.json` as authoritative when it is present. In that case, `loadWeights` loads only the unique safetensors files referenced by `weight_map`. If no index file is present, the existing behavior is preserved: all `.safetensors` files in the directory are loaded.

A regression test covers the indexed case with extra safetensors in the same directory, ensuring only the indexed model weight file is selected.

Tested with:

- `xcodebuild test -scheme mlx-swift-lm-Package -destination 'platform=macOS' -only-testing:MLXLMTests/LoadWeightsTests -skipPackagePluginValidation`
- `swift test --filter LoadWeightsTests`
- `swift format lint Libraries/MLXLMCommon/Load.swift Tests/MLXLMTests/LoadWeightsTests.swift`
- Local smoke test with `mlx-community/Qwen3.5-4B-OptiQ-4bit`: generation produces coherent Chinese text after the loader change.

## Test plan

- [x] `xcodebuild test -scheme mlx-swift-lm-Package -destination 'platform=macOS' -only-testing:MLXLMTests/LoadWeightsTests -skipPackagePluginValidation`
- [x] `swift test --filter LoadWeightsTests`
- [x] `swift format lint Libraries/MLXLMCommon/Load.swift Tests/MLXLMTests/LoadWeightsTests.swift`
- [x] Local smoke test with `mlx-community/Qwen3.5-4B-OptiQ-4bit`: generation produces coherent Chinese text after the loader change.

## Checklist

Put an `x` in the boxes that apply.

- [X] I have read the [CONTRIBUTING](https://github.com/ml-explore/mlx/blob/main/CONTRIBUTING.md) document
[...]

PR #535, opened 2026-08-14
Named image attachments in MLXLMCommon and the Foundation Models adapter

thechriswebb:
## The problem

Foundation Models lets an app attach labeled images to a prompt, ask about one by name, and get an answer that points back at a specific picture. None of that worked through this adapter. Labels were thrown away, so the model never saw a name and could not return one. Orientation was thrown away, so a photo from a camera arrived sideways. An image attached to the instructions was sent on the system message, which the vision models mishandle, so those requests failed or quietly misbehaved. Attachments the adapter could not handle were ignored.

## What this PR does

- Names each image in the prompt text, in the order the images are sent, using the same brackets the on-device model uses. An unlabeled image still holds its place in the list, so later names stay attached to the right picture.
- Rotates an image to its recorded orientation before sending it.
- Ignores an image attached to the instructions and warns that it belongs on a prompt instead. The on-device model ignores it too.
- Rejects a label the model's tokenizer would read as one of its own special tokens rather than as text. Such a label adds a phantom image to the prompt, and the answer comes back plausible instead of failing. Which labels are dangerous depends on the model, so the check asks the loaded tokenizer instead of carrying a list.
- Restores the build and the test suite. Those three commits are unrelated repair work that belongs in its own PR, and they are here only because nothing compiles or tests without them.

It deliberately does not restrict what the model writes as a label. The framework's own lookup already recovers from an imprecise one, and echoing its bracket form is what lets it.

## Tests

Unit tests cover each change. Integration tests run against two real vision models. One asks which of two named pictures is the blue one, so a model that simply repeats the first name it saw fails. The other does the full round trip: the model names a picture and the framework looks it back up. That passed on both models in five runs, every time on an exact name match.

### DIFF
--- a/IntegrationTesting/IntegrationTestingTests/VisionIntegrationTests.swift
+++ b/IntegrationTesting/IntegrationTestingTests/VisionIntegrationTests.swift
@@ -167,8 +167,8 @@ struct VisionIntegrationTests {
 
         // Greedy for the same reason as the test above: which of two labels the
         // model picks is a choice, and sampling makes that choice unstable.
-        // Constraining the label to the two present ones stops the model
-        // inventing a name, but it does not stop it choosing the wrong one.
+        // Nothing constrains the label to the two present ones, so this only
+        // rules out sampling noise, not the model inventing or misspelling a name.
         let response = try await session.respond(
             generating: ColorReport.self,
             options: GenerationOptions(samplingMode: .greedy)
@@ -178,9 +178,14 @@ struct VisionIntegrationTests {
             Attachment(VisionTestImages.solidColor(.blue)).label("Photo_D4E5F6")
         }
 
-        // The schema pins the label to the transcript's labels, so this must be
-        // one of them and must resolve. Without that constraint the model is
-        // free to paraphrase and the lookup returns nil.
+        // The prompt names each image as `[label]` (see
+        // AttachmentLabelRenderer), matching what FoundationModels itself emits,
+        // and `ImageReference.resolved(in:)` matches that label exactly before
+        // falling back to extracting a bracketed substring. Nothing constrains
+        // what the model writes here: `attachmentLabel` is an unconstrained
+        // `String` in FoundationModels, so a model that paraphrased the label
+        // instead of echoing it would fail the assertion below rather than being
+        // prevented from paraphrasing.
         #expect(response.content.image.attachmentLabel == "Photo_D4E5F6")
         let resolved = response.content.image.resolved(in: session.transcript)
         #expect(resolved != nil, "expected the reference to resolve to an attachment")
--- a/IntegrationTesting/IntegrationTestingTests/VisionIntegrationTests.swift
+++ b/IntegrationTesting/IntegrationTestingTests/VisionIntegrationTests.swift
@@ -10,6 +10,38 @@ import Testing
 
 #if FoundationModelsIntegration && canImport(FoundationModels, _version: 2)
 
+/// The two vision models this coverage runs against. Both were used to choose
+/// the label rendering, so both are worth holding to it.
+///
+/// Known issue: `mlx-community/Qwen3-VL-4B-Instruct-4bit` currently fails to
+/// load, and the failure is in the published repo, not here. Its
+/// `model.safetensors.index.json` references `model-00001-of-00002.safetensors`
+/// and `model-00002-of-00002.safetensors`, neither of which the repo ships,
+/// and declares a total size of about 8.9 GB while the repo actually contains
+/// a single 3.1 GB `model.safetensors`. The 8.9 GB figure is the unquantized
+/// size, so the index looks like it was carried over from the source repo and
+/// never regenerated for the quantized upload. The error surfaced through
+/// `loadModelContainer`'s factory fallback chain reads as
+/// `.unsupportedModelType("qwen3_vl")`, which is a red herring: the real error
+/// is a file-not-found on the missing shard, masked because the LLM factory
+/// (which has no qwen3_vl entry) is what reports last. This also affects the
+/// pre-existing color-naming test above, so the condition predates these two
+/// vision-labeling tests.
+let labeledVisionModels = [
+    "mlx-community/Qwen3-VL-4B-Instruct-4bit",
+    "mlx-community/gemma-4-e4b-it-4bit",
+]
+
+/// Response type for ``VisionIntegrationTests/generatedImageReferenceResolvesBackToTheAttachment``.
+/// The `ImageReference` field is the round trip under test: the model names an
+/// input image and the app resolves that name back to the pixels.
+@available(iOS 27.0, macOS 27.0, visionOS 27.0, *)
+@Generable
+struct ColorReport {
+    var image: ImageReference
+    var color: String
+}
+
 /// Opt-in end-to-end VLM test: drives a real `Qwen3-VL-4B-Instruct-4bit` through
 /// the FoundationModels adapter with a labeled image attachment and `.vision`
 /// declared, proving the labeled-attachment path reaches the already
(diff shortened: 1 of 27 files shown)


davidkoski:
This seems pretty neat -- I wonder if we need to augment the `UserInput` in MLXLMCommon with similar capability.  Not necessarily in this PR, but if it is worthwhile we can file an issue for it.

aleroot:
> This seems pretty neat -- I wonder if we need to augment the `UserInput` in MLXLMCommon with similar capability. Not necessarily in this PR, but if it is worthwhile we can file an issue for it.

Yes, I think this would be great. I am definitely interested for my app.

Libraries/MLXFoundationModels/TranscriptConverter.swift
```diff
@@ -161,24 +185,95 @@ struct TranscriptConverter {
...
         }
+        return ordered
+    }
+
+    /// The distinct attachment labels present in `entries`, in first-seen order.
+    ///
+    /// `ImageReference.attachmentLabel` is an unconstrained `String`, so nothing
+    /// requires a generated label to be one of these; this is just the set an
+    /// app can look an image up by.
+    static func attachmentLabels(in entries: some Collection<Transcript.Entry>) -> [String] {
```
davidkoski: This is only called from tests -- is this just introspection for that purpose?  It isn't public so not callable outside the repo.

IntegrationTesting/IntegrationTestingTests/VisionIntegrationTests.swift
```diff
@@ -10,22 +10,76 @@ import Testing
...
+/// cannot load: its `model.safetensors.index.json` names
+/// `model-00001-of-00002.safetensors` and `model-00002-of-00002.safetensors`,
+/// neither of which the repo ships, and declares about 8.9 GB while the repo
+/// contains a single 3.1 GB `model.safetensors`. The 8.9 GB figure is the
+/// unquantized size, so the index was carried over from the source repo and
+/// never regenerated for the quantized upload. The same is true of
+/// `Qwen3-VL-8B-Instruct-4bit` (names four shards, ships two) and
+/// `Qwen3-VL-4B-Instruct-8bit` (names two, ships one), so it is the
+/// quantization batch rather than one repo. `Qwen3-VL-2B-Instruct-4bit` is
+/// packaged correctly if a Qwen3-VL model is ever wanted here.
```
davidkoski: This is interesting, but perhaps should be an issue rather than a comment here.  This honoring of the index was added in #408.  I suspect it _did_ work before that.

Filed #554 to track this.

Libraries/MLXFoundationModels/AttachmentLabelRenderer.swift
```diff
@@ -0,0 +1,63 @@
...
+/// model.
+///
+/// Labels reach the tokenizer as ordinary prompt text, so one whose rendered
+/// form is a tokenizer special token would corrupt the prompt's image
+/// placeholders. ``AttachmentLabelValidator`` rejects those once the tokenizer is
+/// known, and it checks the bracketed form written below rather than the bare
+/// label, because the brackets are part of what gets tokenized: Mistral-family
+/// image tokens are bracketed, so a label of `IMG` renders as `[IMG]`.
+@available(iOS 27.0, macOS 27.0, visionOS 27.0, *)
+struct AttachmentLabelRenderer {
```
davidkoski: This seems like a workaround for the fact that the attachments don't carry a label.  I think that is ok for this PR, but I think `UserInput`/`LMInput` probably need to carry the label so it can get proper handling in the transcript.

davidkoski: The downside of delaying this is that the behavior from LLMs may change when this is fixed -- the input pattern would be different.  Hopefully the behavior change would be for the best, but who knows.

davidkoski: How does this work in the template?  Here is `mlx-community/Qwen2.5-VL-7B-Instruct-4bit`:

```
{% set image_count = namespace(value=0) %}
{% set video_count = namespace(value=0) %}

{% for message in messages %}
    {% if loop.first and message['role'] != 'system' %}
        <|im_start|>system
        You are a helpful assistant.<|im_end|>
    {% endif %}
    
    <|im_start|>{{ message['role'] }}
    
    {% if message['content'] is string %}
        {{ message['content'] }}<|im_end|>
    {% else %}
        {% for content in message['content'] %}
            {% if content['type'] == 'image' or 'image' in content or 'image_url' in content %}
                {% set image_count.value = image_count.value + 1 %}
                {% if add_vision_id %}Picture {{ image_count.value }}: {% endif %}
                <|vision_start|><|image_pad|><|vision_end|>
            {% elif content['type'] == 'video' or 'video' in content %}
                {% set video_count.value = video_count.value + 1 %}
                {% if add_vision_id %}Video {{ video_count.value }}: {% endif %}
                <|vision_start|><|video_pad|><|vision_end|>
            {% elif 'text' in content %}
                {{ content['text'] }}
            {% endif %}
        {% endfor %}
        <|im_end|>
    {% endif %}
{% endfor %}

{% if add_generation_prompt %}
    <|im_start|>assistant
{% endif %}
```

It looks like this is labeling it `Picture 1`.

`mlx-community/gemma-4-e4b-it-4bit` is much longer:

https://huggingface.co/mlx-community/gemma-4-e4b-it-4bit/blob/main/chat_template.jinja

the pertinent part looks like this:

```
            {%- elif message['content'] is sequence -%}
                {%- for item in message['content'] -%}
                    {%- if item['type'] == 'text' -%}
                        {%- if role == 'model' -%}
                            {{- strip_thinking(item['text']) -}}
                        {%- else -%}
                            {{- item['text'] | trim -}}
                        {%- endif -%}
                    {%- elif item['type'] == 'image' -%}
                        {{- '<|image|>' -}}
                        {%- set ns.prev_message_type = 'image' -%}
                    {%- elif item['type'] == 'audio' -%}
                        {{- '<|audio|>' -}}
                        {%- set ns.prev_message_type = 'audio' -%}
                    {%- elif item['type'] == 'video' -%}
                        {{- '<|video|>' -}}
                        {%- set ns.prev_message_type = 'video' -%}
                    {%- endif -%}
                {%- endfor -%}
            {%- endif -%}
```

That doesn't seem to mark it with anything.

Libraries/MLXLMCommon/Chat.swift
```diff
@@ -174,6 +190,52 @@ extension MessageGenerator {
...
+
+        func appendImages() {
+            for image in message.images {
+                if let marker = image.labelMarkerCharacter {
+                    // Emitting it would add a placeholder the model counts against pixels.
+                    messageContentLogger.warning(
+                        "Leaving an image name out of the prompt: it holds `\(marker, privacy: .public)`, which vision models build their image placeholders from"
+                    )
+                } else if let label = image.label {
+                    appendText("[\(label)]")
```
davidkoski, line 217: I wonder if this needs to be configurable?  Mistral 3's special image token is `[IMG]`.  See labelRendersAnImagePlaceholderAsText and oneImageYieldsTwoImagePlaceholders

Have to find the right place to attach it -- it seems like it belongs with the model, though perhaps it is technically part of the template/tokens.  An override in ModelConfiguration?

Libraries/MLXLMCommon/Chat.swift
```diff
@@ -1,5 +1,21 @@
 // Copyright © 2025 Apple Inc.
 
+import os
+
+private let messageContentLogger = Logger(
+    subsystem: "mlx-swift-lm", category: "MessageContent")
```
davidkoski, line 6: This isn't platform neutral -- it won't build on linux.  I see that there are some uses in the MLXFoundationModels code already and I wonder if we should change those?

Anyway, we don't have a logging solution picked but perhaps we should.  I think `apple/swift-log` is the standard, so probably that.

Libraries/MLXLMCommon/Chat.swift
```diff
@@ -174,6 +190,52 @@ extension MessageGenerator {
...
+            parts.append(["type": "text", "text": text])
+        }
+
+        func appendImages() {
+            for image in message.images {
+                if let marker = image.labelMarkerCharacter {
+                    // Emitting it would add a placeholder the model counts against pixels.
+                    messageContentLogger.warning(
+                        "Leaving an image name out of the prompt: it holds `\(marker, privacy: .public)`, which vision models build their image placeholders from"
+                    )
```
davidkoski, line 215: See other comment on os.Logging.  swift-log doesn't support privacy markers.  The standard approach is a constant string with metadata:

```swift
messageContentLogger.warning(
    "Leaving an image name out of the prompt",
    metadata: [
        "marker": .string(marker)
    ]
)
```

or interpolation :-)

davidkoski (changes requested):
Changes look good.  One issue, one potential issue:

- os.Logging won't build on linux -- it may be time for us to adopt swift-log
    - look at MLXFoundationModels too

- we can inject image markers into the stream, e.g. Mistral3, see repro case attached below

[ReproTests.swift.zip](https://github.com/user-attachments/files/32264354/ReproTests.swift.zip)

Libraries/MLXLMCommon/Chat.swift
aleroot: If possible please find a solution that does not require adding dependencies that can potentially increase the size of the executable. Thanks.

davidkoski: Yeah, that is a good point.  We have pretty minimal logging needs.  Maybe something very thin in mlx-swift would do the trick.

I have something sketched out and will put up a PR in the morning.

davidkoski: See https://github.com/ml-explore/mlx-swift/pull/484

PR #538, opened 2026-08-15
Add an Xcode 27 compile check and handle Generation.rejectedToolCall

### LINKED ISSUE #499: FoundationModels integration suite never executes on the Xcode 26 runner
## Context
The self-hosted macOS runner uses Xcode 26.5. The entire `IntegrationTesting/IntegrationTestingTests/MLXFoundationModelsIntegration/` tree — plus `VisionIntegrationTests` — is gated on `#if FoundationModelsIntegration && canImport(FoundationModels, _version: 2)` (see #478), which is false on the macOS 26 SDK. The `MLXFoundationModels` adapter itself is gated the same way. So roughly 190 of ~290 integration test methods compile out and never run.

The `Select Xcode` step in `.github/workflows/integration_tests.yml` currently pins Xcode 26 deliberately (#477). Xcode 27 **beta 2** was tried and rejected: its FoundationModels SDK lacks the `GenerationOptions.SamplingMode.Kind` cases the adapter uses (`.randomTopK`, `.randomProbabilityThreshold`, added in #431), so the adapter fails to compile:

```
MLXLanguageModel.swift:819: error: type 'GenerationOptions.SamplingMode.Kind' has no member 'randomTopK'
```

## Problem
The FoundationModels integration coverage is currently unreachable in CI. Nothing signals this beyond the tests silently not existing.

## Proposed
Decide if/when to exercise the suite:
- **Enable:** install Xcode 27 beta 3+ (or the release) on the runner and extend the Select-Xcode step to prefer it over 26, keeping beta 1/2 excluded.
- **Defer:** document that this coverage is intentionally deferred until the 27 SDK is available on the runner.

## Acceptance
- Documented decision. If enabling: the FoundationModels tests build and run on the runner.

Refs: #431, #477, #478, #481

### LINKED ISSUE #536: IntegrationTesting project no longer compiles: Generation.rejectedToolCall unhandled in 7 switches
The `IntegrationTesting` Xcode project no longer builds. #512 (`d7dc03d`) added a
`rejectedToolCall` case to `Generation` (`Libraries/MLXLMCommon/Evaluate.swift:2603`,
unconditional) and updated `IntegrationTestHelpers`, but seven exhaustive `switch`
statements over `Generation` in
`IntegrationTesting/IntegrationTestingTests/MTPIteratorEndToEndDiagnosticTests.swift`
were not updated:

- lines 150, 371, 495, 511, 692, 925, 1003

None of these are behind a compilation condition, so this breaks on Xcode 26 as well as
Xcode 27.

### Why CI did not catch it

Two independent gaps:

1. **The PR path never builds this project.** `pull_request.yml`'s `mac_build_and_test`
   job builds only `-scheme mlx-swift-lm-Package`. The `IntegrationTesting` Xcode project
   is not part of that scheme, so nothing on the pull-request path compiles these files.
2. **The nightly that does build it ran before the merge.** `integration_tests.yml` is
   the only workflow that builds `IntegrationTesting.xcodeproj`. Its last run was
   2026-08-14T09:39:34Z at `d667610`; #512 merged at 2026-08-14T21:12:16Z. The next
   nightly fails at the build step, which means zero integration coverage rather than one
   failing test.

### Fix

The seven sites are text-accumulating diagnostics that already discard `.toolCall`, so the
new case joins it: `case .toolCall, .rejectedToolCall: break`.

This is filed as the record rather than as open work: the fix ships in the PR referenced
below, together with a new `pull_request.yml` job that builds this project so the first
gap cannot recur.

See also the companion issue for the three sites that only break against the Xcode 27 /
macOS 27 SDK and are therefore invisible to the nightly by construction.

### LINKED ISSUE #537: MLXFoundationModels does not compile against the Xcode 27 / macOS 27 SDK
Building against the Xcode 27 / macOS 27 SDK fails in `MLXFoundationModels`. #512
(`d7dc03d`) added `rejectedToolCall` to `TokenStreamEvent`
(`Libraries/MLXLMCommon/Tool/TokenStreamDecoder.swift:12`) and to `Generation`, leaving
three exhaustive `switch` statements unhandled:

| File | Line | Enum |
|---|---|---|
| `Libraries/MLXFoundationModels/MLXLanguageModel.swift` | 1890 | `TokenStreamEvent` |
| `Libraries/MLXFoundationModels/MLXLanguageModel.swift` | 1943 | `TokenStreamEvent` |
| `IntegrationTesting/.../GuidedGeneration/GuidedGenerationBenchmarkTests.swift` | 426 | `Generation` |

### Why the nightly cannot see these

All three sit inside `#if ... canImport(FoundationModels, _version: 2)`, which is only
true on the macOS 27 SDK. `integration_tests.yml` deliberately selects Xcode 26 (see
#499), where these blocks compile out entirely, so the nightly is green on code that does
not exist in its build. The same gating hides them from `pull_request.yml`.

Evidence that they are real: a trial run on GitHub's hosted `xcode-27` image (Xcode 27.0
beta 4, `27A5228h`, macOS 27.0 SDK) surfaces them:
https://github.com/CharlieTLe/mlx-swift-lm/actions/runs/31849906590

That run also establishes that #499's original blocker is gone. #499 rejected Xcode 27
beta 2 because its FoundationModels SDK lacked
`GenerationOptions.SamplingMode.Kind.randomTopK`; beta 4 compiles the adapter past
`MLXLanguageModel.swift:819` with no `SamplingMode.Kind` errors.

### Fix

- `GuidedGenerationBenchmarkTests.swift:426` is a benchmark that already discards
  `.toolCall`, so the new case joins that group.
- The two `MLXLanguageModel.swift` sites are **not** grouped with `.toolCall`. #512
  established in this same file that a rejection is significant: the allowed-tool path
  collects into `rejectedToolCalls` and then throws `RejectedToolCallError` (lines
  1598-1600, 1613, 1633). These two sites are on the plain/reasoning streaming path, whose
  decoder is built with `tools: nil` and whose closures are non-throwing
  (`decoder.push(token) { ... -> Bool }`), so throwing would require restructuring. They
  log the rejection (reason and tool name, never `rawTextPreview`) through the
  `Self.protocolLogger.error(...)` channel the same switch already uses for
  `.protocolError`. That surfaces the anomaly without silently dropping it and without
  changing control flow.

Filed as the record rather than as open work: the fix ships in the PR referenced below,
which also adds a `pull_request.yml` job that compiles both this library and the
`IntegrationTesting` project against the macOS 27 SDK, so this class of break becomes
visible on the PR path.

CharlieTLe:
Adds a compile-only CI job on GitHub's hosted `xcode-27` image, and fixes the ten
exhaustive `switch` sites that job immediately found.

### The job

`pull_request.yml` gains `integration_build_xcode27` (`needs: lint`, `runs-on: xcode-27`,
60-minute cap). It builds `IntegrationTesting.xcodeproj` with `build-for-testing` against
the macOS 27 SDK.

It closes two distinct gaps:

- `mac_build_and_test` builds only `-scheme mlx-swift-lm-Package`, so the
  `IntegrationTesting` Xcode project is **never built on the PR path**. That is how #536
  landed.
- The nightly `integration_tests.yml` pins Xcode 26, where everything behind
  `canImport(FoundationModels, _version: 2)`, the whole `MLXFoundationModelsIntegration/`
  tree plus the `MLXFoundationModels` adapter, roughly 190 of ~290 integration test
  methods per #499, compiles out and so cannot break the build. That is how #537 landed.

Two guards in the job are load-bearing rather than defensive, and both were confirmed on
the runner:

- **It asserts Xcode 27 and SDK 27 and fails loudly otherwise.** `xcode-27` is a preview
  label ([actions/runner-images#14404](https://github.com/actions/runner-images/issues/14404))
  whose default toolchain has moved before. On a 26 SDK the entire gated tree compiles out
  and the job goes uselessly green.
- **It installs `MetalToolchain`, parsing `Status:`.** The hosted image ships without it
  (the green run below logs `MetalToolchain reports 'uninstalled'; downloading`) and
  mlx-swift compiles `.metal` sources. `xcodebuild -showComponent MetalToolchain` exits
  **0** while printing `Status: uninstalled`, so an exit-code guard passes silently and the
  build dies much later at the first `CompileMetalFile`.

The job is **build-only**, and not by choice, see below.

### The fixes (#536, #537)

#512 added `rejectedToolCall` to `Generation` and `TokenStreamEvent` and updated
`IntegrationTestHelpers`, but ten exhaustive switches were missed.

Eight are text-accumulating diagnostics and a benchmark that already discard `.toolCall`,
so the new case joins that group: `MTPIteratorEndToEndDiagnosticTests.swift` (7 sites) and
`GuidedGenerationBenchmarkTests.swift:426`.

The two `MLXLanguageModel.swift` sites are **not** grouped with `.toolCall`. #512
established in this same file that a rejection is significant: the allowed-tool path
collects into `rejectedToolCalls` and then throws `RejectedToolCallError`. These two sites
are on the plain/reasoning streaming path, whose decoder is built with `tools: nil` and
whose closures are non-throwing (`decoder.push(token) { ... -> Bool }`), so throwing would
require restructuring. They log the rejection through the `Self.protocolLogger.error(...)`
channel the same switch already uses for `.protocolError`, carrying `reason` and
`toolName` and never `rawTextPreview`. That surfaces the anomaly without dropping it
silently and without changing control flow.

**Maintainers may prefer stricter parity with the allowed-tool path**: restructure so
this path can throw `RejectedToolCallError` too. That is a larger change to the streaming
closures, so this PR takes the logging option; happy to switch if you would rather have
the throw.

### Why the job cannot also run tests (new: #539)

The original intent was to also run the tier-aware `PlatformCompatibilityProbes` suite,
which loads no models. That is impossible on this image, and the reason is worth reading:

A test bundle built against the macOS 27 SDK **segfaults before running any test** on the
image's macOS 26.5.2 host. XCTest resolves test classes by name, which calls
`objc_copyClassList` and therefore realizes *every* Swift class in the image. That
realizes `@available(macOS 27, *) TestResponseStream`, completing its metadata instantiates
`AsyncThrowingStream<MLXLanguageModel.Executor.GenerationEvent, Error>`, and
`GenerationEvent`'s metadata lives in weak-linked (null on macOS 26) `FoundationModels`.
Result: `EXC_BAD_ACCESS` at `pc = 0` inside
`+[XCTestSuite prepareContainerSuite:toRunTestIdentifiers:]`. `-only-testing:` cannot dodge
it, because class enumeration precedes test selection. Full backtrace and `otool` evidence
in #539.

So the job builds and stops there. That is not a downgrade of the coverage this PR claims:
compiling the `_version: 2` tree is what found #536 and #537.

### What this does and does not do for #499

It addresses #499 by making the gated tree **compile** in CI, which is the half a public
runner can deliver. Also relevant: Xcode 27.0 beta 4 (`27A5228h`) resolves the blocker that
caused beta 2 to be rejected there: it compiles the adapter with no
`GenerationOptions.SamplingMode.Kind.randomTopK` errors. Provisioning the self-hosted mac
is no longer needed for compile coverage.

It does **not** close #499's runtime goal. #539 sharpens why: on a macOS 26 host the FM
suites do not merely early-return and pass vacuously, the runner does not survive test-suite
construction. Real FM runtime coverage needs a macOS 27 **host**, which no hosted image
offers yet. #499 should stay open for that half.

### Verification

- `pre-commit run --all` clean (swift-format 603.0.0).
- Local build on Xcode 27.0 (`27A5228h`) / macOS 27.0 SDK reaches
  `** TEST BUILD SUCCEEDED **`. Test enumeration confirms the gated suites are genuinely in
  the built bundle (231 tests, including `GuidedGenerationBenchmarkTests`), so the
  `_version: 2` tree compiled rather than compiling out. `PlatformCompatibilityProbes`
  passes locally in 0.028s with no network, this machine is a macOS 27 host, which is
  exactly the difference #539 describes.
- **The job itself, green on the hosted image:**
  https://github.com/CharlieTLe/mlx-swift-lm/actions/runs/31854218828 logs `Xcode 27.0`,
  `Build version 27A5228h`, `macOS SDK: 27.0`, host `26.5.2`,
  `MetalToolchain reports 'uninstalled'; downloading`, then `MetalToolchain installed`, then
  a clean `build-for-testing`. (Run on a fork branch with the repository guard inverted, so
  the job would execute outside `ml-explore`.)
- Earlier runs showing the failures this PR fixes and diagnoses:
  [31849906590](https://github.com/CharlieTLe/mlx-swift-lm/actions/runs/31849906590) (the
  compile breaks),
  [31853566691](https://github.com/CharlieTLe/mlx-swift-lm/actions/runs/31853566691) (the
  #539 crash backtrace).

Relates to #499 and #539. Fixes #536, fixes #537. Origin of the breaks: #512.

### DIFF
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -108,3 +108,126 @@ jobs:
             /Users/runner/Library/Developer/Xcode/DerivedData/**/Logs/Test/*.xcresult
             ~/Library/Logs/DiagnosticReports/*
           retention-days: 7
+
+  # Compile the IntegrationTesting project against the macOS 27 SDK.
+  #
+  # Two gaps this closes. First, mac_build_and_test above only builds
+  # `-scheme mlx-swift-lm-Package`, so the IntegrationTesting Xcode project is
+  # never built on the PR path at all and a change that breaks it merges green
+  # (see #512, which left seven exhaustive switches over `Generation` unhandled
+  # in MTPIteratorEndToEndDiagnosticTests.swift). Second, the nightly
+  # integration_tests.yml pins Xcode 26, where everything behind
+  # `canImport(FoundationModels, _version: 2)` -- the whole
+  # MLXFoundationModelsIntegration/ tree and the MLXFoundationModels adapter --
+  # compiles out and therefore cannot break the build. This job builds both.
+  #
+  # This is a COMPILE check, and cannot be anything more on this image. The
+  # hosted `xcode-27` image carries the macOS 27 SDK on a macOS 26 host, and a
+  # test bundle built against the 27 SDK cannot run *any* test there: XCTest
+  # calls objc_copyClassList at startup, which realizes every class including
+  # the @available(macOS 27, *) `TestResponseStream`, whose metadata instantiates
+  # AsyncThrowingStream<MLXLanguageModel.Executor.GenerationEvent, Error> and so
+  # calls through weak-null FoundationModels metadata. The runner segfaults
+  # during test-suite construction, before any test body or #available guard
+  # runs, so no `-only-testing:` selection can dodge it. FM runtime coverage
+  # needs a macOS 27 host, which no hosted image offers yet (see #499).
+  integration_build_xcode27:
+    needs: lint
+    if: github.repository == 'ml-explore/mlx-swift-lm'
+    # Public preview label, see actions/runner-images#14404.
+    runs-on: xcode-27
+    timeout-minutes: 60
+    steps:
+      - uses: actions/checkout@v6
+        with:
+          submodules: recursive
+
+      - name: Report toolchain and host
+        shell: bash
+        run: |
+          xcodebuild -version
+          swift --version
+          echo "macOS SDK: $(xcrun --sdk macosx --show-sdk-version)"
+          sw_vers
+          # DeviceTier classifies on the *runtime* OS version rather than the
+          # SDK, so printing it here records why this job builds but does not
+          # test: on an OS 26 host the FM surface is unavailable at runtime.
+          case "$(sw_vers -productVersion)" in
+            27.*|2[89].*|3*) echo "Host DeviceTier would be: full (OS >= 27)" ;;
+            26.*)             echo "Host DeviceTier would be: partial (OS 26): the FM surface is @available(macOS 27, *)" ;;
+            *)                echo "Host DeviceTier would be: absent (OS < 26)" ;;
+          esac
+
+      - name: Assert Xcode 27 and the macOS 27 SDK
+        shell: bash
+        run: |
+          # `xcode-27` is a preview label and its default toolchain has moved
+          # before. On a 26 SDK the entire `_version: 2` tree compiles out and
+          # this job goes green while proving nothing, so fail loudly instead.
+          case "$(xcodebuild -version | head -1)" in
+            "Xcode 27"*) ;;
+            *)
+              echo "::error::Expected Xcode 27, got '$(xcodebuild -version | head -1)' (DEVELOPER_DIR=${DEVELOPER_DIR:-default})."
+              exit 1
+              ;;
+          esac
+          case "$(xcrun --sdk macosx --show-sdk-version)" in
+            27.*) ;;
+            *)
+              echo "::error::Expected the macOS 27 SDK, got $(xcrun --sdk macosx --show-sdk-version). The MLXFoundationModels sources are gated on canImport(FoundationModels, _version: 2) and would compile out."
+              exit 1
+              ;;
+          esac
+
+      - name: Install MetalToolchain
+        shell: bash
+        run: |
+          # The self-hosted mac is provisioned with this already, so the jobs
+          # above only assert it. The hosted image ships WITHOUT it, and
+          # mlx-swift compiles .metal sources during the build (mlx-swift_Cmlx).
+          #
+          # Parse `Status:` rather than trusting the exit code: -showComponent
+          # exits 0 and prints "Status: uninstalled" when the component is
+          # absent, so an exit-code guard silently succeeds and the build then
+          # dies much later on the first CompileMetalFile with
+          # "cannot execute tool 'metal'".
+          component_status() {
+            xcodebuild -showComponent MetalToolchain 2>&1 \
+              | awk -F': *' '/^Status:/ { print tolower($2); exit }'
+          }
+          if [ "$(component_status)" = "installed" ]; then
+            echo "MetalToolchain installed."
+          else
+            echo "MetalToolchain reports '$(component_status)'; downloading (839 MB)."
+            xcodebuild -downloadComponent MetalToolchain
+            if [ "$(component_status)" != "installed" ]; then
+              echo "::error::MetalToolchain still reports '$(component_status)' after -downloadComponent."
+              exit 1
+            fi
+            echo "MetalToolchain installed."
+          fi
+
+      - name: Build IntegrationTesting for testing (Xcode, macOS)
+        shell: bash
+        run: |
+          # `build-for-testing` rather than `test`: this compiles both the
+          # ungated suites and the `_version: 2` tree, which is the coverage this
+          # job exists for. Running them is not an option here (see the crash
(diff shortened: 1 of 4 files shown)


davidkoski (approved):
Looks great, thank you!

Issue #543, opened 2026-08-16
MLXFoundationModels does not compile against Xcode 27 beta 5 (27A5237l): LanguageModelCapabilities.init and metadata dictionaries changed


abra-code:
### Summary

`MLXFoundationModels` does not compile against the macOS 27 SDK in Xcode 27.0 beta 5 (27A5237l). Two `FoundationModels` API shapes the repo uses no longer exist in that SDK.

This is separate from the `Generation.rejectedToolCall` exhaustive-switch failures, which are covered by issue #537 (MLXFoundationModels does not compile against the Xcode 27 / macOS 27 SDK), issue #536 (IntegrationTesting project no longer compiles: Generation.rejectedToolCall unhandled in 7 switches), and PR #538 (Add an Xcode 27 compile check and handle Generation.rejectedToolCall). On beta 5 both sets appear together: PR #538 fixes the switch half, and I could not find anything covering this half. I checked that PR's diff and it changes neither of the two call sites below.

### Change 1: `LanguageModelCapabilities.init` has no argument label

The SDK declares exactly one initializer, unlabeled:

```swift
@available(iOS 27.0, macOS 27.0, visionOS 27.0, watchOS 27.0, *)
public struct LanguageModelCapabilities : Sendable {
  public init(_ capabilities: [LanguageModelCapabilities.Capability])
}
```

`Libraries/MLXFoundationModels/MLXLanguageModel.swift:565` passes a label:

```
error: extraneous argument label 'capabilities:' in call
```

`IntegrationTesting/.../PlatformAvailability/CompatibilityProbes.swift:83` has the same call shape.

Both spellings already coexist in the tree, which suggests this went unnoticed rather than being a recent regression on your side: `Tests/MLXFoundationModelsTests/GenerationOptionsForwardingTests.swift:42` has used the unlabeled `LanguageModelCapabilities([])` since PR #456 (Support multi-round tool calling in MLXFoundationModels), while `MLXLanguageModel.swift` kept the labeled form. Nothing caught the disagreement because everything behind `canImport(FoundationModels, _version: 2)` compiles to nothing on the Xcode 26 CI.

### Change 2: metadata dictionaries take `ConvertibleToGeneratedContent`

28 metadata parameters across the framework interface are now `[String: any ConvertibleToGeneratedContent]` rather than `[String: any Sendable & Codable & Equatable]`, including:

[...]

PR #544, opened 2026-08-16
Fix MLXFoundationModels compilation against the Xcode 27 beta 5 SDK

abra-code:
Two FoundationModels API shapes the adapter uses are not present in the macOS 27 SDK shipped with Xcode 27.0 beta 5 (27A5237l).

LanguageModelCapabilities declares a single unlabeled initializer, init(_ capabilities:). Two call sites passed `capabilities:`. The unlabeled spelling was already in the tree - GenerationOptionsForwardingTests has used it since PR #456, "Support multi-round tool calling in MLXFoundationModels" - so the two forms disagreed without anything noticing, because everything behind canImport(FoundationModels, _version: 2) compiles to nothing on the Xcode 26 CI.

Metadata dictionaries across the framework are now [String: any ConvertibleToGeneratedContent] rather than [String: any Sendable & Codable & Equatable], covering Action.updateMetadata, Action.updateUsage,
LanguageModelExecutorGenerationRequest.init, LanguageModelSession.Usage.init and the streamResponse family.

ConvertibleToGeneratedContent does not refine Sendable, so the internal GenerationEvent mirror cannot adopt the framework's exact spelling and stay Sendable. It takes [String: any ConvertibleToGeneratedContent & Sendable] instead, which upcasts implicitly at the SDK boundary. Every emitMetadata call site passes a String or Bool literal, both Generable and therefore ConvertibleToGeneratedContent, so no call site changes.

This is orthogonal to the Generation.rejectedToolCall exhaustive-switch failures, which are reported in issue #537, "MLXFoundationModels does not compile against the Xcode 27 / macOS 27 SDK", and issue #536, "IntegrationTesting project no longer compiles", and fixed by PR #538, "Add an Xcode 27 compile check and handle Generation.rejectedToolCall". Both changes are needed for a green beta 5 build: with this one applied, the only errors left from `xcodebuild -scheme MLXFoundationModels` are the two switch sites PR #538 covers.

Fixes issue #543, "MLXFoundationModels does not compile against Xcode 27 beta 5 (27A5237l)".

Co-Authored-By: Claude

## Proposed changes

## Checklist

### DIFF
--- a/IntegrationTesting/IntegrationTestingTests/MLXFoundationModelsIntegration/PlatformAvailability/CompatibilityProbes.swift
+++ b/IntegrationTesting/IntegrationTestingTests/MLXFoundationModelsIntegration/PlatformAvailability/CompatibilityProbes.swift
@@ -80,7 +80,7 @@ struct PlatformCompatibilityProbes {
             lmAvailable = true
             #if canImport(FoundationModels, _version: 2)
             // Touch the OS-27 surface to prove it is genuinely reachable here.
-            _ = LanguageModelCapabilities(capabilities: [])
+            _ = LanguageModelCapabilities([])
             _ = (any LanguageModel).self
             #endif
         }
--- a/IntegrationTesting/IntegrationTestingTests/MLXFoundationModelsIntegration/Support/FMTestHelpers.swift
+++ b/IntegrationTesting/IntegrationTestingTests/MLXFoundationModelsIntegration/Support/FMTestHelpers.swift
@@ -164,7 +164,7 @@ func makeExecutorRequest(
     schema: GenerationSchema? = nil,
     generationOptions: GenerationOptions = GenerationOptions(),
     contextOptions: ContextOptions = ContextOptions(),
-    metadata: [String: any Sendable & Codable & Equatable] = [:]
+    metadata: [String: any ConvertibleToGeneratedContent & Sendable] = [:]
 ) -> LanguageModelExecutorGenerationRequest {
     LanguageModelExecutorGenerationRequest(
         id: id,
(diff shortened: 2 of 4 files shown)


ctymoszek:
This looks good!

I noticed one more stale location: at `IntegrationTesting/IntegrationTestingTests/Support/FMTestHelpers.swift:167`, there's another usage of the old metadata types.

Would you mind fixing that?

abra-code:
@ctymoszek: Good catch, thank you - fixed.

abra-code:
@davidkoski looks like the Xcode 27 verification pipeline is running Xcode 27 beta 4 so I am afraid this change will not succeed until it upgraded to beta 5

davidkoski:
> @davidkoski looks like the Xcode 27 verification pipeline is running Xcode 27 beta 4 so I am afraid this change will not succeed until it upgraded to beta 5

Yeah, we can still approve and merge.  That runs on a github runner so we don't control when it updates.

davidkoski:
@ctymoszek will verify locally and merge

thechriswebb (approved):
Verified locally on Xcode 27.0 beta 5 that no `MLXFoundationModels` errors remain on beta 5 with these changes.

nit: feel free to ignore this but `& Sendable` is required for `GenerationEvent.updateMetadata` and `emitMetadata` but unnecessary for the two `makeExecutorRequest` test helpers. Including it just makes the test helpers stricter than they need to be, but honestly not a big deal.

LGTM

aleroot:
@thechriswebb  the problem with merging this PR is that the hosted xcode-27 runner is still using Xcode 27 beta 4... Consequently, PRs tested against the new main now fail integration_build_xcode27 at the beta-4 metadata signature, for example: #565 and #571 ... 

@davidkoski FYI

davidkoski:
Yes this was known -- the problem without merging is people on current SDK can't build.  Ugh!  Perhaps input for the next time we try to pick up an API that is still moving.

We don't control the SDK on the Xcode runners (GitHub provided).  If this becomes too noisy we will disable the PR check (we still have the nightly).

PR #545, opened 2026-08-17
Add Qwen 3.8 support for text, VLM, and MTP speculative decoding

aleroot:
## Proposed changes

Adds support for **Qwen 3.8** (e.g. `mlx-community/Qwen3.8-27B-4bit` and `mlx-community/Qwen3.8-27B-8bit`) across text LLM, multimodal VLM, and Multi-Token Prediction (MTP) speculative decoding pipelines.

## Checklist

### DIFF
--- a/Libraries/MLXLMCommon/MTPDrafterModel.swift
+++ b/Libraries/MLXLMCommon/MTPDrafterModel.swift
@@ -26,6 +26,10 @@ import MLXNN
 /// instance. Drafters that need their own per-stream state additionally
 /// conform to ``StatefulMTPDrafterModel``.
 public protocol MTPDrafterModel: BaseLanguageModel {
+    /// Whether the drafter can consume hidden states and model-owned values
+    /// from this target architecture.
+    func isCompatible(with target: any LanguageModel) -> Bool
+
     /// Largest total verification block the drafter can produce efficiently.
     /// `nil` means the caller may choose any block size.
     var maximumBlockSize: Int? { get }
@@ -85,6 +89,7 @@ public protocol MTPDrafterModel: BaseLanguageModel {
 }
 
 extension MTPDrafterModel {
+    public func isCompatible(with target: any LanguageModel) -> Bool { true }
     public var maximumBlockSize: Int? { nil }
     public var requiresSharedTargetKV: Bool { true }
     public var requiresPromptPrefill: Bool { false }
--- a/Libraries/MLXVLM/Models/Qwen35.swift
+++ b/Libraries/MLXVLM/Models/Qwen35.swift
@@ -47,6 +47,9 @@ public struct Qwen35Configuration: Codable, Sendable {
         public var headDim: Int?
         public var ropeParameters: [String: StringOrNumber]?
         public var fullAttentionInterval: Int = 4
+        public var layerTypes: [String]?
+        var resolvedLayerTypes: [String] = []
+        var fullAttentionLayerIndex: Int? = nil
         public var mtpNumHiddenLayers: Int = 0
         public var mtpUseDedicatedEmbeddings: Bool = false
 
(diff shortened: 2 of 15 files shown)


thechriswebb:
Ha, I was just about to look into this myself. Not surprised you beat me to it!

Libraries/MLXVLM/Models/Qwen35.swift
```diff
@@ -756,7 +769,7 @@ enum Qwen35Language {
             _norm.wrappedValue = RMSNorm(dimensions: args.hiddenSize, eps: args.rmsNormEps)
 
             self.ssmIdx = 0
```
davidkoski: Does `ssmIdx` need adjusting too?  It looks like the layers can now be configured and with it set at 0 it requires that layer 0 be `isLinear`:

```swift
            let ssmMask = createSSMMask(h: hiddenStates, cache: cacheArray?[ssmIdx] as? MambaCache)
```

and

```swift
        func makeCache(capacity: KVCacheConfiguration.Capacity?) -> [KVCache] {
            model.layers.map { layer in
                if layer.isLinear {
                    return MambaCache()
                }
```

Given the rest of the attention changes, is this still true?

Libraries/MLXVLM/Models/Qwen35.swift
```diff
@@ -756,7 +769,7 @@ enum Qwen35Language {
...
-            self.faIdx = args.fullAttentionInterval - 1
+            self.faIdx = args.fullAttentionLayerIndex ?? (args.fullAttentionInterval - 1)
```
davidkoski: If HybridAttentionSchedule is configured for all linear attention, this will be nil:

```swift
            self.fullAttentionLayerIndex = layerSchedule.firstFullAttentionIndex
```

and

```swift
    package var firstFullAttentionIndex: Int? {
        layerTypes.firstIndex(of: Self.fullAttention)
    }
```

and will fall back to `args.fullAttentionInterval - 1` -- is that still correct?  I am not sure what `faIdx` should be in that case.

davidkoski (changes requested):
Change looks good, but see questions -- those may or may not be real issues.  Thanks!

Libraries/MLXVLM/Models/Qwen35.swift
aleroot: Good catch. yes, ssmIdx = 0 incorrectly assumed that the first layer was always linear attention.

Libraries/MLXLMCommon/MTPDrafterModel.swift
```diff
@@ -26,6 +26,10 @@ import MLXNN
 /// instance. Drafters that need their own per-stream state additionally
 /// conform to ``StatefulMTPDrafterModel``.
 public protocol MTPDrafterModel: BaseLanguageModel {
+    /// Whether the drafter can consume hidden states and model-owned values
+    /// from this target architecture.
+    func isCompatible(with target: any LanguageModel) -> Bool
```
davidkoski, line 31: Does Gemma4AssistantDraftModel need to implement this?  It is the only other MTPDrafterModel.  Is `true` a safe default?  Should it have a default?

davidkoski (changes requested):
See question about Gemma4AssistantDraftModel.  Aside from that I think this looks ready to go.  Thanks!

Libraries/MLXLMCommon/MTPDrafterModel.swift
aleroot: You are right, the unconditional true default was unsafe. Fixed it.

davidkoski (changes requested):
I think this is ready to go -- just needs conflicts resolved on the `ssmIdx` / `faIdx`.  Thanks!

aleroot:
@davidkoski this should be really ready to go now, I see other forks integrating it already, so I guess it is good...

PR #546, opened 2026-08-17
Add Qwen3.5 VLM processor config fallback

ngutech21:
## Proposed changes

Adds a narrowly scoped processor-config fallback for Qwen VLMs (`qwen3_5` and `qwen3_5_moe`) when both processor metadata files are missing. It derives vision geometry from `vision_config` and applies the official Qwen preprocessing defaults.

## Motivation

Some valid Qwen3.6 OptiQ repositories include the complete vision configuration and weights but omit `preprocessor_config.json` and `processor_config.json`, causing VLM loading to fail. This fallback restores vision support for those models while keeping existing files authoritative, preserving malformed-file errors, and avoiding changes to weight or MTP loading.

Includes coverage for dense and MoE models, file precedence, malformed configurations, and unrelated model types.

## Checklist

- [x ] I have updated the necessary documentation (if needed)

### DIFF
--- a/Libraries/MLXVLM/VLMModelFactory.swift
+++ b/Libraries/MLXVLM/VLMModelFactory.swift
@@ -426,9 +426,12 @@ public final class VLMModelFactory: GenericModelFactory {
         // Note: loadProcessorConfig does synchronous I/O but is marked async to enable
         // parallel scheduling. This may briefly block a cooperative thread pool thread,
         // but the config file is small and model loading is not a high-concurrency path.
+        let processorFallback = try qwenProcessorFallback(
+            modelType: baseConfig.modelType, model: model)
         async let tokenizerTask = tokenizerLoader.load(
             from: configuration.tokenizerDirectory)
-        async let processorConfigTask = loadProcessorConfig(from: modelDirectory)
+        async let processorConfigTask = loadProcessorConfig(
+            from: modelDirectory, fallback: processorFallback)
 
         try loadWeights(
             modelDirectory: modelDirectory, model: model,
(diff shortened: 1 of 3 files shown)


Libraries/MLXVLM/VLMModelFactory.swift
```diff
@@ -426,9 +426,12 @@ public final class VLMModelFactory: GenericModelFactory {
         // Note: loadProcessorConfig does synchronous I/O but is marked async to enable
         // parallel scheduling. This may briefly block a cooperative thread pool thread,
         // but the config file is small and model loading is not a high-concurrency path.
+        let processorFallback = try qwenProcessorFallback(
+            modelType: baseConfig.modelType, model: model)
         async let tokenizerTask = tokenizerLoader.load(
             from: configuration.tokenizerDirectory)
-        async let processorConfigTask = loadProcessorConfig(from: modelDirectory)
+        async let processorConfigTask = loadProcessorConfig(
+            from: modelDirectory, fallback: processorFallback)
```
davidkoski, line 434: We shouldn't have model specific code in the model factory.  I think you could get something equivalent by modifying:

```swift
public enum VLMTypeRegistry {
...
        "qwen3_5": create(Qwen35Configuration.self, Qwen35.init),
        "qwen3_5_moe": create(Qwen35Configuration.self, Qwen35MoE.init),
```

This registers the handler for these model types and the `create()` returns a closure that converts Data -> Configuration -> Model.  I think `qwenProcessorFallback()` could be expressed in that way, and since this is a registry, it is extensible to models outside this library.

Also I think the factory for this probably belongs in the Qwen files (perhaps package scoped so it is visible here).

davidkoski (changes requested):
Change looks ok, but see my note on how to integrate this.  Thanks!

Libraries/MLXVLM/VLMModelFactory.swift
davidkoski: Oops, sorry, this is the processor.  Same idea but `VLMProcessorTypeRegistry`

davidkoski: Well, just a sec.  My concern remains but the details of how this works don't match quite what we would need here.

davidkoski (approved):
OK, different approach.  I don't think you can do what you are doing with the existing registries, so filed:

- https://github.com/ml-explore/mlx-swift-lm/issues/547

we can work on improving that later.

In the meantime we can take this.  Thank you!

PR #548, opened 2026-08-17
Add bounded cross-dialect tool-call recovery

### LINKED ISSUE #259: Gemma 4: tool calls never extracted (ToolCallFormat.infer misses "gemma4" + GemmaFunctionParser uses outdated tags)
## Summary

Tool-calling against `mlx-community/gemma-4-26b-a4b-it-8bit` via `VLMModelFactory.shared` runs cleanly but `mlx-swift-lm`'s streaming pipeline never extracts any `ToolCall` events from the generation. The model emits valid Gemma 4 tool-call text (`<|tool_call>call:NAME{...}<tool_call|>`) per its own `chat_template.jinja`, and Python `mlx_vlm` accepts the same weights, but in Swift the tool calls flow through as `.chunk` text and `toolCalls` ends up empty.

Two compounding causes:

1. **`ToolCallFormat.infer(from:)` never matches Gemma 4.** `Libraries/MLXLMCommon/Tool/ToolCallFormat.swift:174` does `if type == "gemma" { return .gemma }` — exact equality. Gemma 4's `config.json` has `model_type: "gemma4"`, so inference returns `nil` and the runtime falls back to the default `.json` parser. Other families in the same function use `hasPrefix(...)` (`lfm2`, `glm4`, `nemotron`, `qwen3_5`, `qwen3_next`, `mistral3`); the Gemma branch is the only exact-equality one.

2. **`GemmaFunctionParser` uses tags from the older Gemma format.** Even with `.gemma` forced (e.g. via `context.configuration.toolCallFormat = .gemma` post-load), `Libraries/MLXLMCommon/Tool/Parsers/GemmaFunctionParser.swift:8-9` declares `startTag = "<start_function_call>"` and `endTag = "<end_function_call>"`. Gemma 4's actual chat template renders tool calls as `<|tool_call>call:NAME{...}<tool_call|>` (asymmetric `stc_token` / `etc_token` special tokens defined in `tokenizer_config.json`). `ToolCallProcessor.processTaggedChunk` watches for `startTag.first` and `partialMatch`es the prefix; for Gemma 4's emission the second character mismatches (`|` vs `s`), state resets to `.normal`, and the buffer flushes as plain text. EOS is also a no-op since `processEOS()` only runs when state is non-`.normal`.

Net effect: every tool-using turn against Gemma 4 returns with `stopReason == .stop`, `toolCalls == []`, and the tool-call text intact in the regular generated text.

## Reproducer

Standalone SwiftPM package (already exists from a prior issue): https://github.com/angelsbrood/Gemma4SwiftRepro

`ExperimentC` in that repo demonstrates the bug end-to-end. It loads `gemma-4-26b-a4b-it-8bit` via `VLMModelFactory.shared`, builds a `UserInput(chat:tools:)` with a system + user message and two `ToolSpec` definitions, and prints every stream event:

```swift
for await event in stream {
    switch event {
    case .chunk(let text):
        // Receives Gemma 4's `<|tool_call>call:insert_section{...}<tool_call|>` text here
        print("CHUNK: \(text)", terminator: "")
    case .toolCall(let call):
        // Never fires
        print("\nTOOL CALL: \(call.function.name) args=\(call.function.arguments)")
    case .info(let info):
        print("\nINFO: stop=\(info.stopReason)")
    }
}
```

Observed output (truncated): the `<|tool_call>...<tool_call|>` text appears verbatim in `.chunk` events; no `.toolCall` events are ever emitted; `info.stopReason == .stop`.

## Suggested fix

Two-line change covers most cases:

**`Libraries/MLXLMCommon/Tool/ToolCallFormat.swift:174`** — match Gemma family by prefix to align with the other families in the same function:

```swift
// before
if type == "gemma" {
    return .gemma
}

// after
if type.hasPrefix("gemma") {
    return .gemma
}
```

**`Libraries/MLXLMCommon/Tool/Parsers/GemmaFunctionParser.swift:8-9`** — update tags to match Gemma 4's actual emission:

```swift
// before
public let startTag: String? = "<start_function_call>"
public let endTag: String? = "<end_function_call>"

// after (or via a separate Gemma4FunctionParser if back-compat with older Gemma matters)
public let startTag: String? = "<|tool_call>"
public let endTag: String? = "<tool_call|>"
```

The `parse(content:tools:)` body itself works against Gemma 4's `call:NAME{...}` format unchanged — the wrapper-tag strip on lines 17-23 is no-op when the tags don't appear in the buffer (the streaming detection wraps the content before parse), and the `call:`-anchored extraction is the same. Only the wrapper changes.

If Gemma 1/2/3 with `<start_function_call>` is still supported and tag back-compat matters, adding a separate `Gemma4FunctionParser` and dispatching from `ToolCallFormat.createParser()` based on a sub-format would be the cleaner shape.

## Related — chat-template `role:tool` handling

Slightly downstream from this report but worth flagging in case it's the same maintainer surface: Gemma 4's `chat_template.jinja` line 215 wraps the entire message-rendering loop body in `if message['role'] != 'tool'`. `role:tool` messages are skipped at the top level — they only render via a forward-scan from an assistant message that carries a structured `tool_calls` field. The "Legacy" path needs `message['tool_responses']` on the assistant; the "OpenAI" path needs `message['tool_calls']` on the assistant.

`MLXLMCommon.Chat.Message` (Libraries/MLXLMCommon/Chat.swift:4) is `role + content + images + videos` — no `tool_calls` or `tool_responses` field exposed. So a tool-use loop that builds history as `[.user(prompt), .assistant(text-with-encoded-tool-call), .tool(result), …]` produces a flattened message stream where every `role:tool` message is silently dropped by the template, and the model keeps re-emitting the same tool call until it max-iterates.

We worked around this downstream by inlining tool-response markup into the assistant message text body (the template's `strip_thinking(message['content'])` rendering then surfaces it verbatim, mirroring the native legacy-path output). But it would be cleaner if `Chat.Message` exposed structured tool-call/tool-response fields and the message generators knew how to render them via the template's intended paths. Lower priority than the parser-tag fix above; flagging since it's adjacent.

## Environment

- **mlx-swift-lm:** `3.31.3` (resolved revision: `1c05248bb0899e2a7a4962b84d319cf12f4e12aa`)
- **mlx-swift:** `0.31.3` (resolved revision: `61b9e011e09a62b489f6bd647958f1555bdf2896`)
- **swift-transformers:** `1.3.0`
- **Swift toolchain:** `Apple Swift version 6.3.1 (swiftlang-6.3.1.1.2 clang-2100.0.123.102)`, target `arm64-apple-macosx26.0`
- **Xcode:** `26.4.1`
- **macOS:** `26.4.1`
- **Hardware:** Apple M5 Pro, 18-core CPU, 64 GB RAM
- **Model:** `mlx-community/gemma-4-26b-a4b-it-8bit`, snapshot `1382fb7268fa62c07b83ffa174ff6fa455fb0686`

## Notes

- This is the same downstream-app + repro environment as [#258](https://github.com/ml-explore/mlx-swift-lm/issues/258) (the rep-penalty broadcast crash). Filing this as a separate issue because the trigger and fix surface are independent.
- Haven't verified whether the 4-bit Gemma 4 variant (if it exists at the same architecture point) hits the same paths the same way; same repro should reproduce against any Gemma 4 quant since the tags + `model_type` come from the chat template + `config.json`, not the weights.

aleroot:
## Proposed changes

This PR follows #531 and adds bounded, streaming recovery for textual tool calls that the selected native parser does not recognize. Fixes #259 

Some checkpoints can emit valid tool calls using a common alternate textual dialect. Previously, these calls could pass through as ordinary response text even when the function was explicitly declared by the caller.

The recovery layer supports:

- framed `<tool_call>…</tool_call>` JSON or XML calls
- `[TOOL_CALLS]name{...}` and `name[ARGS]{...}`
- Qwen-style `<function=name>…</function>` calls
- Gemma `<|tool_call>…<tool_call|>` calls

The native parser remains authoritative. Recovery only promotes a structurally complete call whose function name exactly matches a declared tool. It is disabled when no nonempty tool declaration is provided.

Before a native or recovered call becomes executable, its arguments are checked against the declared tool schema. Validation is deliberately conservative: supported constraints reject only definite violations, while unsupported, malformed, or ambiguous schema features produce an unknown result and remain fail-open.

This prevents clearly invalid arguments such as missing required properties, incorrect types, unexpected properties, invalid enum values, or out-of-range values from reaching application tools without claiming complete JSON Schema support. Rejections use the existing `invalidArguments` reason with bounded, path-specific diagnostics that do not include argument values.

Schema validation is applied consistently across standard tool formats, Harmony, Onyx, and recovered cross-dialect calls.

The implementation also:

- preserves thinking and response text in its original order while extracting embedded calls
- treats native protocol frames and ordinary JSON responses as opaque
- bounds incomplete recovery input to 64 KiB of UTF-8 data
- resets recovery state at EOS
- uses a delimiter-gated ordinary-text fast path with fixed-string structural scanning, avoiding regular expressions and speculative JSON decoding
- keeps schema validation package-scoped with explicit `valid`, `invalid`, and `unknown` outcomes
- prevents rejected candidates from being dispatched or counted as successfully recovered calls
- bounds validation depth and diagnostic output
- preserves existing public rejection semantics and parser entry points

Together, these changes make tool use more reliable and safer: alternate valid tool-call formats can be recovered, while calls that definitively violate their declared contract are stopped before execution.

## Checklist

### DIFF
--- a/Libraries/MLXLMCommon/ChatSession.swift
+++ b/Libraries/MLXLMCommon/ChatSession.swift
@@ -315,6 +315,12 @@ public final class ChatSession {
 
     public var additionalContext: [String: any Sendable]?
     public var tools: [ToolSpec]?
+
+    /// Optional automatic dispatcher for accepted tool calls.
+    ///
+    /// Automatic dispatch is fail-closed: only calls naming a function declared
+    /// in ``tools`` can reach this callback. If `tools` is `nil` or empty, every
+    /// tool-call-shaped model output is rejected without invoking the callback.
     public var toolDispatch: (@Sendable (ToolCall) async throws -> String)?
 
     /// Speculative decoding configuration, nil if disabled.
@@ -898,6 +904,12 @@ public final class ChatSession {
                 speculativeDecoding
             ] in
             do {
+                // Automatic dispatch must always be schema-authorized. Treat a missing
+                // declaration list as an empty allowlist when a dispatcher is installed,
+                // rather than allowing the processor's schema-less parsing mode.
+                let toolValidationSchemas: [ToolSpec]? =
+                    toolDispatch == nil ? tools : (tools ?? [])
+
                 try await cache.update { cache in
 
                     // these are all Sendable
@@ -1201,7 +1213,8 @@ public final class ChatSession {
                                     modelConfiguration: modelConfiguration,
                                     tokenizer: tokenizer,
                                     iterator: iterator,
-                                    tools: tools)
+                                    tools: toolValidationSchemas,
+                                    recoveryPolicy: generateParameters.toolCallRecoveryPolicy)
                             )
                         }
 
@@ -1324,7 +1337,9 @@ public final class ChatSession {
                                             modelConfiguration: modelConfiguration,
                                             tokenizer: tokenizer,
                                             iterator: iterator,
-                                            tools: tools))
+                                            tools: toolValidationSchemas,
+                                            recoveryPolicy: generateParameters
+                                                .toolCallRecoveryPolicy))
                                 }
                             }
                         } else {
--- a/Libraries/MLXLMCommon/ChatSession.swift
+++ b/Libraries/MLXLMCommon/ChatSession.swift
@@ -1214,7 +1214,7 @@ public final class ChatSession {
                                     tokenizer: tokenizer,
                                     iterator: iterator,
                                     tools: toolValidationSchemas,
-                                    recoveryPolicy: generateParameters.toolCallRecoveryPolicy)
+                                    toolCallPolicy: generateParameters.toolCallPolicy)
                             )
                         }
 
@@ -1338,8 +1338,7 @@ public final class ChatSession {
                                             tokenizer: tokenizer,
                                             iterator: iterator,
                                             tools: toolValidationSchemas,
-                                            recoveryPolicy: generateParameters
-                                                .toolCallRecoveryPolicy))
+                                            toolCallPolicy: generateParameters.toolCallPolicy))
                                 }
                             }
                         } else {
--- a/Libraries/MLXLMCommon/Tool/TextToolCallRecoveryScanner.swift
+++ b/Libraries/MLXLMCommon/Tool/TextToolCallRecoveryScanner.swift
@@ -1240,16 +1240,13 @@ struct TextToolCallRecoveryScanner: Sendable {
             call = parseMistralStyle(raw)
         }
 
-        guard let call, allowedToolNames.contains(call.function.name),
-            satisfiesDeclaredRequiredArguments(call)
-        else { return nil }
+        guard let call, allowedToolNames.contains(call.function.name) else { return nil }
         return call
     }
 
     private func recoverDeclaredArgs(_ raw: String, name: String) -> ToolCall? {
         guard raw.hasPrefix(name + "[ARGS]"),
-            let call = parseMistralStyle(raw, expectedName: name),
-            satisfiesDeclaredRequiredArguments(call)
+            let call = parseMistralStyle(raw, expectedName: name)
         else { return nil }
         return call
     }
@@ -1267,23 +1264,6 @@ struct TextToolCallRecoveryScanner: Sendable {
             raw[...],
             tools: tools,
             allowMissingFunctionClose: allowMissingFunctionClose)
-    }
-
-    /// A recovered call must satisfy the declared schema's required arguments
-    /// before promotion; an empty-arguments promotion for a tool that declares
-    /// required parameters is never valid recovery.
-    private func satisfiesDeclaredRequiredArguments(_ call: ToolCall) -> Bool {
-        for tool in tools {
-            guard let function = tool["function"] as? [String: any Sendable],
-                function["name"] as? String == call.function.name
-            else { continue }
-            guard let parameters = function["parameters"] as? [String: any Sendable],
-                let required = parameters["required"] as? [any Sendable]
-            else { return true }
-            let requiredNames = required.compactMap { $0 as? String }
-            return requiredNames.allSatisfy { call.function.arguments[$0] != nil }
-        }
-        return true
     }
 
     private func rejection(
--- a/Libraries/MLXLMCommon/Tool/TextToolCallRecoveryScanner.swift
+++ b/Libraries/MLXLMCommon/Tool/TextToolCallRecoveryScanner.swift
@@ -6,7 +6,8 @@ import Foundation
 ///
 /// The selected `ToolCallParser` remains authoritative. This scanner only owns
 /// dialect markers that the selected parser cannot consume, plus exact
-/// `declaredTool[ARGS]{...}` rehearsals. It deliberately uses fixed-string
+/// `declaredTool[ARGS]{...}` rehearsals under the permissive policy. It
+/// deliberately uses fixed-string
 /// searches and structural JSON scanning instead of regular expressions or
 /// speculative JSON decoding on ordinary response text.
 ///
@@ -217,7 +218,7 @@ struct TextToolCallRecoveryScanner: Sendable {
                 prefixes.insert(String(signal.text.prefix(length)))
             }
         }
-        if policy != .disabled {
+        if policy == .permissive {
             for name in allowedToolNames {
                 let marker = name + "[ARGS]"
                 for length in 1 ..< marker.count {
@@ -646,8 +647,13 @@ struct TextToolCallRecoveryScanner: Sendable {
     /// heals Mistral's observed `[TOOL_CALLS]name{json}` variant.
     mutating func recoverEOSPayloads(_ raw: String) -> [ToolCall] {
         guard policy != .disabled else { return [] }
-        if raw.contains("[TOOL_CALLS]") {
-            return raw.components(separatedBy: "[TOOL_CALLS]").compactMap { segment in
+        if let firstMarker = raw.range(of: "[TOOL_CALLS]") {
+            // Only text that actually followed a `[TOOL_CALLS]` marker may be
+            // re-framed as a Mistral call. Re-prefixing the prose before the
+            // first marker would fabricate a frame the model never emitted
+            // and could promote plain response text shaped like `name{json}`.
+            let framed = raw[firstMarker.lowerBound...]
+            return framed.components(separatedBy: "[TOOL_CALLS]").compactMap { segment in
                 guard !segment.trimmingCharacters(in: .whitespacesAndNewlines).isEmpty else {
                     return nil
                 }
@@ -699,7 +705,11 @@ struct TextToolCallRecoveryScanner: Sendable {
             }
         }
 
-        guard policy != .disabled else { return earliest }
+        // Markerless `name[ARGS]{...}` rehearsals carry no protocol marker,
+        // so prose that merely mentions a declared call is indistinguishable
+        // from an intended call. Promotion of this dialect is opt-in via the
+        // permissive policy; the conservative default leaves it as text.
+        guard policy == .permissive else { return earliest }
 
         var argsSearchStart = text.startIndex
         while argsSearchStart < text.endIndex,
--- a/Libraries/MLXLMCommon/Tool/TextToolCallRecoveryScanner.swift
+++ b/Libraries/MLXLMCommon/Tool/TextToolCallRecoveryScanner.swift
@@ -1263,22 +1263,10 @@ struct TextToolCallRecoveryScanner: Sendable {
         _ raw: String,
         allowMissingFunctionClose: Bool
     ) -> ToolCall? {
-        guard
-            case .complete(let payload) = QwenXMLPayloadScanner.scan(
-                raw[...], allowMissingFunctionClose: allowMissingFunctionClose),
-            raw[payload.end...].allSatisfy(\.isWhitespace)
-        else { return nil }
-
-        var arguments: [String: any Sendable] = [:]
-        for parameter in payload.parameters {
-            var value = String(parameter.value)
-            // Trim a single leading/trailing newline (matching XMLFunctionParser).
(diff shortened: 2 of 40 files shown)


Libraries/MLXLMCommon/ChatSession.swift
```diff
@@ -898,6 +904,12 @@ public final class ChatSession {
                 speculativeDecoding
             ] in
             do {
+                // Automatic dispatch must always be schema-authorized. Treat a missing
+                // declaration list as an empty allowlist when a dispatcher is installed,
+                // rather than allowing the processor's schema-less parsing mode.
```
davidkoski, line 909: This seems like a good change.  Per the doc on `toolDispatch` you can still get unverified tool calls at the stream level if you want (e.g. to print them).

aleroot:
While stress-testing the recovery layer for false positives, I found two ways ordinary response text could still be promoted into a call: the end-of-stream `[TOOL_CALLS]` repair split on the marker and re-prefixed it, which meant prose *before* the first marker got a frame the model never emitted; and markerless `name[ARGS]{...}` had no protocol marker at all to anchor intent, so a model narrating "I could call search[ARGS]{...}" looked identical to a real call.

The same pass surfaced the opposite failure too: valid calls being rejected. Fixed everything with the last commit. 
I think this is now ready for intensive usage.

aleroot:
Fixes also https://github.com/ml-explore/mlx-swift-lm/issues/609 , if this lands the new #610 is superseded.

Libraries/MLXLMCommon/Tool/TextToolCallRecoveryScanner.swift
```diff
@@ -0,0 +1,1389 @@
...
+        self.maximumBufferedByteCount = maximumBufferedByteCount
+
+        var signals: [Signal] = []
+        if Self.primaryOwnsToolCallFrame(primaryFormat) {
+            signals.append(
+                Signal(
+                    text: "<tool_call>",
+                    kind: .nativeFrame(endMarker: "</tool_call>")))
+        } else if policy != .disabled {
+            signals.append(Signal(text: "<tool_call>", kind: .explicit(.toolCallFrame)))
```
davidkoski: For glm4 in the default policy (conservative) this will register `<tool_call>` for _recovery_.  See glm4NativeFrameWithTools in attached tests.

Libraries/MLXLMCommon/Tool/TextToolCallRecoveryScanner.swift
```diff
@@ -0,0 +1,1389 @@
...
+
+        self.primaryFormat = primaryFormat
+        self.policy = policy
+        self.tools = tools
+        self.allowedToolNames = allowedToolNames
+        self.allowedToolNamesByLength = allowedToolNames.sorted { $0.count > $1.count }
+        self.maximumBufferedByteCount = maximumBufferedByteCount
+
+        var signals: [Signal] = []
+        if Self.primaryOwnsToolCallFrame(primaryFormat) {
```
davidkoski: See below -- potentially this could instantiate the parser for the format and see if the tags match.

Libraries/MLXLMCommon/Tool/TextToolCallRecoveryScanner.swift
```diff
@@ -0,0 +1,1389 @@
...
+    private mutating func appendProtectedText(_ text: String, to output: inout [Output]) {
+        guard !text.isEmpty else { return }
+        output.append(.protectedText(text))
+        previousSourceCharacter = text.last
+    }
+
+    /// Bare top-level JSON remains eligible only when JSON is the selected
+    /// native tool-call format. For every other dialect it is response data and
+    /// must bypass native parsing, including protocol-shaped strings inside it.
+    private mutating func appendJSONData(_ text: String, to output: inout [Output]) {
```
davidkoski, line 763: From Llama3ToolCallParser:

```
/// Parser for Llama 3 tool calls.
/// Llama 3 often outputs inline JSON without standard start/end tags, or preceded by `<|python_tag|>`.
/// It may also output native python function calls like `get_weather(location="San Francisco")`.
```

I wonder if the parser should be able to opt-in to this behavior rather than encoding it here?

See `llama3BareJSONWithTools`

Libraries/MLXLMCommon/Tool/TextToolCallRecoveryScanner.swift
```diff
@@ -0,0 +1,1389 @@
...
+            }
+        }
+
+        // Ordinary JSON values and Markdown code are data: protocol-shaped
+        // text inside them must never be promoted. Single-character signals
+        // are validated structurally when matched, so prose punctuation that
+        // cannot begin a JSON value or code span falls through as text.
+        signals.append(Signal(text: "{", kind: .opaqueJSONValue))
+        signals.append(Signal(text: "[", kind: .opaqueJSONValue))
+        signals.append(Signal(text: "\"", kind: .opaqueJSONValue))
```
davidkoski, line 192: This is potentially a problem since single `"` are possible in legit text:

- inches
- seconds
- arc seconds in lat/lon
- closing quote in German style quotes

I am not sure what the correct handling is, but an unpaired double quote can consume legit tool calls.  See strayQuoteBeforeNativeFrame

Libraries/MLXLMCommon/Tool/ToolCallProcessor.swift
```diff
@@ -802,28 +1000,42 @@ public class ToolCallProcessor {
...
                 callID: call.id,
                 detail: RejectedToolCall.Reason.undeclaredTool.diagnosticDetail)
-            return
+            return false
+        }
+
+        if case .invalid(let violations) = ToolSchemaValidator.validate(
+            arguments: call.function.arguments,
+            forToolNamed: call.function.name,
+            in: tools)
```
davidkoski: There are some differences in tool call validation vs python.  Maybe intended, maybe not.  I don't think there is a way to opt-out (if you want looser validation).

For JSON tool calls the python code simply parses the JSON and returns it -- no validation.

Using a string where an int was desired just passed through.  That might be desired.  I am not sure if typical tool calling handlers would deal with it.  Using a double (e.g. 1.0) where an int was desired also passes through and that _probably_ works fine on the python side.

For xml tool calls, the reference implementation seems to be qwen3_coder.py in mlx_lm:

```python
def _convert_param_value(param_value: str, param_name: str, param_config: dict) -> Any:
    """Convert parameter value based on its type in the schema."""
...
    elif (
        param_type.startswith("int")
        or param_type.startswith("uint")
        or param_type.startswith("long")
        or param_type.startswith("short")
        or param_type.startswith("unsigned")
    ):
# dkoski - wants an int, have a float but the float has no fractional part, coerce to an int
        float_param_value = float(param_value)
        int_param_value = int(float_param_value)
        if float_param_value - int_param_value != 0:
            raise ValueError(f"Invalid integer literal {param_value!r}")
        return int_param_value
...
```

This will actually coerce the 1.0 to an int for xml.

On the XML side, asking for a list or dict but giving bare values does (same function as above):

- tried to parse as JSON and return that (without regard for the type)
- parse it as python text (e.g. `"1", "2"` would turn into a tuple of strings)
- return the raw value

This now fails validation (and indeed it is invalid) but does differ from the python side which tries to recover it.

I don't know if we _want_ it to handle all of these, but given the code on the recovery side that tries to help out troubled models, we might.

See stringifiedIntegerArgument, integerValueDiffersByDialect, nonJSONArrayArgument

davidkoski (changes requested):
Overall looks good, but some questions/comments on different behavior.

I am trying something new here -- see if this is useful and let me know.  I had claude make repro cases for all of these to help me explore them and maybe be useful to you:

[ReproTests.swift.zip](https://github.com/user-attachments/files/32134817/ReproTests.swift.zip)

let me know if that helps -- I can try and generate these when it makes sense in the future.

Libraries/MLXLMCommon/Tool/ToolCallProcessor.swift
aleroot: @davidkoski Thanks for those tests, nice idea to ask the agent to generate those tests! I started my own agent with Qwen 😀 and asked to refactor the code the way I wanted so now there are more changes: I pushed fixes for native-format parsing, quotes/brackets in prose, and argument normalization. I also grouped recovery and validation under ToolCallPolicy to keep configuration consistent and easier to maintain.


Schema validation is now permissive by default, with strict validation opt-in. Declared-tool authorization remains enforced.

Side note for context: this feature is something I originally planned to be inside my app, but then open sourced because it really belongs here at this layer, but probably wasn't the smartest move for me as anyone now building a coding agent can take advantage of this, and not many have this... so in a nutshell it is not really comparable with mlx-lm... 

Could you retest your examples against the latest commit? 
Thanks.

aleroot:
> Overall looks good, but some questions/comments on different behavior.
> 
> I am trying something new here -- see if this is useful and let me know. I had claude make repro cases for all of these to help me explore them and maybe be useful to you:
> 
> [ReproTests.swift.zip](https://github.com/user-attachments/files/32134817/ReproTests.swift.zip)
> 
> let me know if that helps -- I can try and generate these when it makes sense in the future.

Replied in another thread, anyway thanks for the tests suite.

Libraries/MLXLMCommon/Tool/ToolCallProcessor.swift
davidkoski: Changes look good!  I really appreciate the work you have done on tool calling kvcache and more -- hopefully the community takes advantage of what you have contributed!

davidkoski (approved):
Changes look good, thank you!

Libraries/MLXLMCommon/Tool/ToolCallProcessor.swift
aleroot: @davidkoski thank you very much, this pull request is what I need for my next release. appreciate it.

PR #549, opened 2026-08-18
Fix Qwen3VL prompt cache reuse for text-only inputs

ngutech21:
## Proposed changes

- Stop attaching a redundant all-ones attention mask to text-only `Qwen3VLProcessor` inputs.
- Return `LMInput(tokens:)` when no images or videos are present.
- Preserve the existing attention mask and media payload behavior for multimodal inputs.
- Add regression tests for both text-only and image inputs.

## Why

For a single unpadded text sequence, the all-ones mask does not change attention or position semantics. However, `ChatSession` interprets any explicit attention mask as unsafe for prompt-cache suffix reuse, causing the complete prompt to be prefetched again on every continuation.

Removing the redundant mask allows text-only conversations using the VLM loading path to reuse their cached prompt prefix.

## Checklist


davidkoski (approved):
Nice find, thank you!

PR #550, opened 2026-08-18
Fix compiled decoding with fixed-capacity KV cache

### LINKED ISSUE #406: KVCacheSimple.offset (Swift Int) breaks MLX.compile() decode — cache write position frozen across steps
## Summary

`MLX.compile()`-wrapped B=1 decode fails correctness when using the standard compile-with-mutable-state pattern (`inputs:` / `outputs:` on KV cache `innerState()`). Token sequences enter repeating loops; EOS is never reached.

Root cause: `KVCacheSimple.update()` in `Libraries/MLXLMCommon/KVCache.swift` uses a Swift `Int` for `offset` and for MLXArray slice bounds. Those integers are embedded as **constants** at compile-trace time and do not update across decode steps when the compiled graph is reused.

## Observed behavior

Consumer: [macprovider](https://github.com/Augustas11/macprovider) (MLX provider runtime), pin **mlx-swift-lm 3.31.4** (`bd4b743`).

Model: `Qwen2.5-7B-Instruct-4bit`, greedy decode (`temperature=0`), `--prefill-tokens 512 --decode-tokens 64`.

| Path | Result |
|------|--------|
| Uncompiled `TokenIterator` / `generate()` | 42 tokens, EOS hit (ID 151645) |
| `MLX.compile()` decode step | 64 tokens (max), **4-token cycle repeats**, EOS never hit |

Example compiled-path token IDs (steps 1–13): `3535, 11, 432, 4977, 1075, 11, 432, 4977, 1075, 11, 432, 4977, 1075, …`

## Root cause

In `KVCacheSimple.update()`:

```swift
let previous = self.offset          // Swift Int → constant in compiled graph
self.offset += keys.dim(2)
self.keys?[.ellipsis, previous ..< self.offset, 0...] = keys
let returnedKeys = self.keys![.ellipsis, ..<self.offset, 0...]
```

We thread cache state through compile using the documented pattern:

```swift
MLX.compile(inputs: updatables, outputs: updatables, shapeless: false) { token in
    model(token, cache: cache)
}
```

where `updatables` wrap each `KVCache` and forward `innerState()` → `[keys, values]`.

**Problem:** `innerState()` does not include `offset`, and slice indices derived from Swift `Int offset` are not graph nodes. With `shapeless: false`, recompilation triggers on **input array shape** changes, not on integer-constant changes. After the first decode step:

1. **Write position is frozen** at the offset from trace time (every step overwrites the same slot).
2. **Attention window is frozen** at the length from trace time.

This cannot be fixed in a downstream consumer without forking KV cache behavior.

## Suggested fix directions (maintainer choice)

1. **Graph-traceable offset** — store/read `offset` as an `MLXArray` scalar so index arithmetic participates in the graph.
2. **Functional cache step** — `update()` returns updated keys/values/offset tensors suitable as compile I/O.
3. **Dedicated compile-friendly cache type** — e.g. alongside `KVCacheSimple`, for B=1 compiled decode.

Other cache types in the same file that slice on Swift `Int offset` may need the same treatment (rotating/windowed variants).

## Repro sketch

Minimal pattern (after prefill populates cache):

```swift
let updatables: [any Updatable] = cache.map { /* forward innerState() */ }
let compiled = MLX.compile(inputs: updatables, outputs: updatables, shapeless: false) { token in
    model(token, cache: cache)
}
// Repeated compiled(token) calls → wrong tokens when offset uses Swift Int slices
```

Full consumer repro + token trace: macprovider `beta/throughput-engineering/T2-01-compiled-decode-wire-in.md` (merged with `MACPROVIDER_COMPILED_DECODE=1` scaffolding, flag default OFF).

## Impact

Blocks compiled per-token decode optimization for all models using `KVCacheSimple` (Qwen, Llama, etc.). Consumer keeps compile flag **OFF** until upstream fix lands.

Happy to contribute a PR once maintainers indicate preferred API direction.

aleroot:
## Proposed changes

Fixes #406.

This PR fixes incorrect token generation when autoregressive decoding is wrapped in `MLX.compile`. The failure could cause generated text to enter repeating-token loops even though the same model decoded correctly without compilation.

### Root cause

During generation, a model keeps a KV cache containing information from tokens it has already processed. Each new token must be written into the next position in that cache.

The existing `KVCacheSimple` stores this write position as a Swift `Int`. When MLX compiles the decode function, it traces the function once and reuses the resulting computation graph. The Swift integer and the array slices derived from it are captured as constants during that trace.

As a result, later decode calls can continue using the position observed during compilation instead of the current position. In simpler terms, the compiled decoder keeps writing to and reading from the wrong place in its memory. This is the root cause of the repeating generation reported in #406.

### How this PR fixes it

This PR introduces `FixedCapacityKVCache`, a cache designed for reusable compiled graphs:

- It allocates the complete cache shape before compilation, keeping tensor shapes stable between tokens.
- It stores the current write position as an `MLXArray` instead of a Swift integer.
- The position is passed through the compiled graph as both input and output, so it advances on every call.
- New keys and values are written using tensor-based indices rather than Swift array slices.
- Unused cache positions are masked so the model cannot attend to memory that has not been written yet.

A `CompiledDecodeSession` owns the cache and provides the safe high-level interface. It:

- Performs prompt processing eagerly before compiling the single-token decode step.
- Verifies prompt and token shapes.
- Tracks capacity on the Swift side.
- Prevents an out-of-range cache write before it reaches the compiled graph.

### Why this is a separate cache

`KVCacheSimple` remains the better default for normal, non-compiled generation. It grows as needed and only performs attention over tokens that have actually been written.

A compiled graph benefits from the opposite behavior: fixed shapes and tensor-managed state. `FixedCapacityKVCache` therefore allocates the requested capacity up front and performs attention over that capacity, masking unused positions. This uses more attention work when the cache is only partially filled, but it makes compiled decoding correct and allows the graph to be reused.

Keeping these implementations separate avoids making ordinary eager generation slower to solve a compilation-specific problem.

### Supported models

Support is capability-based through `FixedCapacityKVCacheProviding`. A model should conform only after its position handling and cache topology have been audited for compiled execution.

This PR initially enables:

- Qwen 2 and Qwen 2.5 text models through `Qwen2Model`.
- Llama and the Mistral checkpoints implemented by `LlamaModel`.

Models with sliding-window, recurrent, hybrid, vision-language, or custom cache layouts are not assumed to be compatible.

### Testing

The new tests cover:

- Compiled versus eager logit and token parity.
- Qwen2 and Llama model paths.
- Cache position advancement through repeated compiled calls.
- Prompt prefill and multi-token continuation.
- Causal and unused-capacity masking.
- Cache copying, trimming, and serialization.
- Invalid shapes and capacity exhaustion.
- An opt-in regression test using the Qwen2.5-7B checkpoint from #406.

The Qwen2.5-7B regression was also run locally for 16 decode steps. Compiled and eager decoding produced matching logits and selected tokens at every step.

### Longer-term direction

This issue exposes a broader distinction between ordinary Swift state and state that must change inside a reusable MLX computation graph.

Long term, compiled generation should probably move toward:

- Representing all decode-time state that affects computation as tensors carried through the graph.
- Making compiled-decoding support an explicit model capability rather than assuming every cache and model architecture is compatible.
- Keeping compilation lifecycle, prompt prefill, cache ownership, and capacity validation behind a session-style API.
- Adding compiled-versus-eager parity tests whenever a new model family adopts this capability.
- Exploring more efficient fixed-shape or paged cache strategies later, without weakening the correctness guarantees introduced here.

## Checklist

### DIFF
--- a/Libraries/MLXLLM/Models/Llama.swift
+++ b/Libraries/MLXLLM/Models/Llama.swift
@@ -150,7 +150,9 @@ public class LlamaModelInner: Module {
 }
 
 /// Model for Llama and Mistral model types.
-public class LlamaModel: Module, LLMModel, KVCacheDimensionProvider {
+public class LlamaModel: Module, LLMModel, KVCacheDimensionProvider,
+    FixedCapacityKVCacheProviding
+{
 
     public let vocabularySize: Int
     public let kvHeads: [Int]
--- a/Libraries/MLXLLM/Models/Qwen2.swift
+++ b/Libraries/MLXLLM/Models/Qwen2.swift
@@ -164,7 +164,9 @@ public class Qwen2ModelInner: Module {
     }
 }
 
-public class Qwen2Model: Module, LLMModel, KVCacheDimensionProvider {
+public class Qwen2Model: Module, LLMModel, KVCacheDimensionProvider,
+    FixedCapacityKVCacheProviding
+{
     public let vocabularySize: Int
     public let kvHeads: [Int]
 
--- a/Libraries/MLXLMCommon/CompiledDecodeSession.swift
+++ b/Libraries/MLXLMCommon/CompiledDecodeSession.swift
@@ -1,0 +1,121 @@
+// Copyright © 2026 Apple Inc.
+
+import Foundation
+import MLX
+
+/// Errors raised before a compiled decode step can issue an invalid cache write.
+public enum CompiledDecodeSessionError: Error, LocalizedError, Equatable {
+    case invalidCapacity(Int)
+    case invalidPromptShape([Int])
+    case emptyPrompt
+    case promptExceedsCapacity(promptTokens: Int, capacity: Int)
+    case invalidTokenShape([Int], expectedBatchSize: Int)
+    case capacityExceeded(capacity: Int)
+
+    public var errorDescription: String? {
+        switch self {
+        case .invalidCapacity(let capacity):
+            "Compiled decode capacity must be positive and fit in Int32; got \(capacity)."
+        case .invalidPromptShape(let shape):
+            "Compiled decode requires prompt tokens shaped [batch, sequence]; got \(shape)."
+        case .emptyPrompt:
+            "Compiled decode requires at least one prompt token."
+        case .promptExceedsCapacity(let promptTokens, let capacity):
+            "Prompt length \(promptTokens) exceeds compiled decode capacity \(capacity)."
+        case .invalidTokenShape(let shape, let expectedBatchSize):
+            "Compiled decode requires one token per row with shape [\(expectedBatchSize), 1]; got \(shape)."
+        case .capacityExceeded(let capacity):
+            "Compiled decode capacity \(capacity) is exhausted."
+        }
+    }
+}
+
+/// Owns a fixed-capacity cache and a compiled, capacity-safe decode step.
+///
+/// Construction performs prefill eagerly, evaluates the allocated cache state,
+/// and only then creates the compiled closure. ``step(_:)`` accepts exactly one
+/// token per batch row and checks capacity on the host before dispatching the
+/// graph. This prevents out-of-range `putAlong` writes without synchronizing the
+/// graph-valued cache position on every decode step.
+///
+/// This session supports text-only, same-length batches. Keep it within the same
+/// serialized model access that owns `model`; it is intentionally not `Sendable`.
+public final class CompiledDecodeSession {
+    public let capacity: Int
+    public let batchSize: Int
+    public let prefillLogits: MLXArray
+
+    /// Number of prompt and decode-input tokens consumed by the model.
+    public private(set) var processedTokenCount: Int
+
+    /// Number of additional single-token decode calls that are safe.
+    public var remainingCapacity: Int {
+        capacity - processedTokenCount
+    }
+
+    /// Per-layer positions for diagnostics. Reading this synchronizes the
+    /// graph-valued positions to the host; do not poll it in a decode loop.
+    public var cacheOffsets: [Int] {
+        caches.map(\.offset)
+    }
+
+    private let caches: [FixedCapacityKVCache]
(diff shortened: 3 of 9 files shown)


Augustas11:
Thanks for putting this together. I tested PR head `edcda687fb4a2a61ae6028e2ff81a077c14a03e6` in our downstream macprovider spike by pinning `mlx-swift-lm` to that revision and replacing our bench-only manual compiled decode path with `CompiledDecodeSession`.

Environment/config:
- macOS arm64, Apple Swift 6.3.3 toolchain
- `mlx-swift-lm`: PR head `edcda687...`
- `mlx-swift`: resolved to `0.31.6`
- model: `mlx-community/Qwen2.5-7B-Instruct-4bit` from local HF cache
- bench: greedy decode, prefill target 128, decode target 32, one warmup plus one measured run

Validation results:
- `swift build --product macprovider-cli`: pass
- focused downstream guards passed: `DecodeBenchFlagTests`, `KVConversationColdTierTests/testEligibleServeNewCacheProducesKVCacheSimple`, `ModelRuntimeSwapTests/testServingDecodeRunsThroughBlockingInferenceExecutor`
- uncompiled `generate(...)` vs `CompiledDecodeSession` parity: pass for both warmup and measured sample
- measured sample emitted 31 token ids on both paths, exact token-id match. The session path produced the same content sequence; the downstream harness needed to apply the same stop-token exclusion convention as `generate(...)` for EOS `151645`.

One measured-run smoke result, not a perf claim: uncompiled prefill/decode `543.0 / 27.9 tok/s`; compiled-session prefill/decode `572.6 / 26.0 tok/s`.

Only downstream compatibility change needed outside the compiled-path replacement was handling the new throwing `LanguageModel.newCache(...)` API in our runtime/tests. This looks good for our T2-01 parity spike and addresses the frozen cache-position failure mode we were tracking. Thanks again for the fix.

PR #553, opened 2026-08-18
Gemma 4: pass kvSharedOnly for shared-KV layers so E2B/E4B load

### LINKED ISSUE #552: Gemma 4 E2B/E4B fail to load: kvSharedOnly never passed for shared-KV layers
### What happens

No Gemma 4 checkpoint with `num_kv_shared_layers > 0` can be loaded. `loadWeights` fails on a key the checkpoint deliberately omits:

```
Key language_model.model.layers.15.self_attn.v_proj.weight not found in
Gemma4.Gemma4TextLanguageModel.Gemma4TextBackbone.Gemma4TextDecoderLayer.Gemma4TextAttention.Linear
```

Reproduced with mlx-swift-lm 3.31.4 / mlx-swift 0.31.6 on macOS 15 (Apple silicon):

| model | layers | `num_kv_shared_layers` | fails at layer |
|---|---|---|---|
| `mlx-community/gemma-4-e2b-it-4bit` | 35 | 20 | 15 |
| `mlx-community/gemma-4-e4b-it-4bit` | 42 | 18 | 24 |

In both cases the failing layer is exactly `hiddenLayers - numKVSharedLayers`.

### The checkpoints are fine

Worth stating, because the symptom reads like a bad download. Both shards are byte-exact against their own safetensors headers, and the tensor index shows the pattern is deliberate: layers `0 ..< firstKVSharedLayerIdx` carry `k_proj`, `v_proj` and `k_norm`; the rest carry only `q_proj` and `o_proj`, because they consume the shared K/V.

### Cause

`Gemma4TextAttention` already supports this — it takes `kvSharedOnly` and skips building the local K/V projections — and `Gemma4TextBackbone.callAsFunction` already threads `sharedKV` to every layer from `firstKVSharedLayerIdx` onward (`Libraries/MLXVLM/Models/Gemma4.swift`, around line 1310).

The construction site is the one place that disagrees:

```swift
self._layers.wrappedValue = (0 ..< config.hiddenLayers).map {
    Gemma4TextDecoderLayer(config: config, layerIdx: $0)
}
```

`kvSharedOnly` defaults to `false`, so every layer builds its own projections and the weight load then requires them. `isKVSharedLayer` is computed inside `Gemma4TextAttention.init` but only used to size the MLP, never to decide whether to build K/V.

### Fix

Pass the flag at construction. PR: #553

```swift
let firstShared = config.hiddenLayers - config.numKVSharedLayers
self._layers.wrappedValue = (0 ..< config.hiddenLayers).map { idx in
    Gemma4TextDecoderLayer(
        config: config, layerIdx: idx,
        kvSharedOnly: firstShared > 0 && idx >= firstShared)
}
```

With this change both models load and generate coherently. Checkpoints declaring zero shared layers are unaffected — `firstShared > 0` preserves the existing behaviour — and `mlx-community/gemma-4-12b-coder-fable5-composer2.5-4bit` (48 layers, 0 shared) and `mlx-community/gemma-3-4b-it-qat-4bit` still load and generate.

Amon20044:
Fixes #552.

`Gemma4TextBackbone` built local K/V projections for the layers that *share* them, so `loadWeights` demanded `k_proj`, `v_proj` and `k_norm` for exactly the layers a Gemma 4 MatFormer checkpoint omits:

```
Key language_model.model.layers.15.self_attn.v_proj.weight not found in
Gemma4.Gemma4TextLanguageModel.Gemma4TextBackbone.Gemma4TextDecoderLayer.Gemma4TextAttention.Linear
```

Everything needed was already present. `Gemma4TextAttention` takes `kvSharedOnly` and skips those projections; `callAsFunction` already threads `sharedKV` to every layer from `firstKVSharedLayerIdx` onward. Only the construction site never passed the flag, so `kvSharedOnly` defaulted to `false` for all layers.

### Change

One call site in `Gemma4TextBackbone.init`. `firstShared` is read into a local because `self` cannot be captured by the closure before every member is initialised.

### Verified

macOS 15, Apple silicon, mlx-swift 0.31.6. Loaded and generated with each:

| model | layers | `num_kv_shared_layers` | before | after |
|---|---|---|---|---|
| `mlx-community/gemma-4-e2b-it-4bit` | 35 | 20 | fails at layer 15 | loads, generates |
| `mlx-community/gemma-4-e4b-it-4bit` | 42 | 18 | fails at layer 24 | loads, generates |
| `mlx-community/gemma-4-12b-coder-fable5-composer2.5-4bit` | 48 | 0 | loads | loads, generates |
| `mlx-community/gemma-3-4b-it-qat-4bit` | — | — | loads | loads, generates |

Output is coherent rather than merely non-crashing — both E-models answer a one-word prompt correctly, which exercises the shared-K/V path rather than just the module tree.

`firstShared > 0` keeps the previous behaviour for every checkpoint that declares no sharing, so the two models that already worked are untouched.

### DIFF
--- a/IntegrationTesting/IntegrationTestingTests/Gemma4EseriesLoadIntegrationTests.swift
+++ b/IntegrationTesting/IntegrationTestingTests/Gemma4EseriesLoadIntegrationTests.swift
@@ -1,0 +1,64 @@
+// Copyright © 2026 Apple Inc.
+
+import Foundation
+import HuggingFace
+import IntegrationTestHelpers
+import MLXHuggingFace
+import MLXLMCommon
+import MLXVLM
+import Testing
+import Tokenizers
+
+// MARK: - E-series (KV-sharing) target load regression
+//
+// The Gemma 4 "effective" targets (E2B/E4B) use num_kv_shared_layers > 0: the
+// last N decoder layers reuse an earlier layer's K/V and ship no k_proj/v_proj.
+// They carry vision_config so they load through VLMModelFactory. Before the
+// loader fix, the VLM Gemma4 backbone declared k_proj on the shared tail and
+// loadWeights failed `keyNotFound layers.{15,24}.self_attn.k_proj.weight`
+// (independent of MTP — plain load). This exercises the real cached checkpoint
+// end to end: load + a short greedy generate (which also drives the runtime
+// shared-KV forward on those layers).
+//
+// Cache-gated: each test enables only when its checkpoint is already in the HF
+// cache, so it runs locally where the weights exist and disables (does not fail)
+// in CI without them — no env var, no multi-GB auto-download.
+
+private let e4bTarget = "mlx-community/gemma-4-e4b-it-4bit"
+private let e2bTarget = "mlx-community/gemma-4-e2b-it-4bit"
+
+@Suite(.serialized)
+struct Gemma4EseriesLoadIntegrationTests {
+
+    @Test(.enabled(if: hfSnapshotDir(modelId: e4bTarget) != nil))
+    func testE4BTargetLoadsAndGenerates() async throws {
+        try await loadAndGenerate(modelId: e4bTarget)
+    }
+
+    @Test(.enabled(if: hfSnapshotDir(modelId: e2bTarget) != nil))
+    func testE2BTargetLoadsAndGenerates() async throws {
+        try await loadAndGenerate(modelId: e2bTarget)
+    }
+
+    private func loadAndGenerate(modelId: String) async throws {
+        let dir = try #require(hfSnapshotDir(modelId: modelId))
+
+        // The load itself is the regression surface — pre-fix this threw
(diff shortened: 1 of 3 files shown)


Amon20044:
Closing — duplicate of #384, which already fixed this on `main` and more thoroughly: the guard belongs in `Gemma4TextAttention.init` (`!kvSharedOnly && !isKVSharedLayer`), where it holds regardless of the caller, rather than only at the construction site.

My mistake was checking `git ls-remote --tags`, finding 3.31.4 as the newest, and concluding no fix existed — without looking at `main`. Apologies for the noise.

One note that may still help anyone arriving from the same error: the fix is not in any tagged release, so a version requirement cannot reach it. 68947cc cherry-picks cleanly onto 3.31.4.

davidkoski:
Thank you for the PR anyway!  We are prepping for a new tag, hopefully this week, but if not, _soon_!

PR #556, opened 2026-08-19
Muse Glimmer: honor the requested KV-cache capacity

aleroot:
## Proposed changes

Today I was trying the new Muse Glimmer and had some bad surprises: 
`newCache(parameters:)` dropped its `parameters`, so the full-attention layers were always unbounded `StandardKVCache`. Any request carrying a capacity, `GenerateParameters.maxKVSize` or `kvCache`, was then rejected by `validateKVCacheCompatibility` with `incompatibleCapacity`, one count per full-attention layer (13 on the 52-layer checkpoint).

Route both layer kinds through `makeHybridAttentionKVCache` instead, as every other hybrid model does. Behavior is unchanged when no capacity is requested: sliding layers keep the architecture window with `keep: 0` and full layers stay unbounded.

@CharlieTLe FYI

## Checklist

### DIFF
--- a/Libraries/MLXVLM/Models/MuseGlimmer.swift
+++ b/Libraries/MLXVLM/Models/MuseGlimmer.swift
@@ -544,13 +544,14 @@ private class MuseGlimmerLanguageModel: Module, KVCacheDimensionProvider {
         super.init()
     }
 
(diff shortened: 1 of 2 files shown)


davidkoski (approved):
Change looks good, thank you!

PR #557, opened 2026-08-19
Read Gemma's brace-form object and array values

aleroot:
## Proposed changes

> Gemma writes a nested value in the dialect's brace form:
>
> ```
> <|tool_call>call:search{filters:{city: "Paris", limit: 1}}<tool_call|>
> ```
>
> That is JSON in every respect except that the keys carry no quotes, which `JSONSerialization` rejects. `GemmaFunctionParser` falls back to the literal text, so `filters` arrives as `.string("{city: \"Paris\", limit: 1}")`. When the schema declares `filters` an object, `convertParameterValue` takes the same strict path and returns the same string: the tool receives a string where it declared an object, and the call fails at the far end — with an MCP server, `-32602 Invalid params` — with an error the model has no way to act on.
>
> #531 made the argument *list* structural, so a value is no longer truncated at an inner comma. This is the remaining half: the value itself is still read as text.
>
> **Fix.** `BareKeyJSONParser` parses JSON, retrying once with bare object keys quoted. It walks the literal with the `StructuredTextScanner` added in #531, so quoted spans stay opaque and nesting is respected. `GemmaFunctionParser` uses it for parameters the schema declares structured (`object`, `array`, `dict…`, `list…`) and for parameters the schema does not declare at all. Every other path is unchanged.
>
> **Keys only, by design.** A bare *value* is not quoted; the payload is refused and keeps today's literal-text behavior. `{ok: true}` is a boolean while `{ok: yes}` is nothing JSON can name, so quoting there would invent meaning, and a parser that guesses is worse than one that hands back the text.

### DIFF
--- a/Tests/MLXLMTests/ToolTests.swift
+++ b/Tests/MLXLMTests/ToolTests.swift
@@ -1364,6 +1364,69 @@ struct ToolTests {
         #expect(toolCall.function.arguments["urgent"] == .string("true"))
     }
 
+    @Test("Gemma reads a nested object whose keys are unquoted")
+    func testGemmaBareKeyObjectValue() throws {
+        let parser = GemmaFunctionParser(
+            startTag: "<|tool_call>", endTag: "<tool_call|>", escapeMarker: #"<|"|>"#)
+        let content = #"<|tool_call>call:search{filters:{city: "Paris", limit: 1}}<tool_call|>"#
+
+        let toolCall = try #require(parser.parse(content: content, tools: nil))
+
+        // The dialect writes nested objects without quoting their keys, which strict JSON refuses.
+        #expect(
+            toolCall.function.arguments["filters"]
+                == .object(["city": .string("Paris"), "limit": .int(1)]))
+    }
+
+    @Test("Gemma reads a bare-key object into a parameter the schema declares an object")
+    func testGemmaBareKeyObjectValueWithSchema() throws {
+        let parser = GemmaFunctionParser(
+            startTag: "<|tool_call>", endTag: "<tool_call|>", escapeMarker: #"<|"|>"#)
+        let tools: [[String: any Sendable]] = [
+            [
+                "function": [
+                    "name": "search",
+                    "parameters": [
+                        "type": "object",
+                        "properties": [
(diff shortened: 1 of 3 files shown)


Tests/MLXLMTests/ToolTests.swift
```diff
@@ -1364,6 +1364,69 @@ struct ToolTests {
         #expect(toolCall.function.arguments["urgent"] == .string("true"))
     }
 
+    @Test("Gemma reads a nested object whose keys are unquoted")
+    func testGemmaBareKeyObjectValue() throws {
+        let parser = GemmaFunctionParser(
+            startTag: "<|tool_call>", endTag: "<tool_call|>", escapeMarker: #"<|"|>"#)
```
davidkoski, line 1370: Does the BareKeyJSONParser need to understand escapeMarkers as well?  I am not sure this can happen, but something like:

```
<|"|>a b<|"|>: 10
```

The GemmaFunctionParser consumes these escapeMarkers (and they are important) but the BareKeyJSONParser doesn't know about them.

Can it happen?  Do we care?

davidkoski (changes requested):
Looks good but see the question on the escapeMarker -- if that isn't a concern I think it is ready to merge.  Thanks!

Tests/MLXLMTests/ToolTests.swift
aleroot: I have double  checked, and this does not look like a concern.

Google's parser only matches `\w+` keys (`(\w+):(?:<\|"\|>(.*?)<\|"\|>|([^,}]*))`), and the canonical chat template renders tool-call arguments with `escape_keys=False` at every nesting level. Nested keys stay bare identifiers; `<|"|>` is only used for string *values*. Something like `<|"|>a b<|"|>: 10` is not in the dialect.

`BareKeyJSONParser` would refuse that key anyway (not a bare identifier, not JSON-quoted), so even if it appeared the nested object would fall back to the raw text same as today.

Nested *values* can be `<|"|>…<|"|>` (`{city:<|"|>Paris<|"|>}`), which this parser still does not rewrite, but that is probably a separate follow-up.

davidkoski (approved):

PR #559, opened 2026-08-20
Report prompt-cache reuse in GenerateCompletionInfo

aleroot:
## Proposed changes

Expose the cached tokens so that it can be actually compared to the total tokens, with this new value exposed  it is going to be possible to calculate the cache efficiency and see how it performs. 

### Notes

Reading the new fields correctly takes two things:

**Pick the right prompt number.** Three now exist, with distinct meanings:

| Property | Meaning |
|---|---|
| `promptTokenCount` | tokens actually prefilled this pass, the correct denominator for `promptTokensPerSecond` |
| `cachedPromptTokenCount` | tokens served from the reused KV prefix |
| `totalPromptTokenCount` | the rendered prompt (`cached + prefilled`) |

A UI labelled "Prompt tokens" wants `totalPromptTokenCount`. Displaying `promptTokenCount` there makes the prompt appear to shrink as the conversation grows.

## Checklist


Libraries/MLXLMCommon/Evaluate.swift
```diff
@@ -2488,9 +2488,21 @@ public enum GenerateStopReason: Sendable {
...
+    /// ``cachedPromptTokenCount`` and ``totalPromptTokenCount``.
     public let promptTokenCount: Int
 
+    /// The number of prompt tokens served by a reused KV-cache prefix instead
+    /// of being prefilled, or `0` when the whole prompt was prefilled.
+    ///
+    /// Only a cache-owning caller can know this: the generation loop receives
+    /// an already narrowed prompt and has no notion of a prompt cache.
+    /// ``ChatSession`` attributes it from its cache reuse decision.
+    public internal(set) var cachedPromptTokenCount: Int
```
davidkoski, line 2504: The `internal(set)` restricts use of this to code inside mlx-swift-lm -- if you had a custom generation loop in your own app/library it couldn't use this.  Not a blocker for merge, but it may be a use case to think about.

davidkoski (approved):
Change looks good to me.  Thanks!

PR #562, opened 2026-08-21
Load weight files a safetensors index leaves out

### LINKED ISSUE #554: [BUG] unable to load mlx-community/Qwen3-VL-4B-Instruct-4bit
See #408 -- this causes the load to honor the `model.safetensors.index.json` but per this comment in #535 

```
/// Qwen3-VL is deliberately not in this list. `mlx-community/Qwen3-VL-4B-Instruct-4bit`
/// cannot load: its `model.safetensors.index.json` names
/// `model-00001-of-00002.safetensors` and `model-00002-of-00002.safetensors`,
/// neither of which the repo ships, and declares about 8.9 GB while the repo
/// contains a single 3.1 GB `model.safetensors`. The 8.9 GB figure is the
/// unquantized size, so the index was carried over from the source repo and
/// never regenerated for the quantized upload. The same is true of
/// `Qwen3-VL-8B-Instruct-4bit` (names four shards, ships two) and
/// `Qwen3-VL-4B-Instruct-8bit` (names two, ships one), so it is the
/// quantization batch rather than one repo. `Qwen3-VL-2B-Instruct-4bit` is
/// packaged correctly if a Qwen3-VL model is ever wanted here.
```

it doesn't work.  It would be good to fall back to just loading what is present, either automatically or with ModelConfiguration policy.

### LINKED ISSUE #560: Nightly integration tests are failing
The nightly `Integration Tests` run failed.

- Run: https://github.com/ml-explore/mlx-swift-lm/actions/runs/32468213114 (attempt 1)
- Commit: [`130e3f0`](https://github.com/ml-explore/mlx-swift-lm/commit/130e3f0cd68e120019b0864c390d0309b0ec5618) on `refs/heads/main`
- First seen: 2026-08-21

The failing run uploads an `integration-test-results` artifact:
the `IntegrationTesting.xcresult` bundle plus anything in
`~/Library/Logs/DiagnosticReports` (crash logs). Artifacts are
kept for 7 days, so download it from the run page before it
expires, then open the `.xcresult` in Xcode.

Consecutive failures comment on this issue instead of opening a
new one. Close it once the nightly is green again; the next
failure after that opens a fresh issue.

aleroot:
## Proposed changes

Fixes #560. Also fixes #554.

The nightly `RerankerModelIntegrationTests.jinaV3RelevanceScores` fails with:

```
Module.swift:568: Caught error: .keyNotFound(path: ["projector", "linear1", "weight"],
    modules: ["JinaRerankerModel", "JinaRerankerProjector", "Linear"])
```

Two checkpoints, one cause. `jinaai/jina-reranker-v3-mlx` keeps its reranking head in `projector.safetensors`, which its `model.safetensors.index.json` never names, so the head was not loaded and `update(parameters:verify: [.all])` failed. `mlx-community/Qwen3-VL-4B-Instruct-4bit` (#554) ships one 3.1 GB `model.safetensors` but kept the index of its unquantized source repo, which names two shards it does not ship, so loading throws
`[load_safetensors] Failed to open file ... model-00001-of-00002.safetensors`.

(For the record on how #560 got in: #375 was written and run against a base predating #408, and rebasing it onto main changed the loader underneath it. Since that PR touches no loading code there was no conflict, and the integration tests are nightly rather than on the PR path.)

### The change

`mlx_lm.utils.load_model` selects weight files with `glob("model*.safetensors")` at the top level and never reads the index. That is why all three checkpoints load in Python, and it is the rule this PR adopts, in order:

1. the files named by `model.safetensors.index.json`, when it exists and every file it names exists a complete index is precise for a repo shipping both a consolidated file and shards
2. otherwise the conventional `model*.safetensors` names, then `weight*.safetensors`, then whatever is present, so an unconventional local directory still loads

Only the top level of the directory is considered. The checkpoint that motivated #408, `mlx-community/Qwen3.5-4B-OptiQ-4bit`, keeps its auxiliary weights in `optiq/`, so the recursive walk was why they were loaded at all and that failure mode is silent: a stray tensor whose name `sanitize(weights:)` rewrites corrupts generation rather than raising. `ParoQuantLoader` already restricts itself to the top level for the same reason. This is also why the fallback is the `model*` convention rather than "load every file present", which would reintroduce that corruption on a repo that is both stale-indexed and OptiQ-shaped.

Two additions on top:

- **`AdditionalWeightFilesProviding`**, an optional protocol next to `ModelConversionMetadataProvider`, lets a model name files no convention selects; `JinaRerankerModel` declares `projector.safetensors`. The checkpoint's own `rerank.py` loads that file by name for the same reason. Naming the file rather than widening the selection is what keeps unrelated tensors out.
- **`ModelConfiguration.weightFileSelection`** is the other half of #554: a caller that knows a checkpoint's index omits weights no registered model declares can ask for `.allFilesPresent`. It carries through `ResolvedModelConfiguration` into the LLM, VLM, embedder, and MTP drafter factories. The caveats are documented on the case.

### Verification

Both failures reproduce on `main` and pass with this change, on the real checkpoints:

| | before | after |
|---|---|---|
| `jinaV3RelevanceScores()` | `keyNotFound(["projector","linear1","weight"])` | passes, scores within 3e-3 of `rerank.py` |
| `qwen3VLWithStaleIndexLoadsAndGenerates()` | `Failed to open file ... model-00001-of-00002.safetensors` | passes, loads and generates |

- `LoadWeightsTests` (12 cases): index selection including an index that names a nested file,
  stale and partially stale indexes, the `model*` / `weight*` / everything fallbacks, nested
  files excluded under every policy, additional files appended and deduplicated, the caller
  policy, and `loadWeights` end to end with and without the declaration.
- `RerankerTests` asserts the Jina conformance, so the PR path catches what the nightly caught
  this time.
- `xcodebuild test -scheme mlx-swift-lm-Package -only-testing:MLXLMTests`: 490 XCTest + 690
  swift-testing tests pass. Full `RerankerModelIntegrationTests` suite (BGE, Qwen3, Jina) and the
  new `StaleWeightIndexIntegrationTests` pass locally on Apple silicon.
- `swift-format` clean; `scripts/verify-docs.sh` passes.

The new integration test pins the Qwen3-VL revision, since the index/weights mismatch is a
property of that upload and a re-quantized revision would silently stop covering the fallback.
It adds a 3.1 GB download to the nightly runner on first run.

## Checklist


davidkoski:
I think the model declared method sounds good, but I am not sure it will work here:

- https://huggingface.co/jinaai/jina-reranker-v3-mlx/tree/main
- https://huggingface.co/jinaai/jina-reranker-v3/tree/main

Same model, but the latter (what I presume to be the root) doesn't have the `projector.safetensors` file.

I wonder if we need both (config and model)?  The only consistency here is that there is no consistency :-)

aleroot:
> I think the model declared method sounds good, but I am not sure it will work here:
> 
> * https://huggingface.co/jinaai/jina-reranker-v3-mlx/tree/main
> * https://huggingface.co/jinaai/jina-reranker-v3/tree/main
> 
> Same model, but the latter (what I presume to be the root) doesn't have the `projector.safetensors` file.
> 
> I wonder if we need both (config and model)? The only consistency here is that there is no consistency :-)

Good catch, and you were right that something breaks  though not where either of us expected. I pulled the tensor names out of both repos:

**`jinaai/jina-reranker-v3`** (upstream): no index, single `model.safetensors`, 312 tensors, projector stored as `nn.Sequential(Linear, ReLU, Linear)`:

```
projector.0.weight  [512, 1024]  BF16
projector.2.weight  [512, 512]   BF16
```

**`jinaai/jina-reranker-v3-mlx`**: index naming 310 tensors (all → `model.safetensors`, neither the projector nor its file mentioned), plus `projector.safetensors` holding the same two layers renamed and unprefixed:

```
linear1.weight  [512, 1024]
linear2.weight  [512, 512]
```

So there are two independent axes here, and I had only fixed one:

1. **Which files to load.** This one already worked for the upstream repo: a declared file that is not present is ignored, and with no index the loader falls back to the conventional `model*.safetensors` names, which carry the projector in. The declaration costs that layout nothing.
2. **What the tensors are called.** This did not. `sanitize` only knew the MLX spelling, so the upstream repo would have failed with the same `keyNotFound(["projector","linear1","weight"])`  for a completely different reason. Fixed by accepting both spellings.

That is also my answer to "do we need both": yes, but they solve different problems, and the model needs *two* hooks rather than one.

- `AdditionalWeightFilesProviding`: packaging the loader cannot infer, for a checkpoint we ship support for. A config flag here would mean every user of `jina-reranker-v3-mlx` has to know how the repo is laid out.
- `ModelConfiguration.weightFileSelection` (`.allFilesPresent`): the escape hatch for a checkpoint no registered model knows about( #554 ).
- `sanitize`: tensor naming.  No config flag can help: only the model knows that `projector.2.weight` and `linear2.weight` are the same layer.

Both repos are now covered by the same integration assertions, and both reproduce the reference scores from the checkpoint's own `rerank.py` within 3e-3:

```
✔ jinaV3RelevanceScores()                     (jinaai/jina-reranker-v3-mlx)
✔ jinaV3SourceCheckpointRelevanceScores()     (jinaai/jina-reranker-v3, pinned d7d7e73)
✔ qwen3VLWithStaleIndexLoadsAndGenerates()    (mlx-community/Qwen3-VL-4B-Instruct-4bit, #554)
✔ bgeV2M3ReferenceScores(), qwen3ReferenceScores()
```

The upstream test adds a 1.2 GB download to the nightly on first run. I kept it unconditional rather than cache-gated so the "one model class, two packagings" claim keeps being checked rather than being true only on the day I ran it.

davidkoski (approved):
This looks like a great fix, thank you for the followup!

PR #565, opened 2026-08-22
Add extensible VLM processor loading rules

### LINKED ISSUE #547: allow external code to configure model loading
Provide hooks for external libraries to fully configure model loading.

See #546 -- this will provide base processor configuration if it is missing for certain models.

See also:

```swift
        // Override processor type based on model type for models that need special handling
        // Mistral3 models ship with "PixtralProcessor" in their config but need Mistral3Processor
        // to handle spatial merging correctly
        let processorTypeOverrides: [String: String] = [
            "mistral3": "Mistral3Processor",
            "gemma4_unified": "Gemma4UnifiedProcessor",
        ]
```

Both of these encode model specific information into the factory -- models outside mlx-swift-lm can't do this but should be able to.

aleroot:
## Proposed changes

Fixes #547.

External model packages could register their models and processors, but could not control how processor metadata was loaded. This forced model-specific fixes to be hard-coded inside `VLMModelFactory`.

This PR introduces a public processor-loading resolver registry. External packages can now:

- Provide processor configuration when it is missing from a model repository.
- Correct an incorrectly declared processor type.
- Override built-in behavior without modifying MLXVLM.

Existing Qwen3.5, Mistral3, and Gemma4 special cases now use the same public mechanism. Repository configuration files remain authoritative, and fallbacks are only evaluated when those files are absent.

## Checklist

### DIFF
--- a/Libraries/MLXVLM/README.md
+++ b/Libraries/MLXVLM/README.md
@@ -419,6 +419,43 @@ public class VLMProcessorTypeRegistry: @unchecked Sendable {
                 YourModelProcessorConfiguration.self, YourModelProcessor.init),
 ```
 
+### Processor metadata supplied by an external package
+
+Some repositories omit processor metadata, or declare a generic processor that is not the
+right implementation for their model type. An external model package can handle either case
+without changing `VLMModelFactory`:
+
+```swift
+struct YourModelProcessorLoadingResolver: VLMProcessorLoadingResolving {
+    func processorConfigurationFallback(
+        for context: VLMProcessorLoadingContext
+    ) throws -> VLMProcessorConfiguration? {
+        guard context.modelType == "your_model" else { return nil }
+
+        let model = try JSONDecoder().decode(
+            YourModelConfiguration.self, from: context.configurationData)
+        let processor = YourModelProcessorConfiguration(model: model)
+        return VLMProcessorConfiguration(
+            data: try JSONEncoder().encode(processor),
+            processorType: "YourModelProcessor")
+    }
+
+    func processorTypeOverride(
+        for context: VLMProcessorLoadingContext,
+        declaredProcessorType: String
+    ) throws -> String? {
+        context.modelType == "your_model" ? "YourModelProcessor" : nil
+    }
+}
+
+VLMProcessorLoadingRegistry.shared.register(YourModelProcessorLoadingResolver())
+```
+
+Implement only the hook your model needs. A checkpoint's `preprocessor_config.json` or
+`processor_config.json` always wins over a generated fallback. Type overrides are applied
+afterward. For isolated applications and tests, pass a separate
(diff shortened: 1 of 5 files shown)


aleroot:
This PR Action check is failing because 1441444 targeting Xcode 27 beta 5 while the hosted xcode-27 runner is still on beta 4. It now affects every PR tested against main.
@davidkoski, would you like me to open a temporary revert PR and reapply the change once GitHub upgrades the hosted runner, or should we treat integration_build_xcode27 as non-blocking until then?

davidkoski:
We can just treat the xcode27 as non-blocking for now (I will just verify that it is the expected error).

Libraries/MLXVLM/README.md
```diff
@@ -419,6 +419,44 @@ public class VLMProcessorTypeRegistry: @unchecked Sendable {
                 YourModelProcessorConfiguration.self, YourModelProcessor.init),
 ```
 
+### Processor metadata supplied by an external package
+
+Some repositories omit processor metadata, or declare a generic processor that is not the
+right implementation for their model type. An external model package can handle either case
+without changing `VLMModelFactory`:
```
davidkoski, line 426: It is curious that there are several VLMs in this situation and _no_ LLMs.

Libraries/MLXVLM/README.md
```diff
@@ -419,6 +419,44 @@ public class VLMProcessorTypeRegistry: @unchecked Sendable {
...
+}
+
+VLMProcessorLoadingRegistry.shared.register(YourModelProcessorLoadingResolver())
+```
+
+Implement only the hook your model needs. A checkpoint's `preprocessor_config.json` or
+`processor_config.json` always wins over a generated fallback. Processor type resolution is
+applied afterward and can supply a missing `processor_class` or correct an incorrect one.
+For isolated applications and tests, pass a separate
+`VLMProcessorLoadingRegistry` to `VLMModelFactory` instead of registering globally.
```
davidkoski, line 458: Nice!

Libraries/MLXVLM/VLMProcessorLoadingRegistry.swift
```diff
@@ -0,0 +1,158 @@
...
+    }
+}
+
+/// Supplies model-specific processor loading policy that is not present in a checkpoint.
+///
+/// Both hooks are optional. A fallback is consulted only when neither
+/// `preprocessor_config.json` nor `processor_config.json` exists. Processor type resolvers are
+/// consulted after a configuration has been selected, so they can supply missing metadata or
+/// correct metadata shipped by a checkpoint.
+public protocol VLMProcessorLoadingResolver: Sendable {
```
davidkoski, line 39: This looks great -- if we need more hooks we can easily add more methods with nil-returning defaults in the future.

davidkoski (approved):
This looks awesome!  Thank you!

PR #567, opened 2026-08-22
Share fused Qwen3.5 router top-k across LLM and VLM

ngutech21:
## Proposed changes

This is a follow-up port of [PR #469](https://github.com/ml-explore/mlx-swift-lm/pull/469), which introduced a bit-identical fused MoE router Top-K kernel for Qwen3.5/3.6 but only wired it into the text-model implementation. The Qwen3.5 vision-language path continued using the unfused `argPartition` → `takeAlong` → normalize chain.

This change:

- Moves the fused kernel, reference implementation, and dispatch policy into `MLXLMCommon`.
- Uses the shared `moeRouterTopK` function in both `MLXLLM` and `MLXVLM`.
- Selects the fused kernel only for single-row decode with at most 1024 experts.
- Preserves the original chain for prefill, batched inputs, and larger expert counts.
- Keeps the shared function package-internal, without adding public API.

## Testing

- Extended the bitwise tests to cover the fused and fallback dispatch paths.
- Confirmed the fused output remains bit-identical to the reference chain.
- Full package test suite passes.
- `MLXLMCommon` documentation build passes.

This PR does not claim a measured VLM speedup yet; it ports the previously validated text-path optimization and removes the implementation divergence between the two Qwen3.5 paths.
## Checklist

### DIFF
--- a/Libraries/MLXLLM/Models/Qwen35.swift
+++ b/Libraries/MLXLLM/Models/Qwen35.swift
@@ -12,109 +12,6 @@ import MLX
 import MLXLMCommon
 import MLXNN
 
-// MARK: - Fused router top-k
-
-/// One-kernel replacement for the decode router tail: `chainRouterTopK`
-/// fully sorts all `E` experts (`ArgPartition::eval_gpu` delegates to
-/// `gpu_merge_sort`) just to name `K` — three serial dispatches, three
-/// encoder-wide barriers, where barrier-bound decode needs one.
-///
-/// Bit-identical to the chain by construction: the sort is stable
-/// (`sort.h`'s `LessThan` compares values only, ties keep input order), so
-/// counting the elements ranked strictly above `i` — with the index packed
-/// into the low bits of a monotone bit key as the tie-break — reproduces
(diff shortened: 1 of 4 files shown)


aleroot:
Wow 😮 it seems like we were working on the same thing, I can guarantee it was a couple of days working on this 😀
Sorry about that, maybe we should merge the work @ngutech21 see #568

aleroot:
@ngutech21 probably you can merge first and I can rebase on top of your work, which is still very useful for VLM. What do you think @davidkoski ?

ngutech21:
> Wow 😮 it seems like we were working on the same thing, I can guarantee it was a couple of days working on this 😀 Sorry about that, maybe we should merge the work @ngutech21 see #568

No worries. I was going through the merged PRs for the LLM path and thought this could be applied to the VLM as well.

davidkoski:
> @ngutech21 probably you can merge first and I can rebase on top of your work, which is still very useful for VLM. What do you think @davidkoski ?

Sounds good -- I will review this one first on monday.

davidkoski (approved):
Very clean, thank you!

PR #568, opened 2026-08-22
Share fused MoE router top-k across models

aleroot:
## Proposed changes

Mixture-of-Experts models contain several feed-forward networks called “experts.” For every generated token, a router selects a small number of those experts and combines their outputs.

The existing generic router performs this work as several consecutive GPU operations:

1. Select the top experts.
2. Gather their routing scores.
3. Optionally normalize those scores.

These operations process very little data during single-token generation, so repeatedly launching separate GPU operations can cost more than the calculations themselves. Qwen 3.5 already avoids this overhead with a decode-specific Metal kernel.

This PR moves that kernel into `MLXLMCommon`, generalizes it, and shares it with:

- Qwen 3 MoE
- OLMoE
- Mixtral
- Jamba
- Granite MoE Hybrid
- Qwen 3.5, which retains its existing optimized behavior

The shared implementation supports the differences between these model families, including:

- Selecting experts from one tensor while gathering weights from another.
- Preserving ascending or descending expert order.
- Optional score normalization.
- Stable handling of equal routing scores.

The optimized path is deliberately limited to single-token decoding with at most 1,024 experts. Prompt processing and other multi-row inputs continue using the existing generic MLX implementation, where its sorting algorithm is more appropriate.

## Checklist

- [x] I have updated the necessary documentation (if needed) — No user-facing documentation change is required because this is a package-internal optimization; explanatory API documentation was added alongside the shared implementation.

### DIFF
--- a/Libraries/MLXLLM/Models/GraniteMoeHybrid.swift
+++ b/Libraries/MLXLLM/Models/GraniteMoeHybrid.swift
@@ -281,8 +281,16 @@ class GraniteMoeHybridTopKGating: Module {
 
     func callAsFunction(_ hiddenStates: MLXArray) -> (MLXArray, MLXArray) {
         let logits = layer(hiddenStates)
-        let indices = MLX.argPartition(-logits, kth: topK - 1, axis: -1)[.ellipsis, ..<topK]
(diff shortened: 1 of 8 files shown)


davidkoski (approved):
Really nice!  Thank you!

PR #569, opened 2026-08-22
Generalise compiled decode segments to Qwen 3 Next

aleroot:
## Proposed changes

Qwen 3 Next previously executed each decoder layer separately during single-token generation, which adds dispatch overhead and slows token generation. Qwen 3.5 already reduced this overhead using compiled decode segments, but that implementation was model-specific.

This PR:

- Extracts the decode-segment schedule and thread-safe compiled-function cache into `MLXLMCommon`.
- Updates Qwen 3.5 to use the shared implementation without changing its behaviour.
- Adds compiled single-token decode segments to Qwen 3 Next.
- Keeps changing KV-cache operations outside compiled graphs.
- Supports fp16 activations, including models with 4-bit quantised weights.
- Falls back to the existing implementation for bfloat16, float32, quantised KV caches, and TurboQuant KV caches where exact equivalence has not been established.

Tests compare generated logits and every cache tensor byte-for-byte between the compiled and existing paths. They also cover quantized weights, unsupported-cache fallback, segment scheduling, and model deallocation.

On a warmed synthetic 16-layer Qwen 3 Next model with 4-bit weights, interleaved measurements showed approximately 35–38% lower median decode latency, corresponding to 1.55–1.60× throughput. Compilation time was excluded.

## Checklist

### DIFF
--- a/Libraries/MLXLLM/Models/Qwen3Next.swift
+++ b/Libraries/MLXLLM/Models/Qwen3Next.swift
@@ -88,24 +88,11 @@ public final class Qwen3NextAttention: Module {
     public func callAsFunction(
         _ x: MLXArray, mask: MLXFast.ScaledDotProductAttentionMaskMode, cache: KVCache?
     ) -> MLXArray {
-        let B = x.dim(0)
-        let L = x.dim(1)
-
-        let qProjOutput = qProj(x)
-        let qSplit = qProjOutput.reshaped(B, L, args.attentionHeads, -1).split(parts: 2, axis: -1)
-        var queries = qSplit[0]
-        let gate = qSplit[1].reshaped(B, L, -1)
-
-        var keys = kProj(x)
-        var values = vProj(x)
-
-        queries = qNorm(queries).transposed(0, 2, 1, 3)
-        keys = kNorm(keys.reshaped(B, L, args.kvHeads, -1)).transposed(0, 2, 1, 3)
-        values = values.reshaped(B, L, args.kvHeads, -1).transposed(0, 2, 1, 3)
+        let (queriesPreRope, gate, keysPreRope, values) = projectPreRope(x)
 
         let offset = cache?.ropeOffset
-        queries = applyRotaryPosition(rope, to: queries, offset: offset)
-        keys = applyRotaryPosition(rope, to: keys, offset: offset)
+        let queries = applyRotaryPosition(rope, to: queriesPreRope, offset: offset)
+        let keys = applyRotaryPosition(rope, to: keysPreRope, offset: offset)
 
         let output = attentionWithCacheUpdate(
             queries: queries,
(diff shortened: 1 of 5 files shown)


ngutech21:
@aleroot Would it make sense to use the new shared `CompiledDecodeSegment` and `CompiledDecodeSegmentCache` infrastructure for the Qwen 3.6 MoE VLM path in `MLXVLM/Models/Qwen35.swift` as well?
That path still lacks the compiled single-token decode segments and GDN decode-convolution folding introduced by #467 and #468. 
Would extending this PR to cover that path be in scope, or would you prefer a separate follow-up PR?

aleroot:
@ngutech21 I think you are right, the VLM path should reuse the new shared infrastructure. I would go with a new PR because this is much more than swapping in CompiledDecodeSegmentCache. The VLM implementation still needs: the traced layer bodies, GDN state plumbing and decodeConv folding from #467/#468, plus VLM-specific handling for M-RoPE/positionIds around the KV-cache update.
Once #569 lands, the follow-up should build on the shared infrastructure rather than duplicating the scheduler. I would  include bitwise logits/cache parity for normal text decode and image-prefill → text-decode continuation, fallback coverage for unsupported cache types.

Is it something you can take care of?

ngutech21:
Sure I can tackle this once #569 lands

Libraries/MLXLLM/Models/Qwen3Next.swift
```diff
@@ -466,6 +590,133 @@ public class Qwen3NextModelInner: Module {
...
+        for (segmentIndex, segment) in decodeSegments.enumerated() {
+            var arguments = [carry] + pendingAttention
+            for layerIndex in segment.linearLayers {
+                let mambaCache = mambaCaches[layerIndex]!
+                arguments.append(mambaCache[0]!)
+                arguments.append(mambaCache[1]!)
+            }
+
+            let outputs = compiledSegments.call(
+                at: segmentIndex, arguments: arguments
```
davidkoski, line 672: I was recently working on some `compile` code in mlx-swift and came across something that might matter here: this is (I think) capturing all MLXArrays that are not passed as arguments.  Specifically the layers referenced from `segmentBody`.

This is fine for inference, but if combined with LoRA it likely doesn't work as expected.

I don't think this is new to this PR -- I suspect we have other code in here that does the same thing.

davidkoski: Filed https://github.com/ml-explore/mlx-swift-lm/issues/586 to track this.

davidkoski: I think we are ok to take this, but see that issue.  I have a repro in mlx-swift and it is a potential problem.  It looks like the only instances of this are here and in Qwen35.swift via #469 

Probably worth followup.

davidkoski (changes requested):
Change looks good but see the comment on compile -- I think this is ok to merge as-is and deal with it in a followup, but want to get your take on it.

Thanks!

Libraries/MLXLLM/Models/Qwen3Next.swift
aleroot: Sure I will clean this up tomorrow.

aleroot:
> Change looks good but see the comment on compile -- I think this is ok to merge as-is and deal with it in a followup, but want to get your take on it.
> 
> Thanks!

Yes, this is worth processing as a separate issue, tomorrow I think I will have the time to tackle this properly. Thanks.

davidkoski (approved):
Changes look good.  Compile item to be addressed later -- does not affect inference path and is not new.

aleroot:
Added #589 as follow up to fix the related bug #586

PR #570, opened 2026-08-22
Tune TurboFlash query grouping for short decode

aleroot:
## Proposed changes

TurboFlash currently processes two query rows together (`NR0=2`) to reuse decoded key/value data. This helps at longer context lengths, but benchmarks showed that it reduces GPU parallelism immediately after the single-dispatch path ends.

This PR introduces a conservative scheduling policy:

- Uses one query row per GPU group (`NR0=1`) for contexts of 129–256 tokens, 16–32 query rows, and head dimensions of 64 or 128.
- Keeps the existing `NR0=2` behavior for all other shapes and longer contexts.
- Preserves the `TURBO_FLASH_NR0` environment override for manual tuning.
- Adds bit-exact tests covering 2-bit, 3-bit, 4-bit, and mixed key/value configurations, including fused value rotation.

On my Apple M2, the selected region was approximately 2–13% faster in the interleaved benchmarks. The optimization does not change model calculations or output values; it only changes how identical work is scheduled on the GPU. More aggressive `NR0=4` scheduling was tested but intentionally excluded because it regressed some shapes.

No corresponding issue.

## Checklist

### DIFF
--- a/Libraries/MLXLMCommon/TurboQuantKernels.swift
+++ b/Libraries/MLXLMCommon/TurboQuantKernels.swift
@@ -1726,23 +1726,37 @@ enum TurboQuantKernelOps {
     nonisolated(unsafe) private static var flashSinglePassKernels: [String: MLXFast.MLXFastKernel] =
         [:]
 
-    /// NR0: number of query rows processed per SIMD group in the multi-row amortized kernel.
-    ///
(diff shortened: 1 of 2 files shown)


davidkoski (approved):
Change looks good, thank you!

PR #461, opened 2026-07-23
Add DeepSeek-OCR VLM support (MLXVLM)

jyauxi:
## Summary
- Ports DeepSeek-OCR from mlx-vlm (Python) into MLXVLM (Swift): SAM + CLIP dual vision, MoE language decoder, gundam processor, registry wiring.
- First-class crop-mode API (`gundam` / `base`) and grounding / ref / det token surface (`DeepseekOCRSpecialTokens`).
- Cache-gated IntegrationTesting example + supported-model docs.
- Adapted from [mzbac/deepseek-ocr.swift](https://github.com/mzbac/deepseek-ocr.swift) (MIT), refactored toward mlx-swift-lm patterns (`GlmOcr`, `DeepseekV3` sanitize / `@ModuleInfo`).
- Closes #15

## Review guide

Diff shape: **+3286 / −1 across 15 files** — 7 new files (model, processor, tests, smoke tool); the 8 touched existing files are registration, docs, and one shared-code change (detailed below). No existing model paths are modified.

- **Start at** `Libraries/MLXVLM/VLMModelFactory.swift` — the full integration surface (model type + processor registration) is visible there.
- **Mostly pattern-following:** `DeepseekOCR.swift` mirrors the `GlmOcr` / `Qwen25VL` house conventions — config structs with snake_case `CodingKeys`, `@ModuleInfo(key:)` matching safetensors keys, `DeepseekV3`-style `sanitize(weights:)` for MoE expert stacking.
- **The one behavioral change to shared code:** `ArgMaxSampler` in `MLXLMCommon/Evaluate.swift` now casts logits to bf16 before argmax, so greedy ranking matches MLX Python models that keep decode logits in bf16. It only changes outcomes where float32 accumulation breaks a bf16 exact tie. This is the part that affects other models and deserves the closest look.
- **Highest-scrutiny new code:** (1) the MoE gate — `topk_method` semantics + conditional `e_score_correction_bias`, verified against the Python mlx-vlm oracle with gate regression tests; (2) the gundam tiling path in `DeepseekOCRProcessor` — the most parity-sensitive piece, covered by tensor-level parity tests.
- **Parity evidence:** L1 layer-level parity and L3 greedy-decode exact-match against Python `mlx-vlm==0.6.6` with a pinned HF weights revision.

## Commits (fork `feature/deepseek-ocr-v2` @ 8fd30db)

[...]

PR #571, opened 2026-08-23
Fix SSM gradient transforms across chunks

aleroot:
## Proposed changes

Falcon-H1 fine-tuning crashed because `ssmAttn` forced evaluation while MLX was tracing a gradient transformation.

This removes the internal evaluation barriers so gradients remain connected across SSM chunks and adds a regression test covering chunked gradient computation.

## Checklist

### DIFF
--- a/Libraries/MLXLLM/Models/SSM.swift
+++ b/Libraries/MLXLLM/Models/SSM.swift
@@ -238,7 +238,6 @@ public func ssmAttn(
         )
         ys.append(y)
         state = nextState
-        asyncEval(y, nextState)
         if let currentLengths = lengths {
             lengths = currentLengths - step
         }
@@ -248,7 +247,6 @@ public func ssmAttn(
     guard let state else {
         fatalError("SSM update did not produce a state")
     }
-    eval(y, state)
     return (y, state)
 }
 
--- a/Tests/MLXLMTests/SSMTests.swift
+++ b/Tests/MLXLMTests/SSMTests.swift
@@ -4,6 +4,36 @@ import MLX
 import MLXLLM
 import Testing
 
+@Test func testSSMAttnSupportsGradientTransformsAcrossChunks() {
+    let x = MLXArray.ones([1, 5, 2, 2])
+    let aLog = MLXArray.zeros([2])
+    let inputMixing = MLXArray.ones([1, 5, 1, 3])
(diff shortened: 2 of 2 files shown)


davidkoski:
I am curious what the underlying issue is here - could it be related to https://github.com/ml-explore/mlx-swift/pull/461?

Do we have a vulnerability in general when doing fine tuning + eval with the async evals?  I wonder if we need a _synchronous only_ bit on `GenerateParameters`?

Or perhaps this is just eval-ish code inside something that is going to run through compile (this seems more likely looking at where this code lives)?

davidkoski:
Reading the description more closely it isn't compile (but the same issue) it is `grad()` or `valueAndGrad()`.

This change looks good.

I wonder if we should consider something in mlx-swift with a task local that would either allow `eval` / `asyncEval` to NOP in these cases?  `item()` can't be called, but the eval variants should be.  They could be treated as advisory under grad/compile.

davidkoski (approved):
Looks good to me.

aleroot:
> I am curious what the underlying issue is here - could it be related to [ml-explore/mlx-swift#461](https://github.com/ml-explore/mlx-swift/pull/461)?
> 
> Do we have a vulnerability in general when doing fine tuning + eval with the async evals? I wonder if we need a _synchronous only_ bit on `GenerateParameters`?
> 
> Or perhaps this is just eval-ish code inside something that is going to run through compile (this seems more likely looking at where this code lives)?

#461 and this share the tracing/evaluation surface, but they fix different failures. #461 fixes a cross-thread lock inversion while here `ssmAttn` was directly materializing tracer arrays inside grad/valueAndGrad. A synchronous generation flag would not fix this because synchronous eval is also invalid on tracers, and training does not use GenerateParameters. I agree there may be room in mlx-swift for an explicitly advisory evaluation hint that no-ops during transforms, but I would keep ordinary eval strict.

aleroot:
> Reading the description more closely it isn't compile (but the same issue) it is `grad()` or `valueAndGrad()`.
> 
> This change looks good.
> 
> I wonder if we should consider something in mlx-swift with a task local that would either allow `eval` / `asyncEval` to NOP in these cases? `item()` can't be called, but the eval variants should be. They could be treated as advisory under grad/compile.

See [this](https://github.com/ml-explore/mlx-swift-lm/pull/574)

PR #572, opened 2026-08-24
Fuse Qwen GDN input projections

aleroot:
## Proposed changes

Qwen 3.5’s Gated Delta Network currently runs four separate quantized projections over the same input. Each projection starts another GPU operation, adding overhead during text generation.

This PR combines those four projections into one operation and then splits the result into the expected outputs. It applies to both text-only and vision-language Qwen 3.5 models.

The optimization is enabled only when the projections use compatible standard quantization. Models using adapters, custom quantizers, or mixed quantization settings automatically keep the existing path. Checkpoint parameter names remain unchanged.

On my Apple M2 16GB RAM running `Qwen3.5-9B-4bit`, using alternating same-process comparisons:

- Prompt processing improved by approximately 3.3%.
- Token generation improved from 16.62 to 17.39 tokens/second, approximately 4.6%.
- Both paths generated identical output tokens.

The optimization can be disabled with `MLX_QWEN_FOUR_GDN=0` if needed.

## Checklist

  - No public API documentation changes were required; the complete DocC warnings-as-errors check passed.

### DIFF
--- a/Libraries/MLXLLM/Models/Qwen35.swift
+++ b/Libraries/MLXLLM/Models/Qwen35.swift
@@ -1180,7 +1180,7 @@ public class Qwen35TextModel: Module, LLMModel, KVCacheDimensionProvider {
         }
     }
 
-    public func prepareForInference() throws {
+    public func prepare() throws {
         for layer in model.layers {
             if let linearAttn = layer.linearAttn {
                 _ = try linearAttn.prepareFusedInputProjection()
@@ -1289,8 +1289,8 @@ public class Qwen35Model: Module, LLMModel, KVCacheDimensionProvider {
         try languageModel.newCache(parameters: parameters)
     }
 
-    public func prepareForInference() throws {
-        try languageModel.prepareForInference()
+    public func prepare() throws {
+        try languageModel.prepare()
     }
 
     public func sanitize(weights: [String: MLXArray]) -> [String: MLXArray] {
--- a/Libraries/MLXLLM/Models/Qwen35.swift
+++ b/Libraries/MLXLLM/Models/Qwen35.swift
@@ -168,7 +168,7 @@ public struct Qwen35TextConfiguration: Codable, Sendable {
 
 // MARK: - GatedDeltaNet
 
-final class Qwen35GatedDeltaNet: Module {
+final class Qwen35GatedDeltaNet: Module, InferenceStatePreparable {
     let hiddenSize: Int
     let numVHeads: Int
     let numKHeads: Int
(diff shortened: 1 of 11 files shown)


Libraries/MLXLLM/Models/Qwen35.swift
```diff
@@ -336,6 +342,104 @@ final class Qwen35GatedDeltaNet: Module {
...
+    }
+
+    var hasFusedInputProjection: Bool { fusedInProj != nil }
+
+    /// Build one physical quantized projection while retaining the four named
+    /// module paths as storage-sharing views. This runs at most once between
+    /// parameter/module updates; failed eligibility checks are not repeated on
+    /// every token.
+    @discardableResult
+    func prepareFusedInputProjection() -> Bool {
```
davidkoski: This ends up mutating self in a call chain from callAsFunction (inference).  Generally we wouldn't want mutation during inference, but I think this is really just lazily building this fused layer (so in theory it fits, if done in a thread safe manner).

I think this either needs a lock OR this would have do be done from `Module.update(parameters:)`.  The latter isn't great because LoRA and other code would go through that as well.  We could potentially add a `Module.didLoad()` in mlx-swift to be called by the `load()` code.

What do you think?

Libraries/MLXLLM/Models/Qwen35.swift
```diff
@@ -336,6 +342,104 @@ final class Qwen35GatedDeltaNet: Module {
...
+                    "in_proj_b": .value(projection.sourceViews[2]),
+                    "in_proj_a": .value(projection.sourceViews[3]),
+                ]), verify: [])) != nil
+        else {
+            return false
+        }
+
+        // updateModule invalidates while the views are installed. Publish the
+        // fused projection only after the stable source topology is in place.
+        fusedInputProjectionAttempted = true
```
davidkoski: There is a case where:

```swift
    override func updateModule(key: String, _ value: Any) throws {
        try super.updateModule(key: key, value)
        if key == "in_proj_qkv" || key == "in_proj_z"
            || key == "in_proj_b" || key == "in_proj_a"
        {
            invalidateFusedInputProjection()
```

could have marked this as false in one call and there was a throw on another key.  That would leave `fusedInputProjectionAttempted = false` and it would retry this logic over and over.  Perhaps this needs to be in a `defer` block (if I am understanding the purpose correctly that this is used to prevent retrying)?

davidkoski (changes requested):
See questions about mutation.

Libraries/MLXLLM/Models/Qwen35.swift
aleroot: Agree. The fused representation should be prepared after quantization and weight loading, not lazily from callAsFunction. The benchmark already prepares it eagerly, so this preserves the measured gain. I’ll add a post-load preparation hook, make forward read-only, and make invalidation exception-safe for partial parameter/module updates.

davidkoski: Maybe it should be at the LanguageModel level instead of Module -- that might make more sense as the LanguageModel has a lifecycle defined by the library.

aleroot: I moved preparation to the LanguageModel lifecycle and simplified the API to a parameterless prepare() throws overload. The default implementation is a no-op, while the Qwen 3.5 LLM and VLM models explicitly prepare their derived fused GDN projections.
This is a much bigger footprint than before, but the refactor should hopefully turn out useful in the future

davidkoski (approved):
Changes look good -- new prepare lifecycle looks good!  Thank you!

PR #573, opened 2026-08-25
Qwen direct reduction

aleroot:
## Proposed changes

Qwen 3.5 Mixture-of-Experts models currently create a large temporary tensor while combining the selected expert outputs during prompt processing. This consumes extra memory and adds GPU work.

This PR adds a direct reduction path that combines the expert outputs into their final result without creating that intermediate tensor. Unsupported model configurations and small decode workloads automatically retain the existing implementation.

The optimization:

- Is enabled by default without an environment flag.
- Preserves the existing numerical results exactly in the tested configuration.
- Applies only to compatible frozen, quantized Qwen 3.5 models.
- Retains automatic fallbacks for training, custom layers, unsupported data types, and unsupported tensor shapes.
- Adds correctness and fallback coverage without adding a performance microbenchmark to CI.

There is no corresponding issue.

## Benchmarks

Benchmarks were run locally on an **Apple M2 Mac with 16 GB of RAM** using a production-shaped workload:

- 2,048 prompt tokens
- 8 selected experts per token
- 2,048 hidden dimensions
- 5 warm-up runs and 25 alternating measurements

Median results:

| Implementation | Time |
| --- | ---: |
| Existing reduction | 2.9663 ms |
| Direct reduction | 1.8422 ms |
| Improvement | **1.61× faster** |

This represents approximately a **38% reduction in reduction-stage latency**. The benchmark measures this operation in isolation, not complete model generation, so the overall model speedup will be smaller.

The microbenchmark was removed from the submitted test suite; correctness and fallback tests remain.

## Checklist

  - No user-facing documentation changes are required because this is an internal optimization. The documentation validation passed.


Libraries/MLXLMCommon/SwitchLayers.swift
```diff
@@ -120,11 +187,87 @@ public class SwitchGLU: Module {
...
+        }
+
+        let projected = projectExperts(x, indices)
+        guard let inverseOrder = projected.inverseOrder,
+            projected.output.ndim == 3,
+            projected.output.dim(-2) == 1,
+            projected.output.dim(-1) == inputDims,
+            projected.output.dtype == .bfloat16
+        else {
+            return legacyWeightedReduction(projected, indices: indices, weights: weights)
```
davidkoski, line 264: I like this pattern -- having a fallback in plain mlx code if the metal kernel doesn't always work.  I realized this pattern will be important when we get the cuda version of this running.

Filed #588 to capture that idea.

davidkoski (approved):
Code looks good, thank you!

PR #574, opened 2026-08-25
Strengthen chunked SSM gradient tests

aleroot:
## Proposed changes

Follow-up to #571.

The regression test added in #571 verified that `grad` completed and produced a finite, nonzero gradient for `x`. However, `ssmAttn` includes the residual path `x * D`, so that assertion could still pass even if the recurrent gradient graph were disconnected across chunk boundaries.

This PR strengthens the regression coverage by:

- comparing chunk sizes 1, 2, and 3 against a single-chunk reference;
- verifying output, final recurrent state, and loss equivalence;
- comparing gradients for `x`, `ALog`, `B`, `C`, `D`, `dt`, `dtBias`, and the incoming recurrent state;
- adding a state-only loss with `D = 0`, proving that inputs in the first chunk influence the final recurrent state through the cross-chunk recurrence.

This is a test-only change and does not alter any behavior.

## Checklist

- [x] I have updated the necessary documentation (not needed; test-only change)

### DIFF
--- a/Tests/MLXLMTests/SSMTests.swift
+++ b/Tests/MLXLMTests/SSMTests.swift
@@ -4,17 +4,103 @@ import MLX
 import MLXLLM
 import Testing
 
-@Test func testSSMAttnSupportsGradientTransformsAcrossChunks() {
-    let x = MLXArray.ones([1, 5, 2, 2])
-    let aLog = MLXArray.zeros([2])
-    let inputMixing = MLXArray.ones([1, 5, 1, 3])
-    let outputMixing = MLXArray.ones([1, 5, 1, 3])
-    let residualScale = MLXArray.ones([2])
-    let dt = MLXArray.ones([1, 5, 2])
-    let dtBias = MLXArray.zeros([2])
+private func expectAllClose(
+    _ actual: MLXArray, _ expected: MLXArray, label: String, rtol: Double = 1e-4,
+    atol: Double = 1e-5
+) {
+    guard actual.shape == expected.shape else {
+        Issue.record("\(label) shape \(actual.shape) != \(expected.shape)")
+        return
+    }
+    #expect(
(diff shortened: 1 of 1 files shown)


aleroot:
@davidkoski Following up on your comment in [#571](https://github.com/ml-explore/mlx-swift-lm/pull/571#issuecomment-5402792492), I agree that a framework-level mechanism may be worth exploring.
Rather than making ordinary eval or asyncEval silently advisory, would an explicitly named API in mlx-swift make sense?

`public func asyncEvalHint(_ arrays: some Collection<MLXArray>)`


The intended contract would be:
- schedule evaluation normally outside graph transformations;
- become a no-op when the arrays are tracers;
- never be required for correctness;
- keep eval, asyncEval, and item() strict.

Or maybe adding just an optional hint parameter to the asyncEval ? 

What do you think?

davidkoski:
Yes `asyncEvalHint` sounds like a good idea -- it lets the developer express intent directly.

Tests/MLXLMTests/SSMTests.swift
```diff
@@ -4,17 +4,103 @@ import MLX
...
+        Issue.record("\(label) shape \(actual.shape) != \(expected.shape)")
+        return
+    }
+    #expect(
+        allClose(actual, expected, rtol: rtol, atol: atol).item(Bool.self),
+        "\(label) values differ")
+}
+
+@Test func testSSMAttnChunkedMatchesUnchunkedValuesAndGradients() {
+    MLXRandom.seed(11)
```
davidkoski, line 21: This is perhaps a wider concern with swift testing -- it can run tests in parallel and this operates on global state.  I think we need to move everything to use `withRandomState()`.  I will file an issue as this isn't the only spot.

davidkoski (approved):
Looks good, thanks!  See #577 on the random state.

PR #575, opened 2026-08-25
Parallelize weight loading with byte-balanced contiguous groups(up to ~1.8x faster model loads)

aleroot:
## Proposed changes

**Load model weight files concurrently in byte-balanced groups -> up to ~1.8× faster cold loads for large models.**

### Issue

When loading a model, the weights in each `.safetensors` file were read and materialized **one after another, in a single stream**. Modern Apple silicon SSDs can read at 6+ GB/s, but a single sequential reader tops out around 3–4 GB/s, so for big models (e.g. an 18 GB 30B 4-bit model) roughly half the disk's speed was left on the table, and users just sat waiting.

### Solution

- **Small files aren't split.** A minimum of 256 MiB per group means a small model loads exactly as before no overhead, no regression.
- **Concurrency scales with the machine:** `clamp(coreCount, 4...16)` workers, so it behaves well from an iPhone to a Mac Studio.
- **Malformed/exotic headers fall back** to loading the whole file as a single group identical to the old behavior.
- **Duplicate tensor names across files** resolve in file order (later file wins), exactly matching the previous serial loader.

### Benchmarks

Measured results: Apple M4 Pro MacBook Pro 24GB Memory - 1TB Disk(NVMe ~6.1 GB/s raw)

| Scenario | Before | After |
|---|---|---|
| Qwen 9B (5 GB), cold cache | ~2.2–2.6 s | ~1.6–1.8 s |
| Muse-Glimmer-30B-4bit (18 GB), cold cache | ~5.6–6.1 s | ~3.1 s (weight load) |
| Warm cache (both) | — | up to ~2× (>10 GB/s effective) |

Loaded weights were verified **bit-identical** to the serial loader (per-tensor checksums over 1,547 tensors).

## Checklist

- [ ] I have updated the necessary documentation (if needed) — no user-facing API changes; behavior is a drop-in speedup

### DIFF
--- a/Libraries/MLXLMCommon/Load.swift
+++ b/Libraries/MLXLMCommon/Load.swift
@@ -4,6 +4,53 @@ import Foundation
 import MLX
 import MLXNN
 
+#if canImport(Darwin)
+import Darwin
+#endif
+
+// `F_RDADVISE` queues real I/O. Keep at most 128 MiB in flight; larger windows
+// can compete with the reads MLX is about to issue instead of hiding their latency.
+private let maximumPrefetchShardCount = 2
+private let maximumPrefetchBytesPerShard: Int64 = 64 * 1024 * 1024
+
+/// A small lead window overlaps useful I/O with model setup without flooding the
+/// unified buffer cache or competing with MLX's own reads.
+func safetensorReadAheadByteCount(fileSize: Int64) -> Int32 {
+    guard fileSize > 0 else { return 0 }
+    return Int32(Swift.min(fileSize, maximumPrefetchBytesPerShard))
+}
+
+#if canImport(Darwin)
+/// Ask Darwin to asynchronously read the selected safetensors into its unified
+/// buffer cache. Advice is best-effort: an open, stat, or `fcntl` failure leaves
+/// the normal MLX loading path unchanged.
+private func prefetchSafetensorLeadWindows(_ urls: [URL]) {
+    guard !urls.isEmpty else { return }
+
+    let shardCount = Swift.min(urls.count, maximumPrefetchShardCount)
+    DispatchQueue.concurrentPerform(iterations: shardCount) { index in
+        urls[index].withUnsafeFileSystemRepresentation { path in
+            guard let path else { return }
+
+            let descriptor = open(path, O_RDONLY | O_CLOEXEC)
+            guard descriptor >= 0 else { return }
+            defer { close(descriptor) }
+
+            var status = stat()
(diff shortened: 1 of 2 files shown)


Libraries/MLXLMCommon/Load.swift
```diff
@@ -4,6 +4,220 @@ import Foundation
...
+struct SafetensorSpan {
+    let name: String
+    let byteCount: Int64
+}
+
+/// The tensors of the safetensors file at `url`, ordered by their position in the file.
+///
+/// Reads the 8-byte header length and the JSON header only. Throws when the file is not a
+/// well-formed safetensors file; callers fall back to loading the file whole.
+func safetensorSpansInFileOrder(url: URL) throws -> [SafetensorSpan] {
```
davidkoski, line 30: Under the hood the mlx API is `std::unordered_map`:

```
using SafetensorsLoad = std::pair<
    std::unordered_map<std::string, array>,
    std::unordered_map<std::string, std::string>>;
```

too bad -- `load_safetensors` is iterating in file order at the bottom of the call stack.

I think maybe should eventually land in mlx-swift as a concurrent synchronous load function, but it is an implementation detail so it is fine starting here.

davidkoski:
Another thought, somewhat orthogonal to this: I have been thinking about what we might be able to do with mapped tensor files.  Can you map a file and make an IOSurface/MTLBuffer out of it?  Claude says yes.  Would it be interesting on iOS where the memory is backed by a file and becomes pageable (I think)?  Is this fast enough to do anything interesting?

I am not 100% sure this would even work.

Libraries/MLXLMCommon/Load.swift
```diff
@@ -4,6 +4,220 @@ import Foundation
...
+                }
+            } else {
+                items.append(WorkItem(file: file, url: url, names: nil))
+            }
+        }
+        return items
+    }()
+
+    let state = ConcurrentLoadState(fileCount: urls.count)
+    DispatchQueue.concurrentPerform(iterations: items.count) { index in
```
davidkoski, line 194: Just musing here...

I think it is an anti-pattern to call blocking GCD apis inside a swift asynchronous task (maybe some risk of thread starvation).  The entire Load API set is all synchronous, but it is called from async contexts.  Probably not an issue -- we were doing the same thing with `eval()`.  No GCD, but was blocking on I/O calls.

davidkoski (approved):
This looks great, very nice addition!

PR #579, opened 2026-08-26
Suspend instead of blocking cooperative threads while weights load

aleroot:
## Proposed changes

Follow-up to a review comment on #575: the `loadWeights` API is synchronous and blocks its thread on file I/O and `DispatchQueue.concurrentPerform`, but every model factory calls it from an async context. That parks a Swift concurrency cooperative thread for the whole load, which is an anti-pattern and a potential source of thread starvation, since the cooperative pool only has one thread per core and expects them to never block.

This PR adds an `async` overload of `loadWeights` that hops to a global GCD queue and suspends the caller instead of blocking it. The four async model factories (LLM, VLM, embedders, MTP drafter) now use it. The synchronous overload is unchanged and remains for synchronous callers such as model conversion.

No performance change.

Note: downstream code that calls `loadWeights` from an async context will now resolve to the async overload and needs to add `await`.

## Checklist

### DIFF
--- a/Libraries/MLXEmbedders/ModelFactory.swift
+++ b/Libraries/MLXEmbedders/ModelFactory.swift
@@ -231,7 +231,7 @@ public final class EmbedderModelFactory: GenericModelFactory {
         async let tokenizerTask = tokenizerLoader.load(
             from: configuration.tokenizerDirectory)
 
-        try loadWeights(
+        try await loadWeights(
             modelDirectory: modelDirectory, model: model,
             perLayerQuantization: baseConfig.perLayerQuantization,
             weightFileSelection: configuration.weightFileSelection)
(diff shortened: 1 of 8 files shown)


aleroot:
@thechriswebb just to understand how this labelling system is working, when you mark something as needs-changes does it mean I need to make some changes to meet the gate ? Thanks.

thechriswebb:
Hey @aleroot! Yes, that's the goal. The two contributor-facing labels will be `needs-changes` and `needs-lint` whereas the others are more maintainer-facing labels that help us keep track of PR state. It's something I'm working on now, so would love to hear any feedback you have.

aleroot:
> Hey @aleroot! Yes, that's the goal. The two contributor-facing labels will be `needs-changes` and `needs-lint` whereas the others are more maintainer-facing labels that help us keep track of PR state. It's something I'm working on now, so would love to hear any feedback you have.

Sounds good, they give a bit of color which is always good 😀.

davidkoski (approved):
Looks great, thank you!

davidkoski (approved):
Change looks good.  Once mlx-swift is released with https://github.com/ml-explore/mlx-swift/pull/462 we can do followup work to use that instead.

PR #580, opened 2026-08-26
Defer vision weight loading by default for MuseGlimmer

aleroot:
## Proposed changes

Today while I was using Lampo for helping me with some coding tasks at work I discovered disappointing performance of the Muse Glimmer model. So I have implemented two decode-path optimizations for Muse-Glimmer-30B, both bit-identical to the reference:

**1. Compiled kernel fusions.** The model runs 4 `MuseCenteredRMSNorm`s × 52 layers per decoded token; eagerly that's ~7 kernel launches each and dispatch overhead dominates. Now fuses the norm body, residual-add+norm, attention sigmoid gate, and MLP `silu·up` with `compile(shapeless: true)`  same ops, same order, same dtypes as the eager code (same technique as `Gemma4Text.swift`). Verified bitwise-identical: 288/288 norm self-test vs the eager reference, plus end-to-end greedy token comparison on the 4-bit checkpoint. (Note: `MLXFast.rmsNorm` with a folded `1+weight` is *not* an equivalent rewrite  it drifts 1 bf16 ulp and changes decode choices; documented in-source.) @CharlieTLe please if you got time: review this.

**2. Vision weights lazy loading by default.** *Added`DeferredWeightsProviding` protocol, `Load.swift` plumbing, MuseGlimmer conformance, and the measurement table 19.42→15.68 GB, 4.55→3.44 s load, ~13.2→~15.0 tok/s decode, identical greedy tokens and vision checksum.)*

### Vision lazy loading

Muse-Glimmer-30B-4bit is a dense VLM whose vision tower ships unquantized (bf16, ~3.7 GB). Text-only users pay for it anyway: on a 24 GB M4 Pro the full 19.4 GB of weights exceeds the Metal wired limit (19.07 GB `recommendedMaxWorkingSetSize`), causing page eviction during decode and dropping throughput to ~13 tok/s.

This change set makes vision weights lazy **by default**, with no API changes required by callers.

#### Benchmarks

Measured on an M4 Pro 24 GB (2048-token prompt, 256 greedy decode, 4-bit checkpoint):

| | Before | After |
|---|---|---|
| Resident memory after load (text-only) | 19.42 GB | **15.68 GB** |
| Load time | 4.55 s | **3.44 s** |
| Decode | ~13.2 tok/s | **~15.0 tok/s** |

Correctness is unchanged: greedy token streams are bitwise-identical to the eager baseline, and a forced materialization of all vision weights produces a checksum identical to eager loading (815 arrays). Models that don't adopt the protocol are unaffected (default is no deferred prefixes).

In the next few days I will try to see if there is some performance that can be squeezed out of it as unfortunately I have got limited hardware at the moment to work with ... 

## Checklist

### DIFF
--- a/Libraries/MLXLMCommon/ChatSession.swift
+++ b/Libraries/MLXLMCommon/ChatSession.swift
@@ -45,6 +45,7 @@ public struct SpeculativeDecodingConfig: Sendable {
     package enum DraftModelSource: Sendable {
         case loaded(ModelContainer)
         case deferred(bytes: Int, @Sendable () async throws -> ModelContainer)
+        case lookup(PromptLookupDraftModel.Configuration)
     }
 
     package let draftModelSource: DraftModelSource
@@ -104,6 +105,40 @@ public struct SpeculativeDecodingConfig: Sendable {
         self.memoryPolicy = memoryPolicy
     }
 
+    /// Initialize model-free speculative decoding by prompt lookup.
+    ///
+    /// Draft tokens come from n-gram matches over the session's own token
+    /// history (see ``PromptLookupDraftModel``), so no draft model is loaded,
+    /// no additional weight memory is used, and no memory policy applies. The
+    /// emitted stream is what the main model would produce on its own — a
+    /// draft is accepted only when it equals the verifier's sampled token.
+    ///
+    /// This shines on agentic/coding workloads, where generation constantly
+    /// copies spans that already appear in the context.
+    ///
+    /// - Parameters:
+    ///   - lookup: n-gram matching configuration; its
+    ///     ``PromptLookupDraftModel/Configuration/vocabularySize`` must cover
+    ///     the main model's vocabulary
+    ///   - numDraftTokens: tokens proposed per verification cycle (vLLM ships
+    ///     3–4 for its ngram speculator)
+    public init(
+        lookup: PromptLookupDraftModel.Configuration,
+        numDraftTokens: Int = 3
+    ) {
+        self.draftModelSource = .lookup(lookup)
+        self.numDraftTokens = numDraftTokens
+        self.memoryPolicy = nil
+    }
+
+    /// The prompt-lookup configuration, when this config drafts by n-gram lookup.
+    public var lookup: PromptLookupDraftModel.Configuration? {
+        if case .lookup(let configuration) = draftModelSource {
+            return configuration
+        }
+        return nil
+    }
+
     package var estimatedDraftModelBytes: Int? {
         guard case .deferred(let bytes, _) = draftModelSource else {
             return nil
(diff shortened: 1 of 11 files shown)


aleroot:
## Update: performance follow-ups for agentic workloads

While experiment with my agentic coding app on top of this branch (M4 Pro, 24 GB, Muse-Glimmer-30B-4bit) I found the dominant cost per turn wasn't decode speed it was that **prompt-cache reuse was completely disabled for this model**, plus two decode-side improvements. 
So I have posted three additional commits:

### 1. Text-only prompts no longer carry an all-ones attention mask (`5aa6e1e`)

`MuseGlimmerProcessor.prepare` attached a dense all-ones int8 mask to *every* prompt, including pure text. `ChatSession` treats any masked input as non-trimmable, so **every turn re-prefilled the entire conversation**  on long tool-calling sessions that meant ~253k cumulative prompt tokens with 0% cache reuse. An all-ones mask is semantically a no-op, so text-only prompts now omit it. Multimodal prompts are unchanged.

| Metric (multi-turn tool-calling session) | Before | After |
|---|---|---|
| KV-cache reuse across turns | 0% (full re-prefill) | e.g. prefill **707**, reused **4261** |
| Follow-up turn TTFT | 35–50 s | **0.87–0.92 s** |

### 2. Speculative decoding rewinds via staged cache rounds (`bc2fbf7`)

`SpeculativeTokenIterator` rewound rejected drafts with a raw `trim`, which is only sound while the cache still holds every position — once a `RotatingKVCache` wraps past its window, trim silently corrupts state. The verify loop now stages each round (`beginRound` / `commit(retaining:)` / `rewindLastRound`) so rejected tokens are rewound exactly even after the cache wraps. Caches that can't support exact rewind trigger a permanent, recorded fallback to plain decoding rather than risking divergence. Cache-less drafters keep working.

New tests verify token-for-token equivalence with plain generation across a rotating-cache wrap (draft widths 2/3/4), exact rewind of rejected rounds after wrapping, and both fallback paths.

### 3. Opt-in prompt-lookup speculative decoding (`0824f39`)

A model-free drafter that proposes continuations by matching the longest recent n-gram against earlier prompt/generation history; no second model, no extra weights, negligible memory. Drafts are verified by the target model through the existing speculative loop, so **output is exactly what plain decoding produces**.

```swift
let session = ChatSession(
    model,
    speculativeDecoding: .init(lookup: .init(), numDraftTokens: 3)
)
```

Measured on my current work machine (Muse-Glimmer-30B-4bit, M4 Pro 24 GB):

| Workload | Plain | Prompt-lookup | Acceptance | Output |
|---|---|---|---|---|
| Repetition-heavy (tool outputs / code quoted back) | 14.8 tok/s | **27.4 tok/s** | 98% | byte-identical |
| Novel prose | ~14.7 tok/s | 8.5–12.4 tok/s | low | identical |

It shines exactly on agentic coding traffic (file contents, identifiers, and tool results being restated) and is neutral-to-slower on novel text, so it is **opt-in** and off by default.

@davidkoski This PR has now drifted a bit 😅 from the original purpose, but it was just more convenient for me to work on this branch, let me know if you want me to split the pull request or if there is something you don't like and you want to change. Apart from the technicalities I think now this model is finally usable on my work laptop with my Lampo app.

CharlieTLe (approved):

PR #581, opened 2026-08-26
Add resumable Qwen MTP speculative decoding to ChatSession

ngutech21:
## Proposed changes

This PR adds MTP drafters as a first-class speculative-decoding strategy for `ChatSession` and enables Qwen text and VLM drafters to resume safely across cached turns.

- Add `SpeculativeDecodingConfig(mtpDrafter:blockSize:)` with typed block-size validation while preserving existing draft-model APIs.
- Route eligible generations through `MTPSpeculativeTokenIterator`, with target-only fallback for unsupported sampling, media inputs, incompatible caches, or missing MTP state.
- Persist target KV cache, model state, transcript, and MTP drafter continuation atomically across turns, tool restarts, EOS, and cancellation.
- Resume Qwen text and VLM drafter caches over prompt suffixes, including M-RoPE state and hidden-state aggregation during warm VLM prefill.
- Validate target family, hidden size, and vocabulary size before MTP prefill.
- Model speculative-cache reuse as an exhaustive capability enum and share shifted Qwen drafter finalization between text and VLM implementations.
- Preserve existing ordinary draft-model behavior and source compatibility.

## Testing

Coverage includes:

- Single-turn and warm multi-turn Qwen MTP generation
- Automatic tool-call restarts
- Early EOS and consumer cancellation
- Stateless, resumable, and non-resumable drafter behavior
- Raw prompt-cache, media, and sampling fallbacks
- Prompt-cache append, rewind, and rebuild decisions
- Qwen text/VLM target-family and dimension validation
- Qwen VLM continuation and M-RoPE alignment
- Existing eager and deferred draft-model configuration compatibility

Latest targeted verification:

- 20 Qwen MTP compatibility, registration, and Metal-backed tests passed
- `git diff --check` passed
- Strict Swift formatting reports no branch-introduced diagnostics; the four `Qwen35.swift` naming diagnostics are also present at the merge base
## Checklist


ngutech21:
@davidkoski could you take a look at this pr when you get a chance?

Libraries/MLXLMCommon/ChatSession.swift
```diff
@@ -1214,18 +1350,48 @@ public final class ChatSession {
...
                         }
 
-                        if speculativeDecoding != nil, requiresMainOnlyContinuation {
+                        if !speculationIsEligibleForParameters || requiresMainOnlyContinuation {
                             generation = try defaultGeneration()
+                        } else if let speculativeDecoding,
+                            case .mtp(_, let blockSize) = speculativeDecoding.strategy,
+                            let drafter = mtpDrafter
+                        {
+                            let iterator = try MTPSpeculativeTokenIterator(
```
aleroot: The continuation ownership looks like the right boundary for applications using ChatSession. One remaining concern: ChatSession directly constructs MTPSpeculativeTokenIterator, while its initial beginRound probe can still throw when the cache cannot stage a speculative verification block. That appears to fail the chat request rather than select target-only generation.
Could unsupported speculative-cache capability be handled explicitly before prefill, preserving the authoritative target cache? Please keep this distinct from catching arbitrary initialization errors, since some errors may occur after cache mutation. A focused test with a cache that supports ordinary generation but refuses speculative staging would establish the fallback contract for both cold and warm sessions.

Libraries/MLXLMCommon/MTPSpeculativeTokenIterator.swift
```diff
@@ -167,9 +217,9 @@ public struct MTPSpeculativeTokenIterator: TokenIteratorProtocol {
 
         let prefillStart = Date.timeIntervalSinceReferenceDate
         var mtpPrefill = parameters.prefill
-        if drafter.requiresPromptPrefill {
+        if drafter.requiresPromptPrefill || continuationToResume != nil {
```
aleroot, line 231: Stateful MTP and warm continuation currently override the caller’s prefill configuration with Int.max and .unchunked. For long RAG prompts or large continuation suffixes, this could increase peak memory and first-token latency substantially.
Could we preserve bounded chunking while aggregating the required hidden states, or provide an explicit admission/fallback policy for cases that require unchunked prefill? It would help to include cold-prompt and warm-suffix measurements covering peak memory and first-token latency alongside decode throughput.

Libraries/MLXLMCommon/ChatSession.swift
```diff
@@ -1214,18 +1350,48 @@ public final class ChatSession {
...
                         if carriesPreparedMedia || !draftCacheIsRecoverable {
                             // Drop the draft cache as `.appendSuffixToMain` does: retaining it
                             // would leave it at an offset the main cache never visits.
                             draftKVCache = nil
+                            mtpDrafterContinuation = nil
                             requiresMainOnlyContinuation = true
                         }
 
-                        if speculativeDecoding != nil, requiresMainOnlyContinuation {
+                        if !speculationIsEligibleForParameters || requiresMainOnlyContinuation {
```
aleroot, line 1371: Could completion metadata expose the reason when ChatSession bypasses speculation before constructing the iterator? Currently the sampling fallback has nil speculative telemetry and nil passthroughReason, making it indistinguishable from acceleration being disabled.
A small typed fallback reason covering unsupported sampling, media, cache capability, and missing continuation would let applications accurately explain why a configured accelerator was inactive. Zero proposals alone is ambiguous, especially for very short generations. This would also make fallback tests more precise.

aleroot:
@ngutech21 this is something I need as well, so thanks for implementing this. See if my comments make sense.

Libraries/MLXVLM/Models/Gemma4Assistant.swift
```diff
@@ -201,6 +201,16 @@ public final class Gemma4AssistantDraftModel: Module, MTPDrafterModel {
         super.init()
     }
 
+    public func validateCompatibility(with target: any LanguageModel) throws {
```
aleroot, line 204: Could Gemma4AssistantDraftModel.validateCompatibility validate the required dimensions and shared-KV layout, in addition to Gemma4BackboneProviding?
Currently a different Gemma variant can pass this check even though preProjection expects inputs sized from the drafter’s backboneHiddenSize.

Libraries/MLXLMCommon/ChatSession.swift
ngutech21: Thanks for the feedback. I have provided a fix:

The iterator now throws a dedicated `MTPInitializationError.unsupportedSpeculativeCache` when its initial staging probe fails. This happens **before target prefill**, while the target cache is still valid.

Libraries/MLXLMCommon/ChatSession.swift
ngutech21: Fixed the missing fallback diagnostics by adding typed reasons to ChatSession completion metadata, so applications can explain why configured speculative decoding was bypassed without changing generation behavior.

Libraries/MLXVLM/Models/Gemma4Assistant.swift
ngutech21: Fixed: `Gemma4AssistantDraftModel.validateCompatibility` now validates backbone/vocabulary dimensions, shared-KV layout, and sliding-window limits. Incompatible pairings are rejected before prefill or drafting. Regression tests cover both `Gemma4` and `Gemma4Unified`; the full package test run and formatting checks passed.

Libraries/MLXLMCommon/MTPSpeculativeTokenIterator.swift
ngutech21: We could preserve the configured chunk size for both target and drafter prefill, aggregating target hidden states across chunks while maintaining correct alignment for warm continuations. Since this primarily affects long prompts and large continuation suffixes, I’d suggest addressing it in a follow-up PR to keep this PR small and focused. What do you think?

aleroot: Yes, I think it is reasonable. 
Thanks, nice work overall.

PR #583, opened 2026-08-27
Replace MLXRandom.seed with task-local withRandomState in tests

### LINKED ISSUE #577: [BUG] tests use MLXRandom.seed
Many tests use MLXRandom.seed.  This is ok in serialized tests, but swift testing allows for parallel execution.  On top of not being thread safe, it is also not reproducible if multiple tests are consuming random state at the same time.

Move all tests to use `withRandomState` and maybe consider marking `MLXRandom.seed` (mlx-swift) as deprecated (separate issue but my concern is that it is rarely safe).

aleroot:
## Proposed changes

Fixes #577.

Many tests seeded the process-wide random stream via `MLXRandom.seed`. That was fine when tests ran serially, but Swift Testing runs suites in parallel by default: any two tests consuming the global stream concurrently interleave their draws, so seeded tests were neither isolated nor reproducible.

## Checklist

### DIFF
--- a/Tests/MLXLMTests/ContinuationAssertions.swift
+++ b/Tests/MLXLMTests/ContinuationAssertions.swift
@@ -88,33 +88,36 @@ struct ContinuationAssertions {
     func assertWarmTextContinuation<M: LanguageModel>(
         _ model: M, file: StaticString = #filePath, line: UInt = #line
     ) throws {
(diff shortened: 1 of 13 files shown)


davidkoski (approved):
Perfect, thank you!

PR #584, opened 2026-08-27
RotatingKVCache: make trim wrap-aware instead of corrupting the ring

aleroot:
## Proposed changes

Continuing with usage of Muse Glimmer, I have found another issue...

`RotatingKVCache` is a fixed-size, circular memory that sliding-window models (like Muse-Glimmer) use to remember recent context: once it fills up, new tokens overwrite the oldest ones ("wrapping").

**The problem:** `trim(_:)`  used by speculative decoding to rewind tokens that were generated ahead of time but rejected assumed the cache had never wrapped. On a wrapped cache it just decremented the counters, which left stale entries in the *middle* of the logical timeline instead of removing the newest ones. Result: silently corrupted context for anything trimming past the window boundary. In practice this capped speculative decoding at `prompt + generation < window size`.

**The fix:** trimming a wrapped cache now first unrolls the ring back into chronological order, then cuts the newest rows always consistent, and clamped so it never touches the pinned prefix (`keep`) region.

## Checklist

### DIFF
--- a/Libraries/MLXLMCommon/KVCache.swift
+++ b/Libraries/MLXLMCommon/KVCache.swift
@@ -568,6 +568,16 @@ public class RotatingKVCache: BaseKVCache, CustomDebugStringConvertible {
     private var step: Int
     private var idx: Int = 0
 
+    /// Whether the buffer is in ring layout: rows at `idx...` chronologically precede
+    /// rows at `keep ..< idx`. `nil` only after restoring a legacy metaState, where the
+    /// layout is derived on demand: of the states legacy code could save, exactly the
+    /// wrapped ring has a mid-buffer write position with more history than rows behind it.
+    private var wrappedFlag: Bool? = false
+
+    private var wrapped: Bool {
+        wrappedFlag ?? (idx < (keys?.dim(2) ?? 0) && offset > idx)
+    }
+
     /// Model-native sliding-window caches deliberately keep their architectural
     /// window and do not participate in requested-capacity validation.
     package var capacityOrigin = CapacityOrigin.modelNative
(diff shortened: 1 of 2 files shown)


aleroot:
Partially supersedes #424.

Fixes the direct trim(_:) footgun: a wrapped ring no longer corrupts when rewound. trim unrolls, cuts the newest rows, and live state is bounded by idx + wrapped.
Does not close #424. isTrimmable stays exact (offset < maxSize), so trimPromptCache / SpeculativeTokenIterator still no-op after wrap, and prefix reuse still falls back to a full re-prefill. That is intentional a post-wrap trim is consistent, not exact.
Leave #424 open for the iterator and reuse paths.

Libraries/MLXLMCommon/KVCache.swift
```diff
@@ -820,14 +835,23 @@ public class RotatingKVCache: BaseKVCache, CustomDebugStringConvertible {
...
                 self.capacityOrigin = .modelNative
             }
+            if newValue.count == 7 {
+                guard let wrappedValue = Bool(newValue[6]) else {
+                    fatalError("Invalid RotatingKVCache wrapped flag '\(newValue[6])'")
+                }
+                self.wrappedFlag = wrappedValue
+            } else {
+                // Legacy metaState: derive the layout lazily via `wrapped`.
+                self.wrappedFlag = nil
```
davidkoski, line 856: I have a concern on this one: we would end up in this state if restoring older metaState.  Leaving this as nil defers to `wrapped`.  I think there is a gap where the saved cache was at `idx == maxCacheSize` and it has wrapped.  Reads work ok, but trim takes the else branch:

```swift
    public override func trim(_ n: Int) -> Int {
        if wrapped, let keys = self.keys, let values = self.values {
...
        }

// ends up here, which I think may be wrong
        let trimmed = Swift.min(Swift.min(offset, idx), n)
        offset -= trimmed
...
    }
```

I wonder if it would be better to set the value here:

```swift
self.wrapped = idx < (keys?.dim(2) ?? 0) && offset > idx
```

`wrapped` is really indicating if the buffers are in temporal order or not (wrapped true == not in temporal order).

If this was set eagerly here, then `wrappedFlag` doesn't need to be optional (since it has a default value).

Please check my math on this -- I think that is right but it is a bit convoluted so I am not 100% sure.

davidkoski (changes requested):
Check my question on the restore of metaState.  If my idea is wrong then I think this is ready for merge.  Thanks!

Libraries/MLXLMCommon/KVCache.swift
aleroot: You’re right. The else branch is initially valid, but changing idx makes the deferred predicate flip, resurrecting trimmed rows.
Fixed by freezing the inferred layout before mutation. I kept inference deferred until then to support either restoration order. This also prevents the layout changing between trimming keys and values.

aleroot:
> Check my question on the restore of metaState. If my idea is wrong then I think this is ready for merge. Thanks!

Done, it should be ready now. Thanks

davidkoski (approved):
Looks good, thank you!

PR #587, opened 2026-08-27
Leverage GitHub workflows to keep PR labels up-to-date

thechriswebb:
## The problem

The triage labels on a pull request are set by hand. Someone opens the run, reads which job failed, and picks a label. That repeats on every push, and gets skipped often enough that the labels drift from the truth. A label can easily go stale: a pull request can sit on `needs-review` after the code a reviewer would open has been replaced.

## What this PR does

- Labels a pull request `ci-running` while its run is going, then replaces it when the run ends: `needs-review` when everything passed, `needs-lint` when only formatting failed, `needs-changes` when something the author owns failed, and `needs-ci` when the run needs repeating. If the label has already been labeled `approved` it becomes `ready-to-merge`.
- Asks for another run rather than a change from the author, when only the build machine failed. Three steps that check the machine now begin with `Infra: `. Without it, a runner missing its Metal toolchain would tell every open pull request author to fix code that is fine.
- When a contributor pushes a new commit, every label about the previous one goes, an approval included, and the pull request waits for CI again.
- Throws away a CI result if the contributor pushed while the run was going. It tested code that is gone.
- Puts all the label rules in one file, with tests that run on every pull request.
- The two workflows that write labels never check out the pull request. They run only code from this repo, so a contributor cannot reach the token that writes labels.

Nothing here changes when CI runs, or who can start it.

## Tests

- 39 unit tests on the rules that pick the label. None touch the network.
- One test per label, plus the awkward cases: formatting failing on its own, formatting failing next to a real build failure, a cancelled run, a failure with no failing step.
- Tried for real on a throwaway repo, with a pull request from a second account.
- No label while a run waits for approval. The label appears once approved.
- Pushing mid-run: old result thrown away, pull request queued again.
- Each kind of failure produced the label it should.

## Checklist

## AI usage

- [x] AI usage disclosure: I used AI for all of this, but can confirm I understand the code.

### DIFF
--- a/.github/scripts/ci-state.js
+++ b/.github/scripts/ci-state.js
@@ -46,6 +46,20 @@ function onStart(labels) {
 // for a change.
 const INFRA_PREFIX = "Infra: ";
 
+// The names these steps carried before they gained the prefix. A fork pull
+// request runs its own copy of the CI workflow, so a copy branched before the
+// rename still reports the old names. Retire this list together with the
+// matching one in the mlx-ops config, once no open pull request predates it.
+const LEGACY_INFRA_STEPS = [
+  "Verify MetalToolchain installed",
+  "Assert Xcode 27 and the macOS 27 SDK",
+  "Install MetalToolchain",
+];
+
+function isInfraStep(name) {
+  return name.startsWith(INFRA_PREFIX) || LEGACY_INFRA_STEPS.includes(name);
+}
+
 // The three orderings below carry the whole correctness of this function.
 // The no-failed-step test comes first, so a timeout or a lost runner reads as
 // "run it again" rather than as a fault. The infrastructure test comes before
@@ -68,7 +82,7 @@ function classify({ conclusion, jobs, labels }) {
       .map((step) => step.name));
 
   if (failedSteps.length === 0) return "needs-ci";
-  if (failedSteps.every((name) => name.startsWith(INFRA_PREFIX))) return "needs-ci";
+  if (failedSteps.every(isInfraStep)) return "needs-ci";
   if (failedJobs.length === 1 && failedJobs[0].name === "lint") return "needs-lint";
   return "needs-changes";
 }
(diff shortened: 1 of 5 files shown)


.github/scripts/ci-state.js
```diff
@@ -0,0 +1,159 @@
...
+
+  const failedSteps = failedJobs.flatMap((job) =>
+    (job.steps ?? [])
+      .filter((step) => step.conclusion === "failure")
+      .map((step) => step.name));
+
+  if (failedSteps.length === 0) return "needs-ci";
+  if (failedSteps.every(isInfraStep)) return "needs-ci";
+  if (failedJobs.length === 1 && failedJobs[0].name === "lint") return "needs-lint";
+  return "needs-changes";
```
davidkoski, line 96: Sometimes we see CI failures where the author can't do anything:

- xcode 27 SDK versions
- failure in the CI machinery (needs a rerun)
- environment issue on the runner (typically requires a fix/patch on main and then rebase)

The last case is tricky because it may be broken for a while and it may take us some time to figure out how to fix it.  In the meantime we might be marking things as "needs-change" but there is no change to make.

This is maybe a frustration with GitHub -- the workflow runs from the branch.  I have seen systems that do the opposite and they are impossible to test, so tradeoffs either way.

The only thing I could think of is we might want a kill switch on main or a fixed branch -- if things were misbehaving we could turn off the automated workflow across all the branches at once.  Otherwise every open PR might be doing "bad things" with no easy way to stop it.

Perhaps a serial number -- that way when we fix it, it will pick up when the workflow code that matches gets into the branch.

What do you think?

thechriswebb: Great idea, I liked the idea so just kinda went with it. Let me know what you think. Just pushed.

davidkoski (approved):
Change looks good -- let's try it out!

PR #589, opened 2026-08-28
Declare module weights as compile state in compiled decode paths

### LINKED ISSUE #586: [BUG] examine and understand calls to compile that capture module parameters
See https://github.com/ml-explore/mlx-swift-lm/pull/569/changes#r3874210950

I am not 100% sure this is an issue, but from recent work on `compile()` I suspect it is.  Any MLXArray that are used in the compiled code without being passed as an argument or `state:` will be captured.

This is ok for inference because the values are not mutated, but if the models are used with LoRA, this would be trouble.

I think we need to:

- understand if this is a real issue -- should be easy to construct a test in mlx-swift for this
- if it is real, identify areas in the current mlx-swift-lm that are likely to fall victim
- come up with a fix, probably passing the layers as `state:`

aleroot:
## Proposed changes

Fixes #586, rightfully raised by @davidkoski on #569. Pairs with ml-explore/mlx-swift#465, which confirms the same behaviour upstream.

### Design

**1. Attempt to make the mistake unbuildable.** `MLXLMCommon` shadows MLX's stateless `compile` overloads with versions that take a `@Sendable` body. Neither `MLXArray` nor `Module` is `Sendable`, so the capture that causes the bug is a
compile error:

```
error: capture of 'self' with non-Sendable type 'Qwen35SparseMoeBlock'
       in a '@Sendable' closure
```

This is the part that prevents the next occurrence. A future model port cannot reintroduce the bug by writing the natural-looking thing; the compiler stops it at the keystroke, not in review.

`MLX.compile(inputs:outputs:)` is deliberately not shadowed. Capturing a module *and declaring it* is correct, and it stays the way to compile a training step, exactly as mlx-swift#465 documents.

**2. `CompiledTrace<Owner>` when a body must read weights.** The body takes its owner as a parameter instead of capturing it, and declares which modules it reads. Those become the compile `inputs:`.

```swift
private let compiledForward = CompiledTrace<Qwen35SparseMoeBlock> { block, arguments in
    [block.forward(arguments[0])]
}

func callAsFunction(_ x: MLXArray) -> MLXArray {
    x.dim(1) != 1 ? forward(x) : compiledForward(self, x)
}
```

Compilation stays lazy, so a trace built in `init` still reads weights loaded later. Traces never retain their owner, so the existing deallocation tests hold.
Default state is the owner itself, which is complete for a body that stays in its own subtree, which is every per-layer trace here.

**3. Invalidate when the module tree changes.** Declared state fixes changing *values*. It does not fix a changing *tree*: after a LoRA load the trace still describes a `Linear` where a `LoRALinear` now sits.
`Module.invalidateCompiledTraces()` resets every trace in a tree, and lives in `replaceLayers`, the one helper all four adapter paths (`from`, `load`, `fuse`, `unload`) already share, plus the two quantization paths. Traces are found by reflection, so a model owns one just by storing it. Nothing to conform to, nothing to remember, no per-model boilerplate to forget in the next port.

### Why snapshot weights instead of re-reading them

`compile(inputs:)` re-flattens its inputs on every call, so passing the `Module` would walk the module tree once per segment per token. Measured on an MacBook M2 machine: flattening one decoder layer takes ~35us, a 48 layer model ~1.6ms.
Passing already-flattened arrays costs ~60ns each per call.

So the snapshot is taken once, when the trace compiles. It stays correct because `Module.update(parameters:)`, the path both optimizers and adapter weight loading take, writes *through* the same `MLXArray` objects rather than replacing them.
Replacing modules does create new objects, which is exactly what invalidation covers.

## Questions

- **Deliberately not included:** a lint hook banning explicit `MLX.compile`
  outside `CompiledTrace.swift`. The type system covers the accidental case and
  the bypass is visible in review. Easy to add later if it ever appears in a diff.
- **Open question.** `ModelConversion` and `ParoQuant` invalidate after replacing
  modules, but both run at load time before any trace exists. I kept the calls as
  cheap insurance against future reordering. Happy to drop them if you would
  rather not pay for the noise.

## Checklist

## AI usage

- AI usage disclosure:

### DIFF
--- a/Libraries/MLXLMCommon/CompileOverloads.swift
+++ b/Libraries/MLXLMCommon/CompileOverloads.swift
@@ -1,0 +1,332 @@
+// Copyright © 2026 Apple Inc.
+
+import Foundation
+import MLX
+
+// The rest of the stateless `compile` surface, mirroring MLX's own overload
+// matrix (1 to 8 inputs, 1 to 4 results). See CompiledTrace.swift for what
+// these are for: a `@Sendable` body cannot capture a Module or an MLXArray, so
+// weights cannot be frozen into a trace by accident. A shape missing here would
+// fall through to MLX's unchecked overload, which is why the matrix is complete
+// rather than only the shapes models happen to use today.
+//
+// Hidden from DocC, as upstream hides the equivalent list.
+
+@_documentation(visibility: internal)
+public func compile(
+    shapeless: Bool = false,
+    _ body: @escaping @Sendable (MLXArray) -> (MLXArray, MLXArray)
+) -> @Sendable (MLXArray) -> (MLXArray, MLXArray) {
+    MLX.compile(shapeless: shapeless, body)
+}
+
+@_documentation(visibility: internal)
+public func compile(
+    shapeless: Bool = false,
+    _ body: @escaping @Sendable (MLXArray) -> (MLXArray, MLXArray, MLXArray)
+) -> @Sendable (MLXArray) -> (MLXArray, MLXArray, MLXArray) {
+    MLX.compile(shapeless: shapeless, body)
+}
+
+@_documentation(visibility: internal)
+public func compile(
+    shapeless: Bool = false,
+    _ body: @escaping @Sendable (MLXArray) -> (MLXArray, MLXArray, MLXArray, MLXArray)
+) -> @Sendable (MLXArray) -> (MLXArray, MLXArray, MLXArray, MLXArray) {
+    MLX.compile(shapeless: shapeless, body)
+}
+
+@_documentation(visibility: internal)
+public func compile(
+    shapeless: Bool = false,
+    _ body: @escaping @Sendable (MLXArray, MLXArray) -> (MLXArray, MLXArray)
+) -> @Sendable (MLXArray, MLXArray) -> (MLXArray, MLXArray) {
+    MLX.compile(shapeless: shapeless, body)
+}
+
+@_documentation(visibility: internal)
+public func compile(
+    shapeless: Bool = false,
+    _ body: @escaping @Sendable (MLXArray, MLXArray) -> (MLXArray, MLXArray, MLXArray)
+) -> @Sendable (MLXArray, MLXArray) -> (MLXArray, MLXArray, MLXArray) {
+    MLX.compile(shapeless: shapeless, body)
+}
+
+@_documentation(visibility: internal)
+public func compile(
+    shapeless: Bool = false,
+    _ body: @escaping @Sendable (MLXArray, MLXArray) -> (MLXArray, MLXArray, MLXArray, MLXArray)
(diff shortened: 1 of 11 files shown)


Libraries/MLXLMCommon/CompiledTrace.swift
```diff
@@ -0,0 +1,193 @@
...
+///     }
+/// }
+/// ```
+///
+/// Compilation is lazy, so a module can create its traces in `init` and they
+/// still read weights loaded later. Traces never retain their owner.
+///
+/// A trace is built from the modules present when it first ran, so replacing
+/// modules invalidates it: see `Module.invalidateCompiledTraces()`.
+public final class CompiledTrace<Owner: Module>: CompiledTraceInvalidating {
```
davidkoski, line 88: This looks good!

`CompiledTrace` doesn't introduce a new concurrency domain -- calling it requires exclusive access to the model.  I was thinking about how we guard the `inputs:` against concurrent mutation and evaluation, but since this will be owned by the model it works out, assuming the model is correctly guarded.

In theory ModelContainer is providing that guarding, but in practice it (and ChatSession) play a little fast and loose and _borrow_ the model during inference.  This allows concurrent model evaluation, which is desirable.

If you were doing LoRA + concurrent evaluation you might run into trouble, but that problem was already there.

There may be interplay with this and https://github.com/ml-explore/mlx-swift/pull/418 + https://github.com/ml-explore/mlx-swift-lm/pull/335.  The first adds a MaterializedModule, which provides a container holding a Module that we promise is immutable.  This works well with the `compile(@Sendable)` change -- the weights won't change so it could be captured by the closure.  

Any `CompiledTrace` instances in the interior would also work fine -- they would never invalidate and the `inputs:` would never mutate.  I need to think through this when #335 is closer to being ready, but I think this works out (and certainly isn't a blocker here).

In the case of LoRA, you would keep it as a normal Module and since it is not Sendable, you would need to use the same care as today thread-wise.

Libraries/MLXLMCommon/CompileOverloads.swift
```diff
@@ -0,0 +1,332 @@
...
+// weights cannot be frozen into a trace by accident. A shape missing here would
+// fall through to MLX's unchecked overload, which is why the matrix is complete
+// rather than only the shapes models happen to use today.
+//
+// Hidden from DocC, as upstream hides the equivalent list.
+
+@_documentation(visibility: internal)
+public func compile(
+    shapeless: Bool = false,
+    _ body: @escaping @Sendable (MLXArray) -> (MLXArray, MLXArray)
```
davidkoski, line 18: I think this is a fabulous idea!

I did something similar for compile and vmap in mlx-swift: https://github.com/ml-explore/mlx-swift/pull/469

The `@Sendable` on the closure guards against exactly the case we are trying to protect.

davidkoski (approved):
Really nice fix, thank you!

PR #591, opened 2026-08-28
Report prompt token counts to FoundationModels again

thechriswebb:
## The problem

An app that runs a model through this adapter reads its token usage from the session. Prompt tokens always came back zero. A fix in July removed the usage update the adapter sends after each response, because the beta SDK declared a call the shipping system library did not contain, and every request crashed the moment generation finished. Completion tokens kept working, because the framework counts those from the streamed text on its own. A total still looked plausible, so the missing half was easy to miss.

## What this PR does

- Sends the usage update again. A caller reads prompt tokens as well as completion tokens, per response and added up across a session.
- Confirms the mismatch that caused the crash is gone on the current toolchain, on macOS and on iOS.

It adds no continuous integration coverage. The nightly job runs an older toolchain, where this code compiles to nothing, and the newer job only compiles.

## Tests

- A new session reports zero.
- One response reports both prompt and completion tokens.
- A session adds usage across three turns.
- Two sessions stay independent.
- The streaming path agrees with its own final snapshot.

Three of the five fail without the change and pass with it, against a real model on macOS 27 with Xcode 27 beta 6. The other two check the zero state, which the bug did not affect.

### DIFF
--- a/IntegrationTesting/IntegrationTestingTests/MLXFoundationModelsIntegration/TextGeneration/SessionUsageAccountingTests.swift
+++ b/IntegrationTesting/IntegrationTestingTests/MLXFoundationModelsIntegration/TextGeneration/SessionUsageAccountingTests.swift
@@ -1,0 +1,120 @@
+// Copyright © 2026 Apple Inc.
+
+#if FoundationModelsIntegration && canImport(FoundationModels, _version: 2)
+
+import Testing
+import Foundation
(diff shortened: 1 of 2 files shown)


thechriswebb:
@davidkoski This is something we'd definitely want to get in before we cut a tag.

davidkoski (approved):
Change looks good.

PR #592, opened 2026-08-29
Fix DeepSeek MTP layer filtering

aleroot:
## Proposed changes

DeepSeek V3 checkpoint filtering assumed that multi-token prediction (MTP) weights always started at layer 61. That is correct for the standard model, but not for variants with a different number of hidden layers. For example, GLM-5.2 has 78 hidden layers and one appended MTP layer: layer 61 is a valid model layer, while layer 78 is the MTP layer. Deriving the boundary from `num_hidden_layers` prevents valid weights from being discarded and keeps checkpoint filtering consistent with the model configuration.

### Note

This PR does not add GLM-5.2 support itself; it removes the hardcoded assumption from the shared DeepSeek V3 stack and prepares it for architectures with different configured depths.

## Checklist

- [x] I have updated the necessary documentation (not needed for this internal fix)

## AI usage

- AI usage disclosure: none

### DIFF
--- a/Libraries/MLXLLM/Models/DeepseekV3.swift
+++ b/Libraries/MLXLLM/Models/DeepseekV3.swift
@@ -426,6 +426,16 @@ public class DeepseekV3Model: Module, LLMModel, KVCacheDimensionProvider, LoRAMo
         return lmHead(out)
     }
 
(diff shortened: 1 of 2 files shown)


davidkoski (approved):
Change looks good, thank you!

PR #594, opened 2026-08-29
Clear the compiler warnings from the library and both test targets

thechriswebb:
## The problem

The compiler printed many warnings. Most came from style diagnostics, and from calls to MLX functions that MLX has renamed. It printed them on every build and every test run, so a warning that mattered was hard to find.

Two warnings came from real defects. Gemma read each video frame through a deprecated property. That property traps when the frame's image is not a CoreImage image. The process then stops. A quantization test passes a model layer into a concurrent closure. Swift cannot prove that this is safe. The test does it on purpose, but the code never said so.

## What this PR does

- Gemma reads each video frame through the supported accessor. If a frame's image is not a CoreImage image, the caller now gets an error instead of a crash. Gemma was the last model still calling the deprecated frame API. Every other vision model already used the current one.
- The integration tests call MLX's Memory type for memory figures and for cache clearing, not the older GPU names. Behavior does not change, because the deprecated calls forward to the new ones.
- The quantization concurrency test now shares its layer through a small wrapper marked unchecked-Sendable, so the compiler no longer warns about the sharing. The test still uses that layer from several threads without synchronization, which is the behavior it checks.
- Removes the style diagnostics from the library and both test targets: an unused local, bindings that never change, and try and await markers on calls that neither throw nor suspend. One conditional cast tested a conformance that the compiler checks statically. If someone removes that conformance, the build still fails.

The compiler still warns about MLX's own Metal kernels. Those kernels come from a vendored dependency, so this repository cannot fix them.

## Tests

No file in this repository now produces a warning.

The unit suite passes: 549 XCTest cases, one skipped, and 938 swift-testing cases in four targets. The two integration suites whose files changed ran against real models. All 15 of their tests passed.

### DIFF
--- a/IntegrationTesting/IntegrationTestingTests/CoherenceIntegrationTests.swift
+++ b/IntegrationTesting/IntegrationTestingTests/CoherenceIntegrationTests.swift
@@ -34,7 +34,7 @@ private func runCoherence(_ configuration: ModelConfiguration) async throws {
 private func releaseModel(_ configuration: ModelConfiguration) async {
     await models.evictLLM(configuration)
     Stream.gpu.synchronize()
-    GPU.clearCache()
+    Memory.clearCache()
 }
 
 /// Fast, reliable, architecturally-diverse subset that runs by default so the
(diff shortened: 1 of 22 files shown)


davidkoski (approved):
Changes look good, thanks!

PR #595, opened 2026-08-30
Expose prepared categorical distributions for samplers

VaderChen:
## Summary

- Add public `PreparedDistribution` with `.none` and `.logWeights(MLXArray)` cases.
- Add defaulted `prepare(logits:)` and `sample(prepared:logits:)` operations to `LogitSampler`.
- Make `TopPSampler` and `CategoricalSampler` expose the exact unnormalized log weights passed to `categorical`.
- Keep `ArgMaxSampler` on the default one-shot path because greedy decoding is not a categorical distribution.
- Keep `sample(logits:)` unchanged as a compatibility convenience.
- Add tests for temperature scaling, filtering, prepared sampling equivalence, and uniform `LogitSampler` use.

## Update since review

Following reviewer feedback, this revision replaces the separate capability
protocol with a uniform two-phase API on `LogitSampler`:

- `PreparedDistribution` represents either reusable log weights or `.none`.
- `prepare(logits:)` and `sample(prepared:logits:)` have defaults, so callers
  do not need capability casts or optional prepared values.
- `TopPSampler` and `CategoricalSampler` consume the exact prepared log weights,
  while `ArgMaxSampler` keeps the default one-shot behavior.
- Existing `sample(logits:)` call sites remain source-compatible.

## Benefits

The prepared value binds sampling to the exact distribution used for the draw.
A speculative decoder can reuse that same distribution for acceptance ratios
and residual sampling without re-evaluating temperature or token filters.

The common protocol interface avoids capability casts and optional prepared
values in callers. Top-p and top-k filtering are performed once, avoiding
repeated sorting and partitioning work. This provides a focused foundation for
lossless stochastic speculative decoding and DFlash without adding
Tanpopo-specific logic or changing the existing generation path.

## Validation

- `pre-commit run --all-files` passed with swift-format 603.0.0.
- `MLXLMCommon` DocC generation passed with warnings treated as errors.
- `SampleTests`: 37 tests, 0 failures.
- The full package unit-test command completed with `** TEST SUCCEEDED **`.
- `scripts/verify-docs.sh` built the changed `MLXLMCommon` target successfully.
  The full script still reports the repository's existing MLXCXGrammar
  ExtractAPI failure under the local Xcode 27 SDK (`<algorithm>` not found).

## Compatibility

Existing `LogitSampler` conformances and `sample(logits:)` call sites remain
source-compatible because the new protocol operations have defaults. Existing
categorical samplers preserve their random-state and filtering behavior.

## Checklist

- [x] I have run `pre-commit run --all-files` to format my code / installed pre-commit

## AI usage

- AI usage disclosure: OpenAI Codex assisted with comparing the Tanpopo prototype against the current `mlx-swift-lm` checkout, designing and implementing the sampler protocol refactor, updating tests and documentation, reviewing API compatibility, and preparing this pull request description. I reviewed and understand every submitted line.

### DIFF
--- a/Libraries/MLXLMCommon/Evaluate.swift
+++ b/Libraries/MLXLMCommon/Evaluate.swift
@@ -14,6 +14,25 @@ public protocol LogitSampler {
     func sample(logits: MLXArray) -> MLXArray
 }
 
+/// A sampler that can expose the exact categorical distribution it samples.
+///
+/// The returned values are unnormalized log weights. Applying `softmax` on
+/// the final axis produces the sampler's token probabilities. Keeping the
+/// representation unnormalized avoids an unnecessary `softmax` followed by
+/// `log` before `categorical`, while allowing speculative decoders to compute
+/// exact acceptance ratios and residual distributions.
+///
+/// `ArgMaxSampler` intentionally does not conform because greedy decoding is
+/// not a categorical sampling distribution.
+public protocol LogitDistributionSampler: LogitSampler {
+
+    /// Returns the unnormalized log weights used by ``sample(logits:)``.
+    ///
+    /// Applying ``softmax(_:axis:)`` on the final axis produces the categorical
+    /// probabilities represented by these weights.
+    func logWeights(logits: MLXArray) -> MLXArray
+}
+
 /// A `LogitProcessor` is an optional visitor of `logits`.
 ///
 /// The ``LogitProcessor`` is called with the input (prompt) before generating tokens:
@@ -338,7 +357,7 @@ public struct ArgMaxSampler: LogitSampler {
 /// Each filter operates on the full vocabulary in original token order, masking
 /// rejected tokens with `-inf`. This matches the composable filter chain in
 /// `mlx_lm.sample_utils.make_sampler`.
-public struct TopPSampler: LogitSampler {
+public struct TopPSampler: LogitDistributionSampler {
     let temp: MLXArray
     let topP: MLXArray?
     let topK: Int?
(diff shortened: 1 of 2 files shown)


Libraries/MLXLMCommon/Evaluate.swift
```diff
@@ -364,27 +383,31 @@ public struct TopPSampler: LogitSampler {
         self.randomState = seed.map { MLXRandom.RandomState(seed: $0) } ?? MLXRandom.RandomState()
     }
 
-    public func sample(logits: MLXArray) -> MLXArray {
+    public func logWeights(logits: MLXArray) -> MLXArray {
```
aleroot: Why the name `logWeights` ? 
would `samplingLogits(logits:)` be more consistent? 

More importantly, since `sample` and `logWeights` are separate requirements, the protocol does not guarantee that the returned token was drawn from the exposed weights. It may also require evaluating stateful or expensive filters twice. Should the capability instead return the token and exact prepared distribution together, or expose a prepared distribution that can be sampled directly?

Would it make sense to model this as a two-phase operation: prepare the categorical distribution once, then sample directly from that prepared value? That would bind the token to the exact distribution, avoid evaluating filters twice, and later allow the residual distribution to be sampled without reapplying temperature or top-p/top-k.

VaderChen: Thanks, this points out a real limitation in the current protocol. I agree that exposing a separate logWeights requirement does not guarantee that sample uses the same prepared distribution and may repeat stateful filters. I’m going to revise this into an explicit prepare-then-sample API, while keeping sample(logits:) as a convenience wrapper.

Libraries/MLXLMCommon/Evaluate.swift
```diff
@@ -14,6 +14,44 @@ public protocol LogitSampler {
...
+/// A sampler that can prepare and sample an exact categorical distribution.
+///
+/// Preparation applies the sampler's temperature and token filters once. The
+/// returned distribution can then be sampled directly, allowing speculative
+/// decoders to reuse the exact distribution for acceptance ratios and residual
+/// distributions without reapplying those transformations.
+///
+/// `ArgMaxSampler` intentionally does not conform because greedy decoding is
+/// not a categorical sampling distribution.
+public struct PreparedCategoricalDistribution {
```
davidkoski, line 26: This is just a question -- I don't know if it is a good idea or not.

Thinking about how this would be consumed, I wonder if the split between LogitDistributionSampler and LogitSampler will cause trouble?

For example, in speculative decoding you might:

```swift
if let sampler as? LogitDistributionSampler {
    // handle prepare for both the normal and speculative paths
} else {
    // the same, but no prepare
}
```

OK, sure, you could store the value in an optional but you need the same logic when you try to use it later on.

What if we made (using some defaults) all samplers be able to prepare and sample?  Something like this:

```swift
   public enum PreparedDistribution {                                                               
       case none                                                                                    
       case logWeights(MLXArray)                                                                    
   }                                                                                                
                                                                                                    
   public protocol LogitSampler {                                                                   
       func sample(logits: MLXArray) -> MLXArray                                                    
                                                                                                    
       func prepare(logits: MLXArray) -> PreparedDistribution
       func sample(prepared: PreparedDistribution, logits: MLXArray) -> MLXArray                    
   }                                                                                                
```

Callers who want the current one shot implementation call the first method.  Callers that can make use of the split call the latter two.  There is no special case.  It always works either way.

davidkoski (changes requested):
This looks good.  See my question about how callers might use it -- is that a good idea?  or not?

Libraries/MLXLMCommon/Evaluate.swift
VaderChen: This is a good suggestion. I had not considered this while working on the change.

The current capability cast through `LogitDistributionSampler` would require
callers to maintain two separate control-flow paths and an additional optional
prepared value.

I agree that the two-phase operations should be available directly on
`LogitSampler`, with a default “no preparation” behavior for samplers that do
not expose a categorical distribution.

I will revise the API so that categorical samplers return prepared log weights,
while `ArgMaxSampler` uses the default one-shot sampling behavior. This keeps
existing `sample(logits:)` call sites unchanged and lets speculative decoding
use one uniform prepare-then-sample flow without requiring a capability cast.

PR #596, opened 2026-08-30
Expose Falcon-H1 encoder surface at @_spi(FalconH1Encoder)

xocialize:
Mirrors the Gemma 3/4 encoder exposures (#387 and the Gemma 4 follow-up) for Falcon-H1, so client code can run the decoder stack over an embedding it built itself rather than over a single token id.

### Motivation

The motivating case is a DualAR TTS model — [Audio8 TTS Preview 0.1b](https://huggingface.co/Audio8/Audio8-TTS-Preview-0.1b), `model_type: arktts`, `slow_backbone: falcon_h1` — whose slow-stack input is a text embedding summed with ten codec-codebook embeddings. Such a client never has a single token id to look up, so `FalconH1ModelInner.callAsFunction(_:mask:cache:)` is unreachable for it. Today the only options are to fork the file or re-implement the mask setup and layer loop against internal details.

### What is exposed

At `@_spi(FalconH1Encoder)` scope, keeping this off the advertised public API of MLXLLM:

- `FalconH1Configuration.hiddenSize` / `.numHiddenLayers` / `.rmsNormEps` / `.embeddingMultiplier`
- `FalconH1DecoderLayer` and its `callAsFunction`
- `FalconH1ModelInner.embedTokens` / `.layers` / `.finalLayerNorm`
- `FalconH1ModelInner.callAsFunction(inputsEmbeds:cache:)` — **new**
- `FalconH1ModelInner.callAsFunction(_:mask:cache:)` — was internal

`callAsFunction(inputsEmbeds:cache:)` is the seam, and the token-id overload now delegates to it so the two paths cannot drift. **No behaviour change for existing callers**: `FalconH1Model` still embeds, runs the stack, and applies the head exactly as before.

Unlike the Gemma 4 change, nothing else had to widen — `FalconH1Model.model` and the mask helpers were already public, and none of the decoder layer's stored property types are `private`.

### Tests

`FalconH1EncoderAccessTests` proves sufficiency the way the Gemma access tests do: it drives the stack using **only** the exposed surface, importing MLXLLM without `@testable` and opting in via `@_spi(FalconH1Encoder)`. It checks shape, determinism, seam-vs-token-path agreement, layer-at-a-time driving for an all-hidden-states tap, and that the stack actually transforms its input — the last so the suite cannot pass on a seam that returns its argument or runs zero layers.

One test earns its place beyond access checking. `sanitize()` folds `embedding_multiplier` into `embed_tokens.weight`, which is correct when the token lookup is the only thing entering the stack, but is a live trap for a composite-embedding client: fold it and then add unscaled contributions, and the lookup half is scaled while the rest is not. Measured against the PyTorch reference for the model above, that is **77% relative error on the embedding and 38% on the final hidden state** — while still producing plausible audio of the right length, so listening does not catch it. `testCompositeEmbeddingMultiplierEquivalence` pins `(raw + other) * m == folded + other * m` and asserts the unscaled-remainder variant is observably different, both at the embedding and after the stack, so the wrong form cannot quietly become the expected one. The doc comment on the new seam spells the same thing out.

Runs on a tiny randomly-initialized model; no weights are downloaded.

`swift test --filter FalconH1`: 14 tests, 0 failures (5 new, 8 existing, 1 registry).

### DIFF
--- a/Libraries/MLXLLM/Models/FalconH1.swift
+++ b/Libraries/MLXLLM/Models/FalconH1.swift
@@ -20,11 +20,11 @@ public struct FalconH1Configuration: Codable, Sendable {
     var attentionInMultiplier: Float
     var attentionOutMultiplier: Float
     var bosTokenId: Int
-    var embeddingMultiplier: Float
+    @_spi(FalconH1Encoder) public var embeddingMultiplier: Float
     var eosTokenId: Int
     var headDim: Int
     var hiddenAct: String
-    var hiddenSize: Int
+    @_spi(FalconH1Encoder) public var hiddenSize: Int
     var initializerRange: Float
     var intermediateSize: Int?
     var keyMultiplier: Float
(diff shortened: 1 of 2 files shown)


davidkoski (approved):
Change looks good, thanks!

PR #597, opened 2026-08-31
Add configuration-based LoRA metadata discovery

aleroot:
## Proposed changes

Adds a public API for discovering a model’s LoRA layer count and default target module names from its configuration.

Reading these names from safetensors is unreliable because model loading may rename or reshape checkpoint weights. The new API instead uses the registered runtime model architecture, which is the source of truth for LoRA targets. It does not load checkpoint weights.

This is useful for applications that need to show valid LoRA options before loading a model, while remaining compatible with custom model registrations.

## Checklist

## AI usage

- AI usage disclosure:

### DIFF
--- a/Libraries/MLXLLM/LLMModelFactory.swift
+++ b/Libraries/MLXLLM/LLMModelFactory.swift
@@ -658,6 +658,18 @@ public final class LLMModelFactory: GenericModelFactory {
     /// by the model itself, e.g. DeepSeek-R1
     public let conventionsRegistry: ChatConventionsRegistry
 
+    /// Returns the model's LoRA metadata from its configuration without loading weights.
+    ///
+    /// The registered model architecture is instantiated so custom `LoRAModel.loraDefaultKeys`
+    /// implementations remain authoritative. No checkpoint files are read.
+    public func loraMetadata(configurationData: Data) async throws -> LoRAModelMetadata? {
+        let baseConfiguration = try JSONDecoder.json5().decode(
+            BaseConfiguration.self, from: configurationData)
+        let model = try await typeRegistry.createModel(
+            configuration: configurationData, modelType: baseConfiguration.modelType)
+        return (model as? LoRAModel)?.loraMetadata
+    }
+
     public func _load(
         configuration: ResolvedModelConfiguration,
         tokenizerLoader: any TokenizerLoader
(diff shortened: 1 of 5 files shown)


Libraries/MLXLLM/LLMModelFactory.swift
```diff
@@ -658,6 +658,18 @@ public final class LLMModelFactory: GenericModelFactory {
     /// by the model itself, e.g. DeepSeek-R1
     public let conventionsRegistry: ChatConventionsRegistry
 
+    /// Returns the model's LoRA metadata from its configuration without loading weights.
+    ///
+    /// The registered model architecture is instantiated so custom `LoRAModel.loraDefaultKeys`
+    /// implementations remain authoritative. No checkpoint files are read.
+    public func loraMetadata(configurationData: Data) async throws -> LoRAModelMetadata? {
```
davidkoski, line 665: Would it make sense to move this into MLXLMCommon?  Maybe here:

```swift
extension GenericModelFactory where ContextType == ModelContext, ContainerType == ModelContainer {
```

as it is VLM's can't use this (I am not sure people use VLMs with LoRA, but as far as I know they could).

Libraries/MLXLLM/LLMModelFactory.swift
```diff
@@ -658,6 +658,18 @@ public final class LLMModelFactory: GenericModelFactory {
...
+        let baseConfiguration = try JSONDecoder.json5().decode(
+            BaseConfiguration.self, from: configurationData)
+        let model = try await typeRegistry.createModel(
+            configuration: configurationData, modelType: baseConfiguration.modelType)
```
davidkoski: I was thinking about Random state here -- creating the mode will mutate the Random state.  That isn't a problem directly, but given the recent attention on thread safety I wonder if this (or maybe `createModel()`) should use a task local `withRandomState()`.

Probably not an issue for this PR, but let me know what you think.

davidkoski (changes requested):
Looks good, ready to merge.  See my question on the placement of the function -- should it be in MLXLMCommon?

aleroot:
> Looks good, ready to merge. See my question on the placement of the function -- should it be in MLXLMCommon?

Moved the shared implementation into MLXLMCommon, on ModelTypeRegistry<LanguageModel>, and exposed it through both factories.
I also isolated construction and metadata access with a fresh withRandomState scope.
Thanks .

davidkoski (approved):
Thank you!  Just waiting on CI

PR #599, opened 2026-08-31
Harden the Foundation Models test fixtures

thechriswebb:
## The problem

An audit of `MLXFoundationModels` against Swift's published style guidance surfaced eight test fixture files that broke concrete rules, not just style preferences. Two channel drains swallowed every error in an empty `catch` block. Six load and warmup tasks discarded their result with `try?`. Five stub types asserted thread safety with `@unchecked Sendable` instead of the compiler proving it. One stub enforced an OS availability requirement with a runtime guard the compiler could enforce instead. Each of these could hide a real failure or a real race and let a broken fixture report a passing test.

## What this PR does

- Extracts one channel-drain helper in `GenerationEventObserverTests` and asserts its result. A drain that fails with anything but cancellation now fails the test, instead of vanishing into an empty `catch {}`.
- Asserts the specific error every load and warmup task throws in `AvailabilityTests` and `ModelCacheEvictionTests`, instead of discarding it with `try?`.
- Converts five stateless downloader and tokenizer-loader stubs from classes marked `@unchecked Sendable` to structs, so the compiler proves `Sendable` instead of the annotation asserting it.
- Replaces `NSLock` and `@unchecked Sendable` on `EventBox` and `SeedRecorder`, the two fixtures that hold mutable state across a task boundary, with the standard library's `Mutex`.
- Spells out a two-closure call in `GenerationEventObserverTests` that swift-format's `OnlyOneTrailingClosureArgument` rule flagged, so it is unambiguous which parameter each closure binds to.
- Marks `EvictBiasStubTokenizerLoader` with the OS 27 availability its returned tokenizer already requires, replacing a runtime `guard` and a `preconditionFailure` that could never fire.

## Tests

- `pre-commit run --all` (swift-format): passes with zero files changed.
- `xcodebuild build-for-testing` for `mlx-swift-lm-Package`: builds clean.
- `xcrun xctest MLXFoundationModelsTests.xctest`: 139 tests in 21 suites pass, run 20 times in a row with no flakes.
- `xcrun xctest MLXLMTests.xctest`: 722 tests in 55 suites pass. This is CI's only required suite, but it covers none of this PR's diff; `MLXFoundationModelsTests.xctest` is the suite that exercises it.
- Mutation check: inverted one `#expect` in each of the six affected test files, confirmed all six produce the expected failure, then reverted and confirmed the suite returns to 139 tests in 21 suites passing.

## Checklist

## AI usage

- AI usage disclosure: I used AI for all of this, but understand the code that I'm contributing.

### DIFF
--- a/Tests/MLXFoundationModelsTests/AvailabilityTests.swift
+++ b/Tests/MLXFoundationModelsTests/AvailabilityTests.swift
@@ -87,12 +87,14 @@ extension FoundationModelsCacheTests {
                         configuration: configuration, progressHandler: progress)
                 })
 
-            let warmTask = Task { try? await model.warmUp() }
+            let warmTask = Task { try await model.warmUp() }
             await gate.waitUntilStarted()
 
             let availability = await model.availability
             await gate.release()
-            _ = await warmTask.value
+            await #expect(throws: BlockingDownloaderReleased.self) {
+                try await warmTask.value
+            }
 
             #expect(
                 availability == .available,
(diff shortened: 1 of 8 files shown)


davidkoski (approved):
Changes look good to me, thanks!

PR #602, opened 2026-09-03
Gemma4Text: expose decoder layers as loraLayers so MLP LoRA targets load

cives93:
loraLayers returned only selfAttn, so adapters targeting mlp.* failed LoRAContainer.load with unhandled keys. Gemma3Text/Qwen35/Llama (and the Gemma4 VLM wrapper) return the decoder layers; match them. Adds a regression test. Fork SHA db6928356e86281dbb60a4ed080909959e61637f.


davidkoski (approved):
change looks good, thank you!

PR #603, opened 2026-09-03
Extract the model cache and stop whole-cache eviction abandoning an in-flight load

thechriswebb:
## The problem

Reviewing the FoundationModels adapter against the [Swift API Design Guidelines](https://www.swift.org/documentation/api-design-guidelines/) and the [Swift Foundation contribution guidelines](https://github.com/swiftlang/swift-foundation/blob/main/CONTRIBUTION_GUIDELINE.md) surfaced a defect behind several naming and layout gaps.

`MLXLanguageModel.evictAll()` cleared its load registry without cancelling, discarding the only handle to a load that was still running. Nothing could reach that load afterwards. The single-model `evict()` already cancelled, so two calls that differ only in scope behaved differently in kind, and the public documentation described the opposite of the shipped behavior by claiming the in-flight load completes.

## What this PR does

- `evictAll()` cancels every registered load, matching `evict()`. No load is left running with its handle discarded.
- Corrects the eviction documentation, which claimed an in-flight load completes.
- A failed constraint clone swallows only the fork-unavailable error. Any other clone failure now reaches the caller.
- Moves the cache into `ModelCache.swift`, so the file name matches its primary type.
- Renames three cache members off the `XG` prefix. `XG` names the C shim in `MLXCXGrammar`, and the Swift wrappers are already `GrammarTokenizer` and `GrammarConstraint`. These three were left behind by that earlier rename.

Cancellation is cooperative, so the effect depends on the injected loader. It reaches a load that is still downloading. A load already inside `loadWeights` finishes and allocates, because that path suspends on a continuation that does not observe cancellation. That limit applies to `evict()` too and is not introduced here. `evictAll()` and `remove(modelID:)` still enumerate the same caches separately; consolidating the per-model state is the durable fix and is not in this PR.

## Tests

- `MLXFoundationModelsTests`: 139 tests in 21 suites at the move and at the catch narrowing, matching main. 141 in 21 after the cancellation change.
- `evictAllCancelsRegisteredLoads` covers the new behavior. It substitutes the whole container loader, so it proves the registered load is cancelled, not that a real weight load is interruptible.
- `evictCancelsRegisteredLoad` already passes on main and documents the existing per-model path.
- `MLXLMTests`: 722 tests in 55 suites pass, matching main. That bundle holds no MLXFoundationModels test.
- `pre-commit run --all`: swift-format passes with zero files changed.
- `xcodebuild build-for-testing` succeeds at each commit, each with its own derived data path.

## Checklist

- [X] I have read the [CONTRIBUTING](https://github.com/ml-explore/mlx-swift-lm/blob/main/CONTRIBUTING.md) document

## AI usage

- [X] I have read this PR description in full and approve it as my own, and it
- AI usage disclosure: I used AI for all of this, but fully understand the changes and can speak to them.


Libraries/MLXFoundationModels/ModelCache.swift
```diff
@@ -0,0 +1,298 @@
...
+
+    /// Evicts all cached state: model containers, tokenizers, constraint
+    /// templates, and per-model tokenizer biases. Cancels every registered load
+    /// task, as `remove(modelID:)` does for a single model. No GPU-stream
+    /// synchronization is required — in-flight callers retain their own
+    /// `ModelContainer` and free it via ARC on completion.
+    func evictAll() {
+        containers.removeAll()
+        for loadTask in loadingTasks.values {
+            loadTask.task.cancel()
```
davidkoski, line 264: Not sure if this is an issue, but looking at MLXLangageModel, I am not sure a cancel will correctly signal through MLXDownloadProgress that it is done -- I think `isActive` may stay true:

```swift
            let container = try await load(configuration) { progress in
                MLXDownloadProgress.report(progress: progress, modelID: configuration.name)
            }
            MLXDownloadProgress.reportCompleted()
```

I think the cancel will signal an error on the load and it won't hit the last line.

Actually I wonder about MLXDownloadProgress.  ModelCache is clearly set up to handle multiple simultaneous downloads, but because MLXDownloadProgress is a singleton, the first completion will mark it as `isActive = false`.

I think it could remain a singleton, but it would need to track state keyed on the modelId (also ConstraintKind?) and provide aggregating accessors for the singleton.

davidkoski (changes requested):
Check comment on interaction with MLXDownloadProgress and see what you think.

Libraries/MLXFoundationModels/ModelCache.swift
thechriswebb: Good catch. I addressed both.

`reportCompleted()` ran after `load(configuration)` returned. load(configuration) throws when a download fails and when a caller cancels the load, so `reportCompleted()` never ran on either path. I renamed `reportCompleted()` to `reportEnded()`, and the loader closure now calls it from a defer, so `reportEnded()` runs whether `load(configuration)` returns or throws. Tests cover a successful load, a failed download, and a cancellation.

I now key the progress entries on the model ID, so when one load ends, every other entry stays. `completedBytes` and `totalBytes` add every entry.

I found two more defects.

First, the old code started a new main-actor task for each report. Those tasks run in  no fixed order, so a progress task could run after the `reportEnded()` task. `isActive` then stayed true after `reportEnded()` ran. Every report now goes through one `AsyncStream`, and its consumer applies the reports in the order the loader closure sent them.

Second, `evictAll()` cancels a load, but `load(configuration)` still transfers bytes until its next cancellation check. A second load for the same model can start before that check. Both loads share one entry. When the cancelled load ended, the old code deleted the second load's entry. The entry also held byte samples from both loads, so the reported rate could go negative. Each load now has its own identifier, and `MLXDownloadProgress` applies a report only when the report's identifier matches the entry's. `ModelCache.load`  rejects a superseded write by comparing its load task.

`ConstraintKind` belongs to the constraint cache key. Nothing in the constraint cache downloads bytes, so I left `ConstraintKind` out of the progress key. Please let me know if you had other thoughts in mind.

`isActive` and `modelName` are gone. `activeDownloads` returns a list, and `download(forModelID:)` returns one model's entry.

davidkoski (approved):
Change looks good, ready to merge!

PR #604, opened 2026-09-03
Rename async version of loadWeights to loadWeightsAsync

### LINKED ISSUE #601: Nightly integration tests are failing
The nightly `Integration Tests` run failed.

- Run: https://github.com/ml-explore/mlx-swift-lm/actions/runs/33614518067 (attempt 1)
- Commit: [`5694a2f`](https://github.com/ml-explore/mlx-swift-lm/commit/5694a2f6705f7c8b9cf195f29ec6d05938d42d22) on `refs/heads/main`
- First seen: 2026-09-02

The failing run uploads an `integration-test-results` artifact:
the `IntegrationTesting.xcresult` bundle plus anything in
`~/Library/Logs/DiagnosticReports` (crash logs). Artifacts are
kept for 7 days, so download it from the run page before it
expires, then open the `.xcresult` in Xcode.

Consecutive failures comment on this issue instead of opening a
new one. Close it once the nightly is green again; the next
failure after that opens a fresh issue.

davidkoski:
## Proposed changes

- swift's selection rules with default parameters are inscrutable
- even with @_disfavoredOverload it still tries to call the async variant sometimes (with no await)
- distinguish by name to get reliable resolution
- fixes #601

## Checklist

## AI usage

- AI usage disclosure: none

### DIFF
--- a/Libraries/MLXLMCommon/Load.swift
+++ b/Libraries/MLXLMCommon/Load.swift
@@ -415,7 +415,7 @@ public func loadWeights(
 /// Swift concurrency's cooperative threads must never block, so this overload runs the load
 /// on a global queue and suspends the caller instead. Async callers resolve to this overload;
 /// the synchronous one remains for synchronous code such as model conversion.
-public func loadWeights(
+public func loadWeightsAsync(
     modelDirectory: URL, model: BaseLanguageModel,
     quantization: BaseConfiguration.Quantization? = nil,
     perLayerQuantization: BaseConfiguration.PerLayerQuantization? = nil,
--- a/IntegrationTesting/IntegrationTestingTests/Gemma4AssistantDraftModelIntegrationTests.swift
+++ b/IntegrationTesting/IntegrationTestingTests/Gemma4AssistantDraftModelIntegrationTests.swift
@@ -93,7 +93,7 @@ struct Gemma4AssistantIntegrationTests {
         // loadWeights enumerates *.safetensors in the directory, sanitizes via
         // the model's sanitize hook, and applies via update(parameters:verify:).
         // If any keys are missing or unexpected, update(verify: [.all]) throws.
-        try loadWeights(modelDirectory: drafterDir, model: model)
+        try await loadWeightsAsync(modelDirectory: drafterDir, model: model)
     }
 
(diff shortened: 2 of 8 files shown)


Libraries/MLXLMCommon/Load.swift
```diff
@@ -415,7 +415,7 @@ public func loadWeights(
 /// Swift concurrency's cooperative threads must never block, so this overload runs the load
 /// on a global queue and suspends the caller instead. Async callers resolve to this overload;
 /// the synchronous one remains for synchronous code such as model conversion.
-public func loadWeights(
+public func loadWeightsAsync(
```
davidkoski, line 418: @aleroot FYI -- integration tests were failing to build because of a missing `await`.  No matter what I tried the swift compiler would usually resolve the asynchronous variant (including `@_disfavoredOverload`).  There was no way to predictably call the sync variant, so I think just renaming it is probably the easiest.

davidkoski:
```swift
func t1(url: URL, model: BaseLanguageModel) async throws {
    // you _have_ to use the asynchronous version here
    try loadWeights(modelDirectory: url, model: model)
}

func t2(url: URL, model: BaseLanguageModel) throws {
    try loadWeights(modelDirectory: url, model: model)
}
```

Actually let's take #605.  This overload situation is fine -- if you are in a sync context it resolves the sync version and if in an asynchronous context it requires the asynchronous version (which is what we want).

aleroot: Uhm, this is really interesting... I think it is related to [this](https://github.com/swiftlang/swift-evolution/blob/main/proposals/0296-async-await.md#overloading-and-overload-resolution ), an async context prefers the async overload, even when await is omitted. 
Renaming is a reasonable API choice after all, but I think for the solving of the issue adding await to the five affected test calls should reliably address the original errors without renaming or using the attribute.

davidkoski: Yes, exactly the conclusion I came to -- we would prefer the async variant be called in this case and without the rename the compiler will enforce that (surprise bonus!)

PR #610, opened 2026-09-06
Keep the text before a `<` that turns out not to be a tool-call tag

### LINKED ISSUE #609: [BUG] ToolCallProcessor.processChunk drops the text before a "<" that is not a tool-call tag
      drafted.

**Describe the bug**

When a model writes a `<` in ordinary text, `ToolCallProcessor.processChunk` has to wait for the next chunk to know whether a `<tool_call>` tag is starting. While it waits it holds back the text in front of the `<`, and once the next chunk shows it was just text, that held-back part is never returned. So `` a `<memory>` block `` comes out as `` a<memory>` block `` and `i < n` as `i< n`.

With token-by-token streaming this happens a lot: Qwen tokenizers have single tokens like `` `< ``, ` <`, `(<` and `"<`, and every one of them loses the characters before the `<`. The same thing happens when a real `<tool_call>` starts in one chunk and ends in a later one: the text before the tag is lost too.

`processChunkOutputs`, which the library's own `generate` uses, is fine. Apps that call `processChunk` directly, as the doc comment at the top of the file shows, hit this.

**To Reproduce**

```swift
import MLXLMCommon

func collect(_ chunks: [String]) -> String {
    let processor = ToolCallProcessor(format: .json)
    var text = ""
    for chunk in chunks {
        if let piece = processor.processChunk(chunk) { text += piece }
    }
    if let rest = processor.processEOS(returnBufferedText: true) { text += rest }
    return text
}

collect([" injected as a", " `<", "memory", ">`", " block"])
// " injected as a<memory>` block"

collect(["if a", " <", " b", " {"])
// "if a< b {"

collect(["Note: <tool_call>{\"name\":\"get_weather\",", "\"arguments\":{\"location\":\"Osaka\"}}</tool_call>"])
// "" (the tool call is parsed, "Note: " is gone)
```

**Expected behavior**

The text before the `<` is kept, as `processChunkOutputs` already does for the same chunks: `` injected as a `<memory>` block ``, `if a < b {` and `Note: `.

 - OS Version: macOS 26.5.2
 - Device: MacBook Pro, Apple M3 Max, 48 GB
 - Version: `main` as of 2026-09-06

**Additional context**

Seen in our app with Qwen3.8-27B. I noticed because we key a prompt cache on the client's echo of the answer, and the echoed text stopped matching what the model had produced. A fix with tests is in PR <PR link>.

spokvulcan:
## Proposed changes

Fixes #609.

`processChunk` held back the text in front of a `<` while it waited to see whether a `<tool_call>` tag was starting, and never returned it once the next chunk settled the question. Now each call returns that text itself, the way it already did when the tag resolved inside the same chunk. Four tests in `ToolTests` feed the chunks from the issue through both `processChunk` and `processChunkOutputs` and check that they return the same text.

## Checklist

## AI usage

- AI usage disclosure: the diagnosis, the fix, the tests and this description were produced with Claude Code under my direction; I reviewed the diff and this text before opening the PR.

### DIFF
--- a/Libraries/MLXLMCommon/Tool/ToolCallProcessor.swift
+++ b/Libraries/MLXLMCommon/Tool/ToolCallProcessor.swift
@@ -595,7 +595,9 @@ public class ToolCallProcessor {
                 } else {
                     recordResponse(leadingToken ?? "")
                     leadingTokenWasRecorded = true
-                    return nil
+                    // Keep buffering the possible tag, but hand back the text
+                    // that preceded it: this call is the only one that knows it.
+                    return leadingToken?.isEmpty ?? true ? nil : leadingToken
                 }
             } else {
                 // Otherwise, return the collected text and reset the state.
@@ -619,7 +621,7 @@ public class ToolCallProcessor {
 
         case .collectingToolCall:
             guard let endTag = parser.endTag else {
-                return nil
+                return leadingToken?.isEmpty ?? true ? nil : leadingToken
             }
 
             if toolCallBuffer.contains(endTag) {
@@ -672,7 +674,9 @@ public class ToolCallProcessor {
                 return combine(leadingToken, trailingToken)
             }
 
-            return nil
+            // The call stays buffered until its end tag; the text that
+            // preceded the start tag in this chunk does not.
+            return leadingToken?.isEmpty ?? true ? nil : leadingToken
 
         case .collectingJSONToolCall:
             return processCollectingJSONToolCall(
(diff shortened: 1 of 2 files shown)


aleroot:
@spokvulcan please see #548 which has already completely reworked this area.
The actual tool call processing is too simplistic and needs to be replaced. 
If #548 land, this PR should be considered superseded.

spokvulcan:
> @spokvulcan please see #548 which has already completely reworked this area. The actual tool call processing is too simplistic and needs to be replaced. If #548 land, this PR should be considered superseded.

Thanks, didn't see #548, happy to close this once it lands

Would keep it open until then though, it's a one line fix plus tests that check `processChunk` and the ordered output API return the same thing, if #548 handles the leading-text case, those tests should pass against it too

Also hit some bugs in tokenization and chat history re-rendering that break the prompt cache, will file those separately with fixes

spokvulcan:
Closing: #548 landed and covers this case. The four tests from this PR pass unchanged on current main.

PR #611, opened 2026-09-06
Fix generation loop blocking Swift cooperative workers

### LINKED ISSUE #600: generateLoopTask parks a Swift cooperative-pool worker inside asyncEval
Generation runs its token loop in an unstructured `Task`, and that loop makes a
blocking call. On Swift's cooperative thread pool — which has one worker per
core and does not grow to replace a blocked one — this removes a worker from
circulation for as long as the wait lasts.

## What we measured

`sample` on a shipping app, Swift 6.3.3, macOS 26.6.2, Mac15,6 (18 GB), against
`mlx-swift-lm` at `bd4b7434e6bdb588c7ef55706ff8904cb7fd4c57` and `mlx-swift` at
`0bb916c67f4b9e5c682cbe02a42c701c93ab5021`:

```
Thread_7994062  DispatchQueue_18: com.apple.root.user-initiated-qos.cooperative
  completeTaskWithClosure                        (libswift_Concurrency)
  closure #1 in generateLoopTask<A>(...)         Evaluate.swift:1932
  IteratorProtocol.next() / TokenIterator.next() Evaluate.swift:738
  eval(_:) -> mlx_async_eval                     transforms.cpp:13
  mlx::core::async_eval                          transforms.cpp:307
  mlx::core::eval_impl                           transforms.cpp:252
  std::condition_variable::wait                  (libc++)
  _pthread_cond_wait                             (libsystem_pthread)
```

Samples parked in that `_pthread_cond_wait`, out of all samples on cooperative
threads in each cut:

| Profile cut | Parked / cooperative samples |
| --- | --- |
| 1 | 553 + 391 (two workers) / 3026 |
| 2 | 733 / 3444 |
| 3 (during an app hang) | 1006 / 2137 |

## Why it reaches the cooperative pool

`generateLoopTask` creates its worker with a plain unstructured task:

```swift
let task = Task {
```

No executor preference is passed, and `Evaluate.swift` contains no occurrence of
`executorPreference`, `TaskExecutor`, or `unownedExecutor`. Under
[SE-0417](https://github.com/swiftlang/swift-evolution/blob/main/proposals/0417-task-executor-preference.md),
an unstructured `Task` inherits no task-executor preference from its creator, so
a caller cannot move this loop off the pool from the outside — we tried, which
is how we found it.

`TokenIterator.next()` is documented as scheduling asynchronously:

```swift
// compute the next state and async eval the next token
let token = step(previous: previousY)
y = .init(tokens: token)
asyncEval(token)
```

but `async_eval` blocks in `eval_impl` when the stream has too much outstanding
work, which is what the stack above captures.

## Impact

While generation runs, unrelated `async` work elsewhere in the process contends
for a pool that is short one or two workers. In our app that showed up as
cooperative-worker thread churn — new thread identities across successive
profile cuts — while the pool created and retired workers under the blocked
ones.

## Suggested direction

Give the generate loop an execution context of its own, so a blocking wait
occupies a thread the runtime is willing to replace rather than a cooperative
worker. Either would do it:

- create the loop with `Task(executorPreference:)` on a dedicated `TaskExecutor`,
  optionally caller-supplied; or
- expose the preference in the public API so an embedder can choose, since an
  unstructured task cannot inherit one.

Happy to test a patch against the profiles above.

aleroot:
## Proposed changes

Fixes #600.

`generateLoopTask` runs its token loop in an unstructured `Task`. Although `asyncEval` schedules GPU evaluation asynchronously, it can block the calling thread when MLX applies backpressure to bound outstanding work. That wait occupies a Swift cooperative worker and prevents it from running unrelated async work.

### The fix

Run the synchronous generation loop on a private actor backed by `DispatchSerialQueue`, using its built-in `SerialExecutor` conformance. Each generation has its own queue. The same Swift task hops onto that executor, preserving cancellation and task-local values. Iteration, speculative iterator finalization, and the final `Stream().synchronize()` all stay on that executor until cleanup finishes. One blocked generation does not serialize other generations.

MLX's backpressure remains necessary. This change gives the blocking work an appropriate execution context. Synchronous iterator construction and prefill before `generateLoopTask` remain in the caller's context.

### Dependency

The investigation also exposed wired-memory ticket lifecycle defects in mlx-swift, addressed separately in [ml-explore/mlx-swift#471](https://github.com/ml-explore/mlx-swift/pull/471). Cancellation could release a ticket before generation cleanup finished, and cancellation while waiting for admission could trigger a debug assertion.

**Complete wired-memory cancellation correctness depends on #471.** The scheduling fix works with the currently resolved mlx-swift 0.31.6, but that release still contains those lifecycle defects.

## Checklist

## AI usage

- AI usage disclosure:

### DIFF
--- a/Libraries/MLXLMCommon/Evaluate.swift
+++ b/Libraries/MLXLMCommon/Evaluate.swift
@@ -2324,9 +2324,11 @@ private func generateLoopTask<
     let handler = SendableBox(handler)
     let tokenCollector = consume tokenCollector
 
-    // Launch a Task to perform iteration asynchronously.
+    // Keep task cancellation and task-local values while moving blocking MLX work
+    // onto a separate queue for each generation.
     let task = Task {
-        let performIteration = {
+        let worker = GenerationWorker()
+        let performIteration = { @Sendable in
             var iterator = iterator.consume()
             var handler = handler.consume()
             var tokenCollector = tokenCollector
@@ -2452,10 +2454,10 @@ private func generateLoopTask<
 
         if let ticket = wiredMemoryTicket {
             return await WiredMemoryTicket.withWiredLimit(ticket) {
-                performIteration()
+                await worker.run(performIteration)
             }
         } else {
-            return performIteration()
+            return await worker.run(performIteration)
         }
     }
 
@@ -2746,7 +2748,7 @@ private enum TokenLoopDisposition {
     }
 }
 
-private protocol TokenLoopHandler {
+private protocol TokenLoopHandler: SendableMetatype {
     associatedtype Output
 
     /// Semantic boundaries contributed by the response protocol handled by
--- a/Libraries/MLXLMCommon/Utilities/GenerationWorker.swift
+++ b/Libraries/MLXLMCommon/Utilities/GenerationWorker.swift
@@ -1,0 +1,18 @@
+// Copyright © 2026 Apple Inc.
+
+import Dispatch
+
+/// Runs one generation's blocking work outside Swift's cooperative pool.
+actor GenerationWorker {
+    private nonisolated let queue = DispatchSerialQueue(
+        label: "ml-explore.mlx-swift-lm.generation")
+
+    nonisolated var unownedExecutor: UnownedSerialExecutor {
+        queue.asUnownedSerialExecutor()
+    }
+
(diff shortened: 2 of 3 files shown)


davidkoski (approved):
Change looks good, thank you!

PR #613, opened 2026-09-06
Emit only the new scalars when a token extends the previous character

### LINKED ISSUE #612: [BUG] NaiveStreamingDetokenizer repeats the previous character when a token adds a combining scalar
      drafted.

**Describe the bug**

`NaiveStreamingDetokenizer.next()` emits the part of the newly decoded text that lies past its common prefix with the text it already emitted, and since #465 it measures that prefix in `Character`s (`commonPrefix(with:)` followed by `dropFirst`). A Swift `Character` is a grapheme cluster. When a token appends a combining scalar to the previous character, such as the variation selector U+FE0F, a zero-width joiner or a combining accent, the previous cluster changes, the common prefix stops one character early, and the whole merged cluster is emitted a second time. The consumer receives the base character twice.

Byte-level BPE tokenizers (Qwen, Llama 3, the GPT-2 family) emit those scalars as tokens of their own, so every emoji with a presentation selector, every ZWJ sequence, every flag and skin-tone modifier, and every decomposed accent is affected. `🏳️‍🌈` reaches the consumer as `🏳🏳️🏳️‍🏳️‍🌈`, `'️` as `''️`, and `e` + U+0301 + `x` as `eéx`.

**To Reproduce**

```swift
import MLXLMCommon

// The simplest byte-level shape: decode is the concatenation of the pieces,
// one scalar per token.
struct PiecesTokenizer: Tokenizer {
    let pieces: [Int: String]
    func encode(text: String, addSpecialTokens: Bool) -> [Int] { [] }
    func decode(tokenIds: [Int], skipSpecialTokens: Bool) -> String {
        tokenIds.map { pieces[$0] ?? "" }.joined()
    }
    func convertTokenToId(_ token: String) -> Int? { nil }
    func convertIdToToken(_ id: Int) -> String? { pieces[id] }
    var bosToken: String? { nil }
    var eosToken: String? { nil }
    var unknownToken: String? { nil }
    func applyChatTemplate(
        messages: [[String: any Sendable]], tools: [[String: any Sendable]]?,
        additionalContext: [String: any Sendable]?
    ) throws -> [Int] { [] }
}

func stream(_ ids: [Int], _ tokenizer: any Tokenizer) -> String {
    var detokenizer = NaiveStreamingDetokenizer(tokenizer: tokenizer)
    var out = ""
    for id in ids {
        detokenizer.append(token: id)
        if let piece = detokenizer.next() { out += piece }
    }
    return out
}

let flag = PiecesTokenizer(pieces: [1: "\u{1F3F3}", 2: "\u{FE0F}", 3: "\u{200D}", 4: "\u{1F308}"])
stream([1, 2, 3, 4], flag)  // "🏳🏳️🏳️‍🏳️‍🌈", expected "🏳️‍🌈"

let quote = PiecesTokenizer(pieces: [1: "'", 2: "\u{FE0F}", 3: "'"])
stream([1, 2, 3], quote)  // "''️'", expected "'️'"

let accent = PiecesTokenizer(pieces: [1: "e", 2: "\u{0301}", 3: "x"])
stream([1, 2, 3], accent)  // "eéx" (e, e, U+0301, x), expected "éx" (e, U+0301, x)
```

**Expected behavior**

Streaming the tokens yields the same text as decoding them at once: each call emits only the scalars the new token added. Measuring the common prefix in Unicode scalars instead of `Character`s does that, compares exactly rather than under canonical equivalence, and keeps the #465 behaviour for tokenizers whose decode is not append-only.

 - OS Version: macOS 26.5.2
 - Device: MacBook Pro, Apple M3 Max, 48 GB
 - Version: `main` as of 2026-09-06

**Additional context**

Seen with Qwen3.8-27B behind our OpenAI-compatible server. A file the agent wrote through a `write` tool call came out with `🏳🏳️🏳️‍🏳️‍🌈` where the model had produced `🏳️‍🌈`, and every later turn that touched the file made it worse, because the model kept repairing characters it had never written. It went unnoticed for a while: a duplicated base character in front of a variation selector renders almost like the original. A fix with tests is in #613.

spokvulcan:
## Proposed changes

Fixes #612.

`NaiveStreamingDetokenizer.next()` measured the common prefix between the previous decode and the new one in `Character`s, so a token that appended a combining scalar (U+FE0F, a zero-width joiner, an accent) to the previous character turned the whole merged grapheme cluster into "new" text and re-emitted its base. The prefix is now measured in Unicode scalars, which compares exactly and keeps the #465 behaviour for tokenizers whose decode is not append-only. Four tests in `StreamingDetokenizerTests` stream a ZWJ flag sequence, a variation selector after an ASCII quote, a combining accent, and a variation selector after a multi-byte character split across two tokens, and compare the scalar sequence with the full decode; they fail on `main` and pass with this change.

## Checklist

## AI usage

- AI usage disclosure: the diagnosis, the fix, the tests and this description were written with Claude Code (Anthropic) under the author's direction, and are posted from the author's account.

### DIFF
--- a/Libraries/MLXLMCommon/Tokenizer.swift
+++ b/Libraries/MLXLMCommon/Tokenizer.swift
@@ -113,8 +113,21 @@ public struct NaiveStreamingDetokenizer: StreamingDetokenizer {
         // boundary, so that holds; a decode that rewrote an already-emitted
         // interior character would re-emit the tail (no worse than the old
         // behavior, and not observed in practice).
-        let common = newSegment.commonPrefix(with: segment)
-        let new = newSegment.dropFirst(common.count)
+        //
+        // The prefix is measured in Unicode scalars, not `Character`s. A
+        // `Character` is a grapheme cluster, and a token routinely appends a
+        // combining scalar to the cluster the previous token ended: a
+        // variation selector (`🏳` then U+FE0F), a zero-width joiner, a
+        // combining accent (`e` then U+0301). Compared as `Character`s the
+        // old text's last cluster no longer equals the merged cluster, so a
+        // `Character`-level prefix (`commonPrefix(with:)`) stops one cluster
+        // early and the whole merged cluster is emitted a second time — the
+        // consumer receives `🏳🏳️`, `''️`, `eé`. Scalars compare exactly (no
+        // canonical equivalence), and the emitted tail is precisely the
+        // scalars the new token added.
+        let common = zip(newSegment.unicodeScalars, segment.unicodeScalars)
+            .prefix { $0 == $1 }.count
+        let new = String(newSegment.unicodeScalars.dropFirst(common))
 
         // if the new segment ends with REPLACEMENT CHARACTER this means
         // that the token didn't produce a complete unicode character
@@ -128,6 +141,6 @@ public struct NaiveStreamingDetokenizer: StreamingDetokenizer {
             self.segment = newSegment
         }
 
-        return String(new)
+        return new
     }
 }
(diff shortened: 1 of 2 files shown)


davidkoski (approved):
Looks great, nice find!  Thank you.

Issue #588, opened 2026-08-28
[BUG] metal kernels should always have a plain mlx fallback for future cuda path


davidkoski:
See #573 -- while reading through this I liked the fact that there was a fallback path to plain mlx if the metal kernel wasn't prepared for the particular configuration.  This is great because future cuda efforts could use the fallback.

The isn't necessarily complete or correct, but some simple claude searches identified potential cases where there is no fallback kernel (because I suppose the metal kernel _always_ worked):

 No fallback found — always dispatch the Metal kernel (⚠️ would need work for CUDA)                 
                                                                                                    
 1. Bitnet.swift — BitLinear.callAsFunction → executeMatmulKernel                                   
    Unconditionally calls BitLinearKernelManager.shared.bitlinearKernel. makeBitLinearKernel()      
    returns a non-optional MLXFastKernel, and the manager is a static let shared built eagerly — on 
    a backend where Metal kernel compilation fails, this would likely crash at first BitLinear use  
    (or even at first static-init touch), with no ops-based matmul path at all.                     
                                                                                                    
 2. InterpolationUtils.swift — bicubicInterpolate / nearestInterpolate                              
    Both public functions always route through                                                      
    InterpolationKernelManager.shared.{bicubicKernel,nearestKernel}. No pure-array (e.g.            
    take/gather-based) implementation exists as a fallback. These are public API, likely used in    
    VLM image preprocessing pipelines.                                                              
                                                                                                    
 3. ParoQuant/RotateQuantizedLinear.swift — rotate(_:) in callAsFunction                            
    Always calls getRotationKernel(tile:). No non-kernel Givens-rotation implementation.            
                                                                                                    
 4. MoERouterTopK.swift — moeRouterTopK                                                             
    Does branch between fusedRouterTopK (kernel) and chainRouterTopK (pure ops) — but the branch    
    condition is purely a shape heuristic (gates.size == e, i.e. single-row decode), not a          
    kernel-availability check. RouterTopKKernel.shared.kernel is non-optional and built eagerly, so 
    on a backend without this kernel it would crash the first time a single-token decode step runs, 
    rather than degrading to chainRouterTopK.                                                       

We should strive for always having a fallback in the future and we can deal with these cases (assuming real) over time.

davidkoski:
See also #616 -- another use case is with a VJP on the custom kernels you can't train them, e.g. LoRA.  Note that it is registering VJP to define that, so not entirely the same.

PR #616, opened 2026-09-09
Let training differentiate the gated-delta recurrence

jhancock1975:
## Proposed changes

`gatedDeltaUpdate` runs the recurrence through a fused Metal kernel whenever
one is available. A `CustomKernel` has no VJP, so the first backward pass of a
fine-tune aborts the process:

```
MLX/ErrorHandler.swift:345: Fatal error:
[Primitive::vjp] Not implemented for CustomKernel.
```

Qwen3.5 and Qwen3-Next use gated-delta layers for three of every four layers,
so this is every LoRA fine-tune of those models rather than an edge case. It
is a fatal error rather than a thrown one, so nothing downstream can catch it
and fall back.

Two commits, separable:

1. **Let training differentiate the gated-delta recurrence.** Adds `useKernel`
   to `gatedDeltaUpdate`, defaulting to `true` so inference is untouched, and
   has the three models that call it pass `!training`. This is what the Python
   models already do: `mlx_lm/models/qwen3_5.py` passes
   `use_kernel=not self.training` for the same reason. The ops path is already
   differentiable and is already the fallback for a key dimension that is not
   a multiple of 32.

2. **Recompute gated-delta chunks on the backward pass.** The ops path keeps
   every step's state for the backward pass. At Qwen3.5-27B's shapes, 48 value
   heads of 128x128 in fp32, that is about 4 MB a step, so a thousand-token
   sequence holds gigabytes per layer. The recurrence now runs in chunks of 16
   steps through a `CustomFunction` whose backward recomputes the chunk, which
   is what `mx.checkpoint` does for the Python trainer. Take commit 1 without
   commit 2 if you would rather; commit 1 stands on its own.

### Tests

Two added to `GatedDeltaTests`:

- `testGatedDeltaOpsPathIsDifferentiable` — a backward pass through the ops
  path returns finite, non-zero gradients for q, k and v at T = 40, which
  spans more than one recompute chunk.
- `testGatedDeltaOpsPathMatchesTheKernel` — the two paths agree on the output
  and on the carried state. Compared relative to the values involved: these
  inputs are random and the recurrent state compounds to about 1.2e6 by
  T = 40, where one part in a million is 1.25 in absolute terms.

The three existing `GatedDeltaTests` still pass. Run on an M-series Mac with
the metallib colocated with the test host, since `swift test` alone cannot
load it.

### How the crash was confirmed

A backward pass through `gatedDeltaUpdate` on unpatched `main` (e3d4a20)
exits with signal 5 and the fatal error quoted above. That is why the new test
can only demonstrate the fix by passing: there is no pre-fix state in which it
fails cleanly.

## Checklist

## AI usage

- [ x] I have read this PR description in full and approve it as my own, and it
- AI usage disclosure: I used an AI assistant to help write the code in this pull request. I have created this pull request manually.


davidkoski:
Very interesting -- I was thinking about this from the Cuda angle for #588 but this the same issue.

Libraries/MLXLMCommon/GatedDelta.swift
```diff
@@ -230,6 +230,59 @@ private func gatedDeltaStepOps(
...
+                mask: mask.map { $0[0..., t] }
+            )
+            ys.append(y)
+            state = newState
+        }
+        return [MLX.stacked(ys, axis: 1), state]
+    }
+
+    // The closure holds a locked state object; two callers serialise on it.
+    nonisolated(unsafe) static let unmasked: ([MLXArray]) -> [MLXArray] = CustomFunction {
```
davidkoski, line 271: This should probably produce `Sendable` function on the mlx-swift side.

davidkoski: https://github.com/ml-explore/mlx-swift/issues/475

davidkoski (approved):
Change looks good.  i filed an item for mlx-swift improvements.  Thank you!

PR #620, opened 2026-09-13
Clear the MLX cache on the first generated token

negativetime:
## Proposed changes

`TokenIterator.next()` clears MLX's buffer cache when `tokenCount % 256 == 0`, but it checks after incrementing `tokenCount`, so the first clear happens at token 256. mlx-lm's `generate_step` checks `n % 256 == 0` before advancing `n`, so it also clears on the first token, right after prefill.

For short generations that difference decides whether memory is bounded. A request under 256 tokens never clears, and each one leaves its prefill and decode buffers in the cache. Measured with Llama 3.1 8B 4-bit, 8 requests of about 1,300 prompt tokens and 120 generated tokens, a fresh `ChatSession` for each:

| | footprint after request 1 | footprint after request 8 | MLX cache after request 8 |
| --- | --- | --- | --- |
| main (604fae7) | 5.54 GB | 11.95 GB | 7.44 GB |
| this PR | 4.65 GB | 4.63 GB | 0.09 GB |

The change moves the existing check above `tokenCount += 1`. Replies had the same lengths in both runs (temperature 0).

`TokenIteratorClearCacheTests` puts 256 MB into the cache, takes one token from a tiny quantized Llama, and checks that the cache was returned. It fails on main (268,859,541 bytes still cached) and passes with this change.

The full unit suite passes apart from two cosine checks in `TurboQuantIntegrationTests` (`testAffineKeyModeMatchesReference`, `testStandardAttentionRepeatFactors`). They also fail intermittently on main without this change: one of them failed in one of two runs.

## Checklist

## AI usage

- AI usage disclosure: Claude Code (Claude Opus 5) found the issue while comparing memory use against Python mlx-lm, wrote the change and the test, ran the measurements, and drafted this description.


davidkoski (approved):
Change looks good, thank you.

It does make me wonder if we should do something to coordinate _across_ evaluation streams, e.g. if you are doing this multithreaded.  It isn't a correctness issue.  At worst you might miss a reuse of a buffer and have to allocate a fresh one.  Maybe that is more trouble than it is worth.

PR #623, opened 2026-09-15
Parse GLM-4-0414 markerless tool calls (#491)

### LINKED ISSUE #491: Support GLM-4-9B-0414 tool calling (markerless function-name + JSON)
## Context
`ToolCallIntegrationTests.glm4EndToEnd` is currently `.disabled` (see #481). Investigation of the raw model output showed GLM-4-9B-0414 emits a tool call with no delimiters:

```
get_weather
{"location": "Paris", "unit": "celsius"}
```

Its chat template only lists available tools; it defines no `<tool_call>`/`<arg_key>` markers. The `.glm4` parser (`Libraries/MLXLMCommon/Tool/Parsers/GLM4ToolCallParser.swift`) targets the GLM-4.5/4.7 `<arg_key>` format, so `ToolCallProcessor` never detects a call and the function name leaks into the response text.

## Proposed
Add a function-name-then-JSON parser path plus schema-anchored streaming detection: only treat leading text as a tool call when the emitted name matches a provided tool and is followed by a valid JSON arguments object. Anchoring on the tool schema avoids misclassifying ordinary `word\n{...}` text as a call.

## Acceptance
- GLM-4-9B-0414 produces a `get_weather` call with `location`/`unit` args.
- `.disabled` removed from `glm4EndToEnd`; it passes.
- No regression in the other tool-call formats (json, lfm2, mistral, qwen, xmlFunction).

Refs: #481

aleroot:
## Proposed changes

Fixes #491. Refs #481.

GLM-4-9B-0414 emits a tool call as the function name on its own line followed by the JSON arguments object, with no delimiters:

```
get_weather
{"location": "Paris", "unit": "celsius"}
```

### Design

GLM-4-0414 is handled as a second dialect of the existing `.glm4` format, opted in through a parser capability flag (`supportsMarkerlessNamedJSON`), the same pattern `supportsBareJSON` uses today. No new `ToolCallFormat`; existing formats are untouched.

Detection is schema-anchored: `NamedJSONCallScanner` matches a declared tool name at the start of a line, followed by a single newline and a balanced JSON object. Without declared tools the scanner does not exist, so ordinary `word\n{...}` prose can never be misread as a call. 

## Checklist

## AI usage

- AI usage disclosure:


IntegrationTesting/IntegrationTestingTests/ToolCallIntegrationTests.swift
```diff
@@ -35,17 +35,9 @@ struct ToolCallIntegrationTests {
...
-    // format, so nothing is detected. Supporting this shape needs a dedicated
-    // markerless (function-name + JSON) parser plus schema-anchored streaming
-    // detection; tracked as a follow-up. (glm4FormatAutoDetection still runs.)
-    @Test(
-        .disabled(
-            "GLM-4-9B-0414 emits markerless funcname+JSON tool calls the .glm4 parser can't detect"
-        ))
-    func glm4EndToEnd() async throws {
+    // GLM-4-9B-0414 emits a markerless `get_weather\n{json-args}` call, which
+    // the .glm4 processor anchors on the declared tool names (#491).
```
CharlieTLe:
```suggestion
    // the .glm4 processor anchors on the declared tool names.
```

Tests/MLXLMTests/NamedJSONToolCallTests.swift
```diff
@@ -0,0 +1,282 @@
+// Copyright © 2026 Apple Inc.
+
+import Foundation
+import Testing
+
+@testable import MLXLMCommon
+
+/// GLM-4-0414's markerless `name\n{json}` dialect (#491).
```
CharlieTLe:
```suggestion
/// GLM-4-0414's markerless `name\n{json}` dialect.
```

CharlieTLe (commented):
Some minor nits, but how should the following be supported:

```
write_file\n{"contents": "<tool_call>hello</tool_call> world"}
```

The current implementation would do:

```
write_file(contents: "hello world")
```

But is this the expected behavior?

aleroot:
> Some minor nits, but how should the following be supported:
> 
> ```
> write_file\n{"contents": "<tool_call>hello</tool_call> world"}
> ```
> 
> The current implementation would do:
> 
> ```
> write_file(contents: "hello world")
> ```
> 
> But is this the expected behavior?

@CharlieTLe thanks. this makes sense. GLM4 wrapper tags are now stripped only at payload boundaries, so literal tags inside JSON arguments are preserved. Added regression coverage as well.

